PR #321, opened 2016-03-25
Wiki contribution

vmayoral:
- Erle Robotics profile updated
- Minor changes into the style. Sublime text detected several conflicts with the spaces  in the `common-commercial-support.rts` document. The output of the changes proposed is available at https://github.com/vmayoral/ardupilot_wiki/blob/master/common/source/docs/common-commercial-support.rst. Let me know if you'd prefer this to be omitted and i'll submit another PR.
- Added a few milestones to the history of APM
- Added Erle Robotics store


hamishwillee:
@rmackay9 Are you happy with the changes to the history here - otherwise technically all good for me.

@vmayoral The images are quite large - we're trying to keep the size down because this is all stored on github. Can you convert to jpg, width about 700px and reduce quality if needed. Gimp or similar can do this. I don't question file sizes below 100Kb.

vmayoral:
@hamishwillee, this is broken now due to some other commits. I'll split the commits above in simpler PRs following your advice with the images. 
Thanks

PR #348, opened 2016-03-31
Update out-dated content

pepevalbe:
* Throttle Arming is now enabled by default.
* ARMING_DIS_RUD parameter changed to ARMING_RUDDER.
* One link is wrong.

### DIFF
--- a/plane/source/docs/arming-throttle.rst
+++ b/plane/source/docs/arming-throttle.rst
@@ -5,10 +5,10 @@ Throttle Arming in Plane
 ========================
 
 A software safety feature that requires the throttle to be explicitly
-armed by a pilot is available for Plane (first introduced in version
-2.77). This is especially applicable to electric planes. The idea is to
-prevent an unexpected throttle up when someone's hand is near a
-propeller.
+armed by a pilot is available for Plane. It was first introduced in version
+2.77 and since 3.3.0 it is enabled by default. This is especially applicable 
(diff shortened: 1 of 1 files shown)


plane/source/docs/arming-throttle.rst
```diff
@@ -19,79 +19,75 @@ propeller.
    there is always a slight possibility that signal could unintentionally
    reach the motor. Even though this is unlikely (and made even less likely
    by safety features such as this) it only takes one time to chew up a
-   finger or hand!
-
+   finger or hand!
+
 Parameters governing throttle arming are introduced in the
-:ref:`ARMING <ARMING_CHECK>` section of the
+:ref:`ARMING <ARMING-PARAMETERS>` section of the
```
hamishwillee: This link would be broken. If you run the docs build it will download the parameters.rst file and you can see that the anchor for this section is defined as below:

```
.. _parameters_ARMING_:
```

That means your link actually has to look like this (ie dump the preceding underscore):
```
:ref:`ARMING <parameters_ARMING_>`
```

hamishwillee:
@tridge Is this otherwise sane?

pepevalbe:
I changed the commit with the right link.
Thank you!

hamishwillee:
@gmorph Does this make sense to you and Tridge?

tridge:
looks good, pushed!
thanks very much for the work on this

PR #349, opened 2016-04-01
Update common-stores.rst

rmackay9:
Added EnRoute, changed order to be alphabetical (I don't know of how else we can fairly list the companies)


hamishwillee:
Yes, this is the only fair option. Though we might arguably divide up otherwise by geographical region.

Note that if I am searching for a specific link/hardware that is broken I will always use jDrones for that link because Jani has made every effort to supply every missing component we've had so far. I don't feel the need to be egalitarian in that case.

PR #374, opened 2016-04-12
Update Download Locations

billbonney:


hamishwillee:
@billbonney Why changes from linking the _latest version of the file? Making this change means that you need to update the wiki every time you release a new version.

If you remain keen, then consider putting both of those new points as bullets.

billbonney:
I'm going to fix it so the download are in a folder marked stable/latest/beta for more clarity. I was going to keep the actual versions and supported platforms listed in discuss so that would be updated (but unwittingly included the version number in the name)

hamishwillee:
@billbonney Sounds good. Let me know when you've made the change and updated this and we can review again.

billbonney:
Well the changes are more right than what is there, which is totally wrong, as they link to old files.

hamishwillee:
OK, I will  merge then. I'm relying on you to remember to update to latest links in a timely manner (ie out of my hands)

billbonney:
Yep, promise to fix to something more robust.

PR #382, opened 2016-04-16
PING ADSB Updates

cmmcdermott:
During the edit it appears some lines were added but I did not do that.


magicrub:
Merged, thanks!

PR #396, opened 2016-04-22
Rover Ready-to-Roll: update image to Erle-Brain2 based one

imuguruza:
Update img, now contains Erle-Brain 2


hamishwillee:
In general we try to use as low res image as possible - having lower res images means that syncing the repo is a lot quicker, and page loads are faster. Do you think you could downsize this a little?

imuguruza:
Ok, will do

imuguruza:
Think I've fixed with PR  #408

PR #401, opened 2016-04-22
Dev boards: add discontinued status to Erle-Brain v1

imuguruza:
Just add **discontinued** status to Erle-Brain1


hamishwillee:
Makes sense to me. Merging.

PR #402, opened 2016-04-22
Common sim: add Gazebo sim + Erle simulations

imuguruza:
Add gazebo simulation and links to tutorials of Erle Robotics

### DIFF
--- a/common/source/docs/common-simulation.rst
+++ b/common/source/docs/common-simulation.rst
@@ -17,3 +17,23 @@ Most users should choose the :ref:`Software In The Loop (SITL) simulator <dev:si
 (dev wiki) as it can simulate Copter, Plane or Rover without the need
 for any vehicle hardware, and can be run on Linux, Windows and Mac OSX.
 It is also the easiest form of simulation to set up.
+
+
+Gazebo-based simulations
+========================
+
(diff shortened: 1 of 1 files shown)


hamishwillee:
@rmackay9, @magicrub @tridge I am not at all sure about this. It is an update to the "common" Simulation page (which previously just pointed to the dev page on SITL). 

It adds information about ROS/Gazebo but quickly links off-wiki to erle-brain. I don't mind the ROS section at all, but IMO it should probably link to http://ardupilot.org/dev/docs/using-rosgazebo-simulator-with-sitl.html and then that might be updated to link on to Erle brain.

It is also possible that ROS-Gazebo should be at a higher level than just part of SITL story.

Any thoughts?

imuguruza:
Well, I could make a new PR adding content here if you think it makes more sense

magicrub:
HI,

Yeah, ROS has it's own page and the Erle stuff seems a bit too
marketing-ish to be in the Wiki.

hamishwillee:
@imuguruza I'm rejecting this PR as too "marketing ish". We would accept an update to this page which has a sentence about ROS Gazebo and a link to the ROS Gazebo page but not a high level single vendor solution.

As far as I know all ArduPilot/ROS integrations work via SITL and so should be covered in http://ardupilot.org/dev/docs/using-rosgazebo-simulator-with-sitl.html# . I'd be happy for you to add information to that page, but it would have to either integrate properly or clearly identify what you do differently. If you do do something different then perhaps an Erle-Robotics section under a heading "Vendor Implementations" at the end of that article?

imuguruza:
I'll try to make it less commertial, and fit it better to actual content asap

hamishwillee:
@imuguruza Thanks Iñigo  - we really appreciate that!

PR #408, opened 2016-04-25
Rover Img: update Erle-Rover img

imuguruza:
Resized img
Linked to this PR
#https://github.com/ArduPilot/ardupilot_wiki/pull/396

Hope is fine now


hamishwillee:
@imuguruza Thank you = looks good to me. Just FYI, you can also reduce the resolution without reducing the x/y dimensions (ie the dpi). Gimp tool is quite good at this, as is the linux "convert" tool that is part of imagemagik

PR #410, opened 2016-04-25
Common Imgs: Add PXFmini board imgs

imuguruza:
Add PXFmini board imgs to repository


hamishwillee:
@magicrub In future, please reject image-only PRs like this. If we reject any PR that uses the images we end up with "orphans". Also, I find it hard to assess the other PRs without the images they use.

magicrub:
Ok, sorry!

hamishwillee:
No worry at all (for either of you). Its like code though - PRs should be "self contained" and ideally limited to minimum reasonable size to aid with integration.

PR #412, opened 2016-04-27
abort-landing instructions

magicrub:
The previous abort-landing instructions were convoluted. The formatting is a bit strange, what looks good/bad here in git preview is not what looks good/bad on the html version in the wiki so this is probably a first draft makeover with cosmetic changes later.

@CraigElder and @hamishwillee can you take a look please?

### DIFF
--- a/plane/source/docs/automatic-landing.rst
+++ b/plane/source/docs/automatic-landing.rst
@@ -346,9 +346,9 @@ How to abort an auto-landing
 A landing-abort mechanism is provided to allow you to abort a landing sequence in a safe, controlled, and expected way. Custom abort behaviour can be pre-programmed as part of the mission or you can use the default abort mechanism. To enable this feature set param LAND_ABORT_THR=1.
  
 There are three steps to this feature:
-Step 1) Trigger an abort
-Step 2) The behavior during the abort
-Step 3) The mission state after the abort completes.
+#. Trigger an abort
+#. The behavior during the abort
+#. The mission state after the abort completes.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


plane/source/docs/automatic-landing.rst
```diff
@@ -300,279 +300,234 @@ minimise barometric error problems with auto-land
...
-sequence in a safe, controlled, and expected way. Custom abort behaviour
-can be pre-programmed as part of the mission or you can use the default
-abort recovery mechanism. To enable, set param LAND_ABORT_THR=1.
-
-.. note::
-
+A landing-abort mechanism is provided to allow you to abort a landing sequence in a safe, controlled, and expected way. Custom abort behaviour can be pre-programmed as part of the mission or you can use the default abort mechanism. To enable this feature set param LAND_ABORT_THR=1.
+ 
+There are three steps to this feature:
+Step 1) Trigger an abort
```
hamishwillee: These should still be list items. IMO should look as below. The spacing below is required.
```
There are three steps to this feature:

#. Trigger an abort
#. The behavior during the abort
#. The mission state after the abort completes.
```

plane/source/docs/automatic-landing.rst
```diff
@@ -300,279 +300,234 @@ minimise barometric error problems with auto-land
...
+.. note::
+
    This section describes the abort behavior introduced in Plane
-   3.4.
-
-Abort triggers
+   3.4.
+
+   
+Step 1) Abort land triggers
```
hamishwillee: This however is IMO fine :-)

plane/source/docs/automatic-landing.rst
```diff
@@ -300,279 +300,234 @@ minimise barometric error problems with auto-land
...
-   command).
-
-The climb-out procedure is complete when the abort altitude is reached.
-Plane will then change the mission index to allow for a predictable
-landing approach:
-
--  Jump to a ``DO_LAND_START`` command if there is one in the mission.
--  Otherwise decrement the mission index to restart the glide slope.
-
+The abort behaviour has a default configuration and does not require a pre-planned mission or special configuring. The default abort behavior is to simulate an auto-takeoff: pitch up at least 10 degrees and set throttle to TKOFF_THR_MAX and hold the heading until it reaches a target altitude of 30m. It is possible to override the pitch and altitude to allow for a customized behavior.
```
hamishwillee: "special configuring" sounds wrong. Probably 
    
    The abort behaviour has a default configuration (by default there is no requirement for a pre-planned mission or special configuration).

hamishwillee:
Yes, this is much improved IMO. There is more that could be done in tidying up (as you note) but worth pushing out as-is.

magicrub:
fixed

magicrub:
I'll go ahead and push for now, we can tweak it some more later. I'd like to add a little flowchart type image too

hamishwillee:
Sounds good to me. Definitely an improvement. Cheers!

CraigElder:
Thanks Tom this is much easier to read. 
One case that does not seem to be covered off is the use case where the operator just sends a landing sequence that does not include a takeoff waypoint.

magicrub:
@CraigElder Does step 2 need clarity?

Step 2) Abort land flight behavior
"The abort behaviour has a default configuration and does not require a pre-planned mission. The default abort behavior is to simulate an auto-takeoff: pitch up at least 10 degrees and set throttle to TKOFF_THR_MAX and hold the heading until it reaches a target altitude of 30m. It is possible to override the pitch and altitude to allow for a customized behavior."

CraigElder:
I missed it the first time but the instructions seem clear on the second read. 

Maybe "The abort behaviour has a default configuration and does not require a pre-planned mission or a mission with a takeoff waypoint"

magicrub:
done

PR #416, opened 2016-04-29
Autopilot boards: change wording

ivereninov:
Closed solution might confuse users that software is closed.


magicrub:
merged, thanks!

PR #423, opened 2016-05-01
Pixracer_production_pics

lvale:
Changed old pictures with retail board


jpkh:
Normally we do not allow SPACEs on file names. I will fix this and approve PR

PR #430, opened 2016-05-04
Add ArduPilot logos with alpha

billbonney:
Should look nicer and the original larger image for reference


hamishwillee:
These look good. Feel free to update the theme to pull these in (currently it uses the discussion board version).

PR #434, opened 2016-05-05
Add Navio in history of Ardupilot

i-smirnov:


hamishwillee:
@rmackay9 @magicrub You OK with this change? The first add of RaPi shield is definitely news. Upgrades of the products perhaps not so much, but I don't see as harmful or unreasonable.

magicrub:
@hamishwillee I have to agree with you, The NAVIO is fine but iterative upgrades start to clutter the history

hamishwillee:
@i-smirnov So there you have it. We'll definitely accept the first part of the update thank you, but would prefer not to have all the updates ... can you make that change? This makes sense since the first addition is ArduPilot "news" but the rest is really Navio news.

i-smirnov:
@hamishwillee Thank you! I removed redundant Navio upgrades from history and updated the PR.

magicrub:
Merge, thanks!

hamishwillee:
Late to the party - thanks @i-smirnov - and @magicrub

PR #446, opened 2016-05-20
Sitl : Update instructions

khancyr:
Update sitl instruction to use sim_vehicule.py instead of .sh version and by that way, use waf build.
Update package install according to https://github.com/ArduPilot/ardupilot/blob/master/Tools/scripts/install-prereqs-ubuntu.sh
Correct instruction for ubuntu 16.04


khancyr:
Please don't merge until I update other variant (windows, VB etc , it is quite a mess...)

hamishwillee:
No worries! Thanks for this.

magicrub:
What's the status of this? It has merge conflicts. Should we just close and submit a new one when you're ready?

khancyr:
Yep should be closed... don't have time to work on that now...

PR #456, opened 2016-05-31
update from oirgin

glody:


hamishwillee:
I see you've closed this. If you do post a change like this would be best to do in smaller chunks that can be reviewed and explain what the changes are about.

PR #475, opened 2016-06-17
add "python-wxtools"

harry169:
Because on the latest version of Ubuntu 16.04 ,you can no longer install 'python-wxgtk2.8' using apt-get command. And this will lead to a failure when you ues 'sim_vehicle.sh' to open the console and map window. Hence, use 'sudo apt-get install python-wxtools' will fix this problem.

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-linux.rst
+++ b/dev/source/docs/setting-up-sitl-on-linux.rst
@@ -69,7 +69,7 @@ this:
 
 ::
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@harry169 - thank you. Will these instructions work on Ubuntu 14.04 LTS? People still work on the LTS version as often as not. If not, then we need to add this change as a "version specific note".

magicrub:
merged https://github.com/ArduPilot/ardupilot_wiki/commit/57ade8752a0e40fac35c5bb1d7c97ad3ee3f180f

PR #486, opened 2016-07-07
improved grammar

alan707:
Apostrophes are only required for it has or it is.  See: http://www.its-not-its.info/


hamishwillee:
Absolutely. Thanks.

PR #510, opened 2016-08-04
firmware update to be released with 3.4

ctech4285:


common/source/docs/common-electro-permanent-magnet-V3.rst
```diff
@@ -57,7 +57,7 @@ Firmware setup
  
 .. note::
 
-    A pull request has been issued to include the changes in the next APM release. This documentation will be update when the request has been accepted. If this annoys you please write a comment `here <https://github.com/ArduPilot/ardupilot/pull/4361>`__
+    The firmware update is in maser and do, to be released with Copter-3.4A.
```
hamishwillee: Spelling errors. Should be 

    The firmware update is in master and is to be released with Copter-3.4A.

common/source/docs/common-electro-permanent-magnet-V3.rst
```diff
@@ -17,7 +17,7 @@ Objects that it grips onto should be metal or have the included `thin steel targ
 attached to them.
 
 The device accepts a standard PWM servo input or UAVCAN and uses a 5V power supply.  
-Significant amounts of power are only required for about about 1 second when it first grips or releases the cargo.
+Significant amounts of power is only required for about 1 second when it first grips or releases the cargo.
```
hamishwillee: "are" is correct when you refer to plural of something - so "amount of power is" or "amounts of power are". I suggest perhaps "A significant amount of power is"

ctech4285: What about power, is it not quantized and should be used in plural as well?

ctech4285:
Thank you Hamish all changes done!

hamishwillee:
Thanks @ctech4285 - merged.

> What about power, is it not quantized and should be used in plural as well?

In this case power is actually "singular". If you were to talk about powers you would be referring to multiple types of power (e.g. I have the following superpowers: grammar, spelling, editing).

PR #514, opened 2016-08-11
Correct references of sim_vehicle.sh to sim_vehicle.py

peterbarker:
Also sneaks in a fix for the way ADSB simulation is triggered.

@magicrub Do the fixes for simulating ADSB in SITL look correct to you?

### DIFF
--- a/common/source/docs/common-ads-b-receiver.rst
+++ b/common/source/docs/common-ads-b-receiver.rst
@@ -168,6 +168,8 @@ to set the ``SERIAL3_`` parameters:
    SERIAL3_PROTOCOL 1
    SERIAL3_BAUD 57600
 
+At the SITL console set the quantity of simulated aircraft to 5 with the command ``param set SIM_ADSB_COUNT 5``. Values 0 - 100 are available. You can also set the altitude ``SIM_ADSB_ALT`` and radius ``SIM_ADSB_RADIUS`` of the simulated aircraft
+
 TODO
 ====
 


common/source/docs/common-ads-b-receiver.rst
```diff
@@ -154,7 +154,9 @@ When starting SITL use the following command:
 
 ::
 
-    sim_vehicle.sh -A --adsb --console --map
+    sim_vehicle.py --console --map
+
+At the SITL console set the ``SIM_ADSB_COUNT`` parameter to idnicate the number of nearby aircraft to simulate.
```
magicrub: spelling: idnicate

magicrub: I would suggest also mentioning the other options:
At the SITL console set the quantity of simulated aircraft to 5 with the command ``param set SIM_ADSB_COUNT 5`` . Values 0 - 100 are available. You can also set the altitude (SIM_ADSB_ALT) and radius (SIM_ADSB_RADIUS) of the simulated aircraft

hamishwillee:
Hi @peterbarker - so is this some different sim_vehicle script? Every other reference on the wiki to SITL uses the .sh version (and that is what I've tested). What is the reason for the change, and does it need to happen elsewhere.

PS, this [link shows just the file changes](https://github.com/ArduPilot/ardupilot_wiki/pull/514/files?w=1).

peterbarker:
On Sun, 14 Aug 2016, Hamish Willee wrote:

> Hi @peterbarker - so is this some different sim_vehicle script? Every other
> reference on the wiki to SITL uses the .sh version (and that is what I've
> tested). What is the reason for the change, and does it need to happen
> elsewhere.

sim_vehicle.sh has been removed in favour of sim_vehicle.py. 
sim_vehicle.py is pretty much a drop-in replacement and has a few 
advantages - including code reuse with autotest.py.  Major advantage being,
  "not shell script", of course :-)

hamishwillee:
@peterbarker Thanks. So would it be correct to say that this change needs to happen in all our docs, and if so, from what "version" should we suggest that the instructions change (ie if you're building Copter 3.4 and earlier do you still use .sh)?
Should instructions include the "python" call or is that implicit?

magicrub:
I swear I've typed in similar decision somewhere.

hamishwillee:
@magicrub haven't seen it, but not looking closely.

magicrub:
Auto correct kills me. Meant description instead of decision. Anyway, I think the change I was thinking of was on that other PR that I just merged. This now conflicts.

hamishwillee:
Fun for @peterbarker to solve :-)

peterbarker:
Moved the sim-vehicle changes to https://github.com/ArduPilot/ardupilot_wiki/pull/519, leaving just a trivial addition to the ADSB change.

peterbarker:
@hamishwillee sim_vehicle.py is applicable across all supported vehicles, so referencing Copter version anything is not right.  

It is fair that the docs should change for developers - anybody that does a git update will be missing sim_vehicle.sh, so there's no sensible cutoff.  Perhaps some sort of note somewhere about sim_vehicle.sh being renamed?  Where exactly that might go I don't know.

hamishwillee:
I don't think we can/should reasoably add a note about the filename change everywhere, so possibly just in these two sections:

http://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#start-sitl-simulator
http://ardupilot.org/dev/docs/sitl-native-on-windows.html#set-up-directories-paths-in-cygwin

peterbarker:
No longer required.

PR #518, opened 2016-08-16
Update link to DJI manuals

tomlauzon:


hamishwillee:
Thanks for the fix @tomlauzon !

PR #537, opened 2016-09-04
Theme: responsive table width cell wrap

tomlauzon:
This adds a custom css style to the RTD theme that makes the tables fit within the page instead of having a scroll bar at the bottom. Text will wrap within cells.

### DIFF
--- a/antennatracker/source/_static/theme_overrides.css
+++ b/antennatracker/source/_static/theme_overrides.css
@@ -1,0 +1,13 @@
+/* override table width restrictions */
+@media screen and (min-width: 767px) {
+
+   .wy-table-responsive table td {
+      /* !important prevents the common CSS stylesheets from overriding
+         this as on RTD they are loaded after this stylesheet */
+      white-space: normal !important;
+   }
+
+   .wy-table-responsive {
+      overflow: visible !important;
+   }
+}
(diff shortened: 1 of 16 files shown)


hamishwillee:
Hi Tom

Couldn't we instead push this into the theme (rather than as an override)? Then it would be in just one place. I would prefer that if we can.

magicrub:
what's the status of this? Close in favor of redoing it?

hamishwillee:
Yes, I think it would be better to close this and redo it. @tomlauzon - is that something you can do? I can look into it, but wouldn't be until Xmas.

tomlauzon:
I tried many things but did not succeed in simplifying it. I think the static theme is the right way to go because when the theme gets updated we would lose the table style if we modified the theme directly (like child themes in Wordpress). I haven't found a way to add the static theme by modifying only one or two files.. I think we would need to write a bit of python code to copy the modification to each site at each compile instead of having the mod in each of the 7 sites. 
If you think that would be a good solution I could work on it. 

I thing this table is hard to read: http://ardupilot.org/copter/docs/common-telemetry-radio-regional-regulations.html

hamishwillee:
Thanks. I've created theme update above for your review.

PR #548, opened 2016-09-12
Update common-history-of-ardupilot.rst

virtualrobotix:
Add in the history some important step in Ardupilot 32 bit support all the info is available on old blog post available on DYD or Virtualrobotix Blog Post.


magicrub:
the April 2011 one, with the perspective changed, sounds like the only one worthwhile to add. The others seem like detail overkill. What does everyone else think about this?

hamishwillee:
I agree with you @magicrub . This is the ArduPilot history, and we don't want to hide it in the virtualrobotix history.

If we do keep these, then they would need a subedit (English is "variable")

davidbuzz:
I've added the 2011 one, as discussed. at [bf4da6a] closing this PR now.

PR #549, opened 2016-09-12
Create introduction_es.rst

donkrip:
Maybe would be interesting a version of the documentation in Spanish, this is a small contribution to begin.  

Introducción de Copter en español.

### DIFF
--- a/copter/source/docs/introduction_es.rst
+++ b/copter/source/docs/introduction_es.rst
@@ -1,0 +1,130 @@
+.. _introducción:
+
+==================
+Introducción Copter
+==================
+
+Copter es un sistema avanzado de autopiloto avanzado para multirotores,
+helicopteros, y otros vehiculos con rotores.
+
+Overview
+========
+
+Copter es una solución open-source de auto-piloto para vehiculos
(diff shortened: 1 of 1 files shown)


hamishwillee:
@donkrip THanks very much. We don't yet have a multilingual strategy. If we do then we'd probably use the approach outlined here: http://www.sphinx-doc.org/en/stable/intl.html

I need to do some work to get this all set up. Can you check it out and see how you might create and build the translation files for Spanish?

OXINARF:
@hamishwillee Related to this: we should probably consider an online service where people can do translations easily.

hamishwillee:
Yes, we've tried transifex before, and indeed everything else set up to support transifex live service. Problem is that it takes a lot of admin, so I've been considering moving to the approach where we use sphinx translation (native) but have round trip synchronisation with transifex for translation of the files.

magicrub:
if we're working toward a multi-language solution, shall we close this single-page translation PR?

rmackay9:
I guess we should close this because we need to add multi-language support to our wiki first.

PR #553, opened 2016-09-18
Update Linux Version to 16.04 LTS

billbonney:
Updated Linux version to 16.04 LTS for pre-built binaries. but older versions are supported with correctly updated dependencies.


hamishwillee:
Hi @billbonney - So it looks like the only change here is to state support for 16.04. Is that correct? (i.e. you haven't actually stated what dependencies are needed for earlier versions right?)

billbonney:
I'll add stuff about 14.04 I have commit access so consider this PR a wip

PR #556, opened 2016-09-22
Enabling serial connection on Raspberry Pi

Eos2350:
On newer versions of the Raspberry Pi, for some bizarre reason UART is disabled by default. This fix tells readers how to fix that in order for the RPi to talk to the Pixhawk.


magicrub:
Merged, thanks!

PR #563, opened 2016-10-03
added missing stem to install gcc

proficnc:


magicrub:
Isn't this installed in the script? If not, it should be added to it. I don't think adding an additional step on the wiki is needed. Closing.

PR #566, opened 2016-10-05
General update / review post Dronecode split

auturgy:
Some general ammendments are suggested, largely related to removal of Dronecode references and some minor terminology tweaks.


billbonney:
Thanks 👍

PR #570, opened 2016-10-10
remove redundant and older text

magicrub:
comparing these two lines, I chose the top versions which seemed more verbose and therefore newer:

AUTOTUNE_AGGR...
 Higher values will produce a more aggressive tune but sometimes results in gains that are too high. More specifically this parameter controls the threshold for D-term bounce back and P-term overshoot. 
vs
Controls the threshold for D-term bounce back and P-term overshoot.

AUTOTUNE_AXES ..
This is useful if the vehicle’s battery life is not long enough to complete all 3-axis).
vs
(useful if the vehicle’s battery life is not long enough to complete all 3-axis).


magicrub:
@rmackay9 please take a look at this before we merge. There's some redundant text in Copter Autotune wiki

PR #582, opened 2016-10-22
dev: added Build Ardupilot on Windows10

taedla01:
Tutorial: How to compile Ardupilot on Windows 10 x64


taedla01:
Hello everyone,

Please take in care that this is my first PR to the new wiki system... Not sure about if it has the correct text format or correct image implementation. Pay attention and feel free to edit.

Regards from Spain,
Dario.

magicrub:
I think someone is working on a script to make installation a lot easier, like we have for ubuntu. Let's hold off on this for another week or so, there may be major changes needed.

billbonney:
Merged. Please verify that it looks ok when published to the site

@magicrub we can always rework when a script is published.

hamishwillee:
I'm with @billbonney  - its better to have something in place.

This has some typos and layout issues. But the biggest change I'd make is to simplify the title to: "Building ArduPilot on Windows with Bash" - Inside the doc make it clear that this is Windows 10, and Bash is the Ubuntu version.

The headings layout is also odd. I'll see what I can do for a quick fix.

hamishwillee:
OK, I fixed SOME of this https://github.com/ArduPilot/ardupilot_wiki/commit/eb6e78df6dc4fcac235e1e8812abca7a5f94a6fd . But the first image isn't properly referenced or is missing.

There are sections created with HR breaks (e.g. lines of ----). These would be better as headings. Any reason why not?

This needs a further subedit, but even as-is is quite useful.

taedla01:
@hamishwillee: Nice to see you again! First image is missing because I forgot to upload it... Sorry :( 
Text is a copy and paste from my post in DIYDrones and then I have merged it into an existing .rst file (taked as reference). Is there any tutorial about how to make docs for the wiki? What is the correct format for text?

Sorry for the inconveniences. Here is the lost image:

windows10_ubuntu_0


Cheers,
Dario.

hamishwillee:
Hi Dario

The wiki editing guide is here: http://ardupilot.org/ardupilot/docs/common-wiki_editing_guide.html

The best thing about the doc is that it shows you how you can build the docs yourself and check out the results.

Can you add the image and consider whether the  horizontal rule breaks could be changed into headings describing each section?

PR #587, opened 2016-10-27
Copter: update MAV_CMD_CONDITION_YAW doc

expntly:
param2 (yaw rate) seems supported in the code


expntly:
Nevermind, it turns out Copter::yaw_look_at_heading_slew is actually present in the code and set but never unused/read.

PR #596, opened 2016-11-09
autotest.py parameter was changed.

credtiger96:
In previous code, 
  ./Tools/autotest/autotest.py build.Copter fly.Copter logs.Copter --map --viewerip=192.168.184.1

build.Copter, fly.Copter, logs.Copter do not work anymore. 

So I tried to make autotest.py work. 
I think 
    ./Tools/autotest/autotest.py build.ArduCopter fly.ArduCopter --map --viewerip=192.168.184.1
works properly.

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-windows.rst
+++ b/dev/source/docs/setting-up-sitl-on-windows.rst
@@ -181,7 +181,7 @@ terminal on the Ubuntu VM:
 
 ::
 
-    ./Tools/autotest/autotest.py build.Copter fly.Copter logs.Copter --map --viewerip=192.168.184.1
+    ./Tools/autotest/autotest.py build.ArduCopter fly.ArduCopter --map --viewerip=192.168.184.1
 
 Next connect with the mission planner after first setting the "COM Port"
 to "UDP".


peterbarker:
This is definitely a worthwhile fix, thanks!

Unfortunately, it looks like your text editor has changed all of the line endings throughout the modified file.

This means that the diff is far larger than it needs to be, and makes it hard to verify what has changed.  You also now get blamed for *every* wrong thing in the file :-)

Could you ask your editor not to play around with line endings and fix the patch, please?

magicrub:
Peter,

Regarding the CRLF thing. That's actually not always the case. I can edit a
single line and GitHub will change them all... sometimes. I think it's a
bug. Obviously, in some cases what you said is right but the wiki randomly
has that problem regardless of what the local file ending in the editor is.

Merging... Thanks!

peterbarker:
@credtiger96 Did you use github to edit the file?

credtiger96:
@peterbarker  yes. I used github to edit.

hamishwillee:
@peterbarker We recommend using github for minor edits, and I wouldn't want to change that because it is so much easier if people are making occasional changes. We can always hide all the spurious line ending when reviewing information by adding ?w=1 to the github url.

magicrub:
In this case, that trick did not work.

hamishwillee:
Really? Looked OK to me - showed a change markup on these lines only.

-    ./Tools/autotest/autotest.py build.Copter fly.Copter logs.Copter --map --viewerip=192.168.184.1
+    ./Tools/autotest/autotest.py build.ArduCopter fly.ArduCopter --map --viewerip=192.168.184.1

PR #600, opened 2016-11-13
new USA store option for Mauch power modules

tomlauzon:


gmorph:
Hi Tom.  We really appreciate the updated information.  However this PR contains a bunch of non-changed changes e.g. it thinks there have been lines removed and then put back in.  Same goes for comments.  From what I can see the only thing you have changed is the 'available for purchase in the USA' bit.  Are you able to remove all the other changes that aren't actually changes?  If not I can just update the wiki page with your 'purchase in the USA' bit.  Let me know.
Thanks, Grant.

tomlauzon:
I just used the github web interface to do this and all these non-changed changes appeared. I'll try this again using the command line tool. Sorry about that and thanks.

gmorph:
Ah thanks Tom.  Yes the github interface does a sub-optimum job.  Its not a big deal as this is only the documentation wiki i.e. if you break it things don't fall out of the sky :-)
The extra admin isn't worth it - I'm going to apply the PR as is - thanks.
Thanks, Grant.

tomlauzon:
yes, I figured for a wiki page it wouldn't make a big difference. I admit I was a bit lazy. Thanks!

hamishwillee:
FYI all, adding ?w=1 at the end of the file url gives you a view where you can ignore line ending changes: https://github.com/ArduPilot/ardupilot_wiki/pull/600/files?w=1

PR #604, opened 2016-11-17
dev: fixed sim_vehicle location

justineaster:


dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -60,12 +60,12 @@ location in the
...
 file.
 
-For example, to start Copter in *Ballarat* (a named location in
+For example, to start the simulator in *Ballarat* (a named location in
 **locations.txt**) call:
 
 ::
 
-    cd ArduCopter 
+    cd ardupilot/Tools/autotest/
```
peterbarker: This is a fair indication that sim_vehicle.py is not in your path.

Being in the same directory as sim_vehicle.py *should not change that*; `.` should never be in your path.

I think this would be better formulated without the `cd`, at all:
```
./Tools/autotest/sim_vehicle.py -v ArduCopter ...
```

peterbarker: Alternatively, the traditional way to run sim_vehicle was:
```
cd ArduPlane
../Tools/autotest/sim_vehicle.py ...
```

peterbarker: Either way - being in the autotest directory is not the correct solution :-)

rmackay9:
Ok, this doesnt seem like a correct change so closing (without merging)

PR #611, opened 2016-11-26
Update common-pixhawk-and-px4-compatible-rc-transmitter-and-receiver-…

OnkelMatt:
…systems.rst

Updated document with MULTIPLEX SRXL connetion and RX test results done with tridge


tridge:
applied, thanks!

PR #617, opened 2016-11-29
Added anchour, fixed and updated links, formatting

OnkelMatt:


magicrub:
All link changes look fine, I'm wondering about the capitalization of "FrSKY". It seems that all the logos are shown as "FrSKY" but all text, even on the the official website, it's written as "FrSky". Either way, you're correct that 'FRSky" is not right.

OnkelMatt:
Not shure what is finaly the right way to write it. Their logo seem to me to be written "FrSKY"...Just thought that everything should be written in the same way. Shall I change to "FrSky" ?

hamishwillee:
Nice one. Works for me.

PR #626, opened 2016-12-01
added links to params and values

magicrub:
This is a follow up to https://github.com/ArduPilot/ardupilot_wiki/pull/625 by adding links to params and values. @rmackay9 and @FabianL77 please confirm this is correct


rmackay9:
Merging as this makes it more convenient for users to find the parameter descriptions.
In terms of content i dont actually think theres any need to change the battery failsafe parameter.

PR #630, opened 2016-12-04
Update common-ublox-gps.rst

WickedShell:
Following a discussion on discuss.ardupilot.org/t/m8n-config-file-in-wiki-corrupt/12979/ it was decided that the u-blox config files are far to out of date to be trusted for reliable loading so this removes all mentions of the config loading. However the passthrough/wiring to the GPS and u-center is still a useful tool for debugging any GPS problems, so I've left that. (And removed the APM2 stuff).

Never worked with the wiki before, and the preview tab on the web interface doesn't work so feel free to critique away!


rmackay9:
There's one typo where "configuration" is missing the starting "c" but merging anyway cuz its an overal improvement.

PR #632, opened 2016-12-05
Update editing-the-code-with-eclipse.rst

R-Lefebvre:
Remove Dead Link


R-Lefebvre:
Will follow up with another change.

PR #634, opened 2016-12-08
Update set-motor-range.rst

adnan-ulhaque:
adding a WARNING comments for a procedure where APM is getting power from both Lipo and USB (while JP1 is OFF).

### DIFF
--- a/copter/source/docs/set-motor-range.rst
+++ b/copter/source/docs/set-motor-range.rst
@@ -18,6 +18,11 @@ Measuring the deadzone
 -  Remove the propellers from the vehicle
 -  Connect the lipo battery
 -  Connect the flight controller to the Mission Planner using a USB cable or telemetry
+
+Warning
+Connecting USB when you have input voltages at the high end of the range (near Abs Max) can damage the board. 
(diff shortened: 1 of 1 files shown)


rmackay9:
Hmm.. this seems to have a lot of white spaces changes which makes it hard to know what's changed..

gmorph:
I suspect its the stupid github tool that causes all the whitespace changes.  There is a way to exclude the whitespace changes when you view the page but I can't recall what it is off the top of my head.

billbonney:
Add ?w=1 i.e. https://github.com/ArduPilot/ardupilot_wiki/pull/634/commits/bb736f20969320e4dd38baa2f6c84d83b8e7c11b?w=1

gmorph:
Thanks Bill!

rmackay9:
I think perhaps this is only true for the APM boards?  We don't actually support these boards anymore but I suppose it's ok to add the warning if we make it more clear that it only applies to the APM boards.

magicrub:
this has a merge conflict

magicrub:
also, the text needs to mention that this fix is only for the old 8bit APM2.x hardware

PR #639, opened 2016-12-22
include a brief description of what we sell

tomlauzon:
Might help people looking for these products.


lvale:
@magicrub Tom, it might be helpful to also fill the products for the other shops ??

PR #648, opened 2016-12-31
Fix dos file endings under plane directory.

grahamjamesaddis:
Fixed the dos line endings under the plane sub directory.


rmackay9:
Merged, thanks!

PR #671, opened 2017-01-08
Fix erroneous CR characters in files

grahamjamesaddis:
Deleting erroneous CRs within files, or replacing single CRs with LFs, as appropriate.


billbonney:
Probably best just to remove the CR in html tags and not replace with a LF as whitespace is ignored when rendered into the final page.

PR #672, opened 2017-01-08
Update index.rst

El-Mango:
Pic broken link fixed


billbonney:
the image is right here https://github.com/ArduPilot/ardupilot_wiki/blob/master/copter/source/images/home_copter.jpg

This fix is not required.

PR #681, opened 2017-01-16
Corrected the python "Roll Example" program

gerrievanzyl:
Python structure is defined by indents.  The indents are completely incorrect and the program would not execute.

Fixed the indents - the program should now run


rmackay9:
Thanks!!

PR #685, opened 2017-01-19
Dev: Change the release version of gcc-arm cross-compiler.

muramura:
Correct to the version provided on the page of http://firmware.ap.ardupilot.org/Tools/PX4-tools/.


rmackay9:
Thank you very much!!

PR #686, opened 2017-01-20
gazebo: added instructions on using pre-build packages

tridge:


dev/source/docs/using-gazebo-simulator-with-sitl.rst
```diff
@@ -182,14 +199,14 @@ In one terminal, enter the ArduCopter directory and start the SITL simulation:
 ::
 
     cd ~/ardupilot/ArduCopter
-    sim_vehicle.py -f gazebo-iris
+    sim_vehicle.py -f gazebo-iris -D --console --map
```
khancyr: I would not add -D parameter as default but your call

dev/source/docs/using-gazebo-simulator-with-sitl.rst
```diff
@@ -136,7 +136,24 @@ Then find the file libgazebo_common.so.1, probably under `/usr/local/lib` or `/u
...
+
+Instructions for installing the "SASC" version of gazebo that works
+with ArduPilot SITL are here:
+
+  https://github.com/osrf/uctf/tree/master/doc/install_binary
+
+That will install gazebo in /opt/sasc, so you need to also do:
+
+::
+   export PATH=/opt/sasc/bin:$PATH
```
khancyr: better to do something like echo 'export PATH=/opt/sasc/bin:$PATH' >> ~/.bashrc or people will complain that is not working after a reboot !

khancyr:
well already merge ... Got only two little comments, but I will correct that on next revision !

PR #688, opened 2017-01-24
fix a couple of grammar mistakes

larsks:
Correction Guy (http://knowyourmeme.com/memes/correction-guy) says
"you're".


rmackay9:
Thanks very much!  every bit helps!

PR #695, opened 2017-01-29
Update compiler version

jmachuca77:
using brew install gcc-arm-none-eabi would install Gcc version 4.7, a version between 4.9 and 5.9 is needed to compile so in order to install a 4.9 version brew install gcc-arm-none-eabi-49 should be used


hamishwillee:
Makes sense.

hamishwillee:
@rmackay9 @magicrub This made sense to me. I have merged.

PR #699, opened 2017-01-31
Update raspberry-pi-via-mavlink.rst

camow7:
Added a note to use the Unarchiver software to correctly extract the .img file on Mac. Currently the default archiver software gets caught in an infinite loop of extracting between a .xz file and a .cpgz file.


magicrub:
@camow7 can you please add a version number to this statement so if/when it gets fixed the same text is useful and correct?

rmackay9:
Merged, thanks!

PR #709, opened 2017-02-12
Link checking+fixes

sslaughter:
link_check.py is a script that reads through all the wiki pages, attempts to parse out all the urls within and then tests them to see if they're active. All of the links changed here were found through this script. I wasn't sure if the script itself should be included in the repo, but I figured I'd put it in just in case.

I think most of the links were simply replaced, but there are some that were deleted because I couldn't find a suitable replacement. Also, there are a couple more I wasn't sure about that I intend to create issues for. Finally, I haven't checked the built versions of all of these files, but I did go over the ones I was most unsure about.


hamishwillee:
Thanks very much - looks good to me! 

The link check file is part of sphinx, so should not be included - can you please remove that and then I'd be happy to merge.

sslaughter:
It's removed, I actually didn't realize that was part of sphinx and wrote my own lesser version... thanks for the heads up.

hamishwillee:
@magicrub - This looks sane to me. I am merging. Thanks Scott!

common/source/docs/common-downloads_advanced_user_tools.rst
```diff
@@ -421,22 +392,6 @@ DATE POSTED	April 25, 2013
 Arduino USB driver for the Atmega32u2 USB interface chip on APM 2
```
hamishwillee: Hold on, did the parameter file disappear? Should this just be added back?

sslaughter: I wasn't able to find anything and it referenced the HIL simulator which i thought wasn't supported anymore. Is this file needed for general flightgear setup w/ SITL?

PR #716, opened 2017-02-16
Rover: sitlmavproxy remove uneeded note

khancyr:
this is unneeded as sim_vehicule.py already auto load the defaults params


hamishwillee:
It wouldn't have mattered to leave this in place, but nor does it hurt to remove it. I'll merge - thanks.

PR #717, opened 2017-02-17
Update Java note to specify 32 bit

b70dub:
The java link automatically took me to the 64 bit download (which I did not catch before installing). This caused errors when trying to open eclipse (Java returned error code 13).


magicrub:
is your OS 32bit?

magicrub:
merged, I'll add a bit about it having to match Eclipse which is the important bit

PR #718, opened 2017-02-17
Subsystem for Linux on Windows 10

dave-dave:
Tested, works as advertised. Hardware interaction for HITL is another question entirely


OXINARF:
We have an article for building in Windows 10 with the Linux Subsystem: https://github.com/ArduPilot/ardupilot_wiki/blob/master/dev/source/docs/building-ardupilot-onwindows10.rst

Do you think this is really needed?

By the way, HITL isn't supported anymore.

dave-dave:
Not desperately. But no one was sure if it would work, so I thought I'd put
a note in there.
After helping various people with Cygwin a couple of weeks ago, it seemed
that this might be an easier way to solve the problem for some of them.

dave-dave:
..

dave-dave:
I didn't realise there was already an article on that.

OXINARF:
The article doesn't specifically talk about SITL, that's why I was asking if you thought it would be necessary to put a note here. In any case if we put a note in here, we should direct to that article.

PR #719, opened 2017-02-18
Added SITL, removed optional steps

dave-dave:
The /etc/hosts edit isn't strictly required.
If running commands with sudo, su to root is not required.
Need to warn users that Windows can and will mess up file metadata with respect to Linux.
Ubuntu prereqs script now pulls in all dependencies.

My apologies for all the changes in one PR, I got carried away.

### DIFF
--- a/dev/source/docs/building-ardupilot-onwindows10.rst
+++ b/dev/source/docs/building-ardupilot-onwindows10.rst
@@ -34,44 +34,11 @@ Since Windows 10 has Ubuntu as his new partner, it is possible to compile Ardupi
 Setup Ardupilot Dev Enviromment for Ubuntu bash on Windows 10
 ===============
 
-#. First, you have to insert your local host to the /etc/hosts file:
-
-   .. code-block:: python
-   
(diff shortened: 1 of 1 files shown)


magicrub:
@dave-dave you've removed many steps here, can anyone else confirm this works?

OlivierJB:
Didn't work for me. I still needed the following in step 9:

_sudo add-apt-repository ppa:terry.guo/gcc-arm-embedded
sudo apt-get update
sudo apt-get install gcc-arm-none-eabi=4.9.3.2015q3-1trusty1

sudo apt-get update
sudo apt-get upgrade_

Also I'd suggest keeping the following (step 1 and 2):
_sudo nano /etc/hosts
127.0.0.1 "hostname"
(where “hostname” is your PC name)_

sudo works without it so agree it's not strictly needed, but cleaner and avoids the "Unable to resolve" nag.

rmackay9:
Looks like OlivierJB had troubles and there was no further follow-up so closing.  We can re-open if we still need to make this change

PR #720, opened 2017-02-20
Update video for Teraranger Tower object avoidance

msadowski:
The previous video was referencing collision avoidance but in fact the setup can be only used for object avoidance at this stage. Hopefully it will make the intended behavior slightly more clear.


msadowski:
I'm closing this pull request and will follow up with fully updated docs on using TeraRanger Tower for object avoidance.

PR #725, opened 2017-02-28
Common: basic information on getting compass orientation correct

peterbarker:


gmorph:
Looks right to me.  Merge I say!

gmorph:
I wonder if we should include what values people should graph in Mission Planner to check this?

PR #746, opened 2017-03-23
More spacing and hyperlink error fix

SwiftGust:
sorry about the small commits. as i see it is on wiki page there was some glitch.
so this is quick fix.


rmackay9:
all fixes are good fixes, thanks!

PR #755, opened 2017-03-30
Updated to the correct, old driver

proficnc:


proficnc:
pull request is redundant, as Michael Fixed this in the mission planner code :)

proficnc:
After consulting with Michael, we would like people using this driver

PR #763, opened 2017-04-08
Instructions for setting up soaring SITL with X-Plane

akolobov:
The goal of this PR is to add a self-contained set of instructions for setting up an X-Plane SITL for playing with the soaring controller recently added to ArduPlane. The instructions for the controller are already available at http://ardupilot.org/plane/docs/soaring.html. 

The purpose of these instructions is to enable as many people as possible with minimal knowledge of and experience with ArduPlane to set up such a simulation, and to minimize the need for the reader to piece together the setup process based on a multitude of different wiki pages. Accordingly, this document regurgitates instructions from other wiki pages whenever necessary, instead of referring the reader to them by only providing a link.

The instructions refer to a few files that are going into ArduPilot with this PR: https://github.com/ArduPilot/ardupilot/pull/6018.


tridge:
thanks!

PR #771, opened 2017-04-12
clarify where vehicle starts in circle mode

larsks:
the documentation on Circle mode wasn't clear about whether the
vehicle position when engaging circle mode was the center of the
circle or the edge of the circle.  This change attempts to make that
less ambiguous.


rmackay9:
Thanks very much!

PR #772, opened 2017-04-12
SITL: Added references to helper scripts for Windows

stephendade:
Added references to an automatic script to download/install the SITL environment, plus a link to example batch files to run SITL from a single double-click of a batch file.


rmackay9:
Thanks!

PR #775, opened 2017-04-18
Update Intel Aero RTF page

lucasdemarchi:
- Clarify the microcontroller model
  - Add version requirement for the Aero side of the software
  - Add instruction for developers to build and update the flight
    controller
  - Make a section for connecting the transmitter


lucasdemarchi:
Changes done without checking if they are rendered correctly. 

- update.sh expects to be in a certain path
- `make html` seems to need files copied from somewhere
- it needs some very specific versions of tools and themes that are not easily available.

How do people test changes locally?

hamishwillee:
>How do people test changes locally?

I use Vagrant, as discussed here: http://ardupilot.org/ardupilot/docs/common-wiki_editing_guide.html#making-a-big-edit

magicrub:
The changes look OK to me. Thanks @lucasdemarchi , merged!

PR #776, opened 2017-04-18

amilcarlucas:
* 64bit cygwin works as well
* python2 is required, python3 does not work yet

### DIFF
--- a/dev/source/docs/sitl-native-on-windows.rst
+++ b/dev/source/docs/sitl-native-on-windows.rst
@@ -95,7 +95,7 @@ that allow us to rebuild ArduPilot on Windows.
    +----------------+----------------------------------------------------------------------------------+
    | libxslt-devel  | Libs \| libxslt-devel: XML template library (development files)                  |
    +----------------+----------------------------------------------------------------------------------+
-   | python-devel   | Python \| python-devel: Ptyhon 2 language interpreter                            |
+   | python-devel   | Python \| python-devel: Python2 language interpreter (python3 does not work yet) |
    +----------------+----------------------------------------------------------------------------------+
    | procps         | System \| procps-ng: System and process monitoring utilities (required for pkill)|
    +----------------+----------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


dev/source/docs/sitl-native-on-windows.rst
```diff
@@ -99,7 +95,7 @@ that allow us to rebuild ArduPilot on Windows.
    +----------------+----------------------------------------------------------------------------------+
    | libxslt-devel  | Libs \| libxslt-devel: XML template library (development files)                  |
    +----------------+----------------------------------------------------------------------------------+
-   | python-devel   | Python \| python-devel: Ptyhon language interpreter                              |
+   | python-devel   | Python \| python-devel: Ptyhon 2 language interpreter                            |
```
hamishwillee: Typo for "Python" (was in previous version too)

hamishwillee:
64 bit is missing one of our dependencies (procps). Any idea what has changed? @rmackay9 ?

OXINARF:
Apparently that package is now available in Cygwin 64-bit. Also, I'm not sure if that package is still needed.

hamishwillee:
Then as long as typo is fixed this could go in. Though would be nice to have confirmation that package is available and whether procps is a requirement.

OXINARF:
Sorry, I should have been more clear. According to Cygwin webpage the package is now available for 64-bit.
Regarding it being a requirement or not, unfortunately the wiki only says it was for pkill, but not what required it - my guess is that it was for *sim_vehicle* script, which has now been converted to Python and tries different methods to kill processes. I'm pretty sure it isn't needed in Cygwin, but it would need to be tested by someone using it.

hamishwillee:
Thanks for clarification. IMO this can be merged then after the typo is fixed. We should check the dependencies/instructions as a separate issue.

amilcarlucas:
I've benn running SITL (rover and copter) on windows for the past 12 months on cygwin64, and it works

rmackay9:
I think I'll merge this in..

PR #788, opened 2017-04-28

smiltee:
Add the line to install uavcan. In my experience it was required to build master 4ce0a8e24eb2d336a682325e332c4a8b772e25d1.


OXINARF:
@rmackay9 This is actually wrong. We have our own uavcan submodule. The correct command is `git submodule update --init --recursive modules/uavcan`

smiltee:
@OXINARF @rmackay9 Should we revert this commit? I don't have anything to test this right now.

PR #797, opened 2017-05-10
All vehicle types document

peterbarker:
This is going to be a *very* long document.


common/source/docs/common-all-vehicle-types.rst
```diff
@@ -0,0 +1,285 @@
+.. _common-all-vehicle-types:
+
+=====================================
+Vehicle Types Supported by Arducopter
```
hamishwillee: Just copter? If this is just copter then probably put inside a block to hide on the other doc sets for a common page.

peterbarker: Gah, my bad.  fixed.

PR #801, opened 2017-05-11
update.py: option to build only selected wikis

peterbarker:
e.g. time ./update.py --cached-parameter-files True --wikis copter

### DIFF
--- a/update.py
+++ b/update.py
@@ -30,7 +30,7 @@ import re
 import os
 from codecs import open
 import subprocess
-
+import sys
 
 DEFAULT_COPY_WIKIS =['copter', 'plane', 'rover']
 ALL_WIKIS =['copter', 'plane', 'rover','antennatracker','dev','planner','planner2','ardupilot']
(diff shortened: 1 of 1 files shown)


peterbarker:
This overlaps with `--site`; one or the other should die.

hamishwillee:
@peterbarker FYI Where we need improvements in the script is the way that common files are handled. 

At the moment, (prior to a build) the script deletes all common files from the individual wiki source trees then copies in the current set. While this guarantees that the individual wikis contain the latest common files, it means that incremental builds do not work. Everything gets rebuilt, even though most things haven't changed.

peterbarker:
On Wed, 10 May 2017, Hamish Willee wrote:

> At the moment, (prior to a build) the script deletes all common files from
> the individual wiki source trees then copies in the current set. While this
> guarantees that the individual wikis contain the latest common files, it
> means that incremental builds do not work. Everything gets rebuilt, even
> though most things haven't changed.

Why does the remove/copy step negate incremental building?  Are we not 
preserving timestamps while copying the files?

hamishwillee:
I don't know, I just know that incremental builds no longer happen.

peterbarker:
https://github.com/ardupilot/ardupilot_wiki/blob/master/update.py#L98

I think that's a likely reason for the lack of incremental builds:-)

I think you're well placed to get into the head of the person who added that line....

Removing the line drops build time from ~1m30s to 30s - I've no idea if the build is correct, however.

hamishwillee:
The guy who made that change was probably an idiot :-) Previously this had clean as a separate step.

I think that I wasn't sure how well the system would handle a non-clean build with "old" artifacts. My original plan was to at least see if a common page had been added/removed. Anyway, this is something that certainly is worth testing.

tomlauzon:
this has been my nightmare... can it be fixed?

hamishwillee:
Anything can be fixed, but I don't have the bandwidth to think about it right now.

peterbarker:
On Fri, 12 May 2017, tomlauzon wrote:

> this has been my nightmare... can it be fixed?

Are you in a position to test patches?

It is possible things will just work with 
https://github.com/peterbarker/ardupilot_wiki/tree/no-clean

... but being a kinda hacky build process we might find that things don't 
get rebuilt appropriately.  The clean was almost certainly put in 
originally because someone got annoyed with things *not* getting rebuilt!

hamishwillee:
> The clean was almost certainly put in originally because someone got annoyed with things *not* getting rebuilt!

Probably not actually. I think I put it in because I was finishing my contract and wanted things to be robust.

PR #813, opened 2017-05-20
Added section of companion computers supported and supported languges to FlytOS docs, added content to sample applications section to FlytOS docs

sharvashish:


rmackay9:
Merged!  I'm really sorry about the delay on that, I'll be much faster next time I promise!
It's a very nice looking wiki page by the way.

As a side note, next time could you rebase on master - that should get rid of the "Merge branch" commits and the duplication in commits that we see above.  Also if you could add a little "flytos:" prefix in the commit lines that'll make it a bit more clear that it's just affecting the flytos pages.  Anyway, not trying to make excuses for why it took me this long to review!

sharvashish:
Thank you for accepting it. Yes I will make sure to rebase it before asking for PR. I must have forgot to do it this time, and also add "flytos:" in the header.

PR #817, opened 2017-05-24
common: add custom-power-tether document

peterbarker:


rmackay9:
thanks!  I may hack on it a bit :-).

PR #818, opened 2017-05-24

ncx94:
corrected command to launch mavproxy with the console, as used for the rest of the tutorial


rmackay9:
Thanks very much for these fixes!

PR #820, opened 2017-05-29
Update Sub video in vehicle types page.

rjehangir:
The original video did not properly represent the "vectored" type ROV.


rmackay9:
hard to believe but this small change actually broke my ability to compile the wiki on my ubuntu machine.  I'm not sure what's wrong with my setup but it seems very sensitive to any changes in white space within the table.

hamishwillee:
@rmackay9 If it breaks your whole build that is a problem. However breaking the page a bit is not unusual - tables are a weakness of all markup system.

I never use that type of table. I will sometimes use [list or csv tables](http://docutils.sourceforge.net/docs/ref/rst/directives.html#id4). I also use html tables quite a bit.

PR #822, opened 2017-05-30
Update license-gplv3.rst

auturgy:
trivia - added sub to the list of ardupilot platforms, and added APM Planner and MAVProxy to gcs types.


rmackay9:
Great, thanks!

PR #825, opened 2017-06-01
frsky telemetry on the 9xr pro

larsks:
this adds documentation on how to use FrSky telemetry on a 9xr Pro
transmitter.


magicrub:
Merged, thanks!

PR #827, opened 2017-06-02
Add document explaining how to force px4io flash

peterbarker:


marcmerlin:
Thanks for that.
Now, back to the original point, isn't 3.8 going to require this on every pixhawk that's upgrading?
If so, shouldn't the firmware reliably use the alternate method?
If anyone is looking at this issue, the doc is not live on the main website yet, but can be found here:
https://github.com/ArduPilot/ardupilot_wiki/blob/master/dev/source/docs/pixhawk-force-px4io-flash.rst

Thanks for writing it, it's a good stopgap in the meantime.

marcmerlin:
Mmmh, I've been trying nsh as described on http:// ardupilot.org/dev/docs/interfacing-with-pixhawk-using-the-nsh.html and neither apmplanner 2.0.24 (downloaded binary) or 2.0.25rc (built from git) seem to have the terminal icon anymore. Maybe @billbonney knows more.
nsh from mavproxy works though.

PR #833, opened 2017-06-02
tradheli: make connection diagram less confusing

larsks:
the use of a futaba receiver in the image meant to demonstrate
connections for non-futaba receivers could cause confusion.  This
commit removes the branding from the picture.


rmackay9:
Thanks for the fix!

PR #839, opened 2017-06-06
dev: amend instructions on forcing io firmware flash

peterbarker:


peterbarker:
[12:25:56 PM] Buzz: seems plausible. merge it!

Just allocating blame ;-)

PR #841, opened 2017-06-06
Common-RC: New Taranis Q X7, FlightDeck and replace featured receiver

tomlauzon:


hamishwillee:
Nice one@tomlauzon - So essentially RC compatibility is with the flight controller (Pixhawk-series) rather than something that is associated with the autopilot (ArduPilot/PX4) ?

This is all good. Only thing I would have suggested is adjusting size of FrSky_Ph2-X8R-FLVSS_adj.jpg (nearly 1Mb) since images get served at their stored size. Best way I've found for this is to start with a high res version (not this one, which already has some distortion) and use gimp to export as jpg.

PR #844, opened 2017-06-10
Create setting-up-sitl-on-linux.rst

muhammedozel:
python-wxgtk2.8 has been changed vith python-wxgtk3.0 and writing python-wxgtk2.8 cause error

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-linux.rst
+++ b/dev/source/docs/setting-up-sitl-on-linux.rst
@@ -76,7 +76,7 @@ this:
 
 ::
 
-    sudo apt-get install python-matplotlib python-serial python-wxgtk2.8 python-wxtools python-lxml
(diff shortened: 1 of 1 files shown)


OXINARF:
These instructions should be updated to tell developers to use our installer script, but maybe until that is done it is indeed better to show a command line for the latest OS versions.

I'll let other wiki maintainers share their thoughts.

magicrub:
Does this require python 3.x?

OXINARF:
No, it's the wxgtk that was updated to version 3.0 and 2.8 isn't available in more recent OS versions.

muhammedozel:
It has to be changed because of beginner user gets an error and they won't understand what is the issue?

OXINARF:
@muhammedozel Yes, but a user with an older OS version will get the error too since 3.0 doesn't exist there. Anyway, I've merged, thanks!

PR #850, opened 2017-06-18
ESC and GNSS doc update

pavel-kirienko:
Actualized the documentation entries pertaining to UAVCAN ESC and Zubax GNSS.


peterbarker:
Merged, thanks!

Note that my benchmark for "available" was "credit card number can be supplied and tracking number given in return", thus the original "No" - when I looked initially nobody had stock.

peterbarker:
@pavel-kirienko It occurred to me I should ask you to verify that the "Works" column is accurate - does ArduPilot master currently run the Orel 20 CAN ESCs?

pavel-kirienko:
Yes it does.

PR #858, opened 2017-06-24
Update common-rc-transmitter-flight-mode-configuration.rst

6102Captain:
Futaba T14SG is widely used in China. Here is a way to set the 6-point switch for Futaba T14SG.


rmackay9:
Thanks Captain!

PR #868, opened 2017-07-05
Add distance sensors usage example and small fixes

khancyr:


common/source/docs/common-rangefinder-mavlink.rst
```diff
@@ -0,0 +1,21 @@
+.. _common-rangefinder-mavlink:
+
+===================
+MAVLink Rangefinder
+===================
+
+The MAVLink rangefinder is a way to expend the FCU capability by sending it rangefinder mesurement with MAVLink DISTANCE_SENSOR message.
```
OXINARF: expend -> expand
mesurement -> measurement

common/source/docs/common-rangefinder-mavlink.rst
```diff
@@ -0,0 +1,21 @@
...
+
+Setup through the mission planner
```
OXINARF: Nitpicking: Setup through Mission Planner

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,72 @@
...
+    - Ardupilot Master
+    - Copter >= 3.5.1
+    - Rover >= 3.2.0
+    - Plane >= 3.8.1
+
+
+Distance sensors message
+========================
+
+Ardupilot support both distance sensors message (`MAVLINK definition <http://mavlink.org/messages/common#DISTANCE_SENSOR>`__) both as input and output.
```
OXINARF: Ardupilot support both distance sensors message -> ArduPilot support distance sensor message

I think that putting the link into *distance sensors message* would look better than having a parenthesis just saying *MAVLink definition*.

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,72 @@
+.. _ros-distance-sensors:
+
+==========================
+ROS distance sensors usage
+==========================
+
+.. warning::
+
+    Distance sensors message are currently support on :
```
OXINARF: The message is named DISTANCE_SENSOR so I think using singular here would be better.

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,72 @@
...
+
+    - Ardupilot Master
+    - Copter >= 3.5.1
+    - Rover >= 3.2.0
+    - Plane >= 3.8.1
```
OXINARF: 3.7.1? 3.8 isn't out yet :slightly_smiling_face:

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,72 @@
...
+To do that, setup a Mavlink rangefinder on Ardupilot side (:ref:`see here <common:common-rangefinder-mavlink>`) and simply set a subscriber in mavros plugin :
+
+.. code-block:: none
+
+    Configuration in apm_config.yaml:
+    distance_sensor:       <== plugin name
+      rangefinder_sub:     <== ROS subscriber name
+        subscriber: true   <== set subscriber type
+        id: 1              <== sensor instance number must match the one set on Ardupilot side
+        orientation: PITCH_270  # only that orientation are supported by APM 3.4+
```
OXINARF: APM 3.4+ -> Copter 3.4+

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,72 @@
...
+==========================
+ROS distance sensors usage
+==========================
+
+.. warning::
+
+    Distance sensors message are currently support on :
+
+    - Ardupilot Master
+    - Copter >= 3.5.1
```
OXINARF: 3.4? Downward rangefinder wouldn't send distance sensor message, but it would send for all the other directions. In any case, 3.5 isn't out yet, let alone 3.5.1 :slightly_smiling_face:

dev/source/docs/ros-distance-sensors.rst
khancyr: agree

dev/source/docs/ros-distance-sensors.rst
khancyr: nope, on copter, the input is accepted from 3.4. Even on 3.5, copter won't send the msg for rangefinder. It is only send with proximity that is another usage (Didn't write it yet)

khancyr:
@OXINARF thanks for review ! I will correct that ! 
About the version :
- Copter >= 3.5.1
- Rover >= 3.2.0
- Plane >= 3.8.1
None are out yet, but I hope that the DISTANCE_SENSOR will make it on those version. I prefer specify it to prevent user to complain that it doesn't work when it is just not supported...

dev/source/docs/ros-distance-sensors.rst
khancyr: yep, I should have catch this one.

dev/source/docs/ros-distance-sensors.rst
khancyr: Better remove this comments and update the yaml on Mavros

dev/source/docs/ros-distance-sensors.rst
OXINARF: You should then put the versions separately in Rangefinder and Proximity sections. And if it isn't supported yet in some vehicle it shouldn't be mentioned, otherwise it's possible those versions will get out without support for it and we'll get people complaining about it.

OXINARF:
@khancyr Maybe it is better to do a bigger description of what each version supports. In any case, I'm pretty sure Plane 3.8 will support the distance sensor message for downward rangefinder - it doesn't for all other orientations.

khancyr:
@OXINARF @rmackay9 I think this is ready for now !

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,79 @@
...
+Receive from FCU
+----------------
+
+.. warning::
+
+    Distance sensor message is currently support on :
+
+    - Ardupilot Master
+    - Copter >= 3.5.1
+    - Rover >= 3.2.0
```
OXINARF: Can we at least add *(Planned)* or something like that?

dev/source/docs/ros-distance-sensors.rst
```diff
@@ -0,0 +1,79 @@
...
+    - Plane >= 3.8.0
+
+By default, Ardupilot will send distance sensors message for each rangefinder connected on FCU.
```
khancyr: True !

dev/source/docs/ros-distance-sensors.rst
khancyr: ok

rmackay9:
I'm find with this going in except for the "Common: add rangefinder mavlink" commit.  I think without a specific example or a sensor that makes use of this protocol, it's confusing for the end user.

khancyr:
yep, I don't have another example than the ROS one. I add it as some people where asking if it was possible !
Maybe it can be move to dev instead?

khancyr:
rebase, remove addition of mavlink rangefinder, correct some type on ROS files

auturgy:
Thanks Pierre!

PR #873, opened 2017-07-08
remove link to obsolete ublox config files

### LINKED ISSUE #871: wrong link
https://github.com/ArduPilot/ardupilot_wiki/blob/master/common/source/docs/common-ublox-gps.rst

 https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_GPS/config  not found 404

larsks:
the ublox configuration files referenced by this document
were removed in ardupilot/ardupilot@d91f89d.

Closes #871.

### DIFF
--- a/common/source/docs/common-ublox-gps.rst
+++ b/common/source/docs/common-ublox-gps.rst
@@ -4,11 +4,7 @@
 UBlox GPS Configuration
 =======================
 
-:ref:`3DR uBlox <common-installing-3dr-ublox-gps-compass-module>`
-modules are shipped with a `custom configuration <https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_GPS/config>`__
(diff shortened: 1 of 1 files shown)


hamishwillee:
Change is great thanks. 

Do we know if there is any advice on configuration anywhere that we could link to ? (ie it was also a flaw in previous version that we provide no guidance on how you might modify the configuration.). Should not block inclusion of this change.

WickedShell:
@hamishwillee I nuked the configs as they aren't maintained and go stale. At this point ArduPilot does a better job configuring users GPS's then those configs do, and there are actually a number of messages in that configuration that we actually don't want turned on.

That said the ability to use u-center does have value, so I'd like to see that part (the mavserial pass/mission planner stuff) maintained.

hamishwillee:
@WickedShell I get that :-) The question was really "if I did need to update this config, where would I go for additional information".

PR #874, opened 2017-07-08
update docs re: auto mode and arming

### LINKED ISSUE #804: Copter: documentation needs info about auto-takeoff
There is no documentation on autotakeoff, (as part of mission or otherwise) - nowhere users can see info about required throttle stick position, or how to do it at all

larsks:
This commit implements suggestions from @hamishwillee in issue #804.
Specifically, it clarifies that a copter must be armed before engaging
auto mode, and it updates the list of armable modes.

This also removes the phrase "...and the propellers will begin to spin
slowly", because the propellers when armed are still spinning fast
enough to cause injury.

I have elected not to include Guided mode in these changes, even
though it is an armable mode.

Closes #804

### DIFF
--- a/copter/source/docs/arming_the_motors.rst
+++ b/copter/source/docs/arming_the_motors.rst
@@ -7,21 +7,29 @@ Arming the motors
 Arming the motors
 =================
 
-Before arming the motors, make sure all people and objects are clear of
-the propellers. Then do the following:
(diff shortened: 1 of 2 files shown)


hamishwillee:
These changes look excellent to me. @rmackay9 merging makes sense!

rmackay9:
Thanks!

Issue #673, opened 2017-01-10
ardupilot_wiki/common/source/docs/common-rc-transmitter-flight-mode-configuration.rst


lvale:
Some of the links that point to diydrones are pointing to wrong articles

larsks:
@lvale can you update this issue with the specific links (and, ideally, with the corrected links as well)? Thanks!

lvale:
https://github.com/ArduPilot/ardupilot_wiki/pull/875

PR #875, opened 2017-07-08
Removed wrong entries

lvale:
Removed wrong entries not pointing to specific articles on dyd
https://github.com/ArduPilot/ardupilot_wiki/issues/673


hamishwillee:
Thanks @lvale

PR #885, opened 2017-07-19
Update common-storm32-gimbal.rst

Kigurai:
Added warning about RF interference from Storm32 boards.

### DIFF
--- a/common/source/docs/common-storm32-gimbal.rst
+++ b/common/source/docs/common-storm32-gimbal.rst
@@ -24,6 +24,9 @@ gimbal has been tested with a `DYS 3-axis brushless gimbal <http://www.hobbyking
    gimbals running
    `v067e <http://www.olliw.eu/storm32bgc-wiki/Downloads>`__ (or higher). 
 
+Be aware that some v1.3x boards has been found to cause significant RF interference on the 433mhz and 915mhz band.
+Use with caution, if you are using either 433/915mhz control or telemetry.
+
 Connecting the gimbal to the Pixhawk
 ====================================
 


OXINARF:
@olliw42 Do you want to comment on this?

lvale:
@OXINARF I can NOT confirm this. I've tested several boards with a RF Explorer and haven't seen any noticeable interference results on the common frequency ranges (433M, 868M, 900M, 1.2G, 2.4G, 5.8G). We have several of these units in industrial environments in production all around the world, so it is important to verify these issues.

olliw42:
hey, thx oxinarf, and thx luis
a comment could be appropriate; some such instance were reported
I however think that the note is somewhat imprecise, and also could point to solutions
the OP should adapt the text to reflect the situation more correctly, IMHO

Kigurai:
It's quite clear from the discussion on RCG that it has been an issue for
several people.
 I added the note as warning (and Ollie even thanked me for it on RCG).

lvale:
My testing does not confirm your statement. I might have misused the testing equipment (RF-Explorer), so please confirm with some measuring equipment.

Kigurai:
Here's Ollie himself confirming it:
https://www.rcgroups.com/forums/showpost.php?p=37930438&postcount=10018

Here's another user testing it on the 433mhz
https://www.rcgroups.com/forums/showpost.php?p=34132760&postcount=579

I have no equpment for testing it against the 915mhz band, but both myself
and another user were unable to reliably connect with the 915mhz telemetry
radios with the gimbal running.

Now, given that safety is always a overriding issue, don't you think it
would be prudent to put up a warning on a product that the developer
himself has repeatedly stated is a hobby project and should not be expected
to perform to professional standards?

Further, if Ollie wants to add solutions, he can add them himself.

lvale:
Had a STorM32 over here and redid a testing (only one unit so not very relevant) and the difference with the RF Explorer antenna right on top of a STorM32 is very small with or without power. Biggest relevant difference is around the 350MHz

img_20170723_030058

lvale:
I just checked and there's a notice on the STorM32 wiki about an issue with RF noise, and also a solution to the problem is pointed, but only on a very specific board variant.
https://www.rcgroups.com/forums/showpost.php?p=34139984&postcount=585

http://www.olliw.eu/storm32bgc-wiki/STorM32_Boards


If you don't mind rephrasing your comment and pointing to the solution it would be beneficial to other users.

goandfix:
Hello all, I have both boards. Someone could confirm if I can use the GLB STorM32 v1.31 without issue into 433mhz band? Thank you

lvale:
Just in case your STorM32 has a RF noise issue follow these recommendations
https://www.rcgroups.com/forums/showpost.php?p=34139984&postcount=585

rmackay9:
Merged, thanks!

PR #889, opened 2017-07-20
--upload works for networked targets

auturgy:
As pointed out by Lucas, --upload for linux boards uses rsync, not USB serial, so is not effected by the WSL limitation.
I've tested it, and it works.


rmackay9:
merged, thanks!

PR #891, opened 2017-07-26
SOLO WIKI: Updates to links

Pedals2Paddles:
Updated links for purchasing the Green Cube and and updated the nav per Randy's suggestion.


rmackay9:
Merged, thanks!

PR #635, opened 2016-12-09
Common: FrSky passthrough telemetry

tomlauzon:
It took a bit longer than expected but there it is. These changes are related to the new release of the passthrough FrSky telemetry protocol. We hope this will help people with installing FrSky telemetry equipment and how to configure ArduPilot and OpenTX.

PR #892, opened 2017-07-26

auturgy:
Advertisement for craft and theory proprietary software doesn't belong.

### DIFF
--- a/common/source/docs/common-telemetry-landingpage.rst
+++ b/common/source/docs/common-telemetry-landingpage.rst
@@ -22,7 +22,6 @@ information based upon your set-up.
     Configuring a Telemetry Radio <common-configuring-a-telemetry-radio-using-mission-planner>
     Bluetooth <common-mission-planner-bluetooth-connectivity>
     FrSky telemetry <common-frsky-telemetry>
-    FlightDeck FrSky Transmitter App <common-frsky-flightdeck>
     XBee <common-telemetry-xbee>
     SiK Radio — Advanced Configuration <common-3dr-radio-advanced-configuration-and-technical-information>
     3DR Radio v1 <common-3dr-radio-v1>


rmackay9:
I'm less sure about this change.  When it comes to external stuff that works well with ardupilot, i'm not sure we discriminate against open or closed.

OXINARF:
I think the PR itself is OK, since the Flightdeck page is still linked from where it should, inside the general FrSky page. I think that what @auturgy means is that leaving it here gives it an unfair advantage over other FrSky solutions that are only linked in the FrSky specific pages.

lucamax:
I agree with auturgy and OXINARF
Craft and Theory  products have a lot of evidence in the Wiki.
In my opinion DIY solutions should always have the priority , it is also a way to grant gratitude to Open Source developers that with their free work make this community strong.
In example the Mavlink Frsky Port telemetry solution  is actually described as "DIY solution for the APM2.x here" , in fact it works perfectly with Pixhawk , readers seeing AMP2.x will never look at that solution.
Craft and Theory products  looks like mandatory , I do not like that and I express the reasons in this post:
https://discuss.ardupilot.org/t/flightdeck-alternative-for-passtrough-telemetry/21784/11

tomlauzon:
@lucamax, the Mavlink Frsky solution is not part of the Ardupilot code, whereas protocols 4 and 10 are. We can't start describing all sorts of things that are not directly connected to Ardupilot. The documentation is already quite large. The converter is an external piece of hardware that is completely independent from Ardupilot and can work other FCs with MAVLink. It's already good that there is a link to it. FlightDeck is made specifically for Ardupilot and protocol 10.

About your comment about the use of the term "APM" that's what these products use. They were the ones who wanted the pages to say that. If they want to change the names of their products, they can. On the flip side, there is no native FrSky telemetry on APM2.X so you have to use a converter.

Obviously, the landing page is a mess. The subjects are very diverse. It could use a FrSky Scripts and Apps page.

The things is we don't want to have to dig through many submenus to find the info.

I think it's fair to give an advantage to functions that are part of Ardupilot such as protocols 4 and 10 vs external solutions (MAVLink converters).

The DIY cable for protocols 4 and 10 is clearly put forward in the X receiver section cables section.

fnoop:
@tomlauzon You put forward the same argument here:
https://github.com/ArduPilot/ardupilot_wiki/pull/635#discussion_r92296498
This is in your interest, as it sells your product.  However you're ignoring several years of community history which uses the 'external solutions' - ie teensy, very successfully.  Whether they purchase a cable from you or a teensy from elsewhere, the end result is the same - telemetry displayed on the receiver.  It works very well, and is a vital opensource safety component that has been used by many people for a long time.  I'm all for change and improvements, but that is still a very valid and active option that should be prominent.  It also has advantages over an 'internal' solution:
 - Potentially more versatile - can use any data available over mavlink.  Two-way communication with the FC.  More like a mini GCS.
 - Development can be (and has been) much faster split into separate component.  Teensy and LUA code has been developed without having to wait for arudpilot PRs and releases
 - Teensy (or other board) has additional processing capacity for analysis.  Can also be used for other purposes, for example the frsky teensy solutions also control LEDs, more complex current sensors etc - you cannot do this with the 'internal' solution.
 - Components widely available - teensy cheap and widely available across the planet (more so than commercial cables or components), easier to put together than the diy cable.
 - Fully opensource.

Not saying that the built-in telemetry support isn't more modern and elegant - it is, particular on newer flight controllers that have a dedicated frsky port - but the existing opensource solutions perform their function and shouldn't be relegated to 3rd rate and invisible in the documentation.

auturgy:
+1

Regards,

James

rmackay9:
I think I'm going to leave the item in the list 'cuz I think it adds value to users.

Issue #867, opened 2017-07-05
Document usage of first (zero-ith) waypoint as home position


akorb:
The first mission item will be used (and overwritten) as a waypoint containing the home position.

This is a common pitfall, see https://groups.google.com/forum/#!topic/mavlink/SL0oTUJx1fE and https://github.com/ArduPilot/ardupilot/issues/1379 .
To avoid this, this should be mentioned in the documentation. Probably here: http://ardupilot.org/copter/docs/common-planning-a-mission-with-waypoints-and-events.html (Planning a Mission with Waypoints and Events).

This might lead to huge problems. Imagine your first mission item is a takeoff specifying the safe height. You start and instead of rising vertically, it will fly straight to the waypoint given afterwards and there might be an obstacle.

magicrub:
This was fixed via https://github.com/ArduPilot/MissionPlanner/commit/97bdafb3c570fa93bf5aa189b81343a08b7798d9

PR #893, opened 2017-07-26
Update common-planning-a-mission-with-waypoints-and-events.rst

auturgy:
Add zeroth waypoint / home position warning.
Link to https://github.com/ArduPilot/ardupilot/issues/1379 and https://github.com/ArduPilot/ardupilot_wiki/issues/867

### DIFF
--- a/common/source/docs/common-planning-a-mission-with-waypoints-and-events.rst
+++ b/common/source/docs/common-planning-a-mission-with-waypoints-and-events.rst
@@ -18,16 +18,22 @@ This page describes generic waypoint setup for all types of vehicles.
 Setting the Home Position
 =========================
 
-For **Copter** the home position is set as the location where the copter
-was armed. This means if you execute an RTL in Copter, it will return to
-the location where it was armed, so arm your copter in the location you
-want it to return to.
+.. warning::
 
-For **Plane** the home position is the location of the plane where the
-GPS was locked. This means if you execute an RTL in Plane, it will
-return to the location where it first acquired the GPS lock, so only
-power up your plane in the location you intend it to return to when
-performing an RTL.
+   When an auto mission is executed, the first waypoint in the mission will
+   be interpretted as the home position, overwriting any other home position
+   that is set. This means that users **MUST** set the first waypoint at 
+   their desired home position to ensure safe Take-Off and RTL behaviour.
+
+For **Copter** except as above, the home position is set as the location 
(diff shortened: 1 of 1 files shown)


WickedShell:
I don't think this is correct. While the behavior specified is not wrong, that is a GCS handling item. Warning users of it will only create confusion and problems. What GCS handled this incorrectly for you?

WickedShell:
To clarify more, I think this deserves documentation in a developer part of the wiki, or the MAVLink docs, but it doesn't belong in the end user facing documentation.

auturgy:
I'm just clearing out issues: but missionplanner definitely does this, and as I understand things it is the intended behavior for gcs software to do so (set home to waypoint zero).
I don't see it as "incorrect", just something to be aware of.
The referenced issues provide more detail, but enough people have made this mistake that I think the note is warranted. It's up to you: but if you reject the pr, you should also close the referenced issues.

Regards,

James

common/source/docs/common-planning-a-mission-with-waypoints-and-events.rst
```diff
@@ -18,16 +18,22 @@ This page describes generic waypoint setup for all types of vehicles.
...
 
-For **Plane** the home position is the location of the plane where the
-GPS was locked. This means if you execute an RTL in Plane, it will
-return to the location where it first acquired the GPS lock, so only
-power up your plane in the location you intend it to return to when
-performing an RTL.
+   When an auto mission is executed, the first waypoint in the mission will
+   be interpretted as the home position, overwriting any other home position
+   that is set. This means that users **MUST** set the first waypoint at 
+   their desired home position to ensure safe Take-Off and RTL behaviour.
```
OXINARF: I'm pretty sure this is incorrect. The first waypoint is indeed home, but it isn't used for anything. So the user doesn't have to set it to anything and it certainly doesn't overwrite the set home position.
If we want to put a warning, it should be a dev warning, saying that in the MAVLink protocol waypoint number 0 is reserved to be the home location for legacy purposes (but it shouldn't be used to retrieve home anymore).

magicrub:
This was a bug in Mission Planner that was fixed via https://github.com/ArduPilot/MissionPlanner/commit/97bdafb3c570fa93bf5aa189b81343a08b7798d9
closing...

auturgy:
Thanks Tom

Regards,

James

PR #897, opened 2017-07-26
Autotest Framework - update names of ArduPlane

stephendade:
Correct the test names from "Plane" to "ArduPlane"


magicrub:
Good catch @stephendade , thank you!

PR #902, opened 2017-08-05
Vehicle parameter override

lvale:
Ever since the QuadPlane there are parameters that are common and aren't overwritten, To ensure this procedure works, it's better to use Rover


rmackay9:
Actually I don't think it depends upon whether there are parameters in common or not.  For example, we've always had AHRS parameters in common but there hasn't been any issue with those being overwritten.  I'm going to merge it anyway but I'm 99% sure that it's not actually an issue.

PR #904, opened 2017-08-07
Fixed link in docu for powershell setup script

al3xst:
Filename changed in the repo, but wasn't fixed in the docu, where the link was invalidated.


rmackay9:
Thanks!

PR #913, opened 2017-08-10
Update license-gplv3.rst

auturgy:
perhaps nit-picky, but the "GNU Operating System", which for want of a better description is "pure" Linux, actually involves a range of different free software licenses.  So it is not strictly "the same license".  As this deals with licensing I think it best to be correct, that's all.


rmackay9:
nice, thanks!

PR #925, opened 2017-08-27
Pixhawk 2.1 URL update

auturgy:
www.pixhawk2.com is just a redirect to proficnc, so I  made it proficnc just in case the redirect dies or is repurposed at some stage


magicrub:
@proficnc ???? Your call.

proficnc:
Either is fine, happy with the redirect to ProfiCNC

PR #928, opened 2017-08-31
Update common-stores.rst

auturgy:
Split partners from non-partners and added some partners who were missing.
A few clarification words to go with that change.


auturgy:
overtaken by events.

PR #935, opened 2017-09-11
STorM32 wiki corrections

olliw42:
- I've actually never tested myself, but from the code it seems clear that the serial protocol is activated with MNT_TYPE=5
- it's STorM32 and not SToRM32 ;)

### DIFF
--- a/common/source/docs/common-storm32-gimbal.rst
+++ b/common/source/docs/common-storm32-gimbal.rst
@@ -1,20 +1,20 @@
 .. _common-storm32-gimbal:
 
 =========================
-SToRM32 Gimbal Controller
+STorM32 Gimbal Controller
 =========================
 
-The *SToRM32-BGC* is a relatively low-cost 3-axis brushless gimbal
+The *STorM32-BGC* is a relatively low-cost 3-axis brushless gimbal
 controller that can communicate with ArduPilot (Copter, Plane and Rover)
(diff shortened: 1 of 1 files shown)


OXINARF:
Merged, thanks!

olliw42:
many thx

but this was what I was a bit afraid of ... the preview didn't helped me sufficiently to get the formatting correct, so, the intended three bullets also ended up in the docs being screwed up
maybe see also https://discuss.ardupilot.org/t/how-to-contribute-to-the-ardupilot-documentation/21025
I understand it's all my fault, but I honestly didn't know better (I actually spend some time on getting it right, so it's not that this happened just because of me being lazy, I tried to follow the other bullets but couldn't determine why "my" bullets show up differently)

I probably should have added a warning "modification by a rst dummy, needs revision". Maybe you could advice rst dummies to put up a certain label. :)

sorry for stealing time for such a small thing

(there is a reason I shied away from PRs LOL)

OXINARF:
That was my bad. I actually looked at the GitHub preview and saw it, but dismissed it as the GitHub viewer being wrong (which occasionally happens). I've committed a fix, I think I got it right this time - the issue was that it is list and you had line breaks which broke the list formatting.

olliw42:
great
THX

lvale:
Thanks @olliw42

Issue #930, opened 2017-09-02
Building ArduPilot on Windows with waf and Bash


nikhilgupta280695:
Step 9. **sudo add-apt-repository ppa:terry.guo/gcc-arm-embedded** 
has been deprecated and a new version is available [here](https://launchpad.net/gcc-arm-embedded/+announcement/13824) 
Kindly update the step as this repository doesn't get added and causes a lot of errors.

OXINARF:
New repository doesn't have GCC 4.9. That instruction is useful for whoever still has Ubuntu Trusty in Windows Linux subsystem. Unfortunately that subsystem can't run 32-bit binaries, so the binary we normally use isn't possible to be used.

For Ubuntu Xenial, you can install the gcc-arm-none-eabi present in Ubuntu repos (although not the official ARM release, it should be close enough), but when Microsoft updates Ubuntu version again, you'll lose that ability again because next Ubuntu versions don't have GCC 4.9 anymore. You can also download the package for Willy (http://ppa.launchpad.net/terry.guo/gcc-arm-embedded/ubuntu/pool/main/g/gcc-arm-none-eabi/gcc-arm-none-eabi_4.9.3.2015q3-1wily1_amd64.deb) and try to install it on Xenial.

In any case, I agree the instructions need to be updated, just not to the new PPA.

jeremyolsen:
I get errors using either of the previous suggestions in this post.  Are there any other options or is using WSL a dead end?

auturgy:
G’day Jeremy,
I apologise in advance: it’ll be at least a fortnight until I can look into this, as I intentionally left my laptop at home to travel without it (both frustrating and liberating).
Apart from uploading via USB, there should be no restrictions using WSL to compile ArduPilot. Can you list down the steps you’ve taken attempting to compile, and the errors you get?
It’s best to put the details on discuss.ardupilot.org and just add a link in here so I can find it. That way others in the community who may not see things in GitHub can assist.

Regards,

James

jeremyolsen:
James,

Cheers!  Thanks for the quick response.  I'll comment here and re-post over at discuss.

It's not critical that the WSL compile work for me.  I'm only utilizing WSL until my TX2 arrives and I can have a dedicated Ubuntu environment.  I have successfully compiled APM via Windows for my Navio2 board in the past.  I'm currently working on a Pixhawk project and am trying to setup a SITL/Gazebo environment.

[...]

PR #936, opened 2017-09-13
wiki: dependancies - Update building-ardupilot-onwindows10.rst

auturgy:
Fixes https://github.com/ArduPilot/ardupilot/issues/6945
Was stepped through with the author of the issue, but otherwise untested: someone should check this in a clean Windows or VM.


OXINARF:
Related to #916 and #930. As I wrote in the latter one, keeping the old instruction is worth it for someone with an old WSL installation still running Trusty.

auturgy:
Short memory. Sorry.
Will rewrite to just make some notes. Will take a few days

EShamaev:
4.9.3.2015q3-1trusty1 is not available on latest WSL.
Also the "mount via the drvfs driver" does not help.

sudo apt-get install gcc-arm-none-eabi is working to build cloned in WSL copy, but still did not find a way to build the copy that is on Windows drive.

auturgy:
Closing, as overtaken by rate of change.

PR #938, opened 2017-09-13
Provide a warning about back-feeding power to the Pixhawk servo rail …

R-Ronald:
…via an RSSI monitoring cable


magicrub:
This looks ok to me. @proficnc can you double confirm?

PR #940, opened 2017-09-19
Added store links

tropnevad:
Added some store links for people to buy from EU distributors


rmackay9:
I think recently we are reducing the use of links from our product pages and/or limiting them to point to stores from ArduPilot Partners at least in cases where an ArduPilot Partners sells the product.  It's open for discussion but my feeling is that it's fair that ArduPilot promotes purchasing from companies that directly support ArduPilot.  Of course I suspect unmanned tech supports ArduPilot indirectly - they've been around a long time and are already on our stores page.

Sorry, I know this is a touchy subject and it's easy for my comments to be interpreted negatively but at the same time, ArduPilot does a lot of good and becoming a Partner is really not expensive and it's hugely appreciated by the dev team.

I leave it at that for now and let others comment.

auturgy:
Noting the other efforts to consolidate consumer links, I'm going to close this.
It does raise a valid point about providing links via region, which I think we should take on board.
Thanks @tropnevad for taking the time to submit this PR though.

PR #946, opened 2017-09-25
TradHeli:update helicopter tuning method for Copter 3.4 and newer

ChristopherOlson:
This is based on information provided by @lthall on tuning procedure. And a technical re-write of the tuning wiki done by @bnsgeyer to reflect the tuning method recommended by Leonard, and an update for the parameters and procedure that is applicable to Copter 3.4 and later.


auturgy:
This is an important update, LGTM.
@tridge if Rob and Leonard are too busy, could Greg test this?

ChristopherOlson:
I think just about everyone is using this PR anyway, as it has been posted to the forum several times in the last few months in short descriptive format:

https://discuss.ardupilot.org/t/550-trad-heli-pitching-could-do-with-some-advice/20006/3

The old wiki is from the APM days and does not properly explain how to tune the commanded rate for a helicopter, and pretty much only concentrates on rate PID's. Tuning the rate control is the most important part, as it is what compensates for different mechanical linkage rates and variations in servo speed. So the new proposed wiki is mostly (if not 100%) being used by people asking questions on how to tune a heli now. If somebody asks on the support forum we just link to the new one, as the old wiki is sadly outdated.

auturgy:
Merging, because even if it needs a couple of tweaks (I'm unsure, as I can't test), it's better than what we have.  Thanks Chris.

PR #951, opened 2017-09-26
Update making-a-mavlink-wifi-bridge-using-the-raspberry-pi.rst

Rusty105:
I  found when using a RasPi 3 you do not need the driver line in hostapd.conf and you must use the standard hostapd files, not the ones from Adafruit, to make the  access point work


rmackay9:
sorry for the delay, thanks!

PR #960, opened 2017-10-04
Edited title image for webapps

lvale:
Edited title image for webapps


OXINARF:
The previous image should be replaced, not insert a new one. It would also be nice to put everything in the same PR.

PR #962, opened 2017-10-06
Update Repurposed FrSky telemetry page

tomlauzon:
Includes Scripts and conflict warning.

### DIFF
--- a/common/source/docs/common-frsky-repurposed.rst
+++ b/common/source/docs/common-frsky-repurposed.rst
@@ -3,6 +3,8 @@
 FrSky Telemetry Using Repurposed Messages
 =========================================
 
+The original solution for transmitting Ardupilot telemetry data over the FrSky telemetry link was to reuse FrSky data messages meant for another purpose and/or reserved by FrSky for their sensors. For example in this protocol the Flight mode is sent using the tmp1 message instead of the temperature. If you are not using a temperature sensor on your vehicle, this will not cause any problem. If you intend to use other FrSky sensors, this telemetry protocol will cause conflicts with them.
+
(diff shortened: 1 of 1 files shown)


tomlauzon:
No, but there is a link to rc groups: https://www.rcgroups.com/forums/showpost.php?p=37213943&postcount=48

OXINARF:
It's the same person, so it's preferable to link to our forum instead.

The only thing I don't like about this PR is the images coming from an external GitHub repo, it opens up the possibility of inappropriate images showing up in our wiki.

tomlauzon:
@auturgy the executable in question is a text lua script, so it is source code as well. Also linking to something is not distributing it. But that's beside the point. The question was whether we should host the screenshots on the wiki to prevent the pictures from being replaced at the source and showing them on the wiki. I can do it when I have a second. It takes a bit more work, and it's a good idea to let the author know that we intend to host the screenshots if that's what we want to do.

auturgy:
Fair point re Lua (being interpretted etc).  My bad.  "linking to something" vs "distributing" is a nuanced discussion, but accepting the point that for lua the executable is the source, it isn't really relevant for here.
More importantly, the download link on the ardupilot forum didn't work, so we either need to host the LuaPlane script ourselves, link to rcg, or put it somewhere else.  It's also a few merges behind LuaPilot, so may have some minor issues with OpenTX >2.1 (I can't test, as my Taranis is 6000 miles away).  Can someone confirm which OpenTX versions the linked LuaPlane (fully) works with?

Hwurzburg:
Tom, I am trying to clean up the wiki PRs, here....if you could drop the images, but leave the LUA script links (checked as valid by me), I would recommend merging this....thanks

PR #966, opened 2017-10-09
Fix typo of AC_WPNav URL

DawoodTaii:
Fix typo of AC_WPNav URL;
Old URL was linking to the wrong library.


DawoodTaii:
I noticed also a typo on library name. So fixed both in another pull request.

PR #969, opened 2017-10-10
AP password change for edison

hdtechk:
installed new image to edison could not access wifi do to password changed.


peterbarker:
Merged, thanks!

PR #972, opened 2017-10-12
Update using-python-scripts-in-mission-planner.rst

eicossa:
Script.getParam() does not exist
It is Script.GetParam()


rmackay9:
@meee1 didn't review but I suspect the change is a good one.

PR #977, opened 2017-10-15
Update common-stores.rst

auturgy:
removed 3DR and bracketed comment from Craft&Theory.

### DIFF
--- a/common/source/docs/common-stores.rst
+++ b/common/source/docs/common-stores.rst
@@ -11,7 +11,7 @@ recommended for use with ArduPilot (in alphabetical order).  You may also want t
 * `AUAV <http://www.auav.co/>`__
 * `Bask Aerospace <http://www.baskaerospace.com.au/>`__
 * `BlueRobotics <https://www.bluerobotics.com/store/>`__
-* `Craft & Theory <http://www.craftandtheoryllc.com/store/>`__ (FrSky telemetry, MAUCH power modules)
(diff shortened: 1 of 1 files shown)


auturgy:
I fully agree with supporting partners, but short descriptions like yours only work for a subset.
We need to treat all partners equally. An alternate approach might be to split the list into labelled groups. It’s an open discussion topic, hopefully we’ll get some clarity and direction in the very near future.

CraigElder:
@auturgy Being listed on this page is not exclusive to partners and having detail about what Craft & Theory sell does not give them an unfair advantage but removing the detail deprives our users of useful detail.

rmackay9:
@CraigElder, actually we made a decision recently that we would only add partners to the stores page going forward.  We're grandfathering in existing stores (i.e. non-partners can stay if they're already on the page).

Perhaps that decision will be viewed unfavourably by some but I hope not.  The thinking is that stores are for-profit entities and we hope the decision will encourage some more companies profit based on ardupilot will in turn support ardupilot a bit financially (the donation to become a partner is very reasonable).

CraigElder:
@rmackay9 Sorry I don't remember that vote

OXINARF:
@rmackay9 I'm in the same situation as @CraigElder, I don't know where that decision was made. I know this page was discussed in a dev call I wasn't in, but from reading the minutes and asking a member that was there, I didn't get any feedback that rule was approved.

PR #987, opened 2017-10-20
Tested "MAV_CMD_DO_SET_ROI"

Daddy-Boomalati:
Tested using DO_SET_ROI with a fixed camera. Traditional heli keeps nose pointed at ROI. Deleted the "probably" from the wiki.


rmackay9:
txs!

PR #993, opened 2017-10-23
Update ATT message details for Copter

Javiercerna:
The ATT message details were outdated. I changed the names of the parameters, fixed some typos and added ErrRP and ErrYaw.


magicrub:
I'm wondering if this is worth having two charts: old vs new

auturgy:
Fair point, but I'd rather not: it's been a while (3.3?) since the 
changes, and I think it's better for users to adopt newer firmwares.

Suggest merge.

OXINARF:
@auturgy Have you checked the changes are correct? I agree with you in just keeping the current explanation, not only the old one is too old, but it would make the wiki more confusing.

rmackay9:
It looks correct to me so merging, thanks!

PR #1002, opened 2017-11-11
Add one required dependency

lvale:
Add one required dependency to build on RPI3 on Stretch


auturgy:
Not opposed to this, but might it be better to simplify this page and just link out to https://docs.emlid.com/navio2/common/ardupilot/building-from-sources/ ?
Emlid have great documentation, and are better positioned to keep it up to date.

lvale:
@auturgy the link to Emlid docs is already on the page. This simplified page is ok, excepto for the omission I added on this PR

Issue #999, opened 2017-11-05
Plane: Note that downgrading from 3.8 to a previous version will lose any RC changes that were done


WickedShell:
http://ardupilot.org/plane/docs/plane-3-7-to-3-8-migration.html should note that if you downgrade from Plane 3.8 to an earlier version of plane then any changes that were made to the RC_* parameters will be lost in the earlier version, and that upgrading back to 3.8 will not copy over any param changes that happened in the earlier version. This is only an issue if the user decides to downgrade. See https://github.com/ArduPilot/ardupilot/issues/6522

marcmerlin:
further improved the warning, PR sent.

PR #1003, opened 2017-11-12
Outline that downgrading is very likely to cause a crash.

marcmerlin:
Existing text does not make it clear that downgrading really requires very careful review of RC settings and that not doing so is very likely to cause a crash due to reversed servos (this did destroy my aircraft).
https://github.com/ArduPilot/ardupilot_wiki/issues/999


auturgy:
sorry this took so long

PR #1005, opened 2017-11-14
CAN bus and UAVCAN guide

EShamaev:
As more and more users start with UAVCAN, there is a need for some documentation in wiki.


hamishwillee:
Perhaps link to http://uavcan.org/ ?

PR #1009, opened 2017-11-18
Steering derating setup

EShamaev:
Based on https://github.com/ArduPilot/ardupilot/pull/5401
With a picture that @gmorph liked so much!


gmorph:
It is a good picture.  Looks fine to me.

PR #1021, opened 2017-11-27
Add SET_ATTITUDE_TARGET documentation to copter-commands-in-guided-mo…

rcywongaa:
Specifically, some intricacies for attitude field and thrust field I felt would be useful for anyone using it in the future


rmackay9:
oops, sorry for the delay, thanks!

PR #1022, opened 2017-11-29
Package name updated to current version

focanag:
wxgtk2.8 is not in Ubuntu 16.04 repositories. Tested 
https://github.com/ArduPilot/MAVProxy/issues/462


rmackay9:
Took us a while but merged! thanks!

PR #1025, opened 2017-11-30
Spektrum pinout

larsks:
There are two commits in this PR.

The first adds information about the Spektrum/DSM port on the pixhawk, based on https://discuss.ardupilot.org/t/add-spektrum-port-pinout-diagram-to-pixhawk-pinout-page/23533/3.

The second adds third-level headings to this document, which at least when rendered on github result in named anchors that can be used to link directly to the pinout for a particular port.

### DIFF
--- a/common/source/docs/common-pixhawk-overview.rst
+++ b/common/source/docs/common-pixhawk-overview.rst
@@ -552,6 +552,37 @@ pinout is standard serial pinout, to connect to a standard FTDI cable
    </tbody>
    </table>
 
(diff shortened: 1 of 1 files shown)


larsks:
@rmackay9 I just realized this was still open...do you want it?

larsks:
@rmackay9 @peterbarker ?

auturgy:
Looks fine. Cable colours are always tricky as there is no uniform standard, but this complies to a common convention.  User review at the linked forum post. Thanks!

PR #1037, opened 2017-12-08
adding the command "mkdir opt" under the root path

pbjacob:
the pre-install script have "OPT=/opt" while WSL doesn't have it.


rmackay9:
txs

Issue #279, opened 2016-02-02
Create SBUS Servo Output Info


R-Lefebvre:
Ardupilot/Pixhawk can be used with Futaba SBUS servos easily.  Need to create an info page.

rmackay9:
I think all that's required is that the user sets BRD_SBUS_OUT to 1.
@R-Lefebvre, anything else required?

hamishwillee:
We'd probably want to add a section to this article: http://planner.ardupilot.com/wiki/common-servo/#connecting_the_servo_to_a_pixhawk

I'm guessing you just plug the servo to the SBUS out pin (as [shown in this diagram](http://planner.ardupilot.com/wiki/common-pixhawk-overview/#pixhawk_pwm_connectors_for_servos_and_escs_and_ppm-sum_in_and_sbus_out)) and then set BRD_SBUS_OUT to 1?

Would also (I guess) be nice to list/link the servos that are known to work.

R-Lefebvre:
Yep, that is pretty much it.  Set BRD_SBUS_OUT to 1.  

You also need to set the SBUS address in each servo, but I don't think we need to explain that, other than to say it's required.  They would follow the manufacturer instructions for that.

The only servos I'm aware that would work, would be Futaba SBUS servos. I don't think anybody else is making SBUS compatible servos?  Not that I'm aware of anyway.

auturgy:
FrSky also make SBus servos, but Tridge's  recent addition, using SBus Out for additional PWM via a downstream converter, is probably worthy of a mention in the wiki too.

ivan95603:
Dear @R-Lefebvre,
Is it possible to use sbus servos without sbus receiver just over telemetry and PC joystick from mission planner?

Thanks

auturgy:
Yes, possible. Just use the sbus output on the Pixhawk. You need an sbus programmer to make sure your servos are numbered correctly.
Preferred to have rc though, if only as a backup.

Regards,

James

hamishwillee:
@auturgy @R-Lefebvre Can you guys make the changes so that this can be closed? If you can't/really don't want to let me know, and I'll take a shot at it.

auturgy:
I'm moving house, but can do it in a week or so.

Regards,

James

hamishwillee:
Thanks James, that would be awesome!

chandra2sekhar2000:
Hello,
[...]

PR #1039, opened 2017-12-10
common: add SBus serial servo output docs

kd0aij:


auturgy:
Ref: https://github.com/ArduPilot/ardupilot_wiki/issues/279

kd0aij:
@auturgy I didn't realize that the PX4IO SBus output was not documented. I've never tested the PX4IO SBusOut pin myself; does it actually work?

auturgy:
It does, although I’m not sure how many use it.
I think merging what you’ve done first is fine, and a good reminder for me to finish off documentation for the other method.
As with your implementation, you can’t set refresh rate by channel/group, and it’s not bidirectional.

kd0aij:
It would be nice to coordinate the servo output docs to make it clear that the PX4IO is a separate module which is only present in the "Pixhawk" flight controllers. There's also some potential confusion between "MAIN" and "AUX" outputs since those terms are used to differentiate between the IO and FMU PWM outputs of Pixhawk1/2 and some FCs have only the FMU, hence no "MAIN" outputs.

kd0aij:
@lvale Here's the first bit of SBusOut documentation. Comments and suggestions appreciated.

lvale:
@kd0aij I would only add on the title the reference to Serial, as to not confuse with the "traditional" SBus output of the full sized PixHawks :)

PR #1040, opened 2017-12-11
merge the two commands into one

pbjacob:
merge  "cd /   mkdir opt"
into "sudo mkdir /opt"

reason 1: "sudo" shall be added under the root path.
reason 2:  The user shall stay in the current ardupilot path, otherwise the command in the next step will fail.

### DIFF
--- a/dev/source/docs/building-ardupilot-onwindows10.rst
+++ b/dev/source/docs/building-ardupilot-onwindows10.rst
@@ -88,12 +88,11 @@ Setup Ardupilot Dev Enviromment for Ubuntu bash on Windows 10
    
        git clone https://github.com/ArduPilot/ardupilot.git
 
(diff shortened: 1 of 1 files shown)


auturgy:
Thanks for doing this!
You shouldn’t need to do this though. Did you follow the previous steps exactly?

Regards,

James

magicrub:
This is an update to his previous PR https://github.com/ArduPilot/ardupilot_wiki/pull/1037/files so merging

auturgy:
Step 3 in the guide gives the user root privileges. Should be no need for sudo.

Regards,

James

pbjacob:
> Step 3 in the guide gives the user root privileges. Should be no need for sudo.

Thank you for this! 
Yes, sudo can be removed.

PR #1043, opened 2017-12-14
Dev: Change unit notation of Hertz to Hz

muramura:
The Hertz notation is Hz.
I changed hz to Hz according to the notation.


OXINARF:
Merged, thanks!

PR #1046, opened 2017-12-18
Update common-sbus-out.rst

kd0aij:
specify Serial Port in title


auturgy:
I need to do the ‘normal’ sbus-out page still, but no need to hold this up.

PR #1057, opened 2018-01-07
Fix path of autotest script

Javiercerna:
Fix the path of autotest.py, since the _scripts_ folder doesn't contain the autotest module


magicrub:
This looks good, just wondering why it hasn't been a problem?

khancyr:
LGTM

PR #1066, opened 2018-01-23
Added note for UTF-8 encode.

hiro2233:


WickedShell:
Why? I don't see why this is relevant at all for the source files. They should all be ASCII to the best of my knowledge, and I've used a couple of systems that did not support UTF-8 without any issues...

magicrub:
yeah, you're right. It actually should be set for ASCII.

magicrub:
reverted

PR #1075, opened 2018-01-31
Fix typo in Copter-Commands-in-Guided-Mode

rcywongaa:
Introduced in 15782b0fd6c9cdee830f825caf7ca7301b6e630b

Appeared in http://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html

Apologies for the carelessness..


rmackay9:
np, thanks for the fix.

PR #1082, opened 2018-02-07
Update isc-dhcp-server file contents to edit

studroid:
Update isc-dhcp-server file contents to edit from "interfaces" to "interfacesV4"


rmackay9:
I'm going to guess this is correct although i don't really know for sure.

PR #1100, opened 2018-02-28
Copter: Add Smart_RTL Documentation

lordneeko:


peterbarker:
Fixed what James found, squashed and merged!

Thanks!

rmackay9:
Not quite right, the link to the new page is incorrect from the landing page.  It's currently <smart-rtl> but it needs to be <smartrtl-mode>

rmackay9:
I'd like to add it's a well written page..

lordneeko:
I can't take any credit, I just pulled off pages and discussion threads which i found through the discuss server.

So I noticed that on the landing page right away. took me a while to figure out what i'd done wrong.  Added new PR to fix.

PR #1109, opened 2018-03-05
Up-to-date on HighSierra 10.13.3

lvale:
Added the link to the ardupilot tap
removed reference to catkin_pkg

@rmackay9 perhaps also editing the doc title from MacOS X just to the correct name MacOS


rmackay9:
Thanks!

PR #1111, opened 2018-03-05
solo_arducopter_upgrade: Fixed GPS Enabled value

lordneeko:


rmackay9:
Thanks!  that must have been a copy-paste error originally.

PR #1112, opened 2018-03-06
TradHeli Wiki Setup Update

ChristopherOlson:
This updates the setup wiki for TradHeli to be more modern and up-to-date.


rmackay9:
very nice.  it needs a rebase on master but i'll take care of that.

ChristopherOlson:
I had a problem with my fork I was working on not being rebased and didn't catch it before I did the PR. Only commit # 106e46d519cb0438e36f7534e96117340e706a93 was supposed to be PR'd.

rmackay9:
OK, just merged the last commit, thanks!

ChristopherOlson:
Thanks Randy. All those old commits were squashed into one for the tuning wiki. Which is also in the log. I do not know how they got into my master branch without being squashed, so GitHub included them in the PR. Sorry about that :-)

copter/source/docs/traditional-helicopter-connecting-apm.rst
```diff
@@ -1,204 +1,98 @@
...
++--------------+-------------+
+| 5            | Flight Modes|
++--------------+-------------+
+| 6            | Tuning      |
++--------------+-------------+
+| 7            | Aux         |
++--------------+-------------+
+| 8            | Throttle    |
++--------------+-------------+
+
```
bnsgeyer: @ChristopherOlson   Great job on all of this!!!  One thing I think can be misleading is this table.  I think a little more explanation needs to be given on the fact that RC inputs are separated from Servo outputs as far as channel number is concerned.   I think you can say that the software expects from the RC receiver for Aileron, Elevator, Collective, Rudder and flight modes on these channels.  The other inputs channels like throttle can be moved but the default is on channel 8.  (not sure about tuning)   Then you can go into the defaults for servo channels out and explain that they are specified by the ServoX_function parameter.  I think this would help the uninitiated to understand the software and how it works.  Let me know what you think.

ChristopherOlson:
That would probably be a good idea. It can certainly be modified to be better and other points skipped over can be added as we go. The old wiki required an almost 100% replacement in this case since it had been neglected for many years and was carried over from the old DIY Drones site. Now that we have a more modern base to work from, improving it, adding to it, and keeping up-to-date will be MUCH easier.

PR #1114, opened 2018-03-06
Update new mavlink message description

samuelctabor:
Add explanation of how to extend mavproxy to make use of a custom message.


magicrub:
@tridge  or @peterbarker can you please sanity the MAVProxy module here?

peterbarker:
@magicrub looks fine as a minimal example.

Pointing people at MAVProxy's "example" module might be worthwhile, however.  Otherwise we'll be fielding more questions about "how to do X in a module"...

Also, I'd recommend people work out of a MAVProxy checkout rather than dropping and modifying modules into /usr/lib.  This means they're always using version control, for example, and will aid them when they need to seek help ("can you push your module up somewhere I can look at it?").

That being said, this is better than what we had before.

rmackay9:
It's a step forward so merged!

PR #1118, opened 2018-03-09
Home Page Quick Format Change

Naterater:
Attempting to fix the 2nd column - firmware that isn't displaying properly.


rmackay9:
It probably needs a space in there after the single quote..

PR #1127, opened 2018-03-14
Update Emlid Edge docs

kobylyanskiy:
- add flight video
- add links to the docs and shop


magicrub:
I've merged, but I'm thinking it will look better if you keep the image in there instead of replacing the image with the video.

PR #1143, opened 2018-03-27
Common: add information about sensor-rate logging

peterbarker:


rmackay9:
Merged, thanks.  As mentioned on the other PR I just merged.  When parameter names are in the docs, I'd prefer if we added links to the parameter description.  so instead of two single quotes use :ref:'param-name \<param-name\>'

PR #1144, opened 2018-03-29
Add Flip-mode to wiki

lordneeko:


rmackay9:
Merged because it's a good start, thanks!

a couple of things we should add:

- when parameters are mentioned we should make them link to the parameter description by adding :ref:'param-name <param-name>' in front.
- add a video at the top (there's one from the skyrocket throw mode page)

PR #1150, opened 2018-03-30
Update common-pixracer-overview.rst

ardubird:


rmackay9:
For this one we definitely need to add mRobotics because they are the inventor of it actually (the original company merged with mRobotics about a year ago).
https://store.mrobotics.io/category-s/112.htm

auturgy:
Suggest not only including mRo, but checking with Jordi and PhilK before merging.

lvale:
Last I heard those vendors didn't have reliable PixRacer boards.

lthall:
Yeh, to the best of my knowledge the pixracer boards that were sold in hobbyking were of an early firmware revision and will not run using master.

Putting BangGood on as a source is never a good idea as they sell a lot fake components.

Close it I say!!

auturgy:
ping @j07rdi

j07rdi:
Also pinging @pkocmoud so we can discuss it. Thanks for the consideration.

pkocmoud:
Thanks - Looks fine to me.

rmackay9:
I'm pretty sure the hobbyking and banggood links at the end won't work because they're missing the final double underline (__)

auturgy:
thanks @ardubird - merging.

PR #1188, opened 2018-04-11
Ph2removal

ThisIsTheOnlyUsernameAvailable:
Updating all remaining wiki entries with references with PH2.1 or links to deprecated pixhawk2-overview page.


auturgy:
Thanks!
Will probably be the weekend before I can go through this.

Regards,

James

auturgy:
Hex website still calls it PH2 fwiw

auturgy:
I fixed up a few conflicts.  I'd prefer we start to just generalise more though.  Differentiating between APM and Pixhawk was important, but less so between the different 32 bit boards - and we'll just end up having to do this type of clean up more and more.

PR #1202, opened 2018-04-16
Plane: add reference to airspeed calibration on starting-up-and-calibrating page

rmackay9:
Added a small new section to the starting-up-and-calibrating page to tell people to calibrate and test the airspeed sensor (if present)


auturgy:
SDP33 sensor claims not to need calibration.
It may not be in the wiki yet though.

Naterater:
I've heard conflicting reports about letting the airspeed sensor "warm up" to avoid drift. It may be worth mentioning.

PR #1206, opened 2018-04-17
Plane: Update & Simplify Elevon Setup

Naterater:
Emphasized correct RC input reversal prior to starting this. Use stabilization to configure rather than transmitter. Adds case-by-case instructions when to reverse or change function.

In response to https://github.com/ArduPilot/ardupilot_wiki/issues/1204


Naterater:
If approved, this should also be copied over to V-Tail Setup as well. http://ardupilot.org/plane/docs/guide-vtail-plane.html

Naterater:
Bump @magicrub

magicrub:
Sorry, missed this last week while I was traveling. Looks, good!

PR #1210, opened 2018-04-20
Revising requirements

lvale:
Revising requirements for TradHeli

### DIFF
--- a/copter/source/docs/traditional-heli-parts-list.rst
+++ b/copter/source/docs/traditional-heli-parts-list.rst
@@ -7,11 +7,9 @@ Traditional Helicopter – Suggested Parts List
 Electronics including flight controller
 =======================================
 
--  :ref:`The Cube <common-thecube-overview>`
+-  :ref:`An ArduPilot supported board <common-autopilots>`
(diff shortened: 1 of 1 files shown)


rmackay9:
sure you want to remove the mention of extra long cables?  seems like a reasonable thing to recommend.

auturgy:
Extra long i2c cables just invite problems. Probably good not to recommend them. The gps serial cable is of less concern.
In general, for long cable runs we should be recommending CANBus.

rmackay9:
@ChristopherOlson you're happy with this change?

lvale:
I'll be using the shortest cables I've ever used on MR on a TradHeli. Also if we want to mention long cables we should discuss good practices for such (like using CAT5, etc)

ChristopherOlson:
I would agree with it. Not every heli requires a long cable, especially if the FBL unit tray is on the rear of the heli, where the vast majority now put them. And a person can't buy a proper cable extension (that I could find). Any extension to the I2C wiring must be twisted pair or it will pick up common mode interference, which is going to be a user-made item if a heli requires it. So LGTM.

rmackay9:
txs, Merged!

PR #1219, opened 2018-04-27
Steps to load Sonix board firmware from SD card

ssatoru:
Adding steps to load Sonix board firmware from SD card, as it was requires at certain occasions.  Discussed here https://discuss.ardupilot.org/t/lost-of-wifi-connection-to-sv-and-wont-arm/28117/19


rmackay9:
fantastic, thanks!

PR #1230, opened 2018-05-02
Update building-setup-windows10.rst

EosBandi:
This tipp was deleted during the rename.


rmackay9:
Is this accurate?  I have been building with WSL and it is fine.. i'm just not sure we need this warning.. not that I will hit every problem out there but.. if people follow the instructions as written on the page I think it works..

PR #1237, opened 2018-05-04
Fix for dev style, and rewrite sitl setting on linux

khancyr:
Change sitl setting to use install script and switch to arducopter by default instead of plane (that need jsbsim)


rmackay9:
merged, thanks!

PR #1240, opened 2018-05-05
Fix dev and common warning

khancyr:
Fix last warning in dev
put file non referenced in exclude patern
add missing image in common
fix some common warning


auturgy:
LGTM.  Thanks Pierre.

khancyr:
yep. I am just unsure if the exclude patern is the good thing. Normally it should be as those files aren't in any doctree

auturgy:
Yeah.  Those pages are actually linked from http://ardupilot.org/dev/docs/building-the-code.html under "Deprecated instructions".  Adding to the exclude pattern would mean they were no longer built.  I'm not against that, but we should probably discuss it. I'll add a dev call tag.

hamishwillee:
I tend to think that the files should either be removed completely or tagged with an even clearer individual deprecation message. Leaving them in the source is potentially confusing.

khancyr:
I drop the commit with the exclude partern for now to keep only th style fix !

hamishwillee:
@auturgy Then perhaps an even bigger warning up the top of all of these pages. APM2.X is NOT SUPPORTED. These instructions may not work due to deprecated/unavailable toolchain. Use at your own risk.

PR #1242, opened 2018-05-09
updated partner list

lvale:


rmackay9:
in general we should zap the logo as well...

PR #1246, opened 2018-05-12
update source code sample of Copter.h

tajisoft:
I think that wiki information is old. Please ignore otherwise.


rmackay9:
Merged, thanks.  I think maybe there is one extra "#endif" in there.. but .. merged anyway!

PR #1247, opened 2018-05-13
Update common-openpilot-revo-mini.rst

ocularmagic:
Changed "mini-revo" to correctly named "revo-mini"
Changed "dfu-utils" to "dfu-util"
Added "sudo" to dfu-util command.


auturgy:
Awesome, thanks very much.

auturgy:
This was a good pickup by a user. We could generalise the waf steps a bit, for plane/rover etc, but the instructions as they were would lead to compile fail, so merging now.

PR #1253, opened 2018-05-17
Dev - updated build instructions for Cygwin on Windows

stephendade:
Recently went through building SITL and Pixhawk/ChibiOS in Cygwin, so updated the instructions.

Also depreciated old (using make instead of waf) instructions for Windows/Cygwin.

### DIFF
--- a/dev/source/docs/building-setup-windows-cygwin.rst
+++ b/dev/source/docs/building-setup-windows-cygwin.rst
@@ -5,11 +5,9 @@
 Setting up the waf Build Environment on Windows using Cygwin
 ============================================================
 
-These setup instructions describe how to setup `Cygwin <http://www.cygwin.com/>`__ on Windows so that waf can run natively and be used for building.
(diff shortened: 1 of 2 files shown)


rmackay9:
Txs.  This generally looks ok but:

- make on Windows still works and is still the most common way that people build on windows so we should leave this pages available.  We still need to show people how to compile with waf using eclipse before we can remove these pages
- I'm a little surprised that zip in particular is required.  .. are procps-ng and ddd really required?

stephendade:
>make on Windows still works and is still the most common way that people build on windows so we should leave this pages available. We still need to show people how to compile with waf using eclipse before we can remove these pages

I've amended my commit to keep these. Also added a clarification that there's a 3 methods (Eclipse, Cygwin, WSL) to build Ardupilot on Windows.

>I'm a little surprised that zip in particular is required. .. are procps-ng and ddd really required?

They are if you're building an ARM-based target as per https://github.com/ArduPilot/ardupilot/pull/8335

PR #1254, opened 2018-05-17
Copter: Change effective range of RC

muramura:
I THINK...
RC minimum value includes 1300.
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Arming/AP_Arming.cpp#L678

RC maximum value includes 1700.
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Arming/AP_Arming.cpp#L684


auturgy:
A good pickup and useful clarification.  Thankyou!

PR #1256, opened 2018-05-23
Add detailed step-by-step

hunt0r:
Per this forum post, the step-by-step process might be useful for new users: https://discuss.ardupilot.org/t/downloading-and-reviewing-logs-from-dropbox-or-google/29194/7?u=hunt0r


auturgy:
We need to be careful not to be too MissionPlanner centric.  MAVProxy has a pretty good documentation set, but QGC and APMPlanner are also useful log analysis tools.  Any chance of being able to include some pointers for them as well?

PR #1257, opened 2018-05-24
Remove Q_THR_MID doc as the param doesn't exist.

expntly:


rmackay9:
This parameter will have become Q_M_THST_HOVER I think.  It looks to me like plane doesn't call motors->update_throttle_hover() function so I think users do need to set this parameter (@tridge would know for sure) so perhaps the comment should just be updated to use the new parameter name.

PR #1259, opened 2018-05-28

geofrancis:


rmackay9:
looks like the same change appeared twice so closing.

geofrancis:
it on both tf lidar pages, i wasnt sure if i sould have put them both together or separate.

rmackay9:
OK.  we're not too strict on the format for wiki PRs.  the same change in one commit or two is fine I think.  The extra link is still on the tf02 page.. I might remove it next time I need to update the page unless you or someone else beats me to it.

PR #1268, opened 2018-06-03
Update land-mode.rst

blaab2:
Added warning, that the documentation may be outdated (Parameter: LAND_SPEED_HIGH is not yet mentioned)


rmackay9:
Yup, I think this sounds more like something that should go into the issues list so we remember to address it rather than putting it on the wiki so I think I'll close this.  txs for pitching in in any case.

PR #1272, opened 2018-06-06
Update DShot and Telemetry ESC documentation

Naterater:
Added correct parameters for plane users to use telemetry ESC's.
Consolidated telemetry section to not repeat topics.

May want additional notes on OneShot and DShot600 from @WickedShell.

### DIFF
--- a/common/source/docs/common-dshot.rst
+++ b/common/source/docs/common-dshot.rst
@@ -99,7 +99,7 @@ supports four speeds:
 
 - DShot150 at 150kbaud (recommended)
 - DShot300 at 300kbaud
-- DShot600 at 600kbaud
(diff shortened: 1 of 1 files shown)


WickedShell:
Haven't reviewed the changes. The main comment on this page from me I think is that the BLHeli_S ESC's that are advertised as doing DShot 600 only seem to do DShot 600, setting DShot 150 leads to all sorts of weird/wrong behaviour (motors constantly twitching/starting/stopping regardless of throttle level). So if you have a BLHeli_S ESC and want to do DShot, then try DShot 600.

Naterater:
@tridge may want to give some input or approve before merge.

PR #1277, opened 2018-06-09
Update  partners list

OlivierJB:
Update partners list

PR #1278, opened 2018-06-10
Update partners list

OlivierJB:
Update Partners list. Replaces  [#1277](https://github.com/ArduPilot/ardupilot_wiki/pull/1277)


rmackay9:
merged, thanks! Next time we cleanup the list we should also remove the logos.

PR #1279, opened 2018-06-11
RTL not to be used for unstable recovery

Naterater:
Added intent of RTL and warning about using it to recover from bad attitudes.
Fixed some grammar by removing possession of plane.
Separated warning information from note information.

### DIFF
--- a/plane/source/docs/rtl-mode.rst
+++ b/plane/source/docs/rtl-mode.rst
@@ -4,28 +4,33 @@
 RTL Mode (Return To Launch)
 ===========================
 
-In RTL mode the plane will return to its home location (the point where the
-plane armed - assuming it had GPS lock) and loiter there until given
-alternate intructions (or it runs out of fuel!). As with :ref:`AUTO <auto-mode>` mode
+Return to Launch mode is intended to bring a plane back to the launch location during a 
+failsafe event or an operator's command. The plane will return to the launch location 
+(the point where the plane armed - assuming it had GPS lock) and loiter there 
+until given alternate intructions (or it runs out of fuel!). As with :ref:`AUTO <auto-mode>` mode
 you can also "nudge" the aircraft manually in this mode using stick
 mixing. The target altitude for RTL mode is set using the
 :ref:`ALT_HOLD_RTL <ALT_HOLD_RTL>` parameter in centimeters.
 
-Alternatively, you may :ref:`configure the plane to return to a Rally Point <common-rally-points>`, rather than the home location.
+Alternatively, you may :ref:`configure the plane to return to a Rally Point <common-rally-points>`, 
+rather than the launch location.
 
 .. warning::
 
-   "Home" position is always supposed to be your Plane's actual
-   GPS takeoff location:
-
-   #. It is very important to acquire GPS lock before arming in order for
-      RTL, Loiter, Auto or any GPS dependent mode to work properly.
-   #. For Plane the home position is initially established at the time the
-      plane acquires its GPS lock. It is then continuously updated as long as
-      the autopilot is disarmed.
-
-      - This means if you execute an RTL in Plane, it will return to the
-	location where it was when it was armed - assuming it had
-	acquired GPS lock.
-      - Consider the use of :ref:`Rally Points <common-rally-points>` to
-	avoid returning directly to your arming point on RTL
+   RTL is not designed to recover aircraft in unusual attitudes. If 
+   an aircraft has become unstable in an autopilot-assisted mode 
+   (FBWA, Auto, etc.), RTL will rarely have desired effects. 
+   Switch to MANUAL mode to level the plane, then engage RTL.
+   
+   "Launch" position should always be the Plane's actual GPS takeoff 
+   location; it is very important to acquire GPS lock before arming in 
+   order for RTL, Loiter, Auto or any GPS dependent mode to work properly.
+   
+   
+.. note::   
+   The home position is initially established at the time a plane 
+   acquires its GPS lock. It is then continuously updated as long as
+   the autopilot is disarmed.
+   
+   Consider the use of :ref:`Rally Points <common-rally-points>` to 
+   avoid returning directly to your arming point on RTL


plane/source/docs/rtl-mode.rst
```diff
@@ -4,28 +4,33 @@
...
+Return to Launch mode is intended to bring a plane back to the launch location during a 
+failsafe event or an operator's command. The plane will return to the launch location 
+(the point where the plane armed - assuming it had GPS lock) and loiter there 
+until given alternate intructions (or it runs out of fuel!). As with :ref:`AUTO <auto-mode>` mode
 you can also "nudge" the aircraft manually in this mode using stick
 mixing. The target altitude for RTL mode is set using the
 :ref:`ALT_HOLD_RTL <ALT_HOLD_RTL>` parameter in centimeters.
 
-Alternatively, you may :ref:`configure the plane to return to a Rally Point <common-rally-points>`, rather than the home location.
+Alternatively, you may :ref:`configure the plane to return to a Rally Point <common-rally-points>`,
```
peterbarker: This isn't the only alternative; it may also jump to a `DO_LAND_START`

plane/source/docs/rtl-mode.rst
```diff
@@ -4,28 +4,33 @@
 RTL Mode (Return To Launch)
 ===========================
 
-In RTL mode the plane will return to its home location (the point where the
-plane armed - assuming it had GPS lock) and loiter there until given
-alternate intructions (or it runs out of fuel!). As with :ref:`AUTO <auto-mode>` mode
+Return to Launch mode is intended to bring a plane back to the launch location during a 
+failsafe event or an operator's command. The plane will return to the launch location
```
peterbarker: Sorry, but this moves it from a factually correct (if not marked up correctly) description to something not actual correct.

Typically where you *arm* is where home is set.  If you arm in the pits, taxi out to the runway then takeoff - and don't set a rally point, reset your home or have `DO_LAND_START` set, you may really annoy people when your vehicle is circling above the pits.  The vehicle will not return to the "launch point" - the runway - but to the "home location".

The words "home location" should really be marked up - it's a real thing in the ArduPilot code.  "launch location" is not.

plane/source/docs/rtl-mode.rst
```diff
@@ -4,28 +4,33 @@
...
-      the autopilot is disarmed.
-
-      - This means if you execute an RTL in Plane, it will return to the
-	location where it was when it was armed - assuming it had
-	acquired GPS lock.
-      - Consider the use of :ref:`Rally Points <common-rally-points>` to
-	avoid returning directly to your arming point on RTL
+   RTL is not designed to recover aircraft in unusual attitudes. If 
+   an aircraft has become unstable in an autopilot-assisted mode 
+   (FBWA, Auto, etc.), RTL will rarely have desired effects.
```
peterbarker: `FBWA` could very well be a better option that MANUAL.  If the navigation controller is demanding things from the attitude controller that it struggles with, `FBWA` might be viable.

plane/source/docs/rtl-mode.rst
```diff
@@ -4,28 +4,33 @@
...
-	location where it was when it was armed - assuming it had
-	acquired GPS lock.
-      - Consider the use of :ref:`Rally Points <common-rally-points>` to
-	avoid returning directly to your arming point on RTL
+   RTL is not designed to recover aircraft in unusual attitudes. If 
+   an aircraft has become unstable in an autopilot-assisted mode 
+   (FBWA, Auto, etc.), RTL will rarely have desired effects. 
+   Switch to MANUAL mode to level the plane, then engage RTL.
+   
+   "Launch" position should always be the Plane's actual GPS takeoff
```
peterbarker: Again, "launch position" isn't an ArduPilot concept.  Home location is.  Note that most GCSs allow you to specify and lock your home location.

plane/source/docs/rtl-mode.rst
```diff
@@ -4,28 +4,33 @@
...
+   Switch to MANUAL mode to level the plane, then engage RTL.
+   
+   "Launch" position should always be the Plane's actual GPS takeoff 
+   location; it is very important to acquire GPS lock before arming in 
+   order for RTL, Loiter, Auto or any GPS dependent mode to work properly.
+   
+   
+.. note::   
+   The home position is initially established at the time a plane 
+   acquires its GPS lock. It is then continuously updated as long as
```
peterbarker: Not strictly true; it can be set-and-locked by a GCS.

Hwurzburg:
Peter,since this has so many errors, should it be closed? is there something worth keeping? thanks

plane/source/docs/rtl-mode.rst
Naterater: The only reason I use "Launch" was because it's called RTL - NOT RTH. Maybe that is what needs changing.

plane/source/docs/rtl-mode.rst
Naterater: It could be, but often FBWA uses built-up i-gains that will result in bad behavior anyway. Manual is the most reliable for out-of-control planes.

plane/source/docs/rtl-mode.rst
Naterater: Then RTL needs to change to RTH.

Naterater:
I think this is important for users who continually try to use RTL to recover unstable aircraft.

plane/source/docs/rtl-mode.rst
Hwurzburg: I suggest both be mentioned with a cautionary note re FBWA use

Hwurzburg:
Can you please make the changes suggested by Peter? thanks

Hwurzburg:
RTL is defined as returning to home in the flight modes docs....why dont we get the other changes made and have RTL vs RTH mode discussion as a different topic....
    On Saturday, June 1, 2019, 8:06:58 PM CDT, Nathan E <notifications@github.com> wrote:  
 
 
@Naterater commented on this pull request.

In plane/source/docs/rtl-mode.rst:
> -      RTL, Loiter, Auto or any GPS dependent mode to work properly.
-   #. For Plane the home position is initially established at the time the
-      plane acquires its GPS lock. It is then continuously updated as long as
-      the autopilot is disarmed.
-
-      - This means if you execute an RTL in Plane, it will return to the
-	location where it was when it was armed - assuming it had
-	acquired GPS lock.
-      - Consider the use of :ref:`Rally Points <common-rally-points>` to
-	avoid returning directly to your arming point on RTL
+   RTL is not designed to recover aircraft in unusual attitudes. If 
+   an aircraft has become unstable in an autopilot-assisted mode 
+   (FBWA, Auto, etc.), RTL will rarely have desired effects. 
+   Switch to MANUAL mode to level the plane, then engage RTL.
+   
+   "Launch" position should always be the Plane's actual GPS takeoff 

Then RTL needs to change to RTH.

Hwurzburg:
and RTL mode is return to home location by flight modes defs...maybe not the be name for the mode, but it is defined correctly...
    On Saturday, June 1, 2019, 8:04:28 PM CDT, Nathan E <notifications@github.com> wrote:  
 
 
@Naterater commented on this pull request.

In plane/source/docs/rtl-mode.rst:
> @@ -4,28 +4,33 @@
 RTL Mode (Return To Launch)
 ===========================
 
-In RTL mode the plane will return to its home location (the point where the
-plane armed - assuming it had GPS lock) and loiter there until given
-alternate intructions (or it runs out of fuel!). As with :ref:`AUTO <auto-mode>` mode
+Return to Launch mode is intended to bring a plane back to the launch location during a 
+failsafe event or an operator's command. The plane will return to the launch location 

The only reason I use "Launch" was because it's called RTL - NOT RTH. Maybe that is what needs changing.

tridge:
I don't think this advice about unusual attitudes is appropriate for the RTL page. I'm guessing you're referring to stall, and really that should be in general advice in flying, discussing stall and what can be done

PR #1280, opened 2018-06-11
V-tail improvements

Naterater:
Sorry for all the PR's - I keep finding little things to correct on the V-Tail page.

Fixed FBWA yaw command to be a roll command - the V-Tail would only work with a yaw damper.
More grammar fixes and consistency.

### DIFF
--- a/plane/source/docs/guide-vtail-plane.rst
+++ b/plane/source/docs/guide-vtail-plane.rst
@@ -1,16 +1,16 @@
 .. _guide-vtail-plane:
 
 ============
-VTail Planes
+V-Tail Planes
 ============
 
-A common alternative to a traditional elevator and rudder is a VTail, or 
-an ATail (an upside down VTail).
+A common alternative to a traditional elevator and rudder is a V-Tail, or 
+an ATail (an upside down V-Tail).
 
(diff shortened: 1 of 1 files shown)


Kija:
I agree with you that when you roll left, your rudders also turn to the left ; but this is because of arduplane mixing.
You might confuse people with that.
I think you should leave it to yaw left = rudders move left.

Naterater:
@kija, you're right, the mixing is done because of KFF_RDDRMIX, and enabled by default.

The edit that I'm making is because of the procedure - FBWA won't do anything if you yaw the plane to the left or right (remember we're talking about moving the plane now, not commanding a move on the transmitter). The only way you would get rudder movement from physically yawing the plane would be if you had  YAW2SRV_DAMP setup, and that is not set by default.

I don't want users to use the transmitter commands to setup their planes; that's how things get double-reversed and a lot of crashes happen. The correct method that I have proposed is to use roll to determine the rudder movement direction. The transmitter confirmation comes later.

Kija:
Sorry I didn't follow that this was meant for while moving the plane, and not the rc controls. Makes total sense to me now.

PR #1282, opened 2018-06-11
Flight Mode Overview Table

Naterater:
Added a table with a quick summary of modes.


rmackay9:
Hard to see it before merging so let's add it and see how it looks.

PR #1299, opened 2018-06-30
Update common-rcoutput-mapping.rst

johncbahr:
Fixed typo for Servo 3 function line


johncbahr:
I believe this change is correct

PR #1300, opened 2018-07-02
Update common-openpilot-revo-mini.rst 

vierfuffzig:
some additions regarding ChibiOs port, firmware flash procedures, hardware options and cross-vehicle references.


auturgy:
Some of the images are probably a little larger than optimal, but this is definitely an improvement, so merging.

auturgy:
Thanks!

vierfuffzig:
thanks for merging! will pay attention to image size subsequently.

PR #1303, opened 2018-07-03
Create common-osd-overview.rst

vierfuffzig:
Adds a basic documentation for AP_OSD library, focussed on use with FCs featuring onboard MAX7456-type chipsets.


rmackay9:
I guess this should be added into the Optional Hardware section?

vierfuffzig:
thanks for checking this!
i was unsure where to link this page, i'd say optional hardware too.
added image and fixed top line.

khancyr:
Nice !
Do you think you can add SITL version too for  people that want to test with material ?

vierfuffzig:
thanks khancyr! 
thorough documentation of SITL implementations is actually somewhat beyond my scope right now, i'd have to work myself through this first. could we maybe leave that for future improvements?

vierfuffzig:
added SITL references

rmackay9:
Someone in the Copter-3.6 thread is asking about the OSD so seems like a good time to merge it.  I think we should probably create a new section for video and OSD stuff under the Optional Hardware section but we can do that as a follow-up change.

PR #1314, opened 2018-07-08
Update dji-f330-flamewheel.rst

huttonsoftware:
http://dl.djicdn.com/downloads/flamewheel/en/F450_User_Manual_v2.2_en.pdf has motors 2 and 3 reversed, and when I wired it that way it caused instability in the pitch axis.

motors_quadx


rmackay9:
Merged, thanks.

PR #1318, opened 2018-07-11
Update common-osd-overview.rst

Hwurzburg:
Got some more time this evenings. Added font and warnings param descriptions. This should make it current...I think.


rmackay9:
Merged, thanks!

PR #1321, opened 2018-07-12
Name P4 to Pixhawk where hardware is meant

hamishwillee:
Tried to fix this up
- Where PX4 means Pixhawk, this has been changed
- Where PX4 means PX4FMU/PX4IO has been changed.
- Minor tidy of other bits.

There are still some problem cases - ie misuses in parameter docs.

### DIFF
--- a/common/source/docs/common-glossary.rst
+++ b/common/source/docs/common-glossary.rst
@@ -229,7 +229,7 @@ camera pointed towards.
 The square-wave pulse train used to transmit multiple channels of
 information between some RC transmitters and receivers. Some RC
 receivers provide a PPM output (sometimes referred to as the PPM Sum)
-which can be used on the APM and PX4. Other receivers convert the PPM
+which can be used on Pixhawk-series boards. Other receivers convert the PPM
 signal and only provide PWM signals for each channel.
 
 **PWM**: Pulse Width Modulation. The square-wave signals used in RC
@@ -237,7 +237,7 @@ control to drive servos and speed controllers. There is one PWM signal
 for each channel. The width varies from 1000 to 2000 microseconds,
 depending on the RC manufacturer.
 
-**PX4 (PX4FMU and PX4IO)**: Flight Controller system providing
+**Pixhawk (PX4FMU and PX4IO)**: Flight Controller system providing
 capabilities for stabilized flight, position maintenance and automated
 mission (waypoint) path following.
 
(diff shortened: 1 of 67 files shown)


common/source/docs/common-glossary.rst
```diff
@@ -229,15 +229,15 @@ camera pointed towards.
...
+which can be used on Pixhawk-series boards. Other receivers convert the PPM
 signal and only provide PWM signals for each channel.
 
 **PWM**: Pulse Width Modulation. The square-wave signals used in RC
 control to drive servos and speed controllers. There is one PWM signal
 for each channel. The width varies from 1000 to 2000 microseconds,
 depending on the RC manufacturer.
 
-**PX4 (PX4FMU and PX4IO)**: Flight Controller system providing
+**Pixhawk (PX4FMU and PX4IO)**: Flight Controller system providing
```
rmackay9: This is actually incorrect because it's referring to the software.  I think we should change this to be just "PX4" and say something like, "a separate flight code stack with a name that is often confused with the Pixhawk hardware".

common/source/docs/common-here-plus-gps.rst
```diff
@@ -116,9 +116,10 @@ When base/rover is already connected to U-center, click View, go to Message View
 	
 As shown in the figure, the current firmware version is FWVER = HPG 1.30 REF, indicating that the current firmware version is 1.30 for base module. 
 
-Basic opperating manual
-=======================
-This part of the tutorial uses Mission Planner ground control software and Arducopter-3.5 flight control firmware for operating instructions. If you are using PX4 firmware and QGroundControl ground station software, please refer to `this link <https://docs.px4.io/en/advanced_features/rtk-gps.html>`__.
+Basic operating manual
+======================
+This part of the tutorial uses Mission Planner ground control software and Arducopter-3.5 flight control firmware for operating instructions. 
+If you are using PX4 firmware and QGroundControl ground station software, please refer to `this link <https://docs.px4.io/en/advanced_features/rtk-gps.html>`__.
```
rmackay9: we should just remove this link to the other team completely.

rmackay9:
Thanks very much.  I found a couple of small things but merging 'cuz don't think those small items are worth holding up the overall improvement, thanks!

common/source/docs/common-glossary.rst
hamishwillee: Not my intention. I mean to identify the hardware. I'll do something like "Early hardware revision of the Pixhawk flight controller board"

PR #1340, opened 2018-07-26
Numerous fixes to docs 

hamishwillee:
- Docs missing from TOC
- Docs in TOC or linked but not in tree
- Etc


hamishwillee:
Hi @khancyr 

This is a big set of minor bug fixes covering a huge number of docs. I have tested the build, and I know it works. I have submitted a PR to make it visible and for easy postmortem by others, but I wouldn't hold back merging because it might be hard to review and would be a pain to later rebase.
@rmackay9 Is OK with this process, because he trusts that I (generally :-0 ) know what I am doing.

For substantive content changes, I would of course need review, and I will certainly do so.

Sorry, I probably should have spelled out my reasoning before self-merging. And if this process doesn't work for the team, Randy can let me know.

khancyr:
I got no problem with fast merge but you change mavlink. org to .io when we aren't agree on that.
It is better to let a small review window like half a day or a day so somebody can spot little mistakes

rmackay9:
Personally I'm OK with @hamishwillee continuing to merge directly and we can review afterwards.  I'm just worried that it will slow him down and the vast majority of these changes are all good.

auturgy:
Me too.

hamishwillee:
Appreciate the goodwill @rmackay9 @auturgy, but I'm OK with what @khancyr is suggesting. If it proves onerous we can revisit.

PR #1349, opened 2018-07-30
Flesh out the using-autotest page with examples

peterbarker:
.... and explanation of more of the options available.

There are a few sections I'll either trim out (temporarily) or flesh out a little before pushing this in.

Suggestions for additional content to this document welcome, while I'm documenting....


peterbarker:
> Why left two spaces after a dot ?

Not sure what you mean here.

peterbarker:
I've force-pushed up fixes to this which I intend to merge.  It's not as extensive as I'd hoped, but it's better (I think).

PR #1190, opened 2018-04-11
Wiki refactor skyrocket

lordneeko:
I left the information in the development section and some of the common sections alone, since in general many of them will apply to the upcoming Skyviper models as well (APWeb, software loading, firmware resets, etc)  This is a first attempt at refactoring.  You may or may not like the format.  The intential was to build a Skyrocket main landing page, with subpages (expandable via the toctree) for skyviper models, starting with a subpage for the v2450 and have all v2450 pages under that branch.

I don't have the ability to build the site....I've tried and failed...so I haven't been able to test this.

PR #1352, opened 2018-07-31
SkyRocket Wiki: Update to introduce 2018 models

lordneeko:
Ultimately, I'd like to see https://github.com/ArduPilot/ardupilot_wiki/pull/1190 put in, but this is a good placeholder, and should help search engine crawlers to be guided to this location for information once they start buying these.


rmackay9:
thanks for this. One thing I'd like us to avoid is too many subjective adjectives like, "excellent".  I'd prefer we just stay more objective.  so instead of, "The Fury and Scout utilize ArduPilot's excellent OpticalFlow stability.." let's say, "The Fury and Scout include an optical flow sensor.."

auturgy:
I suspect we were looking at this concurrently, but I removed the “excellent...” before you merged.

Regards,

James

PR #1358, opened 2018-08-05
Update building-setup-windows10.rst

marcel-h:
Update according to https://discuss.ardupilot.org/t/build-fails-with-wsl-on-windows-10/31555/2
I would like to change the screenshots too, but couldn't accomplish that with the online editor


rmackay9:
Thanks for this!

PR #1371, opened 2018-08-13
make linux instruction more common

khancyr:
I am lazy, so code once and copy everywhere else ! This should prevent contradiction on different pages.


rmackay9:
Merged, thanks!

PR #1372, opened 2018-08-16
Complete documentation of Omnibus F4 V3 pro

kikislater:
- Add picture to remove diode
- Add note on PPM / SBUS PADS
- Add wiring schema
- Add zadig for windows user otherwise in dfu mode, board could not be detected


kikislater:
Don't know how to add images ! Attaching here 

omnibusf4pro_ardupilot_wiring
omnibusf4pro

rmackay9:
Thanks.  I've merged this, I think we've got some duplication on this page when we should instead link to this page: http://ardupilot.org/copter/docs/common-loading-firmware-onto-chibios-only-boards.html.

Still, because these pages are new, I merged because I figure it's probably an improvement.

kikislater:
Hum ok ! I have to check more often PR next time ... I notice that this PR make it better : https://github.com/ArduPilot/ardupilot_wiki/pull/1330

Hwurzburg:
several errors....I2C does not have TX or RX....clock and data line only....receiver input is not limited to SBUS...ALL supported AP receiver protocols are allowed....PPM, DSM,etc.

the need to mention pullups on I2C is critical to having most external compasses work... all contained in prior #1330

PR #1376, opened 2018-08-16
Updated Links

tropnevad:
Added unmannedtech as they sell alot of other ardupilot related stuff too and are a useful resource for supplies (and a ardupilot partner)


rmackay9:
Merged, thanks!

PR #1392, opened 2018-09-04
Fix links copter

hamishwillee:


hamishwillee:
Hi Randy

For advanced-multicopter-design.rst most of the drones and motors are no longer in production and the ones that are have costs that no longer match current costs. The reason I have created this PR is not really to fix all the broken links but to discuss how the doc might be repaired.

There are two options:
1. Someone who knows a lot about currently available drones to go through this and fill in alternatives.
2. Cut it back hugely, and make it more theorectical. Ie remove ALL the images, 
3. Delete it altogether. 

Any thoughts? It is so out of date my temptation is to go with 3.

Cheers
H

rmackay9:
@hamishwillee, deleting these three seems like the best idea to me but I've ping'd some others to see if they object or want to take over the maintaining these pages.

- http://ardupilot.org/copter/docs/build-your-own-multicopter.html
- http://ardupilot.org/copter/docs/improving-the-capabilities-of-your-multicopter.html
- http://ardupilot.org/copter/docs/advanced-multicopter-design.html

hamishwillee:
Thanks @rmackay9 - that would make sense. I'm happy to handle that deletion, so just drop me a line once approved.

PR #1394, opened 2018-09-05
Issue #7242 Added docker build instructions

acwatkins:


rmackay9:
I think we should set this up as a separate page maybe called "Setup the Build Environment using Docker" and link it from the "[Building The Code](http://ardupilot.org/dev/docs/building-the-code.html)" page.

Hwurzburg:
the edit in place feature is broken, at least I cant get it to work without causing merge conflicts....
will bring this down locally and fix and repost and merge as a new PR....

PR #1402, opened 2018-09-10
Irlock broken links

hamishwillee:
IR lock have updated their links from v1.0 to [v2.0](https://irlock.readme.io/v2.0/docs). The new docs do not include the instructions/tool for connecting via linux. 

I have fixed the links to point to the 1.0 docs but it is possible that this is not supported for whatever reason. @rmackay9 Any ideas how we can find out?

### DIFF
--- a/copter/source/docs/precision-landing-with-irlock.rst
+++ b/copter/source/docs/precision-landing-with-irlock.rst
@@ -38,8 +38,8 @@ Connecting to Pixhawk
 
 The IR-LOCK sensor can be connected directly to Pixhawk via an `I2C cable <https://irlock.com/collections/shop/products/pixhawk-cable>`__. If
 you are using multiple I2C sensors, then you will need an \ `I2C splitter <http://store.jdrones.com/Pixhawk_I2C_splitter_p/dstpx4i2c01.htm>`__.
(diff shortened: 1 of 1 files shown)


auturgy:
I’ve pinged Thomas Stone to have a look 

Regards,

James

ThomasSFL:
Thanks @hamishwillee  and @auturgy 

I just saw this. I will check it. 

Best,
Thomas

hamishwillee:
Thanks @ThomasSFL (and @auturgy ). Any update?

ThomasSFL:
Sorry for the delay @hamishwillee 

It is actually better to point to v2.0. That version of the IR-LOCK documentation corresponds with the latest ArduCopter code/features. 

We have not maintained the Linux support; however, it should still work (I think). For the Linux link, you could point directly to this url: https://irlock.readme.io/v1.0/docs/interfacing-sensor-w-linux-and-python

I hope this is helpful!

Best,
Thomas

auturgy:
Looks fine to me, but I can't test or functionally validate it.

hamishwillee:
Thanks @auturgy - this appears to match the "official docs" and presumably this was validated when the original docs went in. All this PR does is remove a small amount of stuff that is no longer maintained according to vendor. 
Upshot better than before, and certainly no less accurate/inaccurate. I have merged :-)

ThomasSFL:
Thank you, @hamishwillee !

PR #1403, opened 2018-09-10
Fix up links in hoverthings flip sport doc

hamishwillee:
@rmackay9 This fixes up broken links in this doc. 

This used Pixhawk 1 link to 3DR but that is no longer sold by them, so I updated the Pixhawk 1 overview to include link to the mRO equivalent, and then linked to that doc instead. This is model we should always use - overview doc should include purchase links, and everything links to that. That allows us to maintain stores in just one place.

BUT this is quite out of date and will doubtless break again. It covers the predecessor boards to Pixhawk 1. It uses the flip, which I think is no longer produced. 

So think we should merge this, but consider archiving doc?

### DIFF
--- a/common/source/docs/common-pixhawk-overview.rst
+++ b/common/source/docs/common-pixhawk-overview.rst
@@ -44,6 +44,16 @@ Specifications
    -  Width 50 mm (2.0”)
    -  Height 15.5 mm (.6”)
    -  Length 81.5 mm (3.2”)
+   
+   
+Purchase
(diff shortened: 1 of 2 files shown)


rmackay9:
Merged, thanks!

hamishwillee:
@rmackay9 Thanks. Should we archive this doc?

rmackay9:
@hamishwillee, yes, I think so..

PR #1404, opened 2018-09-10
Fix broken link in parachute

hamishwillee:
Just one link that was broken here. I think this is correct replacement.


rmackay9:
Merged, thanks!

PR #1405, opened 2018-09-10
Archive the Hoverthings Flip Sport detailed build

hamishwillee:
This build would probably work but airframe is no longer available and it  uses both Pixhawk 1 and PX4FMU/PX4IO (no longer supported). 

This does mean we don't have many detailed airframe builds. Would be good to encourage some more?


rmackay9:
Merged, thanks!

PR #1407, opened 2018-09-11
Fix link in wiki editing guide

hamishwillee:


hamishwillee:
Self merging. Nothing controversial.

PR #1416, opened 2018-09-16
Firmware Cannot be installed for external USB

ardisd:


common/source/docs/common-loading-chibios-firmware-onto-pixhawk.rst
```diff
@@ -30,7 +30,7 @@ Download the appropriate beta firmware from `firmware.ardupilot.org <http://firm
 Uploading as Custom Firmware
 ============================
  
-- connect the ground station PC to the flight controller using a USB cable
+- connect the ground station PC to the flight controller using a USB cable (Don't use external USB only use the USB built into the flight controller)
```
hamishwillee: What do you mean/how would someone connect via external bus to flight controller? Often there is a warning that you must connect the USB directly to the computer (not via a USB external bus) - is that what you mean?

ardisd: On my copter i have a micro USB plug mounted externally on side of the copter that is plugged into the USB prot on top of the pixhawk. ChibiOS cannot be installed using this set-up. You must use the build in USB port on the side of the pixhawk. I wasted a lot of time trying to install before using the other port

Just thought i could save others some time

hamishwillee: Thanks for the clarification.

auturgy:
They’re actually the same peripheral (the Pixhawk can’t tell which of the two physical ports you’re connected to), so this is likely a specific hardware/connection problem.

Regards,

James

PR #1417, opened 2018-09-16
Correction from Tootloaders to Bootloaders

lvale:
Correction from Tootloaders to Bootloaders


dev/source/docs/porting.rst
```diff
@@ -42,7 +42,7 @@ Some boards come with a bootloader pre-installed while others rely on the board
 The source code for the bootloaders can be found in `AP_Bootloader
 <https://github.com/ArduPilot/ardupilot/tree/master/Tools/AP_Bootloader>`__
 but pre-compiled binaries are available for many boards in the
-`Tools/Tootloaders
+`Tools/Bootloaders
```
peterbarker: case is incorrect; lower-case b

PR #1424, opened 2018-09-17
Corrected rc-in signal pin from 1 to 3

MFlghts:


auturgy:
Merging to aid users. Next step is to have a look into the hwdef.dat and the wiki together and evolve/update as necessary.

Issue #1412, opened 2018-09-11
Setting up SITL using Vagrant Broken


hamishwillee:
Followed instructions in [Setting up SITL using Vagrant](http://ardupilot.org/dev/docs/setting-up-sitl-using-vagrant.html). Do not work.

Instructions suggest using rsync. I installed this via cygwin. BUT if you try to uncomment the sharedfolders line `# config.vm.synced_folder ".", "/vagrant", type: "rsync", rsync__auto: true` you note that this is not present any more. The whole vagrant structure appears to have changed and support multiple base OS and init files. 

1. Do you need to install rsync on host still? If so, how is this "enabled"
2. Can I have a brief update on how the new scripts work - ie how do you choose different bases.

Anyway, at end of installation instruction, tried to run SIM. There is some issue related to Python. Thoughts?
```
D:\github\forks\ardu\ardupilot>vagrant ssh -c "sim_vehicle.py -j 2"
Ardupilot environment ready.  Run 'sim_vehicle.py' to start simulating an arducopter instance.
To build for fmuv2:
  cd /vagrant
  ./waf configure --board=px4-v2
  ./waf build --target=bin/arducopter
/usr/bin/env: ‘python\r’: No such file or directory
Connection to 127.0.0.1 closed.
```

3. Thoughts on debugging this?

peterbarker:
On Mon, 10 Sep 2018, Hamish Willee wrote:

> Followed instructions in Setting up SITL using Vagrant. Do not work.
> 
> Instructions suggest using rsync. I installed this via cygwin. BUT if you try to uncomment the sharedfolders line # config.vm.synced_folder ".",
> "/vagrant", type: "rsync", rsync__auto: true you note that this is not present any more. The whole vagrant structure appears to have changed and support
> multiple base OS and init files.

I suggest we nuke those instructions unless someone knows how to get that 
working.

>  1. Do you need to install rsync on host still? If so, how is this "enabled"
>  2. Can I have a brief update on how the new scripts work - ie how do you choose different bases.

vagrant up xenial32
vagrant ssh xenial32

> Anyway, at end of installation instruction, tried to run SIM. There is some issue related to Python. Thoughts?
> 
> D:\github\forks\ardu\ardupilot>vagrant ssh -c "sim_vehicle.py -j 2"
> Ardupilot environment ready.  Run 'sim_vehicle.py' to start simulating an arducopter instance.
> To build for fmuv2:
>   cd /vagrant
[...]

PR #1425, opened 2018-09-17
Remove rsync info from SITL/vagrant docs

hamishwillee:
This removes the Rsync information from the [Setting up SITL using Vagrant](http://ardupilot.org/dev/docs/setting-up-sitl-using-vagrant.html) doc because the referenced lines do not exist in the current vagrant structure. 

Would be better to confirm if this could be updated. Who would know?

This follows from #1412

### DIFF
--- a/dev/source/docs/setting-up-sitl-using-vagrant.rst
+++ b/dev/source/docs/setting-up-sitl-using-vagrant.rst
@@ -67,60 +67,6 @@ Preconditions
    system PATH. SSH is installed with GIT, or you can install it
    independently for your platform.
 
-Building PX4 Firmware - Rsync
------------------------------
-
-If you're using this Vagrant file for the purpose of building PX4
-firmware you will probably also wish to install *Rsync*, as this
-significantly speeds up PX4 builds. The relative build times using
-different approaches (on a 3Ghz i5 Haswell / 16 Gb) are shown below:
-
--  Native windows px4 toolchain (gcc/mingw): 190 minutes
--  Vagrant with shared folders (the default setup described here): 15
-   minutes 30s (12x faster)
--  Vagrant with rsync folders: 1 minute 40s (120x faster)!.
-
-.. tip::
-
-   Everything is set up to use shared folders by default. The only
-   caveat is that ``px4-clean`` *does not work with shared folders*. You
-   either need to use rsync or run ``px4-clean`` from outside Vagrant (you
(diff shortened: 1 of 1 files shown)


hamishwillee:
The removed text is:

```
Building PX4 Firmware - Rsync
-----------------------------

If you're using this Vagrant file for the purpose of building PX4
firmware you will probably also wish to install *Rsync*, as this
significantly speeds up PX4 builds. The relative build times using
different approaches (on a 3Ghz i5 Haswell / 16 Gb) are shown below:

-  Native windows px4 toolchain (gcc/mingw): 190 minutes
-  Vagrant with shared folders (the default setup described here): 15
   minutes 30s (12x faster)
-  Vagrant with rsync folders: 1 minute 40s (120x faster)!.

.. tip::

   Everything is set up to use shared folders by default. The only
   caveat is that ``px4-clean`` *does not work with shared folders*. You
   either need to use rsync or run ``px4-clean`` from outside Vagrant (you
   can also get usable results by cd'ing to each module subdirectory and
   running\ `` git clean -x -d -f`` but don't do this from the top level
   otherwise you will delete your vagrant temporary files.

If you want to use rsync you need to:

-  Uncomment the `appropriate line <https://github.com/ArduPilot/ardupilot/blob/master/Vagrantfile#L37>`__
   in the Vagrantfile:

   ::

       # config.vm.synced_folder ".", "/vagrant", type: "rsync", rsync__auto: true

-  *rsync* must be installed on the host computer and be added to the
   system PATH. According to `the vagrant rsync guide <http://docs.vagrantup.com/v2/synced-folders/rsync.html>`__ you
   can install rsync from either
   `Mingw <http://sourceforge.net/projects/mingw/files/Installer/>`__ or
   `Cygwin <https://cygwin.com/>`__. Assuming you're using Mingw:

   -  `Download the latest Mingw installer <http://sourceforge.net/projects/mingw/files/Installer/>`__
   -  In the installer, select and install the *mingw-developer-toolkit*

      .. figure:: ../images/MinGW-InstallationManager.png
         :target: ../_images/MinGW-InstallationManager.png

         MinGW Installation Manager

   -  Use *mingw-get* to install rsync (you may need to add *mingw-get*
      to your path):

      ::

          mingw-get install msys-rsync

   -  Set the system path to point to rsync (By default this is in
      **C:\\MinGW\\msys\\1.0\\bin**.)
```

PR #1434, opened 2018-10-01
change inline markup

matthewaltenburg:
The double negative is interpreted by sphinx as a one-line I added backwards quotes to correct this.


hamishwillee:
Thanks @matthewaltenburg . Merging.

Just FYI, IMO this would render even better as a code block. In RST that can be done like:
```
::

    ./waf configure --board disco
    ./waf plane
```

Issue #1160, opened 2018-04-05
Radio Control Calibration in Mission Planner


rfengfpv:
[Radio Control Calibration in Mission Planner](http://ardupilot.org/plane/docs/common-radio-control-calibration.html)

This is a minor correction to generally excellent documentation. 

Under heading "Channel Mappings" it says "Channel 8 (default): Flight modes. Mode selection can be mapped to any unused parameter with the MODE_CH parameter." In Arduplane 3.8.4 this parameter is actually called "FLTMODE_CH".

PR #1444, opened 2018-10-07
Fix parameter name (ArduPlane)

umer936:
Closes https://github.com/ArduPilot/ardupilot_wiki/issues/1160


peterbarker:
This looks correct - but the sentence as a whole still doesn't make sense!

"mapped to any unused parameter with the ``FLTMODE_CH`` parameter." should probably just be `mapped to any RC channel using the parameter ``FLTMODE_CH``"

Would you like to make that change as part of this PR or will I just merge this as-is?

peterbarker:
I've squashed the commit together and merged them.  Thanks!

PR #1447, opened 2018-10-16
Update common-autopilots.rst

auturgy:
remove Snapdragon Flight.  Board is no longer supported.


rmackay9:
perhaps we should move the snapdragon page into the archive?

rmackay9:
Yes, instead of completely removing the link we should just move it down a few lines into the "Discontinued boards"

auturgy:
Sure.
We need to check the associated wiki pages too. We stopped supporting this board some time ago.

Regards,

James

rmackay9:
instead of removing it, I've moved the link to the deprecated section.

PR #1462, opened 2018-11-02
Add video showing how to install Mission Planner.

clum:
I did a screen capture video showing how to install Mission Planner.  This also shows how to install older or multiple versions of Mission Planner.  Finally, it shows how to set the log path so that tlogs and dataflash logs are saved appropriately.

### DIFF
--- a/planner/source/docs/mission-planner-installation.rst
+++ b/planner/source/docs/mission-planner-installation.rst
@@ -4,7 +4,7 @@
 Installing Mission Planner (Windows)
 ====================================
 
-A video showing how to install *Mission Planner* is located `here <https://youtu.be/cw7adeD1yeg>`__.  Alternatively, the below instructions show how to install *Mission Planner* on Windows.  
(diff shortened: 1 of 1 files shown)


clum:
It seems like this change was approved.  Can someone with write access merge this pull request?  Thanks!

rmackay9:
Hi there, txs for the video.  I'm not sure I'm going to merge it actually because it seems like it diverges from the wiki advice which is just to download and use the latest MP.  I must admit that I haven't watched the whole video yet but I got to the part where it suggested installing a specific version of MP.. I think in general the videos should match the content in the wiki.  If the wiki content needs to be updated then let's update that.. I'm just a bit concerned that we're giving different advice in writing vs what's in the attached video.

auturgy:
It’s a good, clear video which I think is quite useful. The points you raise are valid though, and why I pushed some changes into this pr.
The question I have is whether it might be better being linked from the advanced setup page rather than here.

auturgy:
(a note clarifying that versions change would be useful).

clum:
These are good points.  The content of the video is applicable to using the latest version installer.  However it is true that it does focus on showing how to install and use any version you want.  I'm happy to move this to advanced installation page of this is more appropriate.

rmackay9:
OK, let's put it in. Txs for this.

PR #1463, opened 2018-11-03
Rover: Sailboat and wind vane documentation

IamPete1:
This adds documentation for the rover sailboat and wind-vane. 

@rmackay9  I think I caught all the changes between my branch when I originally wrote this and what ended up in master. Can you have a quick double check, Thanks!


rmackay9:
Merged, thanks!

PR #1468, opened 2018-11-06
wiki: update trademark.rst

auturgy:
words to indicate an application is underway.  Need to remember to update this when the process is complete.


rmackay9:
MErged, thanks

PR #1473, opened 2018-11-11
Added connections diagrams and notes

Hwurzburg:
Add connection diagrams and notes


auturgy:
Great!  The only query I have isn't so much related to this PR, but a hwdef decision.  The airspeed input is defaulted to RSSI, but the F35 comes with an analog airspeed sensor, so I expect users will think it "just works".  I think we need to make that very clear, or change the default.  Just my 2c.

Hwurzburg:
As far as rssi vs analog speed sensor, I don't think there is any hwdef issue....it just assigns the ADC ch1 input to the pin used(it can be muxed to many pins and depend on which the board designer decides to physically use)...in F-35 its 13, in Omnibus its 0....when the firmware is loaded, I don't think either RSSI or speed sensor is enabled, you have to manually set either up in the params...in fact, most of the params are invisible until you actually enable one or the other, then assign the sensing pin...but its good into to add to the boards wiki

rmackay9:
Merged, thanks!

PR #1490, opened 2018-11-20
Plane: add logo, remove the tip at the top of page, hide bottom menu

matthewaltenburg:
Add some changes to improve website flow.


hamishwillee:
Did you test this on mobile? It may be that the bottom toc is better hidden, but in some themes it is a better user experience if it is not.

PR #1493, opened 2018-11-21
Website flow

matthewaltenburg:
I have started adding changes to improve the flow of the documentation. These are just some minor ones. For example, the tip at the top of the page creates a circular loop for persons when they first come to the main copter, plane, and rover pages, so I removed them. I will create a branch next week for more significant changes. The menu at the bottom and the side creates the effect that the bottom list is a step by step guide and puts the next button way too far at the bottom of the page.


auturgy:
We’re probably dropping the individual platform logos, and just using ArduPilot everywhere.
Don’t add them. I’ll just have to revert them later.

Regards,

James

matthewaltenburg:
Why have we decided to drop them?

matthewaltenburg:
The way I see it the ArduPilot wiki and webpage is about branding and communication. I know it is open source but what I have noticed is that this becomes its most significant problem with users who get confused as to what Ardupilot is. For example is it for, flying, ground, water, stationery. I have been doing a lot of research on this problem since the last developers meeting in Canberra. Also over the last week, I have been looking much closer at the website. While the wiki is very technical, it confuses the average user. Please allow me the leeway to help fix this.

auturgy:
Basically, we have enough trouble promoting one “brand”, let alone 6 or more.
An enormous number of consumers who use “Arducopter” don’t know that they are using “ArduPilot”. Similarly, the hundreds of thousands of Skyvipers users know their copter runs ArduPilot. They won’t know Arducopter is the same thing. 
Arduplane.com and Arducopter.com redirect to ArduPilot.org
Sub is an exception at the moment, in that Rusty, Jacob and their team maintain Ardusub separately, with their own documentation suite at ardusub.com.

I’m really greatful for your assistance:  some things just need discussion and a bit of context though. Feel free to ping me on Skype.

matthewaltenburg:
I have started a new branch on my repository wiki and will have it reviewed later, so I am closing this pull request.

PR #1515, opened 2018-12-04
adding-a-new-mavlink-message: fix parameter typo

MaEtUgR:
I assume this is a typo...

FYI @hamishwillee


auturgy:
Not actually a typo, but it is an ArduPilot specific parameter.
http://ardupilot.org/copter/docs/common-telemetry-port-setup-for-apm-px4-and-pixhawk.html

rmackay9:
yeah, not a typo actually.  The "n" should be replaced by the port number.  so for example SERIAL1_PROTOCOL should be set to change Telem1.

hamishwillee:
:-)

Just to reduce further chance of misinterpretation, I provided a clarifying example: https://github.com/ArduPilot/ardupilot_wiki/commit/a413ebab522e11a3c9b0f734a1f1309d9acb7900

PR #1522, opened 2018-12-17
optitrack: add required ardupilot version and frame info

chobitsfan:
It notes Copter-3.6 or higher and EKF2 is required for ext nav support. It also add some info about RTF frame used in the example image and video. I hope this may help people to duplicate indoor experiment.


rmackay9:
Merged, thanks!

PR #1523, opened 2018-12-17
Rangefinders: Correct Lightware default baud rate

DavidIngraham:
I've discovered that lightware has changed the default baud rate for some devices (at least for the SF11/C). This PR changes the recommended parameters to work with the new default baud rate and adds a note about how to properly set the baud rate for older devices.

One note is that I haven't verified that the default baud rate was changed for devices other than the SF11/C.


rmackay9:
Merged, thanks!

PR #1526, opened 2018-12-20
dev: Fix build instructions for bebop and bebop 2

jberaud:
Use Parrot's version of the linaro armhf toolchain.
Don't build statically for bebop2 since it is not needed anymore since
Parrot has switched to an official software build with the linaro armhf
toolchain.


dev/source/docs/building-for-bebop-2.rst
```diff
@@ -27,7 +27,52 @@ a few noticeable changes, not the least being a much better quality GPS
...
+
+       export PATH=/opt/arm-2016.02-linaro/bin:$PATH
+
+Download and compile ArduCopter
+-------------------------------
+
+#. You need to install git first (see `instructions here <https://git-scm.com/book/fr/v1/D%C3%A9marrage-rapide-Installation-de-Git>`__)
+#. Clone ardupilot repository
+
+   ::
```
khancyr: .. include:: git-clone.rst
    :start-after: inclusion-marker-do-not-remove
    :end-before: Cloning with the GitHub GUI (Windows or MAC)

dev/source/docs/building-for-bebop-2.rst
```diff
@@ -27,7 +27,52 @@ a few noticeable changes, not the least being a much better quality GPS
...
+#. Add the path to the toolchain to the PATH variable
+
+   ::
+
+       export PATH=/opt/arm-2016.02-linaro/bin:$PATH
+
+Download and compile ArduCopter
+-------------------------------
+
+#. You need to install git first (see `instructions here <https://git-scm.com/book/fr/v1/D%C3%A9marrage-rapide-Installation-de-Git>`__)
```
khancyr: that is the french page ^^ 
as for the git clone, we should point on git-install file 
Something like : 
.. include:: git-install.rst
     :end-before: GitHub Desktop

Will look at that tonight

jberaud:
I removed the git part. I think people should be able to understand that they need git to clone the repository with git. The other docs don't link to that and it could be found on google in an instant.

PR #1527, opened 2018-12-21
fix baudrate error and options in raspberry pi page

mnihalm:
Two Changes - 
1. Update the Serial port section to reflect the change in the settings menu.
2.  Fix the baudrate in the connection command to be the same as the one set in pixhawk

### DIFF
--- a/dev/source/docs/raspberry-pi-via-mavlink.rst
+++ b/dev/source/docs/raspberry-pi-via-mavlink.rst
@@ -177,6 +177,11 @@ And then "Serial" to disable OS use of the serial connection:
 .. image:: ../images/dev_RasPi_Config_Utility_SerialSetting_Capture2-300x78.png
     :target: ../_images/dev_RasPi_Config_Utility_SerialSetting_Capture2-300x78.png
 
+
+.. note::
+
+   In the latest Rapbian Stretch, these settings willbe found in  "Interfacing Options"
+
 Reboot the Raspberry Pi when you are done.
 
 Testing the connection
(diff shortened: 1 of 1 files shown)


dev/source/docs/raspberry-pi-via-mavlink.rst
```diff
@@ -177,6 +177,11 @@ And then "Serial" to disable OS use of the serial connection:
 .. image:: ../images/dev_RasPi_Config_Utility_SerialSetting_Capture2-300x78.png
     :target: ../_images/dev_RasPi_Config_Utility_SerialSetting_Capture2-300x78.png
 
+
+.. note::
+
+   In the latest Rapbian Stretch, these settings willbe found in  "Interfacing Options"
```
khancyr: will be

dev/source/docs/raspberry-pi-via-mavlink.rst
```diff
@@ -189,7 +194,7 @@ RPi type:
 ::
 
     sudo -s
-    mavproxy.py --master=/dev/ttyAMA0 --baudrate 57600 --aircraft MyCopter
+    mavproxy.py --master=/dev/ttyAMA0 --baudrate 921600 --aircraft MyCopter
```
khancyr: agree event if not all board support those baudrate

dev/source/docs/raspberry-pi-via-mavlink.rst
mnihalm: Fixed, Thanks!

dev/source/docs/raspberry-pi-via-mavlink.rst
mnihalm: Agreed!

PR #1530, opened 2018-12-22
Common: rework Revolution(mini) hardware page

vierfuffzig:
- fix diagrams to align with ArduPilot defaults
- add default UART order
- take account of fullsize Revolution and RevoMini pinout differences
- remove redundant passages and reference dedicated wiki pages instead
- minor corrections and structural simplifications


auturgy:
Thanks!

PR #1536, opened 2018-12-27
Update index.rst

auturgy:
Move autopilot options up before first time setup


rmackay9:
Ok, we should probably make the same change for all the vehicles..

PR #1537, opened 2018-12-27
Plane: Add a Need TX column and LAND mode

Naterater:
Clarifies what modes expect TX input.


Hwurzburg:
Nate, can't modes like FBWA and CRUISE be flown without a TX from the GCS using joystick? Transmitter-less operation is possible (although not recommended) as I understand it...

Hwurzburg:
GCS+Joystick is considered the same as TX, so merging

PR #1538, opened 2018-12-27

josheeg:
http://ardupilot.org/dev/docs/setting-up-sitl-on-windows.html
has steps like install git that are not on here that would be needed.

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-linux.rst
+++ b/dev/source/docs/setting-up-sitl-on-linux.rst
@@ -26,6 +26,21 @@ Install steps
 
 Please follow each of the steps described below.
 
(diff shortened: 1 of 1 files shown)


khancyr:
Those are really old and not accurate information. The setting page is already complet ...

khancyr:
I will add another explicite link to git install, but it is present on http://ardupilot.org/dev/docs/building-the-code.html#building-the-code .
I will check for jbsim dependencies

josheeg:
What pages should a user fallow in what order install on Linux would be
number one.

khancyr:
Hello, I have update instructions for linux.
I will update instruction for jsbsim soon. 

I am closing this

PR #1540, opened 2018-12-27
ICE: Update ICE library docs

DavidIngraham:


auturgy:
Great!
Do we need to be explicit that this is for gas engines, not nitro?

rmackay9:
Looks like we have a couple of issues on the page with links.. anyway.. a lot more content now!

PR #1543, opened 2018-12-28
Update common-rtf.rst

auturgy:
Add Holybro QAV250 with Pixhawk 4


rmackay9:
As mentioned, not a fan because it doesn't ship with AP and it's not really an RTF.

PR #1545, opened 2018-12-28
Style Guide: Standardize Parameter Conventions

Naterater:
Adds section to keep parameter conventions consistent.

### DIFF
--- a/dev/source/docs/style-guide.rst
+++ b/dev/source/docs/style-guide.rst
@@ -422,20 +422,25 @@ The description is a long text field for a complete description of the parameter
 **Right:**
 
 ::
-	// @Description: Gain added to pitch to keep aircraft from descending or ascending in turns. Increase in increments of 0.05 to reduce altitude loss. Decrease for altitude gain.
-   
+
+    // @Description: Gain added to pitch to keep aircraft from descending or ascending in turns. Increase in increments of 0.05 to reduce altitude loss. Decrease for altitude gain.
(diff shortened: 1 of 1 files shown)


WickedShell:
> I think the parameter name and description should start with the subsystem name so instead of "Maximum Pitch Angle" it should be "Limit Pitch angle maximum". In general I think "Max" or "Min" and units (if necessary) should be at the end

I find that example to be much harder to read/think about/parse and would definitely prefer the former. It's meant to be the human readable form and I think the former does a lot better at providing that for us.

Naterater:
@rmackay9, I’m fine with the first part, but I have to agree with @WickedShell that the original seems easier to use. Honestly it would be great to start referencing parameters by their Display Names instead of CODE_WORDS (that I can never remember). During support, I think I would rather ask users to raise their “Maximum Pitch Angle” as opposed to asking them to raise their “Limit Pitch Angle Maximum.” I think the max and min at the end of the parameter code name.

rmackay9:
@WickedShell, @Naterater, how about we agree that the Display Name start with the subsystem but the Description does not need to?  I think the Display Name is the equivalent of the parameter name but without the 16 character limit.  At the moment, with many Display Names **not** starting with the subsystem it is very hard to know what area they are related to.. the user really needs to look at the parameter name to figure this out.

rmackay9:
squashed and merged, thanks!

PR #1552, opened 2019-01-04
Dev - add warning about different gcc-arm versions

stephendade:
Add a small warning about not using distribution-repo-provided gcc-arm versions. Just in case a user (like myself) goes and does exactly that.

Turns out that the gcc-arm in the Ubuntu 18.04 repo, although is a very close version to the one in http://firmware.ardupilot.org/Tools/STM32-tools/, produces bad binaries as per https://github.com/ArduPilot/ardupilot/issues/8663


rmackay9:
I wonder if instead of a warning above if we could just have a little warning in brackets right beside where it says, "The gcc-arm cross-compiler from here".  Something like, "(warning: gcc installed with apt-get will not work properly)"

stephendade:
Sure - I've fixed that up

rmackay9:
Great, thanks!

PR #1554, opened 2019-01-04
General update

khancyr:
update sphinx to lastest
update youtube plugin repo and plugin
replace pngmath with imgmath because of deprecation
update copyright date to 2019


rmackay9:
Merged thanks!

I'm not qualified to know if this is really correct but it looks convincing!

PR #1564, opened 2019-01-08
Update JSBSim Instructions

josheeg:
Update JSBSim Instructions to make it more user friendly.

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-linux.rst
+++ b/dev/source/docs/setting-up-sitl-on-linux.rst
@@ -133,6 +133,10 @@ model than the default Rascal110 by specifying the model name using the
 ::
 
     sim_vehicle.py -f jsbsim:MyModel --console --map
+    
+    replace MyModel with Rascal It is in the *Tools/autotest/aircraft/** directory.
+    
+    sim_vehicle.py -f jsbsim:Rascal --console --map
 
 the model should be in the **Tools/autotest/aircraft/** directory.
 


rmackay9:
I don't know enough about JSBSim to know if this is the correct thing to do or not.  Does anyone else know?

auturgy:
We need to look at this a bit more.  The build instructions work, but just using the -f jsbsim switch on SITL errors out
''' james@auturgy:~/Development/ardupilot/ArduPlane$ sim_vehicle.py -f jsbsim --console --map
SIM_VEHICLE: Start
SIM_VEHICLE: Get JSBSim version
SIM_VEHICLE: "JSBSim" "--version"

=========================================================
You need the latest ArduPilot version of JSBSim installed
and in your \$PATH

Please get it from git://github.com/tridge/jsbsim.git
See
http://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html
for more details
=========================================================

SIM_VEHICLE: Killing tasks
james@auturgy:~/Development/ardupilot/ArduPlane$ 
 '''
I think we need to look at SIM_VEHICLE.py, then come back to this.

filipsladek:
I did not notice error you mention but other error if using git://github.com/tridge/jsbsim.git
Solution is in this patch ArduPilot/ardupilot#10185

josheeg:
The link http://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html
Does not show the old not working tridge jsbsim.
The newest official jsbsim release works with the Rascal only for me.

tridge:
it should already be clear I think. The Rascal is the default, the MyModel example is for a new model

PR #1566, opened 2019-01-09
updated editing quadplane mission text

josheeg:
updated editing quadplane mission text
Right click on the map mission -> editor read wp change one move selection write waypoints.


rmackay9:
I'm afraid I'm not qualified to say if this is correct or not.. Hopefully a plane person can make the decision

tridge:
i don't think the new text is clear enough to understand

PR #1570, opened 2019-01-11
Updated "PM (performance monitoring)" message details

Brooce:
Removed few outdated fields and added the ones currently seen when reviewing log data, basing on what can be seen in the code and peterbarker's explanations: https://discuss.ardupilot.org/t/help-understanding-a-pm-performance-monitoring-log-data/37100 . Also removed any fixed numbers that depend on scheduler loop rate.


peterbarker:
Merged, thanks!

Not 100% sure on the Load one - but it's better than what we had.

PR #1573, opened 2019-01-14
mission-planner-installation: fix screenshot

vierfuffzig:
we're using an installation screenshot from APM Planner right now. this fixes the screenshot to use one of an actual MP installation process and moves the image source into planner subfolder.


rmackay9:
Merged, thanks!

PR #1575, opened 2019-01-15
PrecisionLanding: Describe landing param options

hamishwillee:
Hi Randy,

This documents the precision landing options - capturing the fact that `Enabled Strict` searches for a missing beacon and then lands anyway if it can't be found. 

If you can provide more details on the search procedure I could add that too? Ie ie you might just hover and wait, or you might circle or whatever.

### DIFF
--- a/copter/source/docs/precision-landing-with-irlock.rst
+++ b/copter/source/docs/precision-landing-with-irlock.rst
@@ -73,10 +73,12 @@ various mounting locations.
 Setup through Mission Planner
 =============================
 
-Set the following parameters through the Mission Planner (or other GCS)
-to enable the precision landing feature and then Reboot the flight controller.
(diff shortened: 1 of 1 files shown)


rmackay9:
I don't think we should talk about "strict" because it actually makes no difference in behaviour.  So I actually don't think we should make this change.  Instead we should change the parameter description so it doesn't talk about a "2" option.

hamishwillee:
> I don't think we should talk about "strict" because it actually makes no difference in behaviour.

The difference in behaviour implied above depends on the timeout/how things are implemented. If your search behaviour is to ascend 100m then loiter for 10  mins looking for a beacon before landing, that is quite different for the end user than just landing as soon a land is enabled. 

So I guess you're saying that the parameters are implemented in the same way or so close it doesn't matter? If so, sure!

PR #1578, opened 2019-01-15
Update common-baro-external.rst

vierfuffzig:
using BMP280 types on various fixed wing setups with good results, tested on plane and copter.
secondary press / temp has been added to MP.
secondary temp panel has been added to OSD (OSDn_BTEMP).


auturgy:
Thanks!

rmackay9:
yeah, nice update, thanks!

PR #1582, opened 2019-01-17
Dev: update JSBSIM install instruction

khancyr:
make clear that cmake should be install (even if part of install-prereqs)
build JSBSIM as release, as most user will use it that way. The instruction is taken from JSBSIM wiki
consider to use make -j2 to speed up think a little (most computer are at least dual core now, event ARM board)
Remove sudo make install

Add mission -v ArduPIane


rmackay9:
I'm not really qualified to say whether this is correct or not but hopefully @khancyr knows better than me

PR #1584, opened 2019-01-21
pypy cylon is needed

josheeg:
pypy cylon is needed


auturgy:
Hi,
You’ve raised a couple of pr’s for dependencies that shouldn’t be required.
Looking at them, I’m going to hazard a guess that you’re trying to use the JSBSim inbuilt scripting.
As far as I’m aware that won’t be useful for testing or learning ArduPilot. You should try scripting through DroneKit or MAVProxy.

PR #1585, opened 2019-01-21
SCIPY install for requirements numpy

josheeg:
SCIPY install for requirements numpy


khancyr:
Hello,

All those dependencies aren't need.

PR #1586, opened 2019-01-21
cylon needed

josheeg:
cylon needed


khancyr:
This isn't need... https://github.com/JSBSim-Team/jsbsim/blob/master/DevelopersDocs.md#building-jsbsim

Issue #1599, opened 2019-01-23
Deprecated Documentation


stephendade:
Is it worth removing deprecated documentation? Or making it more obvious that it's deprecated?

I ask this after looking at the "Building the Code" section of the developer documentation. There's still many pages there that have a small "Deprecated" label in the left-hand navigation bar ... and that's it (example http://ardupilot.org/dev/docs/building-for-navio-on-rpi2.html). Not really obvious.

In order to reduce the chance of new developers/users mistakenly using the old documentation, do we want to:
a) Remove the deprecated pages (the cleanest way, but maybe people want the old documentation for reference?)
b) Adding a "deprecated" Section to the documentation navigation bar and putting all the pages there
c) Just adding a big "deprecated" warming to the top of each page?

khancyr:
Even if those doc are deprecated there are still people using then... A bunch could be removed thus...
I would vote for option C for now and try to make some triage on which on could be removed

rmackay9:
I also vote for (c)

stephendade:
>(b) is already done I think. http://ardupilot.org/copter/docs/common-archived-topics.html

Ahh, I did not see that before. In that case, probably best to extend this across to the dev documentation

hamishwillee:
Using Appendix for deprecation was a good temporary fix. A better long term solution is to provide versioned builds - that would allow us to dump all the cruft and just have relevant/recommended stuff in the head revision.
This is perfectly do-able, but requires a degree of automation of the toolchain.

Hwurzburg:
versioning is beyond the scope of this issue and is always being discussed, so I am closing this specific wiki issue...we are trying to stay current on Archived topics...

PR #1600, opened 2019-01-24
Dev: move deprecated documentation to appendix

stephendade:
As per conversation in #1599 and the Ardupilot Wiki guide (http://ardupilot.org/dev/docs/common-wiki-editing-archiving.html), I've moved the deprecated build documentation to the appendix.


OlivierJB:
+1 on this.

Even if deprecated renaming  "building-px4-with-make.rst"  to "building-with-make.rst"   or something would also make sense,  as the current file name is bound to  confuse many.

stephendade:
@rmackay9 - are you happy to merge this in?

rmackay9:
Merged, thanks!

rmackay9:
By the way, in general let's not rename files because it will just cause external links to break (this PR doesn't rename any which is good)

PR #1605, opened 2019-01-24
common-partners.rst: fix holybro URL

vierfuffzig:
no s in their http


auturgy:
Thanks, appreciated. I actually knew that when I made the pr. Not sure how I messed it up!

PR #1608, opened 2019-01-25
Update common-camera-shutter-triggering-cables-for-purchase.rst

SeagullUAV:
Added other camera triggering devices from Seagull UAV and updated old #REC post.


rmackay9:
I've removed the affected page and instead used the contents of this PR to create individual pages for 4 of the seagull camera trigger products.  This is a lot of pages, perhaps more than is really required but it's an improvement over the current situation I think.

PR #1611, opened 2019-01-25

SeagullUAV:
Have added Seagull UAV #RADIO to the list of telemetry units. 
The common-telemetry-seagull-radio.rst is pending for approval.


rmackay9:
As mention on the other PR (https://github.com/ArduPilot/ardupilot_wiki/pull/1611) I think I won't merge this one because I think it's just a variant of the SiK radio that we already have a page for.

PR #1613, opened 2019-01-27
Update common-camera-shutter-triggering-using-seagull-map2.rst

SeagullUAV:
Fixed links to the cable finder, which was previously displayed incorrectly.


rmackay9:
I think I've fixed these links as part of the cleanup of the seagull pages

PR #1616, opened 2019-01-28
Create common-seagull-gpk-rtk-ppk-gnss-receiver.rst

SeagullUAV:
Added Seagull #GPK RTK / PPK GNSS receiver. The product is designed for navigational and logging purposes, which is integrate-able with Ardupilot based platforms.


rmackay9:
I've added the DevCallTopic so we can discuss this one on the next Wiki dev call.

rmackay9:
Thanks for the contribution.. after some lengthy discussion and thought we think this product is perhaps not different enough from other GPSs that it's worth us adding it to the wiki.

PR #1617, opened 2019-01-28

SeagullUAV:
Adding Seagull UAV #GPK RTK / PPK GNSS Receiver to the list
The 'common-seagull-gpk-rtk-ppk-gnss-receiver.rst' is awaiting approval.


rmackay9:
Thanks for the contribution.. after some lengthy discussion and thought we think this product is perhaps not different enough from other GPSs that it's worth us adding it to the wiki.

PR #1618, opened 2019-01-29
changes I had to make to run this tutorial run

josheeg:
changes I had to make to run this tutorial run


tridge:
rebased and changed to only use the new instructions

PR #1623, opened 2019-02-02
Changing from px4-v2 to Pixhawk1

MohamedAliRashad:
Because of this error message:
"The board target px4-v2 has been removed from ArduPilot with the removal of NuttX support and HAL_PX4."

### DIFF
--- a/dev/source/docs/learning-ardupilot-the-example-sketches.rst
+++ b/dev/source/docs/learning-ardupilot-the-example-sketches.rst
@@ -26,7 +26,7 @@ sketch on a Pixhawk:
 ::
 
     cd $ARDUPILOT_HOME # the top-level of an AruPilot repository
-    ./waf configure --board=px4-v2
+    ./waf configure --board=Pixhawk1
     ./waf build --target examples/INS_generic --upload
 
 waf can list the examples it can build:


rmackay9:
Thanks.   I know Tridge is keen that we move to board names for the builds but I think this is the first case in the wiki where we use "Pixhawk1".  In all other cases it's fmuv3 I think.  Anyway, it's fine but I'm sticking with fmuv3 because I'm not actually confident that any of the devs have moved to "Pixhawk1".

rmackay9:
although not a direct 1-to-1 mapping I think it'll be OK though..

MohamedAliRashad:
i changed this based on the error message i received
 
```
Please use a replacement build as follows:

 px4-v2     Use Pixhawk1 build
 px4-v3     Use Pixhawk1 or CubeBlack builds
 px4-v4     Use Pixracer build
 px4-v4pro  Use DrotekP3Pro build
```

peterbarker:
On Sun, 3 Feb 2019, mohamed ali rashad wrote:

> i changed this based on the error message i received

Fair enough!

   We're likely to get a rush of "firmware doesn't fit on my board?" 
support requests at some future time due to this advice.  *sigh*.

MohamedAliRashad:
Fixing it now might be a good idea XD

> On Sun, 3 Feb 2019, mohamed ali rashad wrote: i changed this based on the error message i received
> Fair enough! We're likely to get a rush of "firmware doesn't fit on my board?" support requests at some future time due to this advice. *sigh*.

Fixing it now might be a good idea ^ ^

PR #1637, opened 2019-02-12
ROS Aruco Board detection wiki page

anbello:
This wiki page describes how to setup a system capable to realize indoor autonomous flight. The system is based on a quadcopter with a Raspberry Pi 3 and a Raspberry Pi Camera Module v2. Images from camera are used to calculate poses estimation on the Raspberry Pi and the result are sent as mavlink messages to the Flight Controller. The camera is downward looking and on the floor there is an Aruco Board.


rmackay9:
Thanks!

anbello:
I’m happy to help

PR #1641, opened 2019-02-12
upgrade to using python3

drnic:
* python2.7 will retire end of this year https://pythonclock.org/
* this PR updates scripts + documentation to switch from `python` to `python3`

I've not tested the Windows `make.bat` scripts. I'm unsure of the assumptions of the environments where they are being run, sorry.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,24 +9,24 @@ Vagrant.configure(2) do |config|
     set -x
 
     apt-get -y update
-    apt-get install -y unzip git imagemagick mercurial python-pip curl
+    apt-get install -y unzip git imagemagick mercurial python3-pip curl
 
-    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && rm -f get-pip.py
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py && rm -f get-pip.py
 
     # Install sphinx
-    pip install -U sphinx==1.8.3
+    pip3 install -U sphinx==1.8.3
 
     # Install sphinx theme from ArduPilot repository
-    pip install git+https://github.com/ArduPilot/sphinx_rtd_theme.git -UI
+    pip3 install git+https://github.com/ArduPilot/sphinx_rtd_theme.git -UI
 
     # and a youtube plugin:
-    pip install git+https://github.com/sphinx-contrib/youtube.git -UI
+    pip3 install git+https://github.com/sphinx-contrib/youtube.git -UI
 
     # and a vimeo plugin:
     hg clone https://bitbucket.org/jdouglass/sphinxcontrib.vimeo
     pushd sphinxcontrib.vimeo
-    python setup.py build
-    python setup.py install
+    python3 setup.py build
+    python3 setup.py install
     popd
 
   SHELL
--- a/common/source/docs/common-wiki-editing-setup.rst
+++ b/common/source/docs/common-wiki-editing-setup.rst
@@ -59,7 +59,7 @@ The main steps for building the docs are:
    .. code-block:: bash
 
        cd /vagrant
-       python update.py
+       python3 update.py
 
 Build the Wiki
 --------------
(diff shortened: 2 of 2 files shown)


peterbarker:
Switching to point to a *specific* version of Python doesn't seem correct here.  We're effectively overriding the user's Python-version decision and claiming we are correct and they must conform (*right now*!) by installing Python3 (where we still happily work with Python2).

Any place specifying python2 should change to `python` - but changing `python` to `python3` doesn't feel right.  There's no need for us to force this on our users ('though devs should definitely be switching!).

drnic:
Would you like the PR closed? Or want python3 installed in vagrant but aliased to `python`?

Or would you link an initial PR that installs both python-pip + python3-pip packages, and performs the same setup in `Vagrantfile` for both pythons?

peterbarker:
On Tue, 12 Feb 2019, Dr Nic Williams wrote:

> Would you like the PR closed? Or want python3 installed in vagrant but
> aliased to python?

I really like that idea - using it in the Vagrant VM i.e

> Or would you link an initial PR that installs both python-pip +
> python3-pip packages, and performs the same setup in Vagrantfile for both
> pythons?

Nope - as you've pointed out, Python2 is going away, and it's more likely 
we'll be broken on Python3 than Python2.  So just changing us over to 
using Python2 on the VM - but accessed simply through "python" - sounds 
great.

OXINARF:
> So just changing us over to using **Python2** on the VM - but accessed simply through "python" - sounds great.

@peterbarker Did you meant Python3 there? Otherwise, I'm confused 😄

drnic:
I don't think ardupilot/dronekit code ecosystem is ready to be forced into python3 (e.g. https://github.com/dronekit/dronekit-python/issues/909)

So I propose we add `python3` and do a dual installation of components into `python` (python 2.7) + `python3` (currently python 3.7) and allow users/devs to pick their python. We can revisit removing python2 later.

peterbarker:
On Wed, 13 Feb 2019, Francisco Ferreira wrote:

>       So just changing us over to using Python2 on the VM - but
>       accessed simply through "python" - sounds great.
> 
> @peterbarker Did you meant Python3 there? Otherwise, I'm confused 😄

Yes, I did.  You'll have to get used to doing these sorts of translations 
;-)

OXINARF:
@drnic This Vagrant machine has nothing to do with Dronekit though (and Dronekit isn't an ArduPilot project, although quite related). If someone wants to have a Vagrant box for everything ArduPilot-related that's perfectly fine, but that's not what this file is providing.

drnic:
@OXINARF good point; I used it as the most recent example in my recent ardupilot/dronekit/python experiences where I'd seen something not working on python3 yet (for dronekit-python it was just pending a new release of the wheel). I still don't know all the ways devs + users of ardupilot might be using the Vagrantfile so I became more cautious for this PR 😄

python 2.7 must be one of software developments greatest success stories 😃

drnic:
I want to acknowledge that I've been regularly disorientated when discussing this PR - each time I come back to it I forget that its for the wiki/docs, not for ardupilot itself. That the Vagrantfile modified in this PR is only for building the wiki static site; whilst the docs changes in the PR are for end users who would be using a different Vagrantfile/Docker/non-containerized environment. I'm sorry about that.

I'm now going to fix this PR - it will just move the Vagrantfile to use python3; and not suggest any `python` -> `python3` changes in the documentation itself. I'm sorry about the confusion I've caused (myself at the very least).

peterbarker:
On Thu, 14 Feb 2019, Dr Nic Williams wrote:

> I want to acknowledge that I've been regularly disorientated when discussing this PR - each time I come back to it I forget that its for the wiki/docs,

Sounds like a standard day for me!

I'm glad someone is improving our Vagrant support and moving us along that 
Python path.

drnic:
Ok, changes made. I'm aware there are some non-vagrant use cases (`update.sh` looks to be called by someone's CI system), so I've tried to stick to only changing the Vagrant story to python3.

Look forward to learning more about how various parts of ardupilot are tested/released in Canberra next month.

drnic:
I'll close this for now. It can be reopened, or someone else can start the work again, when we're ready to move to py3. 👍

PR #1648, opened 2019-02-18
Update ros-slam.rst for RaspberryPi

soerili:
Giving the advice to run Hector SLAM on a different more powerful machine and not on the Pi.


rmackay9:
Merged, thanks!

PR #1651, opened 2019-02-24
Plane: update d-spoiler page

IamPete1:
Corrects typo, @tridge I think I have corrected it the right way? Thanks

### DIFF
--- a/plane/source/docs/differential-spoilers.rst
+++ b/plane/source/docs/differential-spoilers.rst
@@ -1,8 +1,8 @@
 .. _differential-spoilers:
 
-=====================
-Differential Spoilers
-=====================
+=========================================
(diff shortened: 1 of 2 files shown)


IamPete1:
This seems correct, however this does not result in the spoilers as outlined in the diagram, it gives outer up and inner down. I think this is aerodynamically correct to stop tip stall, so maybe the diagram needs a update. 

Also i think the d-spoiler weight parameters are have inner and outer swapped also, its all very confusing.

IamPete1:
This now updates the d-spoiler page to reflect https://github.com/ArduPilot/ardupilot/pull/10614 the pics still have the flap directions reversed pending getting some new pics

PR #1652, opened 2019-02-24
Added instruction to install the headers on Mojave

lvale:
Added instruction to install the headers on Mojave


rmackay9:
Merged, thanks!

PR #1654, opened 2019-02-26
Copter: Update Mavproxy commands in Guided Mode

rajat2004:
The ```speed``` and  ```yaw``` commands are not recognised by Mavproxy.
The current Mavproxy command for setting speed and yaw in Guided Mode are  ```setspeed``` and ```setyaw``` respectively.


rmackay9:
Thanks!

PR #1665, opened 2019-03-05
Update quadplane-flight-modes.rst

kd0aij:
description of new QACRO mode


rmackay9:
I think it would be better to put the QACRO specific description on a new QACRO specific wiki page or add it to the existing ACRO wiki page.  I think adding it to the general FlightModes page perhaps isn't the best place.. imagine if we add this level of detail for all flight modes to the general page?

kd0aij:
Thanks, I should have marked this WIP since I wrote that more for my own use than for users. I haven't yet installed any wiki editing tools or build environment on my new laptop, and didn't see a way to add new pages in the online editor.

kd0aij:
replaced by #1671

PR #1668, opened 2019-03-06
Update code-overview-adding-a-new-log-message.rst

rajat2004:
The DataFlash library has been renamed to AP_Logger which has broken most of the links in this page.
The code and the file names have also been updated


rmackay9:
Thanks for this.  One issue though is that the "master" has been replaced with a particular commit.. I think we should keep it pointing at master even if master changes and makes the links point to slightly incorrect lines

rajat2004:
I have changed the links to point towards the master. Please have a look at it

rmackay9:
Merged, thanks!

PR #1672, opened 2019-03-08
Update stabilize-mode.rst

rolyexpress:
Hello, I couldn't find any solution to wobbles on descent or any literature on i-term buildup. I think this is an important topic for the kind of birds usually built with ArduCopter and for PID tuning info in general.

### DIFF
--- a/copter/source/docs/stabilize-mode.rst
+++ b/copter/source/docs/stabilize-mode.rst
@@ -143,3 +143,5 @@ Common Problems
 -  sudden flips during flight.  This is nearly always caused by
    :ref:`mechanical failures <common-diagnosing-problems-using-logs_mechanical_failures>`
    of the motor or ESCs.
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks for the contribution.

We have the comment below on the stabilize page re wobbles when descending quickly.

> copter wobbles when descending quickly. This is caused by the copter falling through it’s own props wash and is nearly impossible to tune out although raising the Rate Roll/Pitch P values may help.

Tuning is tricky and I'm actually a little reluctant to include links to advice that hasn't come from @lthall (our control expert) or hasn't gone through some kind of verification from other developers.  We have a [Tuning section here](http://ardupilot.org/copter/docs/common-tuning.html) by the way which is perhaps where we should consolidate general rate tuning advice.

rmackay9:
Thanks again for the contribution but no feedback so closing..

PR #1674, opened 2019-03-10
Update code-overview-adding-a-new-mavlink-message.rst

rajat2004:
The file name and the link mentioned in this page for ArduCopter doesn't exist. But the names are still the same for other vehicles. Therefore adding the new file name for ArduCopter in addition to the old name to decrease confusion.


rmackay9:
Merged, thanks!  I hope we can make Rover consistent with Copter soon.  So Rover will also do most of the processing of commands within mode_auto.cpp like Copter does.

PR #1675, opened 2019-03-11
Multiple changes

OXINARF:
- Add me to commercial support list
- change ARACE UAS link in partners page as requested
- acknowledge Helpmonks for providing their service for free
- add D-Makers to schools (will add to partners as soon as we get logo); ~also remove LETU since link doesn't work anymore (I'm going to send them an e-mail asking for updated link/description and re-add if we get an answer)~


rmackay9:
Merged, thanks!

PR #1687, opened 2019-03-28
Update common-holybro-kakutef7aio.rst

Avalonnw:
Changed "InvenSense ICM20689 IMU (accel, gyro, compass)" to "InvenSense ICM20689 IMU (accel and gyro only, no compass)"


rmackay9:
Merged, thanks!

PR #1688, opened 2019-03-29
Add new drone

Dronotique:
Added new French DGAC Certified drone Alfonce H960 from Dronotique


rmackay9:
Merged, thanks!  By the way, the list is meant to be in alphabetical order so they can be moved much higher if you like.

rmackay9:
Ah, but i'm afraid there is no "buy" button on the website right?  This list is for "ready to use" and "easy to buy".  If there's no "Buy" button I'm afraid I'll need to remove the links from the page.  Sorry!

PR #1696, opened 2019-04-06
Dev: Extracted out JSBSim as a seperate simulator page

stephendade:
JSBSim is now has a dedicated page in the SITL section, making it easier to find, and keeping it consistent with all the other SITL backends.

### DIFF
--- a/dev/source/docs/simulation-2.rst
+++ b/dev/source/docs/simulation-2.rst
@@ -14,12 +14,13 @@ Simulation allows for safe testing of experimental code and settings and crashin
 
 The most commonly used simulators are:
 
--  :ref:`SITL (Software In The Loop) <sitl-simulator-software-in-the-loop>` is the simulator most commonly used by developers.  It is used by the :ref:`autotester <the-ardupilot-autotest-framework>` and other simulators below are actually built on top of SITL
+-  :ref:`SITL (Software In The Loop) <sitl-simulator-software-in-the-loop>` is the simulator most commonly used by developers. It is a simple simulator that is built within all SITL builds of ArduPilot. It is used by the :ref:`autotester <the-ardupilot-autotest-framework>` and other simulators below are actually built on top of SITL
 -  :ref:`Gazebo <using-gazebo-simulator-with-sitl>` is the official DARPA virtual robotics simulator
 -  :ref:`XPlane-10 <sitl-with-xplane>` a commercial flight simulator with a rich 3D interface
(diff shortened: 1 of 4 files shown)


dev/source/docs/sitl-with-jsbsim.rst
```diff
@@ -0,0 +1,41 @@
...
+
+In order to build JSBSIM, you need Cmake, install it with :
+
+::
+
+    sudo apt-get install cmake
+
+In the same directory (your home directory) run these commands:
+
+::
```
peterbarker: We still have `Tools/scripts/build-jsbsim.sh` - why aren't we pointing to it here, I wonder...

dev/source/docs/simulation-2.rst
```diff
@@ -14,12 +14,13 @@ Simulation allows for safe testing of experimental code and settings and crashin
...
+-  :ref:`SITL (Software In The Loop) <sitl-simulator-software-in-the-loop>` is the simulator most commonly used by developers. It is a simple simulator that is built within all SITL builds of ArduPilot. It is used by the :ref:`autotester <the-ardupilot-autotest-framework>` and other simulators below are actually built on top of SITL
 -  :ref:`Gazebo <using-gazebo-simulator-with-sitl>` is the official DARPA virtual robotics simulator
 -  :ref:`XPlane-10 <sitl-with-xplane>` a commercial flight simulator with a rich 3D interface
 -  :ref:`RealFlight <sitl-with-realflight>` a commercial flight simulator with a rich 3D interface and ability to design custom vehicles
 -  :ref:`Morse <sitl-with-morse>` a robotics simulation environment commonly used in research
--  :ref:`Replay <testing-with-replay>` has no graphical interfacde but allows re-running master from a dataflash log
+-  :ref:`Replay <testing-with-replay>` has no graphical interface but allows re-running master from a dataflash log
```
peterbarker: We're trying to avoid "dataflash" now-adays.  More "onboard-log"... preexisting, of course, so no need to fix it here.

dev/source/docs/sitl-with-jsbsim.rst
stephendade: Added ref

dev/source/docs/simulation-2.rst
stephendade: That was just me fixing up a typo in "interface". Will need a different PR to fix all the dataflash -> onboard-log refs.

Also - is there a compendium somewhere of what "official" words/terms to use in the wiki, like the above example and other things like using "Ardupilot" and not "APM"?

peterbarker: Not that I know of.

PR #1700, opened 2019-04-07
Rover: update aux function page

IamPete1:
This updates rover aux functions page from the old Ch7 ect opts to RCx_OPTIONS.

Also I lengthened two underlines that were giving build warnings on the windvane page


rmackay9:
Wonderful, txs!

rover/source/docs/rover-auxiliary-functions.rst
```diff
@@ -4,109 +4,93 @@
...
-   </tr>
-   <tr>
-   <td>11</td>
-   <td><strong>Guided</strong></td>
-   <td>Change to Guided mode
-   </td>
-   </tr>
-   </tbody>
-   </table>
++----------------------+----------------------------+
```
peterbarker: I wonder if we should just link to the parameters page so this doesn't get out of date?

IamPete1: probably, I was planning to do a more in-depth explanation of each thing, but I realised I was not really sure what they all did.

PR #1703, opened 2019-04-09
Dev: Change git push command in git-rebase.rst

rajat2004:
```git push origin master``` is more advisable, especially for beginners to Git for whom this is intended.

It won't push to all the branches which I think is safer, and without any warnings which just ```git push``` gives.
This anyways won't affect advanced users who know what they are doing and will know eactly what command to use

### DIFF
--- a/dev/source/docs/git-rebase.rst
+++ b/dev/source/docs/git-rebase.rst
@@ -26,11 +26,17 @@ the "Git Shell" or "Git Bash" utility that should already be :ref:`installed <gi
 
        cd <ardupilot-path>
 
-#. Starting from here we will assume that you want to rebase you master branch on ArduPilot master branch. This process is the same on all branch, just replace master by your branch name. Ensure you are looking at your master branch or the branch that you want to rebase.
+#. Starting from here we will assume that you want to rebase your topic-branch on ArduPilot master branch. This process is the same for the master branch, just replace topic-branch with master. Ensure you are looking at your master branch or the branch that you want to rebase.
 
    ::
 
-       git checkout master
+       git checkout topic-branch
+
+   If you want to create a new branch to work on which you would like to rebase on top of the Ardupilot master branch,
+
+   ::
+
+       git checkout -b topic-branch
 
 #. Ensure your repository is connected to the upstream repository you
    forked from.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by peterbarker** [comment]
This page doesn't really suggest good work practise, and unfortunately the proposed change here makes it worse.

People should almost never be committing work into their own "master" branch.  They should always work on a "topic" branch, and be rebasing that on top of either their own master branch or upstream's master branch.

We're missing steps like, "Create and check out a new topic branch: `git checkout -b topic-branch`.

Your patch then becomes `git push origin topic-branch`

https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows

**Comment by rajat2004** [comment]
@peterbarker I agree with you that everything should be done on a topic branch and not on the master branch. But it has been specifically mentioned in the page regarding the branch which is to be rebased. The following is from the [wiki page](http://ardupilot.org/dev/docs/git-rebase.html):-

Starting from here we will assume that you want to rebase your master branch on ArduPilot master branch. This process is the same on all branch, just replace master by your branch name. Ensure you are looking at your master branch or the branch that you want to rebase.
```
git checkout master
```

This is the reason that I kept the branch as ```master``` since anyone who wants to rebase their own topic branch can just replace master with it as mentioned. This would keep the change consistent with the Wiki page

Would it be better to change all the commands from ```master``` to ```topic-branch``` in the wiki page like you said?

peterbarker:
On Tue, 9 Apr 2019, Rajat Singhal wrote:

> Would it be better to change all the commands from master to topic-branch in
> the wiki page like you said?

I would say "definitely"!

Do you thnk we should take this to a devcall to see if there's consensus 
before you do the work, 'though?

rajat2004:
If you think that's the proper way to go about this. I'm not sure if this is important enough for that

rajat2004:
Should I make the changes first and then see if everyone is fine with it @peterbarker?

peterbarker:
On Sat, 13 Apr 2019, Rajat Singhal wrote:

> Should I make the changes first and then see if everyone is fine with it @peterbarker?

I'll ask to discuss it on the devcall on Tuesday (my time...)

Ta,

Peter

rajat2004:
I have made the changes @peterbarker. Please have a look and see if the updated changes make it better.
After everything is fixed and approved, I'll squash everything to a single commit

rajat2004:
Any updates? Does it require any more changes

peterbarker:
@rajat2004 We didn't get to discuss it, but I'm merging it anyway - thanks!

PR #1711, opened 2019-04-17
Update common-esp8266-telemetry.rst

Rusty105:
Added the initial access point ID and initial password so users don't have to guess at the password.


rmackay9:
Thanks!

Rusty105:
No problem, was setting one of these up, and it took a few to remember most pw seem to be "ardupilot"

PR #1717, opened 2019-04-19
Add JSBSim executable to PATH

rajat2004:
Without setting the environment variable, JSBSim doesn't start
jsbsim

### DIFF
--- a/dev/source/docs/sitl-with-jsbsim.rst
+++ b/dev/source/docs/sitl-with-jsbsim.rst
@@ -33,6 +33,12 @@ In the same directory (your home directory) run these commands:
     cmake -DCMAKE_CXX_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DCMAKE_C_FLAGS_RELEASE="-O3 -march=native -mtune=native" -DCMAKE_BUILD_TYPE=Release ..
     make -j2
 
+Add JSBSim executable to your search path. Add the line at the end of your ".bashrc" in your home directory to automatically set the path whenever a new terminal is opened.
+
+::
+
+    export PATH=$PATH:$HOME/jsbsim/build/src
+
 If using the JSBSim plane simulator you can specify a different JSBSim
 model than the default Rascal110 by specifying the model name using the
 -f parameter to sim_vehicle.py, like this:


rajat2004:
I actually didn't use the script for the installation but just checked it right now, it's not being added to the path - https://github.com/ArduPilot/ardupilot/blob/master/Tools/scripts/build-jsbsim.sh

This should to be added to the script as well

peterbarker:
On Fri, 19 Apr 2019, Rajat Singhal wrote:

> I actually didn't use the script for the installation but just checked it right now, it's not being added to the path -
> https://github.com/ArduPilot/ardupilot/blob/master/Tools/scripts/build-jsbsim.sh
> 
> This should to be added to the script as well

No - a script called "build-" should not be modifying your build 
environment in a permanent way.  "install-*" find, "build-*" not-so-fine. 
It can certainly suggest you do so...

So it would be OK for install-prereqs to do it, for example.

rajat2004:
Ok, so the script after building JSBSim can tell the user to export the path but not actually do it in the script itself.
Is that fine @peterbarker?

Even then, this PR right now is just for adding in the Wiki that the path needs to be set before using JSBSim. I'm not modifying the script, was just suggesting that this could be added to it also.
Should I add another line saying to add it to `~/.bashrc` to set it automatically?

peterbarker:
On Fri, 19 Apr 2019, Rajat Singhal wrote:

> Ok, so the script after building JSBSim can tell the user to export the path but not actually do it in the script itself.
> Is that fine @peterbarker?

Yes.  It could actually check the path first before whinging about it if 
you were keen :-)

> Even then, this PR right now is just for adding in the Wiki that the path needs to be set before using JSBSim. I'm not modifying the script, was just
> suggesting that this could be added to it also.

Yes, and I think this PR could be merged now.

> Should I add another line saying to add it to ~/.bashrc to set it automatically?

If you like :-)

rajat2004:
> Yes. It could actually check the path first before whinging about it if you were keen :-)

That can be done, but the path wouldn't have been added anytime before this script is ran, right?
But definitely, this will be required if someone is doing a fresh install of JSBSim multiple times

> Should I add another line saying to add it to ~/.bashrc to set it automatically?
> If you like :-)

I'll do that

rajat2004:
It's ready to be merged now

peterbarker:
Merged, thanks!

PR #1718, opened 2019-04-21
Update to using the install-prereqs-ubuntu.sh

dyyap:
added to change permissions use chmod

### DIFF
--- a/dev/source/docs/building-setup-windows10.rst
+++ b/dev/source/docs/building-setup-windows10.rst
@@ -44,9 +44,11 @@ Setup steps
     - ``git submodule init``
     - ``git submodule update --recursive``
 
-#.  run the Tools/scripts/install-prereqs-ubuntu.sh script:
+#.  run the Tools/scripts/install-prereqs-ubuntu.sh script
 
     - run the script ``./Tools/environment_install/install-prereqs-ubuntu.sh``.  You will be asked for your Ubuntu root password which is the password provided in step 1 (above).  Respond with "Y" when it asks if you wish to install a package.
+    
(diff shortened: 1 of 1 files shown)


magicrub:
why would you not have access to a repo you just checked out? Can you describe the scenario?

dyyap:
I used windows 10 WSL and followed tre instructions up to the point of using the script however it only had read and one write privilege no executable

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

peterbarker:
Sorry, this is the incorrect fix.  The scripts should be executable, but we lost that when they were moved, sadly.  Do you feel like doing a PR against master making these scripts in the environment_install directory executable?

What would be welcome is a patch to correct the path to the script here - it's correct in one spot but wrong in another...

@stephendade for visibility

dyyap:
I thought that maybe some of the scripts were meant for internal use in some other scripts, when I just changed permissions for one script I did not know if the other scripts wanted executable permission as well. Other than that permission the other steps work fine.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

peterbarker:
On Sun, 21 Apr 2019, dyyap wrote:

> I thought that maybe some of the scripts were meant for internal use in some other scripts, when I just changed permissions for one script I did not know
> if the other scripts wanted executable permission as well. Other than that permission the other steps work fine.

Uh... so is that a yes or no to PRing a set of chmod+x'd scripts? :-)


Peter

dyyap:
Yes

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

stephendade:
FYI, the scripts have been updated in ardupilot to have the executable permission.

tridge:
thanks, I fixed the conflict and added your chmod update

PR #1722, opened 2019-04-29
common-dshot: pass-thru support on blheli_32 only

vierfuffzig:
points out that pass-through currently is supported on blheli_32 only to avoid confusion.
see https://discuss.ardupilot.org/t/blheli-s-passthrough/36346/17 for reference.


rmackay9:
Merged, thanks!

WickedShell:
Did something change? I originally used this with BLHeli_S units....

vierfuffzig:
https://github.com/ArduPilot/ardupilot/issues/9730

Issue #1223, opened 2018-04-28
update.py target directory fixed


fnoop:
COPY_TARGET_DIR_BASE is fixed path, would be nice to be able to set this as a parameter

fnoop:
Also, update.py doesn't create /var/sites/wiki/web if it doesn't exist, so install fails

auturgy:
natively, or inside vagrant?

fnoop:
Natively

auturgy:
Ah. Hmmm.
For better or worse, the build environment is designed to be run inside the vagrant vm.
I’ll defer to @peterbarker for comment I think.

peterbarker:
Extra parameters won't hurt.

fnoop:
PR raised

Hwurzburg:
`update.py` now takes a `--destdir` argument and auto-creates missing target directories.

Feel free to reopen if you feel this has not been addressed.

PR #1731, opened 2019-05-02
Add --destdir argument to replace static COPY_TARGET_DIR_BASE, #1223

fnoop:
COPY_TARGET_DIR_BASE is a static path.  Instead add a `--destdir`parameter that allows this to be specified, which defaults to the existing '/var/sites/wiki/web'.  This allows the wiki to be built in other environments.


fnoop:
PR for issue #1223

peterbarker:
Merged, thanks!

PR #1743, opened 2019-05-14
Added Bask Aerospace Vehicles

d-skinner:
Added link to AeroDrone MR4

### DIFF
--- a/common/source/docs/common-rtf.rst
+++ b/common/source/docs/common-rtf.rst
@@ -21,6 +21,7 @@ Vehicles from Partners
 * `mRobotics Nano Talon (fixed wing) <https://store.mrobotics.io/ProductDetails.asp?ProductCode=mRo-talon0318-mr>`__
 * `SkyRocket's Journey <http://sky-viper.com/journey/>`__
 * `TT Robotix Rhino 6x6 <http://www.ttrobotix.com/product/rhino6x6>`__ and `Base 1 Rover <http://www.ttrobotix.com/product/base1rover>`__
+* Bask Aerospace - `AeroDrone MR4 <http://www.baskaerospace.com.au/aerodrone/mr4>`__
 
 Vehicles from Non-Partners
 ==========================


rmackay9:
Txs for this.  Normally we've asked that there be a "buy" button or at least a price on the web page before we add it.. I see there is at least one exception to this which is the TT vehicles.

I'd like to hear feedback from other devs on whether we want to keep this requirement..

auturgy:
I think that so long as it actually is available for purchase, it shouldn’t matter whether there is a “Buy” button or public pricing.

rmackay9:
Yes, I tend to agree that perhaps we were being too strict.  Merged!

peterbarker:
OK, I'm going to say that if you can't click "Buy Now" and get  tracking number then the vehicle isn't "Ready To Fly" and thus doesn't belong on this page.

(notionally) Having a link to the "Buy" page has the advantage that we know when the vehicle is no longer available for purchase - because somebody notices that the link no longer works.

auturgy:
That’s fine for toys and consumer products. Less so for industrial machines.

Regards,

James

d-skinner:
Thanks for merging this in. @peterbarker I understand where you're coming from but I have to disagree that "Ready to Fly" means it's ready to fly the second you click "Buy Now". I highly doubt that anyone believes that. As I see it "Ready to Fly" means that when it arrives on your doorstep it will be ready to take out of the box and fly. It seems your thinking "Ready to Ship". Having said that, I agree that it's important to ensure we don't end up with "dead" links on the ArduPilot site. I'll personally make sure our vehicles are kept up-to-date to minimise this issue.

peterbarker:
On Tue, 14 May 2019, David Skinner wrote:

> Thanks for merging this in. @peterbarker I understand where you're coming from but I have to disagree that "Ready to Fly" means it's ready to fly the
> second you click "Buy Now". I highly doubt that anyone believes that. As I see it "Ready to Fly" means that when it arrives on your doorstep it will be

By your definition "custom designed to meet your exact needs, 1 month 
leadtime, enter contractual negotiations by clicking here" counts as 
"Ready To Fly"?

Please don't get me wrong here - I love having our partner's products 
visible on our site.  It helps showcase the variety of vehicles that 
ArduPilot at the very least!  I'm just hopeful we can keep "Ready To Fly" 
as close to what "Ready To Fly" means when you go to J.Random.Webstore.

PR #1737, opened 2019-05-05
Create common-partners-program

auturgy:
Moved partners program description to be separate from logos page (the logos were preventing people seeing the text - with more than 50 partners, it got pushed way down).
Borrowed some text from https://www.open-tx.org/newradios.html
A follow up PR removing the relevant text from the existing page is also provided: merge this one first.

PR #1745, opened 2019-05-14
Split Partners page and change logo list into table

OXINARF:
The PR might be divided into three topics:

- the split into 2 pages (started by @auturgy on #1737/#1738 that this superseeds)
- converting list of partner logos to be a table with custom CSS
- other fixes

Regarding 1st: please check the language and whether you think the split is good (the page with instructions is just linked once in the normal page; I have my doubts on the effectiveness of this, but the old end of page text didn't work either)

Regarding 2nd: note that because this is now a table, we need to be more careful when adding and removing partners since we need to keep a specific format - namely:
- this is a table of two columns so every two lines there needs to be an `*` to signal a new line
- a line can't have only one element, in that case, an empty `-` element needs to be added

Regarding 3rd: fixes donation page to point to wiki partners page (no need to get out of the wiki, plus website version doesn't work correctly on mobile) and a fix that is a different way of my PR https://github.com/ArduPilot/sphinx_rtd_theme/pull/8 (which will be closed by this)

Anyone is invited to test (in multiple devices) a temporary deployed version at http://oxinarf.pt/wiki/ardupilot/docs/common-partners.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,16 @@
 /antennatracker/source/docs/common-*
 /dev/source/docs/common-*
 
+# Common CSS except under Common
+/ardupilot/source/_static/common_*
+/copter/source/_static/common_*
+/plane/source/_static/common_*
+/rover/source/_static/common_*
+/planner/source/_static/common_*
+/planner2/source/_static/common_*
+/antennatracker/source/_static/common_*
+/dev/source/_static/common_*
+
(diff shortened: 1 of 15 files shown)


jpkh:
for me, it looks just fine. our concrete5 partners' page creation script might have small issues but it should be fairly easy to fix after this PR is in master.

PR #1749, opened 2019-05-27
[fix] Google groups link updated to forums link.

brunoolivieri:
Fix on the main index.html.

(rendering checked with the recommended Vagrant environment)


auturgy:
Thanks!

Issue #1435, opened 2018-10-02
Document waypoint 0


khancyr:
Currently we make no mention of waypoint 0 beeing the home. We should document that

cursed4ever:
I want to work on this

brunoolivieri:
As the PR #1759 discussed this point and it is closed, could this issue be closed?

Hwurzburg:
discussed and decided not to change docs

PR #1759, opened 2019-06-01
common-planning-a-mission-with-waypoints-and-events.rst: implement #1435

vierfuffzig:
ping @khancyr

### DIFF
--- a/common/source/docs/common-planning-a-mission-with-waypoints-and-events.rst
+++ b/common/source/docs/common-planning-a-mission-with-waypoints-and-events.rst
@@ -21,7 +21,7 @@ Setting the Home Position
 For **Copter**, **Plane** and **Rover** the home position is set as the location where the vehicle
 was armed. This means if you execute an RTL, it will return to
 the location where it was armed, so arm your vehicle in the location you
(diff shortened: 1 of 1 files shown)


rmackay9:
I wonder if this one is useful.. users's can't really set the 0th waypoint anywhere.. I'm not sure if they really need to know that it's the 0th waypoint..

vierfuffzig:
added comment about this not being user settable

khancyr:
@rmackay9 you can set it with custom IHM (or with ROS), or when loading a mission file. Easy mistake when you don't know MAVLink protocol ...

rmackay9:
@khancyr, Not sure what an IHM is but in any case, using ROS isn't a very common use case for users and when loading with a mission file I'm not sure it's ever used.  I think I'll close this if that's OK.

vierfuffzig:
@rmackay9 @khancyr the primary intention of this PR was to adress https://github.com/ArduPilot/ardupilot_wiki/issues/1435 
i'm perfectly fine with this approach being rejected, i'd appreciate alternative suggestions though.

Issue #1742, opened 2019-05-11
Battery Failsafe


AndKe:
[Battery Failsafe](http://ardupilot.org/copter/docs/failsafe-battery.html) - Describe the issue/suggestion and improve the title. Please keep a link to the original article if relevant.

Please describe behaviour if landing in RTL, or LAND mode and the battery failsafe triggers, will it switch back to RTL, climb again, etc ?
Same goes for other failsafe during LAND.

Naterater:
https://github.com/ArduPilot/ardupilot/issues/10320

The craft will continue to land.

I've made the change and proposed it here: https://github.com/ArduPilot/ardupilot_wiki/pull/1762

rmackay9:
@AndKe, think this issue is closed now?

brunoolivieri:
PR was merged. I do believe that can be closed.

PR #1762, opened 2019-06-02
Failsafe clarification during land or RTL.

Naterater:
Solves https://github.com/ArduPilot/ardupilot_wiki/issues/1742


rmackay9:
A small issue with the new wording though is that say it's Landing but the battery failsafe action is set to RTL, then it will actually RTL...

AndKe:
Yes, the PR makes it clear.

rmackay9:
OK, I bow to your better judgement, merged, thanks!

Issue #1432, opened 2018-09-27
SITL Simulator (Software in the Loop)


pkantue:
[SITL Simulator (Software in the Loop)](http://ardupilot.org/dev/docs/sitl-simulator-software-in-the-loop.html) 
Hi, for the sake of clarity, it should be stated that the physics simulation is now using the SIM_Multicopter.cpp instead SIM_Multicopter.py as illustrated on the figure (I'm using Copter 3.4.6). If I had access to the figure, I could modify it :).

Maybe it's just me being pedantic, but I had a hard time realizing where to modify the physics simulation code for my research.

Cheers!

Issue #911, opened 2017-08-09
SITL Architecture Image


hugo-andrade:
Hi all,

I made a figure to my paper to illustrate the SITL Architecture in ArduPlane, In this version of figure in Wiki, the connections are deprecated in old versions. 

I think the image that I made is pretty good, if you want to use in Wiki.

SITL Architecture
------
![sitl](https://user-images.githubusercontent.com/6294519/29145165-5a01d964-7d32-11e7-8f80-f602d297c1ec.png)

Thanks,

[SITL Simulator (Software in the Loop)](http://ardupilot.org/dev/docs/sitl-simulator-software-in-the-loop.html)

gmorph:
Similar to the picture on this page
http://ardupilot.org/dev/docs/sitl-simulator-software-in-the-loop.html#sitl-simulator-software-in-the-loop

hugo-andrade:
Similar, but the ports and connections type in ArduPlane <-> JSBSim, Serial1 and Serial3 are deprecated in this picture. To develop my work using SITL in ArduPlane, I mapped the connections, cause this picture in link were confusing me.

hamishwillee:
So just to be clear, you've removed serial connections from the image because they are deprecated/no longer used in the platform?

I think the image looks really "clean" - which is important for readability.

My concern is that is loses information in the labeling - APM planner is just one GSC that could be used here.  Similarly this is ArduPlane/JBSIM specific, but nice to have images that cover all platforms (like the one Grant points to).

hugo-andrade:
Hi @hamishwillee, I expressed myself wrong, sorry.

I made this figure to mapping the network structure of ArduPlane SITL comunnication  (UDP and TCP data only). What I saw when I mapped the connections, is some connection shown in the wiki figure don't fit reality, like the UDP connection ArduPlane <-> JSBsim, and the open TCP 5762 that dont exists in wiki figure, only small issues in my opinion.

I agree with you about the specificity when I put APM planner as GSC and JSBSIM as FDM. This figure is the diagram that I used to solve my SITL problem, but if you like and want, I can generalize the image, changing the names and adding the serial connection labels to be used in wiki.

PR #1764, opened 2019-06-02
dev sitl-simulator-software-in-the-loop: update illustration

vierfuffzig:
implements #1432


khancyr:
Hello, thank for trying to fix this. But the new version is not true. We don't have udp connexion between sitl default backend and SITL vehicle code ...
I got new illustration for SITL, I need to find some time to submit them

vierfuffzig:
thanks @khancyr maybe just up and link your image somwhere and i'll submit and PR to wiki if that helps?

brunoolivieri:
Hi @kd0aij 

Did you find the illustration? 

If not, please, which UDP connection is not correct? Could you point here on http://www.olivieri.com.br/downloads/ArdupilotSoftwareintheLoopSITL.jpg ?

kd0aij:
Did you mean me?

brunoolivieri:
My bad! Sorry @kd0aij  

I mean @khancyr , please, which UDP connection is not correct? Could you point here on http://www.olivieri.com.br/downloads/ArdupilotSoftwareintheLoopSITL.jpg ?

peterbarker:
@khancyr was referring to the edges for port 5502 and 5501.

ArduPilot sends to JSBSim on port 5505 and received on 5504 (https://github.com/ardupilot/ardupilot/blob/master/libraries/SITL/SIM_JSBSim.cpp#L63)

SIM_Aircraft and friends are not currently external to ArduPilot - we interact in a rather more direct manner with them.

ArduPilot sends directly to FlightGear on port 5503

I've no idea what the extra 5501 edge is from "mavproxy on telemetry port" to ArduCopter.

Having the port 14550 coming from the MAVProxy on port 5763 is quite possibly the case - but most of the time it would be from a MAVProxy on 5760.

Having the source ports and destination ports present in these diagrams may not be accurate - I'd have to research a bit to determine that (so I suggest they could stay at least in the short term).

There are several interesting bits missing, too - where our newer simulation backends plug in, for example.

rajat2004:
Another thing, just pointing it out, most of the newer SITL interfaces (at least all the ones implementing the `set_interface_ports` function) use ports 9001 & 9002 by default
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_HAL_SITL/SITL_cmdline.cpp#L184

brunoolivieri:
Thanks!

So it is not just a matter of a small cut/paste. I will try to create something from zero and propose.

rajat2004:
This one might be interesting - https://github.com/ArduPilot/ardupilot_wiki/issues/911

khancyr:
I will provide some graph I have already done tomorrow.

brunoolivieri:
Thanks!

Hwurzburg:
@khancyr Pierre, still waiting on the graph...

albcem:
The long awaited graph would indeed be helpful but may be there is a more direct answer to the problem I am dealing with:

I am trying to implement an upgrade for ArduPlane SITL with Jsbsim and Flightgear. The improved feature is visible moving control surfaces during the SITL simulation and some added RC channels.,I went as far as making it work but then somehow the terrain interaction broke: the aircraft initiates below the terrain. 

Here is a link to the thread explaining better what I have done: 

https://discuss.ardupilot.org/t/sitl-sim-vehicle-py-aircraft-starts-below-terrain/58369.

Any insight will be greatly appreciated.

Issue #1420, opened 2018-09-17
Replace Tarot as recommended Gimbal


hamishwillee:
The 3DR Tarot Gimbal is no longer manufactured (it is still available, but this will be temporary). I have therefore archived the article in #1409) and cleaned up links.

The Tarot was previously the "recommended" gimbal. This issue is a placeholder requesting clarification of what is recommended  now (and why) and also for information on other gimbals for which documentation might be useful.

auturgy:
@jmachucca77
This might be a candidate for code-in?

Regards,

James

rmackay9:
the tarot gimbal is still available though.  I'm not sure it's really the recommended gimbal anymore but it is still available and it's pretty reliable.

hamishwillee:
It is available in theory, but only from Amazon. As soon as those stocks go. Gone. I archived it as suggested here: https://github.com/ArduPilot/ardupilot_wiki/pull/1409#issuecomment-420122661

We can undo if you strongly disagree - I did wait 6 days for further comment.

jmachuca77:
Writing the wiki page might be a candidate for code-in... but the problem is what hardware should it describe, plus not sure these students will have a budget to buy a gimbal and necessary hardware to actually test it.

rmackay9:
I think we can close this now..

PR #1767, opened 2019-06-02
common-cameras-and-gimbals.rst: remove dead link

vierfuffzig:
tarot website down, linked hardware item outdated


auturgy:
Thanks

vierfuffzig:
thanks @auturgy , maybe #1420 can be closed now?

rmackay9:
hold on.. did we just remove the tarot gimbal page?  The tarot gimbal is pretty commonly used I think.

rmackay9:
I've restored the Tarot 2D gimbal.  This is actually the 2nd time I've had to do this so let's not delete it a 3rd time ok? :-)

auturgy:
It’s a dead link.
 that gimbal was relevant in 2013, but isn’t available in 2019. Better that it’s retired.

Regards,

James

rmackay9:
If someone wants to upgrade the page to the more recent versions of the Tarot gimbals then that's great but I think the page still has some value.

auturgy:
This page exists because 3DR used to sell them. It was never a gimbal we should actually have been recommending over others. I don’t think we should update it: I really think this page should go.

rmackay9:
@auturgy, I'm afraid I can't agree.. There are no other Ready-to-use brushless gimbals on our camera/gimbals wiki page.  Removing the one example that we have, whether it's perhaps more difficult to find than it once was or not isn't a great idea I think.

vierfuffzig:
@rmackay9 would you agree to have the site point out that this is an illustrative sample and have the dead links and references removed or replaced by local backup copies wherever possible?

rmackay9:
@vierfuffzig, sure, that sounds good.

Issue #1714, opened 2019-04-17
Using Gazebo Simulator with SITL


ankitgc1:
[Using Gazebo Simulator with SITL](http://ardupilot.org/dev/docs/using-gazebo-simulator-with-sitl.html) - Describe the issue/suggestion and improve the title. Please keep a link to the original article if relevant.

sim_vehicle.py: command not found

brunoolivieri:
Hi, you could try to refer the entire path: ../Tools/autotest/sim_vehicle.py  OR add it on your $Path.

In any case, I did a PR to insert the path on the documentation.

I believe that this issue could be closed.

Regards,

Hwurzburg:
closing...resolved

PR #1771, opened 2019-06-03
[minor enhance] Adding path to the example

brunoolivieri:
Adding sim_vechicle.py path due questions as in the issue #1714


rmackay9:
Thanks!!

PR #1777, opened 2019-06-05
common-logs.rst: fix the links

tatsuy:
image


rmackay9:
oh my, thanks for fixing this!  It means the items don't appear on the left side menu but maybe that's OK.  Certainly this is an improvement!

PR #1779, opened 2019-06-05
companion-computers.rst: remove LYCHEE

tatsuy:
this topic isn't include on this wiki.


khancyr:
Hello, 

This is here : http://ardupilot.org/dev/docs/companion-computers.html so it is fine !

tatsuy:
I felt strange that there is an external link on the side menu.
But it's ok. thanks.

khancyr:
yep, but that isn't the only one, so it fine for now. And I really don't  like the raspberry pi page, that isn't maintained ...

tatsuy:
ah, Maverick is too.
I see.

PR #1783, opened 2019-06-09
plane-configuration-landing-page: link to updated arming page

vierfuffzig:
currently referenced arming-throttle page contains outdated version-specific information, unfortunate advice ("ARMING_CHECK parameter should probably be left at 0 when at the an airfield without a ground control station") and is a duplicate of the more up-to-date arming-your-plane page. 
i suggest to link arming-your-plane here and retire arming-throttle


peterbarker:
I agree we should eliminate the throttle-arming page, but we should make sure we get any non-common content over first.

tridge:
I agree on combining the two

tridge:
but arming-your-plane is more recent, so merging this change

Issue #1710, opened 2019-04-16
Companion Computers few page out of date


bys1123:
1. Disable bluetooth on RPi3 should use `dtoverlay=pi3-disable-bt` in /boot/config.txt instead of `dtoverlay=pi-disable-bt` descripted in wiki.
http://ardupilot.org/dev/docs/raspberry-pi-via-mavlink.html#testing-the-connection
2. Look like Config page is work on 80 port with apweb, (8000 port is only for video?)
http://ardupilot.org/dev/docs/apsync-intro.html#simple-configuration-web-page

bys1123:
I saw Randy mentioned apweb on #1147

brunoolivieri:
Hi @bys1123 

I did the simple PR #1786 to fix the BT command and PR #1785 to insert item "2" in the Wiki.

Could this issue be closed?

rmackay9:
@brunoolivieri, thanks for all your efforts on the wiki, txs!  i'll close this now.

bys1123:
Thanks!

Issue #1147, opened 2018-03-30
APSync intro should explain that there are two web configuration pages included


rmackay9:
APSync intro: http://ardupilot.org/dev/docs/apsync-intro.html

These two web pages can be accessed to configure APsync:

http://10.0.1.128:8000 <-- video streaming on/off button
http://10.0.1.128 <-- apweb (cut down version of what's in a skyviper)

brunoolivieri:
I did the simple PR #1785 to insert this information in the wiki.  Could this issue be closed?

rmackay9:
yes, closing thanks!

PR #1785, opened 2019-06-10
[minor add] Simple information insertion due to issue #1147 and issue1710 part.2

brunoolivieri:
APsync IP and port for connections as informed in issue #1147 and argued in issue #1710 item 2.

Render verified on Vagrant.


rmackay9:
Merged, thanks!

PR #1795, opened 2019-06-12
[minor fix] Wiki edit guide command

brunoolivieri:
I had problems with the provided command parameters. So, as I am using the updated Git version, I am suggesting some adjusting for the command that works for me in may PRs.

### DIFF
--- a/common/source/docs/common-wiki-editing-submitting-changes.rst
+++ b/common/source/docs/common-wiki-editing-submitting-changes.rst
@@ -56,7 +56,7 @@ There are a number of way of using Git/Github - including a number of different 
 
    .. code-block:: bash
 
-       git push my_well_named_branch
+       git push --set-upstream origin my_well_named_branch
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
git push -u origin my_well_named_branch
 works the same also

brunoolivieri:
@Hwurzburg thanks.

Regarding the original command "git push my_well_named_branch": Should it work? (Did not work for me)

Make sense to maintain the PR or close it?

Hwurzburg:
I think a little expansion is indeed needed, but I am no Git expert...
_git push my_well_named_branch_   will push the current branch to a branch named my_well_named_branch in the origin, if it exists and is tracked....
it isn't initially....so you use git push -u origin my_well_named_branch to create the branch in origin and set the local branch to track it...after that I think you can just use git push, it will now use the origin as destination and assume you are pushing to a branch in it that is named the same as the current local branch...

khancyr:
we don't ask the -u or --set-upstream on other git help. Normally, if git don't have previous branch on origin, it will say you to use --set-upstream.
I think that for wiki editing, it could be worth it to merge this as most contributor won't look at what git does ....

Hwurzburg:
Thanks Bruno/Pierre....

PR #1796, opened 2019-06-14
Add description of VTOL_LAND fixed wing approach option

Hwurzburg:
@WickedShell Michael, there are a couple of embedded questions for you in the text.I hope I got most of it right... thanks


plane/source/docs/quadplane-auto-mode.rst
```diff
@@ -55,6 +55,13 @@ will be done at whatever height the aircraft is at when it starts on
 the NAV_VTOL_LAND waypoint. So you would typically want the previous
 waypoint to have an altitude of about 20 meters above the ground.
 
+If Q_OPTIONS is set for "Use a fixed wing approach" (bit 4), then instead of transitioning to VTOL flight and doing a VTOL landing, it will remain in plane mode (question for WickedShell: what if in VTOL mode already? transition to plane?), climb or descend to the altitude set in the NAV_VTOL_LAND waypoint. When it reaches within Q_FW_LND_APPROACH_RAD of the landing location, it will perform a LOITER_TO_ALT down (question for WickedShell: what if its lower than this alt already by mistake?) to Q__LND_FINAL_ALT, turning into the wind, transition to VTOL mode and proceed to the landing location and land.
```
WickedShell: The loiter to altitude is an artifact of me actually explicitly planning a loiter to altitude first. Without it the vehicle will simply enter a loiter at the planned landing location and climb or descend (both enroute and at the destination point) until it is at the requested altitude, and on the radius.

Worth noting that if Q_FW_LND_APPROACH_RAD isn't set we will use the vehicles normal loiter radius.

From a tuning perspective the Q_FW_LND_APR_RAD (note APR not APPROACH) to be specified to be Larger then the maximum detransition distance on a no wind day + the space it takes the aircraft to turn into the landing.

Q_LAND_FINAL_ALT is only a control on when to slow the vertical descent rate down, it will behave exactly the same. In practice I don't see an impact, as I never fly a large loiter that close to the ground due to ground obstacles/unexpected interference.

If the aircraft was below the altitude it will climb in fixed wing mode before doing the turn in. The climb is intended as we want to be clear any possible ground interference/obstacles.

Apparently if you are in a quad mode then we just quad over then descend like a normal quad landing? It's not what I would have predicted... That probably could stand some refinement to transition back to forward flight if outside of the radius. I haven't thoughly tested this, just ran it once in the simulator for now.

plane/source/docs/quadplane-auto-mode.rst
```diff
@@ -55,6 +55,13 @@ will be done at whatever height the aircraft is at when it starts on
...
+
+.. image:: ../images/VTOL_LAND.jpg
+    :target: ../_images/VTOL_LAND.jpg
+
+(figure courtesy of SRP AERO)
```
WickedShell: Can this include a link to https://srp.aero ?

Hwurzburg: sure...next commit...and thanks...

plane/source/docs/quadplane-auto-mode.rst
```diff
@@ -55,6 +55,15 @@ will be done at whatever height the aircraft is at when it starts on
 the NAV_VTOL_LAND waypoint. So you would typically want the previous
 waypoint to have an altitude of about 20 meters above the ground.
 
+If Q_OPTIONS is set for "Use a fixed wing approach" (bit 4), then instead of transitioning to VTOL flight and doing a VTOL landing, it will remain in plane mode , and proceed to the landing position,climbing or descending to the altitude set in the NAV_VTOL_LAND waypoint. When it reaches within Q_FW_LND_APR_RAD of the landing location, it will perform a LOITER_TO_ALT to finish the climb or decent to that altitude set in the waypoint, then, turning into the wind, transition to VTOL mode and proceed to the landing location and land.
```
WickedShell: `landing position,climbing or descending` should be `landing position, climbing or descending` (missing whitespace after a comma)

`descent` rather then `decent`

plane/source/docs/quadplane-auto-mode.rst
```diff
@@ -55,6 +55,15 @@ will be done at whatever height the aircraft is at when it starts on
...
+
+Note that if Q_FW_LND_APR_RAD=0, the Waypoint loiter radius will be used. Be sure to set the VTOL_LAND waypoint altitude and loiter radius to avoid obstacles. Note also, that if this waypoint is reached while in VTOL mode, it will proceed as a normal VTOL_LAND command, even if the option is set.
```
WickedShell: You can't set the radius on the landing waypoint to the best of my knowledge. It only uses the vehicles loiter radius, or the `Q_FW_LND_APR_RAD` nothing on the waypoint itself.

Hwurzburg: changed to Q_FW_LND_APR_RAD for clarity

Hwurzburg:
Built and verified

PR #1799, opened 2019-06-14
Update qacro-mode.rst

kd0aij:
begin a QACRO flight tips section (suggested by @losawing)


rmackay9:
Merged, thanks.  As a side note, it's best if new branches aren't created in master .. but good to have the update in any case.

kd0aij:
@rmackay9  Thanks, this is what happens if you use the web editor and don't choose commit; perhaps it could be configured differently?
Assuming the answer is no, I'll use my own fork instead whenever I feel a review is warranted.

PR #1802, opened 2019-06-16
Updated FRSky Sport DIY cable (Frsky no longer offers the components)

Hwurzburg:
DIY info out of date and components no longer available. Current DIY solution commonly used is shown.


rmackay9:
I'm a little concerned about this one. It seems like we're removing info about how a nearly ready-to-use option can be purchased and instead pushing people towards a much harder DIY solution.  It seems that EBay still has the component even if FRSky itself does not.

Hwurzburg:
I can certainly just add it as another DIY alternative, but the FRSKY links are dead ...perhaps a note to the effect that its no longer offered and to search Ebay for exisiting inventory.....fyi, I seem to run into this solution a lot more often on RCG and discuss.ardupilot.com than the FRSky one ...

rmackay9:
ok, removing the FrSky links (which is dead) seems like a good idea but the ebay ones still look good so let's leave those in place..

Hwurzburg:
Changed as suggested

PR #1809, opened 2019-06-24
Rover last version of firmware for APM2

brunoolivieri:
I discover in the forums that APM2 is no longer support newer Rover firmwares. Based on firmwares repo I beleive that the last firmware that fits on the APM2 is the 3.4.2.

PR to provide same information on the wiki.

### DIFF
--- a/common/source/docs/common-autopilots.rst
+++ b/common/source/docs/common-autopilots.rst
@@ -56,7 +56,7 @@ Discontinued boards
 The following boards are end-of-life. The documentation is archived, but
 available if you're still working on those platforms:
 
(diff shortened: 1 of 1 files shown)


rmackay9:
Looking good except one small change is the last Rover version on APM2 was 3.1.0

brunoolivieri:
My bad! I fixed.

I  took this conclusion based on the folder http://firmware.ardupilot.org/Rover/stable-3.4.2/apm2/

Where those legacy versions could be downloaded? (Rover/Copter/Plane?)

rmackay9:
Merged, so it seems that in fact, the last APM2 version of Rover was 2.5.1

PR #1812, opened 2019-06-26
[fix] Rover last version of firmware for APM2

brunoolivieri:
Simple correction from PR #1809  and wiki meeting.


rmackay9:
Merged, thanks!

PR #1820, opened 2019-07-01
HWDEF: update OmnibusF7 per ninja-zx11

Hwurzburg:
@rmacky9 Randy, what is the best way to approach the info about FRsky Sport connection since its not in stable releases (as far as I know) yet, and will appear at different times for different vehicles?

### DIFF
--- a/common/source/docs/common-omnibusf7.rst
+++ b/common/source/docs/common-omnibusf7.rst
@@ -19,8 +19,8 @@ Specifications
 
 -  **Processor**
 
-   -  STM32F7 ARM
-   -  512KB flash
+   -  STM32F745VG ARM
+   -  1MB of Flash memory
 
 
 -  **Sensors**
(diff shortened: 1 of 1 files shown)


OXINARF:
@Hwurzburg Some points regarding that:

- what is only available in dev versions are the serial options, the FrSky Sport is available in stable versions for quite some time (since Copter 3.3 I think)
- the protocol it is telling users to set isn't the standard FrSky Sport, but rather an extended protocol created by C&T; this isn't necessarily what users need or want so it's better to direct them to FrSky pages to let them choose what they want
- I'd rather have the FrSky pages mention the SERIAL_OPTIONS tricks since it isn't specific to this board

In conclusion, I think that the board page should only mention that TX1 pad matches SERIAL1_* parameters and direct users to FrSky pages if they want to set it up.

Also, a note on commit messages: it's OK to reference the origin of information, but please make the commit message mention what the commit is actually doing; otherwise, when looking at history, there is no way to tell without opening every single commit.

Hwurzburg:
@OXINARF Francisco, I agree, best not include in bd wiki....also, I will be more careful in the PR titles.... will rename when I squash after the changes...

Hwurzburg:
Checked in Sphinx...

PR #1822, opened 2019-07-03
optitrack: correct the source of mavlink files

chobitsfan:
Mavlink.cs, MavlinkMessage.cs, etc are actually coming from Mission Planner source code, they are not auto generated by MAVLink. I am very sorry about this mistake.

thanks @jbausano for pointing it out


rmackay9:
Merged, thanks!

PR #1824, opened 2019-07-05
add description of Q_TAILSIT_ANGLE

kd0aij:


kd0aij:
I added description for the timeout; thanks for pointing that out.
BTW, this check in tailsitter_transition_vtol_complete:
```
labs(plane.ahrs.roll_sensor) > tailsitter.transition_angle*100
```
seems wrong since plane.ahrs.roll_sensor is reporting roll in the vehicle body frame.  
The roll check in tailsitter_transition_fw_complete does seem to make sense, since it is in the ahrs_view frame.

PR #1827, opened 2019-07-07
lua-scripts: Update battery docs

WickedShell:
Requires https://github.com/ArduPilot/ardupilot/pull/11740 for the behavior to be consistent with what's described here.


WickedShell:
The required PR has gone into master.

PR #1831, opened 2019-07-10
parameter THR_MID changed to MOT_THST_HOVER

gnitzan:
Added note to the fact that parameter THR_MID changed in version 3.6 to MOT_THST_HOVER.


rmackay9:
Merged, thanks!

PR #1836, opened 2019-07-11
Cartographer: fix workspace name and protocol

tajisoft:
catkin -> catkin_ws
https -> http


rmackay9:
cool, thanks for that!

Issue #1844, opened 2019-07-16
Add missing boards to AutoPilot Hardware Options page


rmackay9:
We are missing a number of boards from our [AutoPilot Hardware Options](http://ardupilot.org/copter/docs/common-autopilots.html).  We should go through and make sure all supported boards are listed on this page.

auturgy:
Pixhawk 4 mini is in the list

rmackay9:
Ah, right, I'll change the title to make it more clear that we should just check that we have all boards listed on the wiki

Hwurzburg:
Can we close this now?

rmackay9:
Yes, we probably need to be more specific when an issue is raised like this so we know when it's complete.  Txs,closing!

PR #1845, opened 2019-07-16
Update common-autopilots.rst

auturgy:
Partial remedy for https://github.com/ArduPilot/ardupilot_wiki/issues/1844
I've used the best weblinks I could find for the boards added, but we'll need to work with vendors to improve them / add dedicated pages.  I've also re-organised into alphabetical order (which was partially already the case).
The valid waf targets that I believe are available to consumers but I haven't added are (boards, not rtf's which I excluded): airbotf4, bhat, crazyflie2, dark, erleboard, mindpx-v2,  omnibusf4v6, rst_zynq, VRCore-v10, zynq
Those targets I'm either not familiar with, couldn't find reasonable links for, and/or we don't have OEM datasheets/pages.  There are also a few boards we build for that aren't available anymore, so I've ignored them for now.

### DIFF
--- a/common/source/docs/common-autopilots.rst
+++ b/common/source/docs/common-autopilots.rst
@@ -4,59 +4,74 @@
 Autopilot Hardware Options
 ==========================
 
-This section provides information about the main ArduPilot Flight Controller Hardware options:
+This section provides information about ArduPilot Flight Controller Hardware options:
 
 Open hardware 
 -------------
 
 .. toctree::
     :maxdepth: 1
-
-    Pixhawk <common-pixhawk-overview>
-    The Cube <common-thecube-overview>
-    Pixracer <common-pixracer-overview>
-    CUAV v5 <common-pixhackV5-overview>
+    
+    BBBMini <https://github.com/mirkix/BBBMINI>
     Beagle Bone Blue <common-beagle-bone-blue>
-    Erle-Brain <common-erle-brain-linux-autopilot>
+    CUAV v5 and v5+ <common-pixhackV5-overview>
+    CUAV v5 Nano <https://store.cuav.net/index.php?id_product=98&id_product_attribute=0&rewrite=cuav-new-pixhack-v5-nano-small-flight-controller-for-ardupilot-px4-drone-parts-free-shipping-whole-sale-&controller=product&id_lang=3>
+    Drotek Pixhawk3 <https://drotek.gitbook.io/pixhawk-3-pro/>
     F4BY <common-f4by>
+    Hex/ProfiCNC Cube Black <common-thecube-overview>
+    Hex/ProfiCNC Cube Green <http://www.proficnc.com/all-products/79-the-cube.html>
+    Hex/ProfiCNC Cube Orange <http://www.proficnc.com/all-products/188-the-cube.html>
+    Hex/ProfiCNC Cube Purple <http://www.proficnc.com/all-products/176-the-cube.html>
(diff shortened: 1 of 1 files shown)


rmackay9:
Alrighty.. well, the order is sure to cause some controversy.. I had them in mostly alphabetical order but put the pixhawk and cube at the top because they are the most commonly used boards (I think).

I had been thinking that we would only add links to our own wiki pages instead of directly to the vendor.. although this is not too bad.

auturgy:
we need to continue to push the manufacturers and vendors to maintain pages.  With the number of boards, and rate of change, it's not realistic or sustainable to expect community contributors to build pages for all of them.  It's always a risk linking to external sites, but better than nothing.

rmackay9:
@auturgy, except that Tridge has made it easy to create a wiki page on github.  I think the intention was that developers would maintain that..

Hwurzburg:
my 2cents....I agree a little with both points....whoever ports the code to a platform should setup the board Wiki page...it could be the manufacturer or an interested dev....if there is no board Wiki page, it should not be listed in the hardware summary, even if the code is in the repo...an incentive for manufacturers to create them...it also prevents external links

brunoolivieri:
My 2 seconds:

1) The render/build of "discontinued boards" appears wrong when I built it to check. (The list of Erle and Intel boards.)

2) I agree with alphabetical order moving most used to the top.

Do you know if there is any survey about most used boards? Could it be useful? I could set something in surveymonkey or google spreads.

[Based on the forums seems that Pixhawk is the most used board. However, the aeroclubs where I use to go, I just saw users using APM2, the only Pixhawk that I saw was my single one. I still use a couple of APMs  :-( ]

auturgy:
The only way to ensure proper documentation is not to merge a new board/feature until the accompanying wiki update is contributed. If we aren’t willing to make that stipulation (evidence suggests we aren’t) then the wiki will always be out of date. We’ve had this discussion before, several times.

brunoolivieri:
I believe that documentation stipulation would be better from the wiki perspective, but not in the developer's point of view. 

It is not to easy to push in any direction as an OpenSource project. I do not feel comfortable to "deny" some feature or port of a new board because it does not have proper documentation.

Maybe we could split the boards into different lists, for an example: Well documented, Manufacturer documented and Not documented at all.  

Another possibility would present the boards in a table that classify the available documentation in columns, for an example: community documentation, manufacturer documentation, pinouts, basic assembly manual, user cases, etc...

tridge:
rebased and pushed

rmackay9:
This is in master so closing

proficnc:
If this is supposedly in Alphabetical order, can we please name the products correctly? 
CubeBlack is “CubeBlack”, not “Hex/ProfiCNC CubeBlack”

rmackay9:
@proficnc, they're suppose to be ordered by company name then product.  So one issue is the Pixhawk needs to be renamed to "mRo Pixhawk".  Otherwise I think they are consistent.. i.e. company name / product name.

rmackay9:
I've gone ahead and added the note re ordering and add the "mRo" before Pixhawk (these changes should appear in about 15min).

I've raised a [new issue to cover the enhancement request to change the list into a table here](https://github.com/ArduPilot/ardupilot_wiki/issues/1866).

I understand the new order is not going to make everyone happy but unfortunately we receive complaints from manufacturers no matter what we do so I think alphabetically is the best option.

I don't really think there's much point in leaving this PR open because the changes have already been merged.  Even if there was a decision to back out the changes this PR wouldn't help.

PR #1846, opened 2019-07-18
Dev: Add page for AirSim Simulator

rajat2004:
Adds setup & usage page for AirSim

WIP

### DIFF
--- a/dev/source/docs/sitl-with-airsim.rst
+++ b/dev/source/docs/sitl-with-airsim.rst
@@ -1,0 +1,286 @@
+.. _sitl-with-airsim:
+
+======================
+Using SITL with AirSim
+======================
+
+.. youtube:: -WfTr1-OBGQ
+   :width: 100%
+
+`AirSim <https://github.com/microsoft/AirSim>`__ is a simulator for drones, cars and more, built on Unreal Engine (they also have an experimental Unity release, but right now it's hasn't been implemented with ArduPilot)
+
+It is open-source, cross-platform and provides excellent physically and visually realistic simulations. It has been developed to become a platform for AI research to experiment with deep learning, computer vision and reinforcement learning algorithms for autonomous vehicles.
+
+Currently, only support for ArduCopter vehicle has been developed in AirSim & ArduPilot.
+
+.. note::
+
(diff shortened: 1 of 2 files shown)


dev/source/docs/sitl-with-airsim.rst
```diff
@@ -0,0 +1,281 @@
+.. _sitl-with-airsim:
+
+======================
+Using SITL with AirSim
+======================
+
+.. youtube:: -WfTr1-OBGQ
+   :width: 100%
```
rajat2004: This is a demo video by AirSim, still need to make one with ArduCopter

rajat2004:
As you can see, the installation instructions are quite long, especially since it's for both Windows and Linux
I have linked the setup pages from the AirSim repo as well, should the instructions be kept or removed?

dev/source/docs/sitl-with-airsim.rst
```diff
@@ -0,0 +1,281 @@
...
+        git clone https://github.com/Microsoft/AirSim.git
+
+#. Currently, the `AirSim PR <https://github.com/microsoft/AirSim/pull/2075>`__  for ArduCopter Support hasn't been merged, so you'll need to fetch that particular branch for using Ardupilot.
+
+    ::
+    
+        cd AirSim
+        git remote add <remote-name> https://github.com/rajat2004/AirSim.git
+        git fetch <remote-name>
+        git checkout -b <your-branch-name> <remote-name>/pr-arducopter
```
rajat2004: I'm not sure if this the best way to go about this. Can someone suggest a more suitable method?

rajat2004:
Still need to test and add info about the Image Apis

dev/source/docs/sitl-with-airsim.rst
rajat2004: Another possibility is to have something like ArduPilot's own version of AirSim, since the AirSim repo seems to be very much inactive these days (something similar to this was there for earlier for JSBSim I think?)

tridge: yes, if it doesn't get merged soon we will fork into ardupilot repo

PR #1850, opened 2019-07-19
Copter: update tuning page

khancyr:
Add correct link
Make safety bold
add more comments for newcomer


rmackay9:
@khancyr, thanks for looking at this.  I don't think we can merge this as-is because there's a number of to-do items included.

I think the first thing we need to do is probably get all the listed parameter changed to links to the param docs.

khancyr:
Yep, that's why I made it WIP, but it was easier for review.

I will modify the parameters to have correct links

khancyr:
updated with links for parameters
add matplotlib to be able to generate plot, I think it can be usefull on some other place too.
This will need an update on our server.
example of plots
Capture d’écran de 2019-07-20 18-22-21

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -158,96 +286,95 @@ If the aircraft appears stable enough to attempt autotune follow the instruction
...
+:ref:`ACRO_YAW_P <ACRO_YAW_P>` should be set to be approximately 0.5 x :ref:`ATC_ACCEL_Y_MAX <ATC_ACCEL_Y_MAX>` / 4500 to ensure that the aircraft will get up to full yaw rate and stop again in approximately half a second.
 
-ATC_ANG_LIM_TC may be increased to provide a very smooth feeling on the sticks at the expense of a slower reaction time.
+:ref:`ATC_ANG_LIM_TC <ATC_ANG_LIM_TC>` may be increased to provide a very smooth feeling on the sticks at the expense of a slower reaction time.
 
-Aerobatic aircraft should keep the ATC_ACCEL_X_MAX provided by autotune and reduce ATC_ANG_LIM_TC to achieve the stick feel desired by the pilot. For pilots wanting to fly ACRO the following input shaping parameters can be used to tune the feel of ACRO:
+Aerobatic aircraft should keep the ATC_ACCEL_X_MAX provided by autotune and reduce :ref:`ATC_ANG_LIM_TC <ATC_ANG_LIM_TC>` to achieve the stick feel desired by the pilot. For pilots wanting to fly ACRO the following input shaping parameters can be used to tune the feel of ACRO:
```
khancyr: ATC_ACCEL_X_MAX doesn't exist

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -9,26 +9,138 @@ Setting the aircraft up ready for tuning
...
+Battery setting
+^^^^^^^^^^^^^^^
+Those values a necessary to scale the motor thrust linearisation.
+
+- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>` - 4.2v x No. Cells
+- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>` - 3.3v x No. Cells
+
+Motors setup
+^^^^^^^^^^^^
+Those values are necessary to scale correctly your motor control.
```
lthall: Parameters used to linearise your motor thrust curve.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -9,26 +9,138 @@ Setting the aircraft up ready for tuning
...
-- ACRO_YAW_P - 0.5 x ATC_ACCEL_Y_MAX / 4500
-
-The initial tune of the aircraft should be done in the aircrafts most agile configuration. This generally means that the aircraft will be at its minimum take off weight with fully charged batteries.
+.. warning::
+
+    Values provide here are a starting point to tune your aircraft, they should not be hardly trust as the best values. Before flying please read the entire instruction to understand what you will do, what you should expect and what to do.
+
+Battery setting
+^^^^^^^^^^^^^^^
+Those values a necessary to scale the motor thrust linearisation.
```
lthall: Parameters used to linearise your motor thrust curve.

This should also include the expo value. So we could move that up.

copter/source/docs/tuning-process-instructions.rst
lthall: Parameters used to define the output range sent to the ESC.

This is probably closer but not perfect.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -9,26 +9,138 @@ Setting the aircraft up ready for tuning
...
+- :ref:`MOT_PWM_MAX <MOT_PWM_MAX>` - Check ESC manual for fixed range or 2000us
+- :ref:`MOT_PWM_MIN <MOT_PWM_MIN>` - Check ESC manual for fixed range or 1000us
+- :ref:`MOT_SPIN_ARM <MOT_SPIN_ARM>` - use the motor test feature :ref:`Checking the motor numbering with the Mission Planner Motor test`
+- :ref:`MOT_SPIN_MAX <MOT_SPIN_MAX>` - 0.95  # TODO why this value? give a little headroom on ESC?
+- :ref:`MOT_SPIN_MIN <MOT_SPIN_MIN>` - use the motor test feature :ref:`Checking the motor numbering with the Mission Planner Motor test` # TODO : how do select value betwen ARM and MIN ?
+- :ref:`MOT_THST_EXPO <MOT_THST_EXPO>` - 0.55 for 5 inch props, 0.65 for 10 inch props, 0.75 for 20 inch props. This parameter should be derived by thrust stand measurements for best results (don’t trust manufacture data).
+- :ref:`MOT_THST_HOVER <MOT_THST_HOVER>` - 0.25 or below the expected hover thrust percentage (low is safe)
+
+.. plot::
+
```
lthall: It would be nice to replace those lines with a curve. From the data I have seen they tend to level out at about the 0.75 to 0.8 mark for the larger props (20' and above). The curve rolls over as we move down through approximately .65 at 10 inch and 5.5 ish at 5 inch. It would be nice to represent this as a gradual curve.

When using the same props in coaxial we get a similar shape but the graph is compressed by about 0.1 at the top end and probably just a little lower at the bottom end.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -9,26 +9,138 @@ Setting the aircraft up ready for tuning
...
+                arrowprops=dict(arrowstyle="->"))
+    ax.annotate('20 inches props',
+                xy=(20, 20), xycoords='data',
+                xytext=(-50, 50), textcoords='offset points',
+                bbox=dict(boxstyle="round", fc="w"),
+                arrowprops=dict(arrowstyle="->"))
+    ax.set_xlabel('Props size in inches')
+    ax.set_ylabel('INS_GYRO_FILTER value')
+    ax.set_title('Theorical relationship between INS_GYRO_FILTER value \n and props size in inches')
+
```
lthall: Approximate relationship between INS_GYRO_FILTER value \n and props size in inches

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -9,26 +9,138 @@ Setting the aircraft up ready for tuning
...
+                arrowprops=dict(arrowstyle="->"))
+    ax.annotate('30 inches props',
+                xy=(30, 20000), xycoords='data',
+                xytext=(-50, 50), textcoords='offset points',
+                bbox=dict(boxstyle="round", fc="w"),
+                arrowprops=dict(arrowstyle="->"))
+    ax.set_xlabel('Props size in inches')
+    ax.set_ylabel('ATC_ACCEL_P_MAX or ATC_ACCEL_R_MAX value')
+    ax.set_title('Theorical relationship between \n ATC_ACCEL_P_MAX or ATC_ACCEL_R_MAX value \n and props size in inches')
+
```
lthall: Approximate relationship between \n ATC_ACCEL_P_MAX or ATC_ACCEL_R_MAX value \n and props size in inches

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -70,6 +190,8 @@ The first take off is the most dangerous time for any multirotor. Care must be t
 10. Hover at approximately 1m altitude and apply small (5 degrees) control inputs into roll and pitch.
 11. Immediately land if any oscillation is observed.
 
+# TODO what to do in case of oscillation ?
+
```
lthall: I don't know that there is much more we can say other than "Immediately land".

Tuning instructions in this situation are further down.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -80,14 +202,14 @@ The first priority when tuning an multirotor aircraft is to establish a stable t
...
-f. ATC_RAT_RLL_D
+a. :ref:`ATC_RAT_PIT_P <ATC_RAT_PIT_P__AC_AttitudeControl_Multi>`
+b. :ref:`ATC_RAT_PIT_I <ATC_RAT_PIT_I__AC_AttitudeControl_Multi>`
+c. :ref:`ATC_RAT_PIT_D <ATC_RAT_PIT_D__AC_AttitudeControl_Multi>`
+d. :ref:`ATC_RAT_RLL_P <ATC_RAT_RLL_P__AC_AttitudeControl_Multi>`
+e. :ref:`ATC_RAT_RLL_I <ATC_RAT_RLL_I__AC_AttitudeControl_Multi>`
+f. :ref:`ATC_RAT_RLL_D <ATC_RAT_RLL_D__AC_AttitudeControl_Multi>`
 
-This process is repeated until the aircraft can hover without oscillations being detectable visually or audibly.
+This process is repeated until the aircraft can hover without oscillations being detectable visually or audibly. # explain audibly ? motor sound should be constant and not noisy ?
```
lthall: I don't think we need to explain this. When the aircraft oscillates you can generally clearly hear the pitch of the motor going up and down. This is where a video or two will help out. For now though I think this line was fine the way it was.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -96,16 +218,21 @@ Be aware that in this state the aircraft may be very slow to respond to large co
...
+This test will allow to test the altitude controler and unsure the stability of your aircraft.
+
+1. Check :ref:`MOT_HOVER_LEARN <MOT_HOVER_LEARN>` is set to 2. This will allow the controler to learn by itself the correct hover value when flying.
+
 Take off in Stabilise and increase altitude to 5m. Switch to Alt_Hold and be ready to switch back. If the aircraft is hovering at a very low hover throttle value you may hear a reasonably fast oscillation in the motors.
 
-1. Check MOT_HOVER_LEARN is set to 2
 2. Ensure the aircraft has spent enough time in hover to let the hover throttle value converge to the correct value.
+
+# todo land ? or in flight ?
```
lthall: 2. Take off in Stabilise and increase altitude to 5m. Switch to Alt_Hold and be ready to switch back to Stabilize. If the aircraft is hovering at a very low hover throttle value you may hear a reasonably fast oscillation in the motors. Ensure the aircraft has spent at least 30 seconds in hover to let the hover throttle parameter converge to the correct value. Land and disarm the aircraft.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -117,27 +244,28 @@ Most pilots will look to move to Autotune as quickly as possible once their airc
...
 Manual tuning of Roll and Pitch
 -------------------------------
 
 Manual tuning may be required to provide a stable tune before Autotune is run, or if Autotune does not produce an acceptable tune. The process below can be done on roll and pitch at the same time for a quick manual tune provided the aircraft is symmetrical. If the aircraft is not symmetrical then the process should be repeated for both roll and pitch individually.
 
-The pilot should be especially careful to ensure that ATC_THR_MIX_MAN and MOT_THST_HOVER are set correctly before manual tuning is started.
+The pilot should be especially careful to ensure that :ref:`ATC_THR_MIX_MAN <ATC_THR_MIX_MAN>` and :ref:`MOT_THST_HOVER <MOT_THST_HOVER>` are set correctly before manual tuning is started.
 
 When oscillations start do not make large or sudden stick inputs. Reduce the throttle smoothly to land the aircraft while using very slow and small roll and pitch inputs to control the aircraft position.
+# todo : in flight ? in hover state ?
```
lthall: This is true for any state so I don't think we need to elaborate.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -146,7 +274,7 @@ When oscillations start do not make large or sudden stick inputs. Reduce the thr
 5. Reduce the P term in steps of 10% until the oscillation disappears.
 6. Reduce the P term by a further 25%.
 
-Each time the P term is changed set the I term equal to the P term.
+Each time the P term is changed set the I term equal to the P term.  # TODO I term or I term change ?
```
lthall: I think this line is fine as is. I am not sure what your question is there.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -158,96 +286,95 @@ If the aircraft appears stable enough to attempt autotune follow the instruction
...
+- :ref:`ATC_RAT_YAW_P <ATC_RAT_YAW_P__AC_AttitudeControl_Multi>` should be reduced from 0.5 to 1
+- :ref:`ATC_RAT_YAW_I <ATC_RAT_YAW_I__AC_AttitudeControl_Multi>` = :ref:`ATC_RAT_YAW_P <ATC_RAT_YAW_P__AC_AttitudeControl_Multi>` x 0.1
 
 These values should only be changed if Autotune produces higher values. Small aerobatic aircraft may prefer to keep these values as high as possible.
 
 Setting the input shaping parameters
 ------------------------------------
 
-Arducopter has a set of parameters that define the way the aircraft feels to fly. This lets allows the aircraft to be set up with a very aggressive tune but still feel like a very docile and friendly aircraft to fly. 
+Arducopter has a set of parameters that define the way the aircraft feels to fly. This lets allows the aircraft to be set up with a very aggressive tune but still feel like a very docile and friendly aircraft to fly.
```
lthall: This allows the aircraft to be set up with a very aggressive tune but still feel like a very docile and friendly aircraft to fly.

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -158,96 +286,95 @@ If the aircraft appears stable enough to attempt autotune follow the instruction
...
+Autotune will set the :ref:`ATC_ACCEL_X_MAX <ATC_ACCEL_X_MAX>` parameters to their maximum based on measurements done during the Autotune tests. These values should not be increased beyond what Autotune suggests without careful testing. In most cases pilots will want to reduce these values significantly.
 
-For aircraft designed to carry large directly mounted payloads, the maximum values of ATC_ACCEL_X_MAX should be reduced based on the minimum and maximum take off weight (TOW):
-ATC_ACCEL_X_MAX  x (min_TOW / max_TOW)
+For aircraft designed to carry large directly mounted payloads, the maximum values of :ref:`ATC_ACCEL_X_MAX <ATC_ACCEL_X_MAX>` should be reduced based on the minimum and maximum take off weight (TOW):
+:ref:`ATC_ACCEL_X_MAX <ATC_ACCEL_X_MAX>`  x (min_TOW / max_TOW)
 
-ACRO_YAW_P should be set to be approximately 0.5 x ATC_ACCEL_Y_MAX / 4500 to ensure that the aircraft will get upto full yaw rate and stop again in approximately half a second.
+:ref:`ACRO_YAW_P <ACRO_YAW_P>` should be set to be approximately 0.5 x :ref:`ATC_ACCEL_Y_MAX <ATC_ACCEL_Y_MAX>` / 4500 to ensure that the aircraft will get up to full yaw rate and stop again in approximately half a second.
```
lthall: to ensure that the aircraft can achieve full yaw rate in approximately half a second.

copter/source/docs/tuning-process-instructions.rst
lthall: ATC_ACCEL_P_MAX, ATC_ACCEL_R_MAX and ATC_ACCEL_Y_MAX

khancyr: thanks, done

copter/source/docs/tuning-process-instructions.rst
khancyr: ok, so I just add Next section will explain how to remove the oscillations.

copter/source/docs/tuning-process-instructions.rst
khancyr: could be say a chopped sound ? non constant sound ?

copter/source/docs/tuning-process-instructions.rst
khancyr: ok. So we set the parameter on ground and not in flight !
after setting PSC_ACCZ_I and PSC_ACCZ_P, should we takeoff again to check the behavior ? or just continue to the next section is enough?

copter/source/docs/tuning-process-instructions.rst
khancyr: So we set the parameter when land , right ?

copter/source/docs/tuning-process-instructions.rst
khancyr: We set I = P  or P percentage change = I percentage change ?

copter/source/docs/tuning-process-instructions.rst
khancyr: done!

khancyr:
To plot the curve, it is python that is generated on compile time.
You just need python and install matplotlib ( sudo pip install matplotlib) (it should already be there if you have mavproxy). Then copy paste what is in plot directive and add "plt.show()" at the end , it will create a window with the plot ! That is simplier than trying to rebuild the whole copter section with sphinx!

khancyr:
even simplier : Copy paste code in https://repl.it/repls/CurlyLateCones and add : 
plt.show()
plt.savefig("quad.png")
at end ! it will plot it for you without installing anything !

copter/source/docs/tuning-process-instructions.rst
lthall: I really don't think it needs any further explanation. It sounds like it looks. I am happy to wait for the video's.

copter/source/docs/tuning-process-instructions.rst
lthall: Sorry, I don't understand what you are asking. The feedback there was to replace point 2 to address your todo.

lthall: OH, the todo's are before what you are asking about! I thought they were after :)

They can do this in flight or on the ground but it is easier to do it on the ground if you are on your own. You can also use ch6 tuning for many parameters. If you are tuning with someone it is often easy enough to do it while in flight. I am happy to leave it up to the pilot.

My personal preference is to do parameter changes on the ground.

copter/source/docs/tuning-process-instructions.rst
lthall: Again, there are a number of ways to do this. I am happy to leave it up to the pilot.

copter/source/docs/tuning-process-instructions.rst
lthall: After making a change to the P term set the I = P. If you are using Ch6 tuning you may not be able to set I = P until you land again. In any case it isn't important when I is set equal to P.

lthall:
Sorry but I have not learnt to use python and I tried to give that website a go but the things I tried just resulted in errors.

khancyr:
does this work ? 
https://repl.it/@khancyr/wiki-curves

lthall:
Sorry, no it doesn't. Can you just attach what you have done here or on the discuss?

khancyr:
Approximate relationship between ATC_ACCEL_Y_MAX value _ and props size in inches
Approximate relationship between INS_GYRO_FILTER value _ and props size in inches
Approximate relationship between MOT_THST_EXPO value _ and props size in inches

lthall:
That looks great!!

If we get rid of those todo's this can go straight in!!!

Legend!!

khancyr:
Last todo removed as unneeded.
Thanks for helping me on this.

Before merging, I just need to add matplotlib on our server and it will be good !

copter/source/docs/tuning-process-instructions.rst
```diff
@@ -15,50 +15,63 @@ The following parameters should be set correctly based on the specifications of
...
-- :ref:`MOT_SPIN_ARM <MOT_SPIN_ARM>` - use the motor test feature :ref:`Checking the motor numbering with the Mission Planner Motor test`
-- :ref:`MOT_SPIN_MAX <MOT_SPIN_MAX>` - 0.95  # TODO why this value? give a little headroom on ESC?
-- :ref:`MOT_SPIN_MIN <MOT_SPIN_MIN>` - use the motor test feature :ref:`Checking the motor numbering with the Mission Planner Motor test` # TODO : how do select value betwen ARM and MIN ?
 - :ref:`MOT_THST_EXPO <MOT_THST_EXPO>` - 0.55 for 5 inch props, 0.65 for 10 inch props, 0.75 for 20 inch props. This parameter should be derived by thrust stand measurements for best results (don’t trust manufacture data).
-- :ref:`MOT_THST_HOVER <MOT_THST_HOVER>` - 0.25 or below the expected hover thrust percentage (low is safe)
 
 .. plot::
 
     import matplotlib.pyplot as plt
+    import numpy as np
```
rmackay9: I guess this part is actually building a graph on the web page?  If it's a static view I think we should just turn it into an image..

lthall: > I'll leave the decision to @lthall though.

This is really not one of those decisions that I should be making :)

All I can say is what we have here is better than what we have at the moment. Eventually I would like to get better thrust data and refine these graphs. Then I would be generating them in matlab and probably outputting an image file.

My knee jerk reaction is to use an image but as I said, I am not the person to make that decision.

rmackay9: ok, I'd prefer an image then unless there are strong objections..

khancyr: My idea with this was to not lost the source. I spend some time to produce those plot and I don't want that somebody updating after spend time on this. 
Numpy and matplotlib are pretty commun include in python, so that shouldn't be the blocking point.
Maybe a solution when a plot is provide would be to ask the source on a gist or some place in the PR. That way we won't lose the source ?

rmackay9: @khancyr, if you really want it generated then I'm ok with it..

khancyr:
rebase, squash and remove the plot directives.
One last review to ensure it is ok and we could merge it !

rmackay9:
Merged, let's see how it looks!

lthall:
Well done!!!!

PR #1856, opened 2019-07-23
Dev: Add more Lua docs

WickedShell:
This adds the GPS, Notify, and Vector2f docs.

This needs https://github.com/ArduPilot/ardupilot/pull/11867 for the GPS docs to be entirely correct, but that's probably not a hard requirement for pulling this in, it's fixing the behavior on a method I'm not even fully sure we need to be exposing to scripting users.


rmackay9:
Merged, thanks!

At some point we will probably want to break the page up into a few pages to keep each one shorter.. but not required immedaitely..

PR #1859, opened 2019-07-24
Minor fixes + making the citation explicit.

akolobov:


rmackay9:
Looks good to me but I'll leave @tridge (or someone) to merge 'cuz they probably know better if the citation link is correct.

PR #1864, opened 2019-07-29
Common: update RC calibration page

rmackay9:
I've reworked the RC calibration page in the hopes that it will make the procedure more clear to new users.  I've found a lot of new Rover users who bump into trouble because they have messed up either their RC input configation and/or output configuration.

I think the procedure outlined follows how I normally configure a transmitter and then perform the calibration and a I suspect most experienced users do something similar.

This is a pretty critical page and affects all vehicles so I've raised it as a PR so it can be reviewed by the likes of @tridge.


common/source/docs/common-radio-control-calibration.rst
```diff
@@ -1,28 +1,85 @@
...
+RC Calibration involves capturing each RC input channel's minimum, maximum and "trim" values so that ArduPilot can correctly interpret the input.
 
-Transmitter configuration
--------------------------
+Check the Transmitter's Setup
+-----------------------------
+
+- Ensure the battery is disconnected (this is important because it is possible to accidentally arm the vehicle during the RC calibration process)
+- Ensure the RC receiver is connected to the flight controller
+- Turn on your RC transmitter and if it has "trim tabs" ensure they are in the middle
```
khancyr: shouldn't the trim be middle before turn on ?

common/source/docs/common-radio-control-calibration.rst
```diff
@@ -1,28 +1,85 @@
...
+   -  No lights may indicate that it is incorrectly wired to the autopilot.  Look for connectors that may have been inserted upside down
+   -  A Red or flashing LED may indicate that your RC transmitter/receiver need be bound.  See the manual that came with your RC equipment for instructions
+
+.. figure:: ../../../images/mp_radio_calibration.png
+   :target: ../_images/mp_radio_calibration.png
+
+- Check the channel mapping in the transmitter (ie. check which inputs channels are controlled by the transmitter's sticks, switches and knobs) by moving the sticks, knobs and switches and observing which (if any) green bars move.  If this is the first time the transmitter is being used with ArduPilot it is likely that the transmitter's channel mapping will need to be changed and normally this is done on the transmitter itself using its built-in configuration menu
+
+  - Determine if your transmitter is Mode1 or Mode2 (see below)
+  - Roll stick should control channel 1
```
khancyr: better put a picture for mode 1/ mode 2

common/source/docs/common-radio-control-calibration.rst
Hwurzburg: Most tx now have digital trim...you can't be sure they are centered until you turn on tx

common/source/docs/common-radio-control-calibration.rst
rmackay9: hi @khancyr, I've got one down at the bottom of the page.. I could make it more clear perhaps that the user should look down below in the "Mode1 vs Mode2" section.

common/source/docs/common-radio-control-calibration.rst
```diff
@@ -1,28 +1,85 @@
...
+
+- Open Mission Planner's **INITIAL SETUP \| Mandatory Hardware \| Radio Calibration** screen
+- Click on the green "Calibrate Radio" button on the bottom right
+- Press "OK" when prompted to check the radio control equipment is on, battery is not connected, and propellers are not attached
+
+   .. figure:: ../../../images/mp_calibrate_radio.jpg
+      :target: ../_images/mp_calibrate_radio.jpg
+
+- Move the transmitter's control sticks, knobs and switches to their limits.  Red lines will appear across the calibration bars to show minimum and maximum values seen so far
+
```
Hwurzburg: Should we discuss that fact that you should make sure that the THROTTLE FAILSAFE throttle minimum should NOT be included in this range calibration? most either use throttle low trim or a switch to force throttle below low stick value to implement this...if its included in the low throttle range (RC_MIN) of this cal it causes all sorts of issues...

rmackay9:
fixed a few issues found during the review this morning and merged

PR #1868, opened 2019-08-01
QPlane: correct flying mode description:QRTL

Hwurzburg:
describe differences in QRTL from Copter RTL


Hwurzburg:
added a note about quad motors running in MANUAL mode

Hwurzburg:
Superceded by #1890

PR #1872, opened 2019-08-02
Dev: Add info about Multi-Vehicle sim with AirSim

rajat2004:
Also correct instructions on running on different computers and ports

Would be better to have this go in after the Lidar, RC PR is merged in order to remove extra steps for fetching the branch.
Also, need to add a video for multi-vehicle simulation which will hopefully make it clearer how to use it


rmackay9:
I think Rajat is the one most in the know about this and it look OK so I will merge now..

rmackay9:
Ah I see from the note that perhaps I merged this prematurely.. anyway.. it's in now!

PR #1874, opened 2019-08-04
update omnibusf4 docs with firmware targets

Hwurzburg:


Hwurzburg:
Merging since I have tested this PR and am certain of correctness...

PR #1877, opened 2019-08-05
Insignificant warning removal

brunoolivieri:
Removing a single warning from Sphinx build. (WARNING: Inline strong start-string without end-string.)


rmackay9:
Merged, thanks!

PR #1883, opened 2019-08-07
Wrong  directive on warning tag

brunoolivieri:
The wrong directive was making some text not visible.


rmackay9:
Merged, thanks!

PR #1888, opened 2019-08-08
fix wiki build error and update

Hwurzburg:
fix inverted flight page build error and deprecated param


rmackay9:
This is plane so I think I'll leave this one to @tridge

PR #1889, opened 2019-08-08
fix flaperon and crow descriptions and build errors

Hwurzburg:
part of fixing wiki build errors
obsolete refs removed and text corrected to new params


Hwurzburg:
rebased and also caught one addtional old ref error not showing in wiki build errors...tridge has reviewed and approved...merging

PR #1892, opened 2019-08-10
fix wiki build error

Hwurzburg:
Last plane specific wiki build error fix...now on to the common files used by plane..

### DIFF
--- a/plane/source/docs/automatic-landing.rst
+++ b/plane/source/docs/automatic-landing.rst
@@ -240,8 +240,7 @@ Also note that if you have a longer range rangefinder then it is a very
 good idea to set the minimum range of the rangerfinder well above zero.
 For example, the PulsedLight Lidar has a typical range of over 40
 meters, and when it gets false readings it tends to read ranges of less
-than 1 meter. Setting :ref:`RNGFND_MIN_CM <RNGFND_MIN_CM>`
-to 150 will discard any rangerfinder readings below 1.5 meters, and will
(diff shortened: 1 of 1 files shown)


Hwurzburg:
Bruno, citing params without ref is done extensively throughout the wiki....thats how I found out how to highlight them with backticks...since it had to have been already setup before the use in landing, I didnt try to link....but dont mind changing it...

brunoolivieri:
I just would like to set a standard, I meam: or always link the parameters or just give the name of the parameter and the firmware version. 

But I agree with you, there are other pages like that and this PR make this one up to date.

Hwurzburg:
Sorry,in general I think that a good standard,  but  I have to disagree on one point....I think its acceptable to highlight the param if its already been linked in the page.

rmackay9:
I would prefer if we keep the link to the parameter name if possible.  It does allow the user to look at the parameter details for other information including the official description.

hamishwillee:
FWIW there are pros/cons to both linking strategies. Technical authors typically link to first instance of "thing" and then  simply highlight the others - this has advantage for reader that they only see unique links, and know they don't need to click to find out more information.

The main reason I much prefer to link everything is maintenance. It is easy to forget to search for all instances of a linked parameter - but if they are all linked you get the notification of the broken references on  building.

I also tend to not mind going to the same reference twice - it is more irritating for me to be down in the doc, decide I want to see the param, and then have to navigate up the doc to find the link.

Note that when I last did a round of link fixing there were some params that were unlinkable as they are duplicates. I submitted a PR to ensure all are unique, but don't know if that went in.

brunoolivieri:
For further maintenance, to link every parameter looks safer. I also prefer to be able to click on a parameter whenever it appears.

Hwurzburg:
Merging

PR #1893, opened 2019-08-10
fix wiki build errors: TOC issues

Hwurzburg:
no content change...this fixes almost all TOC errors (assuming previous PRs are merged)....the "common-partners-program" document does not fit into any TOC tree that I can find...its just a link from the "common-partners' page....suggestions on what to do with it to eliminate the error are welcomed...

### DIFF
--- a/common/source/docs/common-archived-topics.rst
+++ b/common/source/docs/common-archived-topics.rst
@@ -16,6 +16,8 @@ value to users with old hardware.
     NAVIO+ Overview <common-navio-overview>
     APM 2.x <common-apm25-and-26-overview>
     PX4FMU <common-px4fmu-overview>
+    PX4IO <common-px4io-overview>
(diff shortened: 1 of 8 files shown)


rmackay9:
Looks pretty good but are you sure we should delete the pixhackv5 board's page?

Hwurzburg:
the board appears to be the predecessor to the CUAV V5 Plus.....it is not referenced by any other document or TOC tree....if its never linked...its never seen...and generates a TOC error

auturgy:
I’ll update the supported autopilots page and stick a reference in to it.

Regards,

James

Hwurzburg:
Thanks..so I will eliminate the delete and merge later tomorrow...

Hwurzburg:
pixhack5 file re-instated....merged

PR #1894, opened 2019-08-11
Wiki build error fix: rangefinder labels

Hwurzburg:
Eliminates rangefinder errors caused by moving to multiple rangefinders.....no content changes...


Hwurzburg:
merging

PR #1895, opened 2019-08-11
wiki build error corrections

Hwurzburg:
remainder of plane wiki build errors that I can fix at this time...will advice on remaining handful from Sphinx expert since I cant get them to resolve in the same way that works elsewhere...

### DIFF
--- a/common/source/docs/common-gripper-servo.rst
+++ b/common/source/docs/common-gripper-servo.rst
@@ -45,7 +45,7 @@ The transmitter's auxiliary switch can control the gripper.  If using the missio
 Controlling Gripper during a mission
 ====================================
 
-The gripper position can be controlled during a mission in AUTO mode using the :ref:`DO_GRIPPER <mission-command-list#do-mount-control>` command.  Set the "drop/grab" column to 0 to close the gripper, 1 to open the gripper.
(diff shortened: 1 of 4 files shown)


common/source/docs/common-toshiba-can-escs.rst
```diff
@@ -47,7 +47,7 @@ Connection and Configuration
     :target: ../_images/toshiba-can-esc-dip-switches.png
     :width: 400px
 
-- Set the dip switches on each ESC to specify its unique number.  This number also maps the ESC to an equivalent PWM output on the back of the flight controller.  For example ESC 1 will react just like a normal PWM ESC connected to the flight controller's MAIN OUT 1.  By default this mapping also matches the :ref:`motor order diagrams <Copter:connect-escs-and-motors>`.
+- Set the dip switches on each ESC to specify its unique number.  This number also maps the ESC to an equivalent PWM output on the back of the flight controller.  For example ESC 1 will react just like a normal PWM ESC connected to the flight controller's MAIN OUT 1.  By default this mapping also matches the :ref:`motor order diagrams <Copter:connect-escs-and-motors3>`.
```
rmackay9: I think the link should remain to connect-escs-and-motors instead of -motors3..

Hwurzburg: For whatever reason, the original creates a build error....the fix does not and points to the correct location on the copter motors page....this same linking is used in several other places that navigate correctly....I dont know enough about Sphinx as to why the original fails, but the change works...must be that off site refs must use a full URL or the hyperlink on the pages subheader....it just works...

Issue #1869, opened 2019-08-02
Resolve wiki page build issues


rmackay9:
When building the wiki pages if seems like there are many issues displayed and some of them are new-ish.

Just one example:

common-appendix.rst:11: WARNING: toctree contains reference to nonexisting document u'docs/common-training-centers'

brunoolivieri:
I hope that I can try to help with those warnings from next week.

khancyr:
Ideally, we should add travis to check for new warning

brunoolivieri:
I understood that it would be a task to find errors in pages and fix them based on the warnings.

khancyr:
To have the warning, you just have to build the wiki, the warning will be displayed on the terminal.

brunoolivieri:
Yes, I was wondering to build, grep some warnings to try fix on the .rst files: https://pastebin.com/X8DFYCmp.

I did not get how Travis could help in this case.

brunoolivieri:
I did a quick check, and there were 466 warnings while building the wiki:

1) 324 warnings from "undefined labels"  (~70%)
2) 93 warnings from documents that are not in any toctree (~20%)
3) 35 warnings from parameters.rst regarding "duplicate explicit"  (~8%)

Regarding 1) This look hard to crosscheck. Is there an easy way to find the missed reference?

I mean: ACRO_EXPO is missing on copter/accro_mode.rst. Would be the better solution to remove such :ref:?

Regarding 2) Does it needs to check one by one? Is it just an alarm do try to show lost pages? I am asking it because of lots of files in this situation are easily accessible.

Regarding 3)  I believe that could be removed at once on the build scripts?

rmackay9:
ACRO_EXPO has been split into ACRO_RP_EXPO, ACRO_Y_EXPO

brunoolivieri:
I did a draft to investigate the parameters that may have had their names changed. So, the 324 pages can be summarized to the 66 labels below. Before read each page I compared the list with [parameters page](http://ardupilot.org/copter/docs/parameters.html) to get some clues. 

@rmackay9 Please, could you check if this is what you mentioned?

| Missing label                                                    | Looks like a parameter?                                                                          | TO-DO                                                                                       |
|------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
| _atc_rate_ff_enab                                                | Is it ATC_RATE_FF_ENAB?                                                                          |                                                                                             |
[...]

PR #1903, opened 2019-08-12
RNGFND1_MIN_CM reference fix

brunoolivieri:
Fix broken parameter reference due issue #1869.


Hwurzburg:
this needs to wait until multiple RNGFNDR appearss in betas....also, you should say something like,

T"he solution is, for example in the case of the first rangefinder,  to reduce the :ref:`RNGFND1_MIN_CM <RNGFND1_MIN_CM>` value (to perhaps "5").

brunoolivieri:
Done and rebased, thanks.

Hwurzburg:
Hey Bruno, do you mind if I close this? we were working at the same time and #1954 fixes this and a bunch more....its easier to close this than extract it from #1954 to avoid a future merge conflict....

brunoolivieri:
Sure, no problem. Closing...

PR #1905, opened 2019-08-12
Fix broken parameter reference

brunoolivieri:
Fix broken parameter reference due issue #1869.


brunoolivieri:
> we need to fix the broken link above this one also...

Request done.



Indeed there were two images not found in Erle server. The first one I found with small resolution and the second is ok. I added both to the wiki images folder to do not lose them again.

Maybe we could set as a standard do not link outsiders images and insert them in out folder when possible.

rmackay9:
Merged, thanks!

PR #1908, opened 2019-08-14
Initial documentation for Holybro Durandal H7 flight controller

auturgy:
I need to iterate and add some additional images, but this is enough to get started


rmackay9:
Merged, thanks!

Hwurzburg:
James....not sure this ever was tested....this has broken links and doesn't appear in the wiki...
the file is not named common-durandal-overview as referenced in the autopilots file....I dont see the images referenced loaded either...

auturgy:
Hmmm. looks like it’s missing a commit.

Regards,

James

PR #1911, opened 2019-08-16
Dev: Add more Lua docs

WickedShell:
This requires https://github.com/ArduPilot/ardupilot/pull/12055 to be accurate.


rmackay9:
Merged before the code change but it's probably ok.  By the way, I think there's a typo in one of the top lines, "Vector3f is a userdata object that holds a 2D vector" but it's probably meant to say, "3D vector"?

PR #1915, opened 2019-08-17
Remove Cube references in tradheli setup wiki

bnsgeyer:
This PR removes the Cube references placed in the Connecting and Calibration wiki for Traditional Helicopters.  More generic language is used and one of the Cube images is replaced with the Pixhack to give more controller diversity to the page.


rmackay9:
Merged, thanks!

PR #1917, opened 2019-08-19
ROS: add apriltag detection

hoangthien94:
This article adds a new wiki page for for non-GPS flight with [`AprilTag`](https://github.com/AprilRobotics/apriltag) in ROS, based on the content of [this blog](https://discuss.ardupilot.org/t/indoor-non-gps-flight-using-apriltags-ros-based/42878/1).

As briefly discussed on [ardupilot_wiki Gitter channel](https://gitter.im/ArduPilot/ardupilot_wiki), we already have [ros-aruco-detection](http://ardupilot.org/dev/docs/ros-aruco-detection.html), which would be similar to the `apriltag` implementation in terms of system description, parameters for ArduPilot, how to do ground test and flight test; but would differ on the tag and ROS sides (the tag board, package installation, how to run all the nodes).

So I think of two options:

1. Create a new wiki page separated from the existing `ros-aruco-detection` page, maybe `ros-apriltag-detection`, uses similar title and content outline for consistency between the two wiki pages, only add changes where necessary.

2. Modify current `ros-aruco-detection` page to include the new `apriltag` support, and change the title to something more inclusive like `ros-fiducial-marker-detection`.

However, once I review and see the number of necessary changes required, I think it would be more clear to separate the two pages. Furthermore, aruco is closed-source now whereas apriltag is being actively developed on, so there are more rooms for future development with apriltag.

Thus, in this PR I am going with **option 1**. Any advice or guidance would be greatly appreciated!


patrickpoirier51:
@rmackay9  @khancyr   Does everyone agree with option 1 ?

rmackay9:
option 1 sounds fine to me, I'll merge this now, thanks!

anbello:
Sorry if I comment in a merged PR.
@hoangthien94 from what you wrote it seems that an optical flow module and rangefinder are necessary, but from what I saw in my test optical flow data are not fused in EKF2 when you have COMPASS_USE 0, instead it works OK with EKF3 even with COMPASS_USE 0 but in EKF3 we have no VISION_POSITION_ESTIMATE message handling.
It seems that is not possible to have vision position estimate and optical flow fused in the EKF, at least till now.
But maybe I'm wrong and I'd be happy because it would be a good thing to have vision position estimate and optical flow fused.

hoangthien94:
Optical flow and rangefinder are not mandatory, as you said. They were part of my setup so I mentioned them and link to the [PX4FLOW wiki](http://ardupilot.org/copter/docs/common-px4flow-overview.html) for reference, but I guess it should be made a bit more clear. I will check out for any problems in the other wikis and make an edit as once.

PR #1921, opened 2019-08-20
Rover: WindVane: split into sub-pages and add NMEA

IamPete1:
This splits the wind vane page into sub-pages for each backend and add documentation for the recently added NMEA backend


rmackay9:
Looks ok from a quick review so merging..

PR #1928, opened 2019-08-21
Copter: Fix build warning (parameters)

brunoolivieri:
For this one, I think to maintain Copter 3.5 as the default parameter would stay in compliance with the rest of the page.  So I removed the ref and added the new parameter name.

### DIFF
--- a/copter/source/docs/solo_arducopter_upgrade.rst
+++ b/copter/source/docs/solo_arducopter_upgrade.rst
@@ -156,7 +156,7 @@ Value   Meaning
 =====   =======
 
 =====   =======
-:ref:`FS_BATT_ENABLE <FS_BATT_ENABLE>` controls the low battery failsafe action. The low battery failsafe kicks in when when the values set in parameters ``FS_BATT_VOLTAGE`` or ``FS_BATT_MAH`` are breached.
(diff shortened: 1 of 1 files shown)


copter/source/docs/solo_arducopter_upgrade.rst
```diff
@@ -156,7 +156,7 @@ Value   Meaning
 =====   =======
 
 =====   =======
-:ref:`FS_BATT_ENABLE <FS_BATT_ENABLE>` controls the low battery failsafe action. The low battery failsafe kicks in when when the values set in parameters ``FS_BATT_VOLTAGE`` or ``FS_BATT_MAH`` are breached.
+FS_BATT_ENABLE controls the low battery failsafe action for Copter-3.5. For the current version check the :ref:`BATT_FS_LOW_ACT <BATT_FS_LOW_ACT>` parameter. The low battery failsafe kicks in when when the values set in parameters ``FS_BATT_VOLTAGE`` or ``FS_BATT_MAH`` are breached.
```
rmackay9: I think we should say, "for Copter-3.5 (and earlier).  For Copter-3.6 (and higher) check the ...

brunoolivieri: Sure. Fixed.

rmackay9:
Merged, thanks!

PR #1933, opened 2019-08-26
Wiki update for vectored thrust tailsitters

SpaceGoast:
Added a sentence with guidance to disable Q_A_ANGLE_BOOST for vectored thrust tailsitters. Found this by crashing this weekend, and discovered this had been discussed some time ago. See tridge's advice below:
https://discuss.ardupilot.org/t/dual-motor-tailsitters/15302/500

### DIFF
--- a/plane/source/docs/guide-tailsitter.rst
+++ b/plane/source/docs/guide-tailsitter.rst
@@ -100,6 +100,8 @@ servo is channel 6, then set:
 you also need to set the right SERVOn_REVERSED values, and the right
 SERVOn_TRIM, SERVOn_MIN and SERVOn_MAX values.
 
+Q_A_ANGLE_BOOST should be disabled for vectored thrust tailsitters. Failure to disable this will cause the throttle to decrease and the nose dips, making the nose dip even further and resulting in a crash. 
+
 Vectored Gains
 ==============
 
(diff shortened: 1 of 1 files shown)


kd0aij:
@lthall Somewhat off topic, but shouldn't "tilt" be the angle between earth Z and vehicle Z here?
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AC_AttitudeControl/AC_AttitudeControl_Multi.cpp#L294

If so, the cosine is just get_rotation_body_to_ned().c.z

tridge:
why would it make the throttle decrease? It should make it increase
@kd0aij maybe we should discuss?
If we do need to disable this then we should just change the defaults in quadplane init code, but I'm not convinced that disabling is correct
@kd0aij your comment on using c.z seems to be just an optimization to avoid a multiply? They should be the same thing

kd0aij:
@tridge Your post on the tailsitter thread (referenced by @SpaceGoast) said:
"I’ve tracked this down to the angle-boost code. If you set Q_A_ANGLE_BOOST=0 then I think you’ll find that you get the full throttle range at any angle.
The angle-boost code doesn’t really make sense for a vectored tailsitter, as it is assuming thrust is to the front of the vehicle."
I didn't attempt to reproduce the problem myself. 

And that trig identity was not obvious to me, so I was wondering whether I was interpreting the tilt angle correctly.

SpaceGoast:
@tridge I found in my test that when trying to take off from a belly stance, the plane had almost no throttle output even when I commanded 100% throttle in QStabilize. If I stood the plane up on its tail I could take off just fine, but as the nose fell closer to horizontal, the throttle would cut again. This issue went away when I disabled that parameter.

PR #1937, opened 2019-08-27
Rover: Duplicated ref and small tip

brunoolivieri:
The main page reference was duplicated. The tip enhances the pointer to the tunning page.


rmackay9:
Merged, thanks!

PR #1939, opened 2019-08-27
Dev: Fix DroneKit URL

brunoolivieri:
Fixing Dronekit URL


auturgy:
Please check with @CraigElder before merging this one.

auturgy:
Sure: the new URL does have a clone of the docs, it’s just that they’re “unofficial” and I don’t want linking to them to have unintended consequences for Craig. I think it should be fine, but I’d like him to ok this.

Regards,

James

rmackay9:
@CraigElder, any thoughts on this one?  You're OK with the new link or is there another?

rmackay9:
I've gone ahead and merged this because it's a better link but hopefully Craig can respond as well and if necessary we can change the link

PR #1943, opened 2019-08-30
fix plane home wiki build errors

Hwurzburg:
after Peter and I have failued to figure out why the ref plane:home doesnt work linking to the plane home page, I just added direct url links to get rid of the build errors....


Hwurzburg:
Merging

PR #1946, opened 2019-08-30
Add CUAV V5 to discontinued list,resolving build error

Hwurzburg:
Since CUAV no longer lists this, nor their distis, and CUAV V5plus and nano replace it....it should go in discontinued list and eliminate it as a floating document


rmackay9:
Merged, thanks!

PR #1947, opened 2019-08-30
fix copter wiki build errors

Hwurzburg:
several TOC related errors.....starting work on copter build errors....
no content change


Hwurzburg:
pushed more non-content error corrections

rmackay9:
Merged, thanks!

PR #1950, opened 2019-08-30
Fix SportMode wiki build errors

Hwurzburg:
fix param links


rmackay9:
Merged, thanks!  took me a while to see what had changed!

PR #1964, opened 2019-09-02
fix rover wiki learn_ch build error

Hwurzburg:


common/source/docs/common-rcmap.rst
```diff
@@ -53,7 +53,7 @@ elevator (throttle) is on channel 2, then the settings should be:
 make them as long as they are not channels 1 and 2.
 
 For Rover you should also select the receiver channels that will be used
-to control the Waypoint select (:ref:`LEARN_CH <rover:LEARN_CH>`)
+to control the Waypoint select (:ref:`CH7_OPTION <rover:CH7_OPTION>`)
```
rmackay9: I think we should probably just remove this sentence regarding the "waypoint select" but leave the sentence about setting the "Mode select".  the ''LEARN_CH" is mentioned again just below it seems so that should go as well.

It was reasonable to have these learn-ch notes back when each feature had a parameter to select which channel should be used but now we've turned that upside-down so that each channel is assigned a feature.

rmackay9:
Merged, thanks!

common/source/docs/common-rcmap.rst
Hwurzburg: Will do..

PR #1969, opened 2019-09-03
New mode is no longer declared in "defines.h"

rishabsingh3003:
New mode is no longer declared in "defines.h". It is now declared it mode.h itself. This PR fixes it.


peterbarker:
Merged, thanks!

PR #1974, opened 2019-09-04
fix wiki build error in DEV

Hwurzburg:


rmackay9:
MAG_ENABLE still exists in the stable branches so I think this should be delayed until the next release or we should just remove the reference for MAG_ENABLE for now.

Hwurzburg:
de-ref'd...thanks

rmackay9:
.. looks like a little typo has crept in though, it's MAG_EN instead of MAG_ENABLE..

Hwurzburg:
duh.....

rmackay9:
Merged, thanks!

PR #1982, opened 2019-09-06
Add a Chinese version of trad-heli-tuning part

MasterBoBo:
A Markdown version of traditional-helicopter-tuning part in Chinese


rmackay9:
We discussed this one and we think we need to setup a system like https://translatewiki.net/.  We are thinking about having a separate call to discuss how we proceed.

MasterBoBo:
> We discussed this one and we think we need to setup a system like https://translatewiki.net/.  We are thinking about having a separate call to discuss how we proceed.
That would be awesome, Thanks for your time.

cuhome:
@rmackay9 Will the Chinese version of the wiki be implemented? We are willing to write a Chinese version for the wiki.

PR #1983, opened 2019-09-06
added a page on Vicon setup for copters

tridge:


rmackay9:
I think this should be put under common-optional-hardware and renamed to "Vicon for Non-GPS navigation" to fit in with similar pages like "OptiTrack for Non-GPS navigation".  It's possible that we have enough of these that we should group all these pages under a single "Non-GPS navigation" landing page.

- Marvelmind for Non-GPS Navigation
- Pozyx for Non-GPS navigation
- OptiTrack for Non-GPS navigation
- VIO tracking camera for Non-GPS navigation
- ZED for Non-GPS navigation

brunoolivieri:
If you prefer,  you may merge this PR itself, and I can rearrange the files in a new landing page.

rmackay9:
@brunoolivieri, that would be good thanks!

Hwurzburg:
@brunoolivieri merging....thanks for volunteering to fix this after the merge....

brunoolivieri:
PR #2001  tries to covers that.

PR #1984, opened 2019-09-06
Add mention of Qplane in booster motor page

Hwurzburg:


Hwurzburg:
probably should make this a common page at some point...

Hwurzburg:
MErge

PR #1987, opened 2019-09-07
Fix wiki build errors in DEV

Hwurzburg:
final build error fix in Dev (that I can fix...the code block error is mystifying me still)


Hwurzburg:
Merged

PR #1988, opened 2019-09-07
Add differential thrust in Plane

Hwurzburg:
adding to table...expanding explanation of output type


Hwurzburg:
built tested Merging

PR #1989, opened 2019-09-07
Add simple overview of operation before mandatory setup

Hwurzburg:
I know this is obvious and pretty simplistic to those of us familiar with Ardupilot. But someone having no autopilot experience or coming from iNav/Betaflight probably has no idea what goes in and what goes out of the system...
I have fielded questions on the forum about the difference between rc mapping, rc functions, and servo functions...
Maybe its not critical info, but I thought it couldn't hurt


Hwurzburg:
Fixed per @tridge  suggestions/comments.....Merged

PR #1993, opened 2019-09-08
Update in "adding a new drive mode" documentation

rishabsingh3003:
Mapping between the mode's number and the instance of the class for rover now done in "mode.cpp" instead of 'control_modes.cpp'. This PR fixes it.


peterbarker:
Merged, thanks!

Issue #1994, opened 2019-09-08
HeliQuads FRAME_CLASS setting throws PreArm error


rdiverdi:
I am trying to build a HeliQuad and after getting all the software running, I am getting this error:
`Critical: PreArm: check firmware or FRAME_CLASS` 
(as displayed by QGroundControl)
The error seems to be specific to the `HeliQuad` (13) setting (I did not test all settings, but I don't get the error with `Quad` (1) or `DodecaHexa` (12))

My setup is a BeagleBone Blue running a debian image, and I downloaded and compiled the current ArduCopter 3.6 branch of ardupilot (3.6.10) by following [this ArduPilot for beginners](https://sudonull.com/posts/2177-ArduPilot-for-beginners-Installation-and-configuration-on-BeagleBone-Blue) guide. I also checked the [HeliQuads (Variable Pitch Multicopters) page](http://ardupilot.org/copter/docs/heliquads.html) to check the settings. Currently I have no motors or servos actually connected - my only peripheral is a receiver - I am just trying to get the software to arm to confirm it is set up correctly.
And just for completeness, my ground station is a PC running Ubuntu 16.04, I am using QGroundControl v3.5.4, and the comm link is over udp.

After compiling the software, setting my frame type, and calibrating the compass, accelerometer, and radio, I tried to arm the board and the only message I got was `Critical: PreArm: check firmware or FRAME_CLASS` 
I tried setting the `FRAME_TYPE` to `QUAD` and everything worked (I was able to arm the board)

auturgy:
Heli quad requires tradheli as the base firmware, so you need to run the arducopter-heli binary (rather than arducopter). Compiling for copter generates both, so it should already be waiting for you :)

Regards,

James

rdiverdi:
Thank you!

PR #1999, opened 2019-09-11
Note for HeliQuad

brunoolivieri:
Note from issue #1994


Hwurzburg:
thanks Bruno....just moved the note higher up....its important....also a minor reword....the rationale was to  not imply that compiling was necessary...

PR #2001, opened 2019-09-11
Non-GPS pages re-arrange.

brunoolivieri:
This intends to re-arrange the Non-GPS pages, as discussed in issue #1983.

ros-vio-tracking-camera and ros-cartographer-slam were removed from because in already in DEV environment.

The pages are filtered for Rover and Copter only.


rmackay9:
Looking good, thanks!

PR #2002, opened 2019-09-12
dev: switched to RealFlight9 for RealFlight instructions

tridge:
RF9 comes with a menu item to enable FlightAxis support


Hwurzburg:
Merged

PR #2007, opened 2019-09-13
Expand explanations of ACRO and STEERING modes

Hwurzburg:
@rmackay9  I hope I got the STEERING mode explanation correct...after our conversation, I had a few more questions but wrote this up making a few assumptions about how lat accel would be handled...let me know what I guessed wrong...

### DIFF
--- a/rover/source/docs/acro-mode.rst
+++ b/rover/source/docs/acro-mode.rst
@@ -4,9 +4,10 @@
 Acro Mode
 =========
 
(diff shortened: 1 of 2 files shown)


rmackay9:
Really good explanation I think.  One small correction is that WP_OVERSHOOT is not used in Steering anymore.  It was in older modes.

Hwurzburg:
Will fix...tnx

PR #2009, opened 2019-09-14
Add information about setting up non-spring throttle in Rover

Hwurzburg:
Also a note about throttle setup with non-spring loaded throttles ...for the Rover site only


Hwurzburg:
changes made per call....merging

PR #1979, opened 2019-09-06
Major update to SERVOx_FUNCTION page

Hwurzburg:
Changed page to apply to ALL vehicles and to allow logical path to find needed output functions
- grouped by vehicle or function
- noted for which vehicle(s) the function applies
- added explanations to links for every function

once merged, I intend to add a DEFAULT ASSIGNMENTS table for SERVO1-8 listing every vehicle type and what it assigns as defaults to each output...

PR #2017, opened 2019-09-17
tradheli update to RC Output mapping wiki

bnsgeyer:
this PR follows up on work by @Hwurzburg merged from this PR #1979.  I'm not quite sure I understand how this page is structured.  Are the tables (except for the last which covers defaults) for the main vehicle types (i.e. plane, copter, rover)?  It seems like individual frames are included like quadplane.  Should I generate another table for tradheli?  I included it in the Copter/Quadplane table.   I think I should add Trad heli to that title.  

It is also confusing that copter is listed in the Plane available functions table.  

Looking for guidance on the titling for the Copter function table.  I included the tradheli frames in the table.


Hwurzburg:
Bill, LGTM....heli is part of the ArduCopter code base...hence my grouping it with  Copter...I dont have an issue if you want to expand that table's title...

I tried to group outputs in logical groups, avoiding duplications as best I could...but I wanted to be complete in terms where each output is available in terms of major firmware group, even if they get duplicated.....quadplane is a bit unique having separate plane and copter operational functions that why its called a  Copter/QuadPlane table vs Copter/Plane table or just putting quadplane's copter function outputs all in the plane table...

If you have plane its the Plane table, a Copter (and it includes heli,etc.) that table, Rovers, etc....the half-breed Quadplane is the exception to the rule...often the copter and plane hardware is almost independent....could have gone either way...

Merging....

Issue #652, opened 2017-01-03
Add how to get involved page


rmackay9:
add a "how to get involved" page which outlines the broad areas that people can get involved in including software development, wiki updating, beta testing, administration (i.e. taking meeting notes) and PR (i.e. blogging about ardupilot)

brunoolivieri:
I did a brief draft in PR #2020

Hwurzburg:
Closing since 202 is about to be merged once bruno does the  PR changes discussed in our wiki call

PR #2020, opened 2019-09-18
Add get involved page

brunoolivieri:
This PR tries to solve issue #652. Would be something like that @rmackay9?

I still need to include it on some toctree. Where could I link it? May be in vehicles home pages?

### DIFF
--- a/antennatracker/source/index.rst
+++ b/antennatracker/source/index.rst
@@ -39,6 +39,13 @@ Below is `Canberra UAV's tracker used during the 2014 OutBack Challenge <http://
 
 This manual will guide you through setup and configuration process.
 
+
+How to get involved with ArduPilot?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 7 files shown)


Hwurzburg:
Bruno, good idea! I think it should be integrated somehow on the ardupilot.org/index.html page though

rmackay9:
By the way, we have this "contact us" page which has some similar info on it. http://ardupilot.org/dev/docs/common-contact-us.html

brunoolivieri:
Indeed the "contact us" pages have a similar idea. Makes sense create another?

Anyway, I updated the PR  with your comments.

What do you think?

brunoolivieri:
Update: As discussed in the last Wikicall, there were many repetitions. We discuss in do not add a new page, but add a link to actual pages.

brunoolivieri:
I guess the [PR 10 on sphinx theme](https://github.com/ArduPilot/sphinx_rtd_theme/pull/10) handles the discussed in last wikicall.

The we merge that I think we could close this PR ignoring these proposals.

Hwurzburg:
yes....we will close this one...the Sphinx PR is best way to go...I requested a small change....I can test it once its done....

brunoolivieri:
sure, thanks

PR #2021, opened 2019-09-18
Fix build warning

brunoolivieri:
It removes some warnings from building the wiki. 

Some code blocks were flagged as python, but they have linux commands.


Hwurzburg:
Merged

brunoolivieri:
That was trick. I tried some python/sphinx changes so I notice that wasn´t about that...

Cheers.

PR #2029, opened 2019-09-23
Rover: add mention of MOT_THR_MIN in setup and loiter pages

Hwurzburg:


Hwurzburg:
no new content (was in balance-bots only), just placing it in more locations...merging

PR #2039, opened 2019-09-30
Link is dead, hence removed.

joygetsit:


brunoolivieri:
I wonder if would be better updated with a proper link instead of removal. Maybe something like https://www.ebay.com/itm/CP2102-USB-2-0-to-TTL-UART-Module-6Pin-Serial-Converter-STC-Replace-FT232-Module-/381374541932  or https://www.sjoerdlangkemper.nl/2019/03/20/usb-to-serial-uart/

auturgy:
We should check to see if a Partner has a relevant product 

Regards,

James

joygetsit:
https://www.sjoerdlangkemper.nl/2019/03/20/usb-to-serial-uart/ looks like a good overview of what one can use. And this link is a good example https://www.amazon.com/Prolific-PL2303-USB-TTL-Module/dp/B008958S7A

brunoolivieri:
Indeed, @auturgy is right, and it would be better using a partner link.

I looked in partners' links, and the majority of them have RTF/PNP solutions or services. I found two pieces of hardware* that can be used. However, the Bask Aerospace looks "closed hardware," and the Blue Robotics is not ready to use and both cost  3x or 4x priced while comparing with generic hardware.

*links
https://baskaerospace.com.au/store/telemetry/ftdi-cable-3-3v/
and 
https://bluerobotics.com/store/comm-control-power/tether-interface/bluart-r1-rp/

The PR "as-is" LGTM as well if it replaces the link for a partner would be as well.

Hwurzburg:
changed to hyperlinks and merged....thanks

PR #2040, opened 2019-09-30
Dev: Changed to correct path for script

bromeara:
Directory Tools/environment_install/ no longer exists. At least on master and the latest stable version of copter. https://github.com/ArduPilot/ardupilot/blob/Copter-3.6.10/Tools/scripts/install-prereqs-ubuntu.sh


bromeara:
Does not apply to master I jumped the gun

PR #2052, opened 2019-10-08
Rover Follow mode: update for offset reset upon mode entry

Hwurzburg:


rmackay9:
Looking good.  I haven't delayed the merge cuz I think we will do another beta release soon.

PR #2054, opened 2019-10-08
Fix Wiki build errors for base index page

Hwurzburg:
remove pages not referenced in Ardupilot site build


Hwurzburg:
Built,tested,merged

PR #2060, opened 2019-10-10
update sitl call commands

fagoth:


Hwurzburg:
The description as is currently provided is correct....if you start sim_vehicle from a non-vehicle directory THEN you must specify the vehicle type....feel free to expand the current instructions, but the they are not incorrect...

fagoth:
> The description as is currently provided is correct

This is windows page, not linux... If you run commands 
```
cd ~/ardupilot/ArduCopter
sim_vehicle.py --map --console
```
bash says that sim_vehicle.py is unknown command.

peterbarker:
I'd like to take this to a devcall.

I actually invoke sim_vehicle.py from the ardupilot root directory: `./Tools/autotest/sim_vehicle.py`.  I really don't like the implicit vehicle-from-directory-name stuff, but it probably has to stay as our vehicle maintainer's muscle memory is long.

That being said, encouraging more people to use the magic-from-vehicle-directory isn't necessarily a good thing.

But I think a devcall discussion would be a good idea.

peterbarker:
On Thu, 10 Oct 2019, fagoth wrote:

> cd ~/ardupilot/ArduCopter
> sim_vehicle.py --map --console
> 
> bash says that sim_vehicle.py is unknown command.

That's actually because the install scripts are supposed to corrupt... I 
mean adjust... your PATH to find sim_vehicle.py

fagoth:
> I actually invoke sim_vehicle.py from the ardupilot root directory: `./Tools/autotest/sim_vehicle.py`

Do you specify a vehicle via the -v param?
What happens in cygwin for me:

$ ./Tools/autotest/sim_vehicle.py
SIM_VEHICLE: Start
SIM_VEHICLE:
** Is (ardupilot.git) really your vehicle type?
Perhaps you could try -v ArduCopter|AntennaTracker|APMrover2|ArduSub|ArduPlane
You could also try changing directory to e.g. the ArduCopter subdirectory

SIM_VEHICLE: Killing tasks

dev/source/docs/sitl-native-on-windows.rst
```diff
@@ -29,12 +29,12 @@ Running SITL
       
 #. Open a Cygwin64 terminal from the start menu
 
-#. Navigate to one of the vehicle directories (in this case Copter) and call ``sim_vehicle.py`` to start SITL.  If you get a windows security alert for the the firewall, allow the connection. 
+#. Navigate and call ``sim_vehicle.py`` to start SITL.  If you get a windows security alert for the the firewall, allow the connection. Specify a vehicle type after ``-v`` parameter.
```
OXINARF:
```suggestion
#. Call ``sim_vehicle.py`` to start SITL.  If you get a Windows security alert for the the firewall, allow the connection. Specify a vehicle type (ArduCopter | ArduPlane | APMrover2 | ArduSub) after ``-v`` parameter.
```

dev/source/docs/sitl-native-on-windows.rst
```diff
@@ -29,12 +29,12 @@ Running SITL
...
 
    ::
 
-       cd ~/ardupilot/ArduCopter
-       sim_vehicle.py --map --console
+       cd ~/ardupilot/Tools/autotest
+       ./sim_vehicle.py --map --console -v ArduCopter
```
OXINARF:
```bash
       cd ~/ardupilot
       ./Tools/autotest/sim_vehicle.py  -v ArduCopter --map --console
```

This is my personal preference. Executing SITL will create several files and hiding them away in autotest folder doesn't seem the best option to me.

fagoth: Applyed your propose... It's almost the same as by @peterbarker

fagoth:
Path fixed by @rmackay9, closing...

Issue #1704, opened 2019-04-09
Combine training centers and school pages


rmackay9:
We apparently have separate training centers and school wiki pages.  we should combine these.

auturgy:
Originally they were intended for different things: flight/operations training centres, and code/development schools.
I think the distinction made sense, until we cross merged and messed it up.
Up to you though.

Regards,

James

brunoolivieri:
I can see the Training Centers in plane/copter wikis but Schools only in dev wiki. None of the others have those links. It is correct? Should be updated?

rmackay9:
@brunoolivieri, yes, they should all be consistent I think.

Hwurzburg:
It would seem that code schools and flight training should be separated... and both pages listed in all wikis...I will do that if you guys agree...

rmackay9:
I think we should just put them together on a single page with some way to see what each school offers.  Otherwise I think we will end up with duplicates as some schools offer both... and i think from the user's point of view they are all schools.

Hwurzburg:
PR #2066 combines

Hwurzburg:
closing

PR #2066, opened 2019-10-11
Combine schools and training into one page

Hwurzburg:
addresses issue #1704


rmackay9:
So we just need to be sure that ardupilot.org's About >> Schools link continues to work (i.e. maybe it needs to be redirected or removed)

Hwurzburg:
where do all these banner menus reside?

rmackay9:
@Hwurzburg, I don't know but @OlivierJB should know I think.

OlivierJB:
Those menus are on ardupilot.org within the concrete5 framework.

The About/schools link to Ardupilot,org currently points to http://ardupilot.org/dev/docs/schools.html, I could change that to the new link, looks like http://ardupilot.org/dev/docs/commong-training-centers.html ?

Hwurzburg:
I think we need to have a call specifically to discuss the entire approach to top banners/ duplications/ links into specific wiki sites, etc.
I will ask on the dev skype when Olivier,Tridge,Randy and myself could get together...

Hwurzburg:
@OlivierJB so at this point, I think changing the main page banner menu ABOUT/Schools to point to this new page is what we should do....when that is done, I will merge this update...and remove the old page...thanks

OlivierJB:
Henry, sounds good. Let's try to coordinate, I'll try to catch you on mumble.

Hwurzburg:
WIKI call tonight at 2300UTC?

OlivierJB:
That'll work for me. See you there.

Hwurzburg:
thank @OlivierJB Merged

PR #2072, opened 2019-10-12
Documentation on configuring the static notch and dynamic harmonic notch

andyp1per:


andyp1per:
@bnsgeyer are you able to comment on how this needs to be extended for TradHeli?

bnsgeyer:
@andyp1per yes I will need to provide input for TradHeli users.  I will try to push to your PR.

bnsgeyer:
@andyp1per I tried pushing my commit to your PR branch but it wouldn't let me.  So maybe you can cherry-pick it into your branch.  Here is the link to my [branch](https://github.com/bnsgeyer/ardupilot_wiki/tree/patch-2).

andyp1per:
Thanks @bnsgeyer I have pulled it in

bnsgeyer:
@andyp1per  I just made another commit.  I realized that I didn't explain the difference in setting up the static and dynamic notch.  Could you check this commit.  For some reason when I preview it, the page doesn't show properly.  it just shows the code version.  If you can fix it, please do and pull into your PR.
Thanks!

andyp1per:
HI @bnsgeyer thanks I've pulled in your changes and added some diagrams

andyp1per:
I'm going to switch this to review as I think it is pretty close

Hwurzburg:
Is this intended only for copters or planes, as well? wie will need to decide which page it is called form and in what TOC tree...

bnsgeyer:
@Hwurzburg and @andyp1per I think there needs to be a wiki on vibrations and filtering.  I think there is a lot of confusion on what these filters actually affect.  Does IMU filtering affect the EKF? (I believe the answer is no).  We know it affects the signals used by the flight controls.  So at a minimum I think a wiki needs built for how to use filtering (Low Pass and Notch) to provide better signal quality for the flight controls.  Maybe this goes into the tuning section?  Leonard mentions setting Low Pass filters in this section and even gives settings for the filters based on prop size.   So I think the tuning sections could reference this wiki as part of a filtering wiki.  I will have to figure out how I plan to address it and where in the tradheli wiki.

andyp1per:
Currently the dynamic notch is only for copters and I can't see the static notch being very useful for anything other than copters and copter-like things. The IMU filtering *does* affect the EKF indirectly, since IMU filtering affects the output of the PID loop which then mechanically affects the input of the EKF.

rmackay9:
I think we should probably link this from the Advanced Configuration area..

andyp1per:
@Hwurzburg this will also need changing when @rmackay9 merges https://github.com/ArduPilot/ardupilot/pull/12544

andyp1per:
@rmackay9 @Hwurzburg I've linked the pages in to a couple of sections and added information about INS_HNTCH_MODE. Be great if you could review.

common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -0,0 +1,146 @@
+.. _common-imu-notch-filtering:
+
+============================================================================
```
Hwurzburg: need to add [copywiki destination="copter"]  since this is only site that references this document and creates wiki build errors otherwise for the other sites

common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -0,0 +1,146 @@
...
+.. image:: ../../../images/imu-batchsampling-fft-mp3.png
+    :target:  ../_images/imu-batchsampling-fft-mp3.png
+    :width: 450px
+
+In most cases, only the rotor speed frequency, the second harmonic and 4th harmonic.
+
+Static Harmonic Notch
+======================
+If there is not an RPM sensor, the harmonic notch reference frequency will be set and will not change.  If the rotor speed in RPM based on settings of the ESC governor or the internal RSC governor is known, then that value converted to Hz would be used to set the harmonic notch reference frequency.  So here are the settings for the static harmonic notch.  Note that the parameter INS_HNTCH_REF is set to zero for the static harmonic notch.
+
```
Hwurzburg: link INS_HNTCH_REF

common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -0,0 +1,146 @@
...
+-----------------
+Multicopter Setup
+-----------------
+
+Pre-Flight Setup
+=================
+
+Arducopter allows you to select the mechanism used for controlling the harmonic notch frequency. By default this mechanism is :ref:`throttle-based<common-imu-notch-filtering-throttle-based-setup>` since that will work on all copters. However, for copters with an rpm sensor or BLHeli ESC telemetry support, setup can be radically simpler:
+
```
Hwurzburg: I think you added this link:  _common-dshot-blheli-telemetry: so it could be a hyperlink on the "BLHeli ESC....." wording...otherwise its not used anywhere that I have found...

Hwurzburg:
Otherwise it builds/looks good...we have a problem with the wiki math checker tagging an error in the build, but I think its a build checker error, not a doc error....will check with @peterbarker  about that

andyp1per:
> 
> 
> Otherwise it builds/looks good...we have a problem with the wiki math checker tagging an error in the build, but I think its a build checker error, not a doc error....will check with @peterbarker about that

Thanks - I'll fix your comments. Re the maths stuff, I found that LaTeX maths does not work even though all the RST docs say it should. In fact I find our environment seems to not work in a myriad of ways I don't quite understand. If you have a way of making the equivalent of \sqrt{A/B} work that would be awesome as I could not get it to work at all!

Hwurzburg:
I suggest just this...looks fine and avoids the MATH submodule issue...
*hover_thrust* * SQUAREROOT(*min_freq / hover_freq*)

common/source/docs/common-imu-notch-filtering.rst
andyp1per: Fixed

common/source/docs/common-imu-notch-filtering.rst
andyp1per: Fixed

common/source/docs/common-imu-notch-filtering.rst
andyp1per: Fixed

Hwurzburg:
small tweak...LGTM

andyp1per:
Hahaha, I just made the same change! Great minds and all that

andyp1per:
squashed

Hwurzburg:
oops...still has the title ====s wrong that I had fixed....they should be as long as the text to prevent wiki build errors being generated...

andyp1per:
> 
> 
> oops...still has the title ====s wrong that I had fixed....they should be as long as the text to prevent wiki build errors being generated...

Do you mean the ones that are slightly too long? All the others look correct

Hwurzburg:
yep...should match length of text

andyp1per:
> 
> 
> yep...should match length of text

Ok fixed

rmackay9:
Merged, thanks!

PR #2073, opened 2019-10-12
Update git-rebase instruction on pushing to origin

Hwurzburg:


Hwurzburg:
@tridge @rmackay9 per our conversation during the wiki call....this change is correct for the workflow currently in place in the WIKI....if we want to change that flow across all the GIT docs in the wiki, an issue can be raised and discussed...until that time, I am merging this as its in line with current docs and addresses an issue raised on gitter by a new coder....

Issue #2081, opened 2019-10-12
ERLE Robitics


Hwurzburg:
Apparently this company no longer exists...should we archive and remove the numberous references to it?

rmackay9:
@Hwurzburg, yes, I think that's probably fine. thanks!

Hwurzburg:
Are there ANY Linux flight controllers being made? no Navio no ERLE...if not, should LINUX stuff be archived also?

Hwurzburg:
PR #2084 addresses this

proficnc:
I’d personally prefer this stayed... as the designer of the PXF, I see it as an important step in our journey.

With the new BB out, I’m actually considering another revision of the PXF

Hwurzburg:
merged

PR #2084, opened 2019-10-13
Remove/Archive ERLE references 

Hwurzburg:
addresses #2081

### DIFF
--- a/common/source/docs/common-archived-topics.rst
+++ b/common/source/docs/common-archived-topics.rst
@@ -26,6 +26,7 @@ value to users with old hardware.
     External Compass (Magnetometer) <common-external-magnetometer-for-improved-performance>
     Recording and Playing Back Missions <common-recording-and-playing-back-missions>
     AttoPilot Current/Voltage Sensor <common-using-a-current-sensor>
+    Erle-Brain2 Wiring Quick Start <common-erle-brain2-wiring-quick-start>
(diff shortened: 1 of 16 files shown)


khancyr:
Hello,

I am pretty sure Navio2 isn't discontinued yet and still supported by Emlid

Hwurzburg:
You are correct...got on a roll...will fix

proficnc:
Please keep the PXF stuff, we are considering a new run of these boards now that the new BB is out. 

PXF is a ProfiCNC product that was built under license by ERLE

Hwurzburg:
@khancyr fixed Navio2 mistake....sorry...

@proficnc  No PXFmini stuff is removed (or will be...at most the links on a few remaining pages will be removed and moved to the archive list), currently the PXFmini is listed as a discontinued board...this PR only adds the archived label to two docs, but does not actually remove the only two links from  vehicle assembly pages to the PXFmini (which actually should be removed and moved to the archived documents list at some point since the product is not available, nor has concrete plans for availability)....in any case, if the PXFmini becomes available again, it can be quickly and easily re-highlited as an available device on the flight controllers and assembly pages, or pages even modified if changes to the product are made...we want to make sure that any available ArduPilot compatible product is promoted and highlited...
no documents are removed, only links relocated for inactive pages...

PR #2090, opened 2019-10-15
Serial options

Hwurzburg:


vierfuffzig:
@Hwurzburg imho it would be great to point out that the Rx / Tx inversion bit is only supported on F7 boards by default, plus a limited number of F4 boards with a GPIO-controlled hardware inverter. afaik these are
- DrotekP3Pro
- fmuv4
- mindpx-v2

i think there's a noticeable amount of confusion regarding inversion support on F4 boards if you skim through forum support requests regarding frsky telemetry.

Hwurzburg:
I was wondering about that....the fmuv4 and mindpx-v2 are not documented as boards in the wiki...
although I think the FMUv4 is the I/O processor on many boards...it would be helpful if there was an exact list that corresponds to the Autopilot Hardware page, for me to include in this ....

vierfuffzig:
afaik only pixracer and drotek pixhawk 3 pro follow fmuv4 design standard. however, the drotek p3pro doesn't use fmuv4 hwdef in ardupilot's build environment, likely due to some hardware differences (MCU / IOMCU).

common/source/docs/common-serial-options.rst
```diff
@@ -0,0 +1,24 @@
...
+Serial Port Hardware Options
+============================
+
+This page describes the hardware configuration options for the serial ports.
+
+
+SERIALx_OPTIONS Parameter
+=========================
+
+Every serial port has in addition, to its baud rate (``SERIALx_BAUD``) and protocol format (``SERIALx_PROTOCOL`), the ability to invert its RX input and/or TX data, operate in half-duplex mode, and/or swap its RX and TX inputs.
```
brunoolivieri: I guess the ``SERIALx_PROTOCOL**`** is missing a **`**.

Hwurzburg:
I believe that the omnibusf4 AIO V6 has controllable inverters on UART6, but that port is normally GPS
OmnibusF4 Pro and NanoV6 have a fixed inverter on SBUS->UART1RX....I probably just copied the original F4Pro stuff when I did the Nano and thats why its there...but I removed the PC8/9 inverter control stuff when I expanded the PWM output to 8 on the OmnibusF4Pro....the original omnibusF4 AIO was first board into CHIBIOS I think...after V3 it had the controlled inverters on UART6...hence it was in the HWDEF that other Omnibus boards copied 

I dont think the F35Lightning is controllable, but may be wrong....

to avoid confusion, I propose that we just say its F7/H7 boards for now(since they are not board dependent and valid for all UARTs, correct?).. since its really only useful for FRSky, having only one pin invertable, like Pixracer UART8TX is not really useful.

Hwurzburg:
I pushed a change to add a note to which boards this option  applies, based on my proposal...

PR #2111, opened 2019-10-21
Change ArduCopter/Plane to just Copter Plane

Hwurzburg:


Hwurzburg:
non functional change, merged

PR #2114, opened 2019-10-21
Copter: Refactor prearm to a common page, and update for 4.0

Hwurzburg:
will move Plane and Rover to this new page once merged....

I guessed at some of the new pre-arm failure messages in terms of their causes...let me know and I can fix as required....


rmackay9:
Merged, thanks!

PR #2117, opened 2019-10-22
Copter: add vibration failsafe page

Hwurzburg:


rmackay9:
I don't see the page itself.. I see the link but not the page..

Hwurzburg:
yeah, cascade failure....I usually forget  to git add a new page, until after I do the PR and re-check it...but this time, I had a build problem that Tridge helped me with...it involved a "clean"...and I lost it....just re-created and pushing it now

PR #2123, opened 2019-10-23
added parallel build of wiki

tridge:
defaults to building all wikis in parallel. You can choose level of
parallism with the --parallel option. Note that this will use quite a
lot of memory

### DIFF
--- a/update.py
+++ b/update.py
@@ -30,7 +30,9 @@ import re
 import os
 from codecs import open
 import subprocess
+import multiprocessing
 import shutil
+import time
 
 
 DEFAULT_COPY_WIKIS =['copter', 'plane', 'rover']
@@ -43,8 +45,9 @@ COMMON_DIR='common'
 import argparse
 parser = argparse.ArgumentParser(description='Copy Common Files as needed, stripping out non-relevant wiki content')
 parser.add_argument('--site', help="If you just want to copy to one site, you can do this. Otherwise will be copied.")
-parser.add_argument('--clean', default='False', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
-parser.add_argument('--cached-parameter-files', default=False, help="Do not re-download parameter files", type=bool)
+parser.add_argument('--clean', action='store_true', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
+parser.add_argument('--cached-parameter-files', action='store_true', help="Do not re-download parameter files", default=False)
+parser.add_argument('--parallel', type=int, help="limit parallel builds, -1 for unlimited", default=-1)
 parser.add_argument('--destdir', default="/var/sites/wiki/web", help="Destination directory for compiled docs")
 args = parser.parse_args()
 #print(args.site)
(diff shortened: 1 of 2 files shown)


update.py
```diff
@@ -43,8 +45,9 @@
 import argparse
 parser = argparse.ArgumentParser(description='Copy Common Files as needed, stripping out non-relevant wiki content')
 parser.add_argument('--site', help="If you just want to copy to one site, you can do this. Otherwise will be copied.")
-parser.add_argument('--clean', default='False', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
-parser.add_argument('--cached-parameter-files', default=False, help="Do not re-download parameter files", type=bool)
+parser.add_argument('--clean', action='store_true', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
+parser.add_argument('--cached-parameter-files', action='store_true', help="Do not re-download parameter files", default=False)
```
OXINARF: Nitpicking: the default argument can be removed since that's added by the new action `store_true`.

update.py
```diff
@@ -84,20 +87,44 @@ def fetchparameters(site=args.site):
...
+    procs = []
+
+    while len(done) != len(wikis):
+        wiki = list(wikis.difference(done))[0]
+        done.add(wiki)
         if site=='common':
             continue
         if not site==None and not site==wiki:
             continue
         print('make and clean: %s' % wiki)
```
OXINARF: This is repeated now, we should remove it here.

update.py
```diff
@@ -43,8 +45,9 @@
 import argparse
 parser = argparse.ArgumentParser(description='Copy Common Files as needed, stripping out non-relevant wiki content')
 parser.add_argument('--site', help="If you just want to copy to one site, you can do this. Otherwise will be copied.")
-parser.add_argument('--clean', default='False', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
-parser.add_argument('--cached-parameter-files', default=False, help="Do not re-download parameter files", type=bool)
+parser.add_argument('--clean', action='store_true', help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).")
+parser.add_argument('--cached-parameter-files', action='store_true', help="Do not re-download parameter files", default=False)
+parser.add_argument('--parallel', type=int, help="limit parallel builds, -1 for unlimited", default=-1)
```
OXINARF: I'd prefer that this defaulted to 1, otherwise this will unexpectedly try to use all resources of the machine.

update.py
tridge: I know, I just like making it explicit

update.py
tridge: in discussion with henry and randy we thought default parallel was better

tridge: after looking at the amount of free memory on ardupilot.org I've changed my mind, and made it default to 1

update.py
tridge: done

tridge:
now defaults to single process, so unchanged behaviour
merging

PR #2124, opened 2019-10-23
partners: added vimdrones

tridge:


rmackay9:
Merged, let's see how it looks on the real page

Hwurzburg:
after six hours, I dont see it yet on the build...did our build frequency drop?

rmackay9:
I see that the "pretty" page has updated: http://ardupilot.org/about/Partners, but as you say it hasn't appeared in the common-partners page.  http://ardupilot.org/copter/docs/common-partners.html

By the way it looks like the common-partners page is not-found if accessed from the index page:
  - index: http://ardupilot.org/copter/index.html (see "Partners" down on the bottom left)
  - http://ardupilot.org/ardupilot/docs/common-partners.html <-- 404 error

proficnc:
On a mobile http://ardupilot.org/about/Partners is missing 2/3 of the partners

PR #2134, opened 2019-10-24
Re-order Ardupilot side menu to match Sphinx top menu

Hwurzburg:


Hwurzburg:
oops....accidentally merged another one....will fix

PR #2147, opened 2019-10-26
Plane: Re-organized setup like Copter...archived a few unused pages

Hwurzburg:
First step in Plane site clean up...


Hwurzburg:
rebased and merging

PR #2152, opened 2019-10-26
Add LUA scripting as common advance config feature

Hwurzburg:
for 4.0


rmackay9:
Looks reasonable but looks like a merge conflict has popped up.

Hwurzburg:
conflict resolved....merging since approved

PR #2153, opened 2019-10-27
add orientation warning to acc calibration page

Hwurzburg:


common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -16,6 +16,8 @@ This article shows how to perform basic accelerometer calibration (using
 Calibration steps
 =================
 
+.. warning:: Be sure the :ref:`AHRS_ORIENTATION<AHRS_ORIENTATION>` is correct before doing the calibration!
```
rmackay9: let's add a warning to this page too: http://ardupilot.org/copter/docs/common-mounting-the-flight-controller.html

rmackay9: let's remove "in Mission Planner" from the title.
Let's remove the note, "Accelerometer calibration is mandatory if you’re using EKF to estimate vehicle position."

rmackay9: maybe let's change the warning to make it more clear that users only need to worry about the orientation if it's non-standard.

If the board is mounted in a non-standard orientation (i.e. arrow is *not* pointing forward) then please ensure the AHRS orientation is set before doing the accelerometer calibration.

Hwurzburg: done

Hwurzburg:
Merged with changes

PR #2155, opened 2019-10-27
Update quadplane-flying.rst

kd0aij:
@Hwurzburg tiltrotors do implement "forward throttle" in QLOITER mode (the new text is taken from a comment in QuadPlane::tiltrotor_continuous_update() in tiltrotor.cpp), but I seem to have screwed up syntax for the note::
The preview shows the "Note" notation as deleted.

### DIFF
--- a/plane/source/docs/quadplane-flying.rst
+++ b/plane/source/docs/quadplane-flying.rst
@@ -39,9 +39,6 @@ The one exception to the forward motor stopping in QuadPlane VTOL
 modes is if you have the :ref:`Q_VFWD_GAIN <Q_VFWD_GAIN>` parameter set to a non-zero
 value. In that case the forward motor will be used to hold the
 aircraft level in a wind. See the description of :ref:`Q_VFWD_GAIN <Q_VFWD_GAIN>`.
-
-.. note::
- Tilt-Rotor Quadplanes do not have the capability of using Q_VFWD_GAIN, since there is no separate foward motor.
  
 Assisted fixed-wing flight
 ==========================


Hwurzburg:
@kd0aij Mark, a bit of confusion I believe, as I understand the code, and from @tridge when I added the comment:
- In VTOL mode Q_VFWD_GAIN has no effect in a tilt-rotor (yet), code comments not withstanding....
the code at https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/tiltrotor.cpp#L109
thru https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/tiltrotor.cpp#L127 determines the base tilt of the motors, with attitude controller inputs added on top in VTOL and assisted modes
Q_FWD_GAIN impacts the forward motor velocity controller, but only for the forward motor in frames that have one...
those comments were in the original code from back in 2016 and the intent was to have this ability I am sure....
unless something changed that I am not aware of since I discussed this with @tridge in late July and added the note....

kd0aij:
Lines 126-127: https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/tiltrotor.cpp#L126
tilt the motors forward from 0 to q_tilt_max as k_throttle ranges from 0 to 100.
It looks to me as though this has always been supported.

k_throttle is set here: https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/servos.cpp#L457
quadplane.forward_throttle_pct() runs a control loop with gain Q_VFWD_GAIN, so if that is non-zero you'll get motors tilting forward in qloiter mode with a tiltrotor.

Works in SITL, Convergence and miniConvergence

My PR https://github.com/ArduPilot/ardupilot/pull/11763 uses the same mechanism to add manual control of forward throttle for tiltrotors in qhover, qstabilize and qacro modes.

kd0aij:
Plot from a SITL run showing that motor tilt tracks the forward throttle channel. (loiter mode with zero groundspeed and wind speed starting at about 7m/sec then ramping up to >25 m/sec):
forward_throttle

PR #2156, opened 2019-10-27
plane aux functions: add support starts with plane4.0

vierfuffzig:


Hwurzburg:
I am working on a common page ...this is included in that page....

vierfuffzig:
ok, i‘ll close then. thanks henry.

PR #2160, opened 2019-10-28
Add Winch page to Copter

Hwurzburg:


rmackay9:
Great, thanks for this.  I've disabled the Winch code by default so users can't actually use it I'm afraid.

I would like to add a page for Winches though in particular the "Daiwa Winch" which is what I used in the Japan Innovation Challenge a few weeks ago.

Hwurzburg:
So should I remove winch control from RCx_OPTIONS and output from SERVOx_FUNCTION?or you re-enabling it?

rmackay9:
@Hwurzburg, i think we should probably remove it from the RCx_OPTIONs and SERVOx_FUNCTIONs parameter values.. thanks for thinking of this.

Hwurzburg:
removed from this PR...fixed the AUX_FUNCTION wiki PR....will do one for SERVO_FUNCTION to remove it

Hwurzburg:
closing since not supported

PR #2161, opened 2019-10-28
Move to common AUX functions page and expand info on all options

Hwurzburg:


Hwurzburg:
removed winch per Randy

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
++----------------------+----------------------------+----------+---------+---------+
+|        29            | Landing Gear               |    X     |    X    |         |
++----------------------+----------------------------+----------+---------+---------+
+|        30            | Lost Vehicle Sound         |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        31            | Motor Emergency Stop       |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        32            | Motor Interlock            |    X     |         |         |
++----------------------+----------------------------+----------+---------+---------+
+|        33            | Brake                      |    X     |         |         |
```
rmackay9: "Brake" should probably become "BRAKE mode"

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
++----------------------+----------------------------+----------+---------+---------+
+|        34            | Relay 2 On/Off             |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        35            | Relay 3 On/Off             |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        36            | Relay 4 On/Off             |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        37            | THROW mode                 |    X     |         |         |
++----------------------+----------------------------+----------+---------+---------+
+|        38            | Avoid ADSB                 |    X     |         |         |
```
rmackay9: We should probably append "mode" to "Avoid ADSB" and maybe it should be all caps?

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
++----------------------+----------------------------+----------+---------+---------+
+|        71            | FLOWHOLD mode              |    X     |         |         |
++----------------------+----------------------------+----------+---------+---------+
+|        72            | CIRCLE mode                |    X     |    X    |         |
++----------------------+----------------------------+----------+---------+---------+
+|        73            | DRIFT mode                 |    X     |         |         |
++----------------------+----------------------------+----------+---------+---------+
+|        74            | Sailboat Motor 3Pos Sw     |          |         |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        75            | Surface Tracking Up/Down   |          |         |    X    |
```
rmackay9: Surface Tracking Up/Down is for Copter (not Rover)

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
++----------------------+----------------------------+----------+---------+---------+
+|        76            | STANDBY mode               |    X     |         |         |
++----------------------+----------------------------+----------+---------+---------+
+|        77            | TAKEOFF mode               |          |    X    |         |
++----------------------+----------------------------+----------+---------+---------+
+|        207           | Mainsail (Sailboat)        |          |         |    X    |
```
rmackay9: By the way, this is not actually an auxiliary function but rather a way to set the control for the mainsail to be done from an auxiliary switch.  Anyway, that distinction is perhaps too detailed for this page.. so it's fine.

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
+
+Supported Features
+==================
+
++----------------------+----------------------------+----------+---------+---------+
+| **RCx_OPTION value** | **Feature Description**    |**Copter**|**Plane**|**Rover**|
++----------------------+----------------------------+----------+---------+---------+
+|        0             | Do Nothing (Default)       |    X     |    X    |    X    |
++----------------------+----------------------------+----------+---------+---------+
+|        2             | Flip                       |    X     |         |         |
```
rmackay9: "FLIP mode" I guess..

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -0,0 +1,633 @@
...
+.. raw:: html
+
+   <table border="1" class="docutils">
+   <tbody>
+   <tr>
+   <th>Option</th>
+   <th>Description</th>
+   </tr>
+   <tr>
+   <td><strong>Flip</strong></td>
```
rmackay9: I'm a big fan of avoiding duplication in the wiki so I think we can skip this description for Flip because we already have it on the Flip flight mode page: http://ardupilot.org/copter/docs/flip-mode.html#flip-mode

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: okay this one really confuses me....the action states that it flips and then returns to the flight mode it was in....that is not a flight mode but a temporary  action....so that is not a flight mode....but yet its listed as a flight mode on the flight mode switch, which begs the question, if I use the flight mode switch to move to this mode, what does it return to? the mode it was in before the switch occurred? what happens if I am a bit slow and the switch moves thru AUTO from    ACRO before getting to FLIP, does it return to AUTO? intuitively to me,  flip or brake should be actions modifying the current  flight modes, not a flight mode itself....but I need to understand the answers to the above question before I agree its a flight mode

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: I have been seeing and using modes in all caps, but we can change going forward...personally i think it differentiates the word as a mode...params are all caps, but almost always have an underscored form...so I think confusion is low probability....
This is indeed a flight mode, but I need to change the mode set wording below a bit since this is not an edge triggered mode switch like most are...

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: Same comment as FLIP...the wiki currently says it should be set as a flight mode, shouldn't we remove it from the FLTMODEx param listing also?

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: oops..will change

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: is the explanation below for it correct?

common/source/docs/common-auxiliary-functions.rst
rmackay9: When flip starts it [records the mode that it was entered from](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/mode_flip.cpp#L61) and then returns to that after it completes the flip.  So, yes, if the pilot was slow on the switch and the vehicle entered Auto before flip then it would return to Auto.

"temporary action" or "flight mode", it's implemented as a flight mode... and it can be activated from a flight mode switch or an auxiliary switch..

common/source/docs/common-auxiliary-functions.rst
rmackay9: Is it not "edge triggered"?  I think it should be just like other flight modes in many respects..

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: agree re duplication, but I think that it really should be an action and not a mode, and I think it should also be removed from the FLTMODEx Params....and this could refer to that page (with a tweak to the page)

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: that not what the existing descritption says on the CH7/8 optin switches in the wiki....its in brake on high, out of brake when low.....out of brake to what is question...previous mode I take it...

common/source/docs/common-auxiliary-functions.rst
rmackay9: It really is a flight mode although it's true that users shouldn't really need to set it up on a flight mode switch nor an RC channel.  On the otherhand it won't do any harm either, the vehicle will just sit there loitering in place and not accepting any pilot input.

We need to keep the mode in the FLTMODEx parameter listing or else the GCSs won't know what string to display to the users when the vehicle does go into this mode.  It would appear as "Unknown".

My guess is that AvoidADSB was added to the RC_Options list as part of a review of all missing modes.  It shouldn't really have been added.  We could remove it.

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: since its a real mode, I will change to refer to the existing mode wiki page...

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: fixed

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: fixed

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: per conv changed name...will PR a code change for RC_Channel tomorrow to match ADSB Avoidance Enable

common/source/docs/common-auxiliary-functions.rst
Hwurzburg: changed to BRAKE mode

Hwurzburg:
okay finished hacking on it....had to make sure all modes were formatted the same....

rmackay9:
Merged, thanks!

PR #2166, opened 2019-10-29
Common: lidar-lite refresh

rmackay9:
This is a "refresh" of the lidar-lite range finder page to simplify and correct various issues I found while testing the PWM interface today in response to a user report that it wasn't working ([issue](https://github.com/ArduPilot/ardupilot/issues/10111#issuecomment-546627151))

I've done the following:

- replaced nearly all the wiring images with new ones based on the manufacturer's manual ([link](https://github.com/ArduPilot/ardupilot/issues/10111#issuecomment-546627151))
- put all the required parameter settings into the I2C and PWM areas.  This causes some repetition but it's easier for the user than jumping around trying to figure out what must be set
- re-ordered the "Problems with Lidar-Lite v1 & v2" section to cover "Blue-Label" issues first (Blue-Label versions are newer than the Black-label).
- changed the wording to be more concise (users don't want to waste time reading unnecessary words)
- formatting changes to remove the carriage return after 80characters


rmackay9:
Merged after a quick review from Tridge

PR #2168, opened 2019-10-29
Add boat and bicopter images/video to all vehicle page

Hwurzburg:
will replace boat image with its video as soon as its edited...


rmackay9:
Great, merged!

PR #2172, opened 2019-10-30
Plane: Update Airspeed sensor pages...add type,use params

Hwurzburg:


rmackay9:
Not really my area but at some point I think we should make an airspeed sensor landing page and then individual pages for the various sensors... a bit more like the range finder section.  this is a bit of a guess because I've used an airspeed sensor exactly twice in my life but I think there are various options and setup methods.

Issue #1896, opened 2019-08-12
Common: add SPL telemetry link to our Telemetry landing page


rmackay9:
I think we should add a link to our Telemetry landing page that points users to the SPL solution for 3G/4G (or whatever it is!) telemetry.  [There is a blog post here about it](https://discuss.ardupilot.org/t/stretching-comm-links-from-indoors-to-the-globe/45896) and we've already done the same for a competing product called UAVCast.

auturgy:
Yup. FYI SPL is an Iridium satcom datalink

adnan-ulhaque:
NAT Traversal through telecomm / sattelite service provider is the key here. I believe this an extremly important topic to connect telemetry+fpv over bvlos (beyond visual line of sight). UAVcast and other alternatives are not free . There are free public cloud services where Raspberry Pi (with Pixhawk) can be connected to MP on TCP via any data cellular SIM e.g. below...

https://discuss.ardupilot.org/t/fpv-over-internet-via-rpi3/22597?u=adnan_ul_haque

Hwurzburg:
PR# 2178 addresses this

PR #2178, opened 2019-11-01
Added SPL telemetry and re-organized telem landing page

Hwurzburg:
addresses #1896


rmackay9:
Do we want to organise them by range?  It's not bad.. I'm just thinking that there are lots of different criteria that people make their decision on including availability, price, bandwidth...

Hwurzburg:
yeah, but range is really the big decision point IMHO.....kinda matches your TX/RX system capability...

Hwurzburg:
I intend to add DL as a long range telem  system also....there may be others, and we can add them as they are proven....I think Crossfire is working on it, but I hear issues still present...

rmackay9:
Merged!

PR #2182, opened 2019-11-04
Add fuel flow and level sensor page

Hwurzburg:


common/source/docs/common-fuel-sensors.rst
```diff
@@ -0,0 +1,59 @@
...
+
+.. image:: ../../../images/fuel-level-sensor.jpg
+   :target: ../_images/fuel-level-sensor.jpg
+
+The displays/logs/limits for maH consumed is now actually the milliliters consumed with 1100 us PWM input indicating empty, and 1900us PWM indicating full.  Voltage will always report as 1.0V. Current is undefined.
+
+Connecting to the Autopilot
+---------------------------
+
+Again, as above, any GPIO capable pin can be used as the PWM input from the sensor.
```
rmackay9: Is it really any GPIO capable pin or only the AUX OUT pins? .. or is that the same thing?  I think it's only possible using the AUX pins.. although the use of the words "Aux Pins" is a bit Pixhawk specific..

rmackay9:
Merged because it's certainly useful.  I didn't know that foxtech had a sensor available.

I was wondering if we should actually put it in the Battery Monitor section?  I suspect, it's better in it's own section as is done in this PR but thought I shouldn't bring up this question at least.

common/source/docs/common-fuel-sensors.rst
Hwurzburg: I dont know for sure...figured @tridge or you would correct if wrong....let me know if it is and the exact restrictions and I can always change it...

PR #2187, opened 2019-11-04
Add Bicopter frame to Copter;update RC output mapping

Hwurzburg:


rmackay9:
Thanks!  especially interesting that you managed to come up with the bi-copter image.

Hwurzburg:
Piece of cake...took 30minutes to alter a quad image

PR #2189, opened 2019-11-04
Expand tailsitter info in qplane frame and tailsitter docs

Hwurzburg:
info on default outputs and how to setup motors in vectored tailsitter


Hwurzburg:
you must have an old version locally....pull current master....that doc is not part of the ardupilot wiki sub-site, but is in all the others....but just navigate to the dev site and it will be on the left menu

Hwurzburg:
changes per @tridge ...merging

PR #2205, opened 2019-11-08
Format only, global ArduPilot vs Ardupilot

Hwurzburg:


dev/source/docs/sitl-simulator-software-in-the-loop.rst
```diff
@@ -58,8 +58,8 @@ vary.  For instance the ports between ArduPilot and the simulator
 on the image are 5501/5502 but they can vary to be 5504/5505 or other
 port numbers depending on your environment.
 
-.. image:: ../images/ArdupilotSoftwareintheLoopSITL.jpg
-    :target: ../_images/ArdupilotSoftwareintheLoopSITL.jpg
+.. image:: ../images/ArduPilotSoftwareintheLoopSITL.jpg
```
rmackay9: This link to an image should probably be left unchanged..

rmackay9:
This LGTM with the possible exception of the link to an image mentioned above

dev/source/docs/sitl-simulator-software-in-the-loop.rst
Hwurzburg: yep..that was the last dir i did and forgot to run the complete wiki after...my daily cron job that pulls the master and builds the complete wiki, for me to check in the AM, would have caught it...but after the fact...
fixed

rmackay9:
a wonderful change ..

PR #2206, opened 2019-11-08
Clairify outputs used for singlecopter/coax

Hwurzburg:


rmackay9:
Merged, thanks.  It's a little unfortunate that we used the motor1, motor2 thing in the code when determining which output to send the signal to.

Issue #2045, opened 2019-10-02
Benewake TFmini / TFmini Plus lidar


adalta87:
[Benewake TFmini / TFmini Plus lidar](http://ardupilot.org/copter/docs/common-benewake-tfmini-lidar.html) - Describe the issue/suggestion and improve the title. Please keep a link to the original article if relevant.

This is wrong as the TFmini plus as far as I have been made aware of by the community is not able to use UART communication and as such needs to be expressed in the wiki as to not confuse potential buyers of this sensor that think it can be used on UART. I am aware that the TFmini has both I2C and UART commo's available but the plus version has yet to implement the uart commo. I was hoping this page could reflect that the tfmin supports both 12c and the tfmini plus only at this time supports i2c. considering these are different sensors they should at least have their own sections on setup even though the setup is almost identical to each other.

rmackay9:
yes, I agree that the tfmini and tfminiplus pages should be split.

peterbarker:
I have a uart-version TFMini Plus here, and it is working just fine and dandy on a UART, using the TFMini driver (`RNGFND1_TYPE=20` (BenewakeTFmini)).

I think "the community" may need poking to say that it does work on UART, you just need to specify the Mini driver, not the Plus driver.  Perhaps we should adjust the paramter description for the Plus to include "i2c" in it - that driver is i2c-specific.

adalta87:
That would make the most since especially considering plenty of people ask if you have read the wiki! Might as well have the correct information or its not really worth having at all. There have been plenty of folks complaining about this on the forum and its a bit counterintuitive if you have a setting for tfmini plus but dont specify that its for i2c only and probably should also provide on the tfmini mini plus setup that to use tfmin plus on uart then set to param 20 mot 25 or whatever the mini plus is. 
-------- Original message --------From: Peter Barker <notifications@github.com> Date: 10/31/19  19:14  (GMT-07:00) To: ArduPilot/ardupilot_wiki <ardupilot_wiki@noreply.github.com> Cc: adalta87 <adalta87@gmail.com>, Author <author@noreply.github.com> Subject: Re: [ArduPilot/ardupilot_wiki] Benewake TFmini / TFmini Plus lidar (#2045) I have a uart-version TFMini Plus here, and it is working just fine and dandy on a UART, using the TFMini driver (RNGFND1_TYPE=20 (BenewakeTFmini)).
[...]

PR #2212, opened 2019-11-10
Add info on I2C vs UART to Benewake TFmini/Plus

Hwurzburg:
addresses #2045


peterbarker:
This is correct.

In fact, my device didn't function with the MiniPlus driver on uart... I think that's a known issue, 'though?

rmackay9:
@peterbarker, my guess is that the MiniPlus will work with the UART but you'll need to update the lidar's firmware.  Sadly we don't print out a message saying that the lidar's firmware needs updating.  Instead we just silently fail :-(

Hwurzburg:
I must have miss-understood....I thought @peterbarker said in the issue "I have a uart-version TFMini Plus here, and it is working just fine and dandy on a UART, using the TFMini driver (RNGFND1_TYPE=20 (BenewakeTFmini))." so it did work with the correct driver
In fact, the Sparkfun datasheet and manual for the Plus ONLY mentions UART connections, nothing showing I2C as mentioned on the product brief...but the Benewake site shows a newer version with the I2C interface detailed, so it was either undocumented or a product revision.....
I suggest we use the latest info....and proceed with two drivers one UART and one I2C labeled as such somehow...

rmackay9:
Merged, thanks!

PR #2215, opened 2019-11-11
Add MatekF765-Wing doc

Hwurzburg:


Hwurzburg:
added RTS and CTS note on UART7

Issue #982, opened 2017-10-19
Hyperlinks need checking throughout the wiki


auturgy:
ref https://github.com/ArduPilot/ardupilot_wiki/issues/964
This issue is opened to flag the requirement to check for broken links throughout the wiki.
Internal links can be checked via sphinx, however many "broken" links are due to external re-directs, so this task will require manual effort to complete.  
Partners/vendors - please check links to your products/sites and submit PR's where appropriate.
All - please add pages with broken or incorrect links into a comment below.

auturgy:
http://ardupilot.org/plane/docs/common-pixracer-overview.html#common-pixracer-overview
Pixracer hyperlink goes to auav.co redirect page.  Needs to point to the mRo product

Naterater:
This page needs a lot of help.
http://ardupilot.org/plane/docs/ecalc-motor-and-prop-efficiency-guide.html

Hwurzburg:
pages mentioned fixed or removed....

brunoolivieri:
I found some GitHub broken links. I guess that was linked for a current version in past time (this phrase makes sense?)

| File | Link |
|------|------|
|.\common\source\docs\common-ppm-encoder-apm2x-atmega32u2.rst    |    https://github.com/ArduPilot/ardupilot/blob/master/Tools/ArduPPM/Libraries/PPM_Encoder.h
|.\common\source\docs\common-qualcomm-snapdragon-flight-kit.rst    |    https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_HAL_Linux/qflight
|.\common\source\docs\common-rangefinder-objectavoidance.rst    |    https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_RangeFinder/RangeFinder.h#L24
|.\common\source\docs\common-zed.rst    |    https://github.com/yankailab/OpenKAI/blob/master/kiss/apmCopter.kiss
|.\dev\source\docs\apmcopter-programming-libraries.rst    |    https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_Buffer
|.\dev\source\docs\building-for-erle-brain.rst    |    https://github.com/ArduPilot/ardupilot/blob/master/mk/targets.mk#L75
|.\dev\source\docs\editing-the-code-with-eclipse.rst    |    https://github.com/ArduPilot/ardupilot/blob/master/mk/px4_targets.mk

hamishwillee:
[...]

PR #2226, opened 2019-11-13
A tool to help to deal with broken links

brunoolivieri:
It can parse the RST files or built HTML files. It is not fast enough to insert in the build script, but help me to find some broken link that I already did a couple of PRs to fix, and I inserted some of them in the issue #982

I guess this is a script like something used in #1382.

Due to some isses[1], it presents lots of false positives. I tried to bypass this [2], but some help would be appreciated regarding web servers responses.

- When using the parser for RST files, it is faster than HTML because it looks less* files and is not able to find any link**.   (* IFF "common" folder is not copied across vehicles yet.)(** Only look for sphinx expected links.)

- When using the parser for HTML, it has a more comprehensive parser for HTML "a" attributes.

[1] https://stackoverflow.com/questions/47506092/python-requests-get-always-get-404
[2] https://stackoverflow.com/questions/27652543/how-to-use-python-requests-to-fake-a-browser-visit

### DIFF
--- a/search_for_broken_links.py
+++ b/search_for_broken_links.py
@@ -66,7 +66,7 @@ def create_list_of_files(initial_path, pattern):
 
     return list_of_files
 
-def fetch_links(uri):
+def fetch_links_on_html(uri):
         """
         Fetch links from a file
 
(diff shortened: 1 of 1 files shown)


khancyr:
Haa what a pain... I am sorry for you ...but Sphinx is able to do it at build time...

peterbarker:
@khancyr how does one get Sphinx to do that?

Are there advantages to doing it externally?  Cases Sphinx doesn't handle?

hamishwillee:
I think it's just `make linkcheck` or  ` sphinx-build -b linkcheck`. It runs a check on every URL. You will need to do it on every wiki.

This is what I used to do occasionally and then copy the broken links into issues for tracking.

brunoolivieri:
@khancyr yes! An unnecessary pain... I should spend time back in parameters versioning. LOL!  But I read in some place that sphinx could not do that. My bad.

@hamishwillee thanks!

@peterbarker and all: Now, I can not see a significant advantage in inserting my script. Looks redundant...  I also like @Hwurzburg idea of the use of an offsite tool (deadlinkchecker.com)

May be to build the wiki locally once a while using the link checkers would be enough to look for deadlinks.

Should I close it?

brunoolivieri:
During the wiki call today we chose close it.

PR #2227, opened 2019-11-14
Add RC Option note to ADSB page

Hwurzburg:


Hwurzburg:
add warning to setup RC switch for in flight disabling of the feature

PR #2230, opened 2019-11-15
Update how-the-team-works-development-fund.rst

auturgy:
a couple of links to SPI following up from https://discuss.ardupilot.org/t/does-ardupilot-disclose-where-donations-are-spent/49283


adalta87:
I find it quite odd being a non profit that all intake/spending isn't listed, it is also quite odd that just finally ardupilot decided to disclose its relationship to spi. Transparency is king, and as of current its not at all transparent.

khancyr:
@adalta87 it was already written that we were part of SPI to manage the fund, we just put direct link for monthly and annual report to help people to find them more easily

rmackay9:
Merged, thanks!

PR #2232, opened 2019-11-15
COPTER: update log CTUN description

fagoth:
updated some descriptions


rmackay9:
Merged, thanks very much!

PR #2251, opened 2019-11-22
Unify all vehicles first time setup and config page toc orders

Hwurzburg:
no content change...just TOC ordering to unify all vehicles into common flow


Hwurzburg:
built and tested, merging

PR #2253, opened 2019-11-23
Make FC wiring Pixhawk agnostic,add fc i/o page

Hwurzburg:
first step in generalizing wiki for all flight controllers...note references to future planned Serial Port and GPIO advance config pages, which I will do in next couple of days...
FC diagram may be useful in making other page Pixhawk agnostic....if this is approved, will continue to work on those pages over time


rmackay9:
I think we should put an arrow on the generic flight controller image..

Hwurzburg:
add  arrow....merging

rmackay9:
A couple of things I've noticed post merge:

- we should reduce the white space in the images.  For example under "RC input" are's a diagram of the flight controller connected to an RC receiver and there's a huge amount of white space around the image.
- we should try and keep images scaled to 450pixels or less I think.  The top image under "Typeical Flight Controller Wiring Connections" is a bit too wide I think.
- ideally the arrow for the GPS and the flight controller should be pointing in the same direction because this is how it should be setup on a real vehicle

NickNunno:
Totally petty request but we support many vehicle types.

This would be a great opportunity to start using "Autopilot" in place of "Flight Controller".

Hwurzburg:
@rmackay9 fixed by #2265

Hwurzburg:
@NickNunno open a wiki issue and we will discuss in the next meeting...thanks

Hwurzburg:
globally changed

PR #2273, opened 2019-11-27
Update UAVCAN docs, add Adapter Node and refs to AP_Periph firmware

Hwurzburg:


tridge:
I think the UART connector showed SDA/SCL instead of RTS/CTS. Current nodes don't have RTS/CTS, and probably won't be added

Hwurzburg:
revised diagram

PR #2278, opened 2019-11-28
Adding last updated date on all pages

brunoolivieri:
It automatically inserts the last date when such a page was edited on the bottom of each page. Two sites as examples: 

[http://www.olivieri.com.br/ardupilot/wiki_last_update_test/dev ](http://www.olivieri.com.br/ardupilot/wiki_last_update_test/dev )[http://www.olivieri.com.br/ardupilot/wiki_last_update_test/copter/](http://www.olivieri.com.br/ardupilot/wiki_last_update_test/copter/)

To respect the actual last edit date, I understand that we cannot update it on the .rst files and commit them because it would update such a time to such a commit date. 

Some thoughts:

1) It is not beautiful, but works. Could be more beautiful if insert it in the bottom of the .html file instead of the bottom of the .rst file. But:

- It might not get the necessary attention;
- I do not know how to do that yet. Would it be possible to insert the date in a variable inside the .rst file and use it in the sphinx theme?

2) I am not sure if the update.sh is the actual script that updates the Wiki in the server but looks like that it is something called by as a cron job.

3) It is yet another external small fix and not an architectural solution regarding versioning. However, it works with a minimal footprint, and it is effortless to remove or revert.

### DIFF
--- a/insert_last_updated_date.sh
+++ b/insert_last_updated_date.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+#===========================================================================================
+#	Automatically add last modified date to .RST files according 
+#	to last commit date of each file.
+#
+#   Caution: do use while updating the Wiki in the server because it alters ALL .RST files.
+#
+#	Run: ./insert_last_updated_date.sh N (where N is the number is the parallel tasks
+#
+#===========================================================================================
+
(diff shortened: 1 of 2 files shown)


brunoolivieri:
From the wiki call today, after a discussion, we opt to decline the idea of present the last update date.

PR #2279, opened 2019-11-28
Add GPIO page as advanced setup

Hwurzburg:


rmackay9:
Looks pretty good so I've merged it.  I wonder slightly what the difference is between Relays and this GPIO page.  I guess relays are always outputs while GPIO can also be inputs?

As a side note, I wonder if perhaps we should make the "relay" words in this document link to the relay wiki page (or at least the first instance that "relay" appears)

Hwurzburg:
also camera trigger, and perhaps find an input like one of the polarity pins and hot link in that intro list of GPIO uses?

rmackay9:
@Hwurzburg, I'm just wondering though isn't a GPIO a combination of a Relay (which is output and is already documented [here](http://ardupilot.org/copter/docs/common-relay.html)) and a Button which is input?  Maybe we should instead modify this new page to describe how to setup buttons?

Hwurzburg:
I added some links for inputs...but the only button I found was the Rover start button and it has setup on its page...I linked it...

PR #2299, opened 2019-12-05
Update LUA scripting page with current bindings and some examples

Hwurzburg:


common/source/docs/common-lua-scripts.rst
```diff
@@ -103,12 +103,18 @@ A Location userdata object supports the following calls:
 
 - :code:`loiter_xtrack([should_xtrack])` - If the location is used for a loiter location this flags if the aircraft should track from the center point, or from the exit location of the loiter.
 
-- :code:`get_distance(other_loc)` - Given a Location this calculates the horizontal distance between the two locations in meters.
+- :code:`get_distance(other Location)` - Given a Location this calculates the horizontal distance between the two locations in meters.
```
WickedShell: bad change, no spaces are possible for argument names, and this would read like it's actually 2 arguments.

common/source/docs/common-lua-scripts.rst
```diff
@@ -103,12 +103,18 @@ A Location userdata object supports the following calls:
...
 
 - :code:`offset(offset_north, offset_east)` - Translates this Location by the specified number of meters.
 
 - :code:`get_vector_from_origin_NEU()` - Returns nil or Vector3f that contains the offset from the EKF origin to this location. This will return nil if the EKF origin wasn’t available at the time this was called.
 
+- :code:`get_bearing(other Location)` - Given a Location this calculates the relative bearing to the location in degrees
```
WickedShell: No space, same comment. This is also a value in radians not degrees.

common/source/docs/common-lua-scripts.rst
```diff
@@ -103,12 +103,18 @@ A Location userdata object supports the following calls:
...
+
+- :code:`get_distance_NED(other Location)` - Returns nil or Vector3f that contains the 3D vector to the other Location.
```
WickedShell: No space, should note that it's a vector in meters.

common/source/docs/common-lua-scripts.rst
```diff
@@ -103,12 +103,18 @@ A Location userdata object supports the following calls:
...
+
+- :code:`get_distance_NE(other Location)` - Returns nil or Vector2f that contains the 2D vector to the other Location.
```
WickedShell: No space, should note that it's a vector in meters.

common/source/docs/common-lua-scripts.rst
```diff
@@ -165,11 +171,17 @@ The following operators are supported on a Vector3f:
...
 - :code:`-` - Subtracts the two vectors by components.
 
-AHRS
-~~~~
+AHRS (ahrs:)
+~~~~~~~~~~~~
 
 The ahrs library represents the Attitude Heading Reference System computed by the autopilot. It provides estimates for the vehicles attitude, and position.
 
+- :code:`get_roll()` - Returns the current vehicle roll angle in degrees.
```
WickedShell: Radians, you need to actually lookup the code when writing the descriptions.

common/source/docs/common-lua-scripts.rst
```diff
@@ -165,11 +171,17 @@ The following operators are supported on a Vector3f:
...
+
+- :code:`get_pitch()` - Returns the current vehicle pitch angle in degrees.
```
WickedShell: Radians, you need to actually lookup the code when writing the descriptions.

common/source/docs/common-lua-scripts.rst
```diff
@@ -165,11 +171,17 @@ The following operators are supported on a Vector3f:
...
+
+- :code:`get_yaw()` - Returns the current vehicle yaw angle in degrees.
```
WickedShell: Radians, you need to actually lookup the code when writing the descriptions.

common/source/docs/common-lua-scripts.rst
```diff
@@ -182,8 +194,28 @@ The ahrs library represents the Attitude Heading Reference System computed by th
 
 - :code:`groundspeed_vector()` - Returns a Vector2f containing the vehicles velocity in meters/second in north and east components.
 
-Battery
-~~~~~~~
+- :code:`get_velocity_NED()` - Returns nil, or a Vector3f containing the current NED vehicle velocity.
```
WickedShell: Units

common/source/docs/common-lua-scripts.rst
```diff
@@ -182,8 +194,28 @@ The ahrs library represents the Attitude Heading Reference System computed by th
...
+
+- :code:`get_velocity_NE()` - Returns nil, or a Vector2f containing the current NE vehicle velocity.
```
WickedShell: Units

common/source/docs/common-lua-scripts.rst
```diff
@@ -182,8 +194,28 @@ The ahrs library represents the Attitude Heading Reference System computed by th
...
+- :code:`home_is_set()` - Returns a true if home position has been set.
+
+- :code:`prearm_healthy()` - Returns a true if current pre-arm checks are passing.
+
+Arming (arming:)
+~~~~~~~~~~~~~~~~
+
+The Arming library provides access to arming status and commands.
+
+- :code:`disarm()` - Attempts to disarm the vehicle. Returns true if successful.
```
WickedShell: This can return false if the vehicle was already disarmed (on some vehicles). The more useful information here is the vehicle will always be disarmed after calling this.

common/source/docs/common-lua-scripts.rst
```diff
@@ -254,13 +286,58 @@ The GPS library provides access to information about the GPS's on the vehicle.
...
++---------------+-----------+
+|4              | Warning   |
++---------------+-----------+
+|5              | Notice    |
++---------------+-----------+
+|6              | Info      |
++---------------+-----------+
+
+
+- :code:`set_message_interval(serial channel, message type, rate)` - Sets the message types update rate on SERIAL (serial channel). For example, gcs:set_message_interval(0, uint32_t(30), math.floor(1000000 / 2.0)) sets SERIAL0 rate for ATTITUDE message stream (30) to 2.0 Hz .
```
WickedShell: Way to complex an example. First note that the rate is actually the time between sending the message in microseconds. Then to keep the example simple it should just be an argument of `500000` the math.floor stuff is to fix a type problem, that you only have by doing division, and it's needlessly making this seem more complicated.

common/source/docs/common-lua-scripts.rst
```diff
@@ -254,13 +286,58 @@ The GPS library provides access to information about the GPS's on the vehicle.
 
 - :code:`first_unconfigured_gps()` - Returns nil or the instance number of the first GPS that has not been fully configured. If all GPS's have been configured this returns 255 if all the GPS's have been configured.
 
-Notify
-~~~~~~
+GCS (gcs:)
+~~~~~~~~~~
+
+- :code:`send_text(severity, "text")` - Will send the text variable or constant string with message severity . Severity level is :
```
WickedShell: Don't put `"text"`, that should just say message or text, no quotes, as that implies a string literal which doesn't make sense. The description is complicated as well, just say that text should be a string to send to the GCS. You don't need to distinguish between a constant one or one from a variable, that just implies complexity that doesn't exist.

common/source/docs/common-lua-scripts.rst
```diff
@@ -254,13 +286,58 @@ The GPS library provides access to information about the GPS's on the vehicle.
...
+
+
+- :code:`set_message_interval(serial channel, message type, rate)` - Sets the message types update rate on SERIAL (serial channel). For example, gcs:set_message_interval(0, uint32_t(30), math.floor(1000000 / 2.0)) sets SERIAL0 rate for ATTITUDE message stream (30) to 2.0 Hz .
+
+Serial LED (serialLED:)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+This library allows the control of WS8212B RGB LED strings via an output reserved for scripting and  selected by SERVOx_FUNCTION = 94 thru 109 (Script Out 1 thru 16)
+
+- :code:`set_num_LEDS(output number, number of LEDs)` - Sets servo output number and number of LEDs in the string on that output. "output number" is servo output number 1-16 that the string is attached with a string having "number of LEDs".
```
WickedShell: `set_num_LEDs` not `set_num_LEDS`.

No space in the variable names. This doesn't set a servo output number, it sets the number of LED's on that output channel. The number of LED's is not a string, it must be a number.

common/source/docs/common-lua-scripts.rst
```diff
@@ -254,13 +286,58 @@ The GPS library provides access to information about the GPS's on the vehicle.
...
+
+- :code:`set_RGB(output number, LED number, r,g,b)` - Set the data for "LED number" (1-32) on "output number" (1-16) output the the "r,g,b" values (0-255)
```
WickedShell: Again, no use of quotes or spaces in names, that makes it more complicated.

common/source/docs/common-lua-scripts.rst
```diff
@@ -254,13 +286,58 @@ The GPS library provides access to information about the GPS's on the vehicle.
...
 
 - :code:`play_tune(tune)` - Plays a MML tune through the buzzer on the vehicle. The tune is provided as a string.
 
-Terrain
-~~~~~~~
+
+Vehicle (vehicle)
+~~~~~~~~~~~~~~~~~
+
+- :code:`set_mode(mode number)` - Attempts to change vehicle mode to "mode number". Returns true if successful.
```
WickedShell: No quotes, probably worth linking to the defines for these, you can use the entries in ArduPilot's MAVLink  XML (IE https://mavlink.io/en/messages/ardupilotmega.html#PLANE_MODE )

WickedShell: Needs to describe the boolean return value if the mode change failed for any reason.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
 
 - :code:`toggle(relay_num)` - Toggles the requested relay on or off.
 
+
+Servo_Channels (SRV_Channels:)
```
WickedShell: No underscore, this is the human friendly output.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- :code:`find_channel(output function)` - Returns first servo output number -1 of an output assigned "output function" (See ``SERVOx_FUNCTION`` parameters ). False if none is assigned.
+
+Servo Output
+~~~~~~~~~~~~
+
+This is special case that is not generated via the binding methods.
```
WickedShell: This is misleading. It is guaranteed to be available, the binding was just done differently.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+
+- :code:`servo.set_output (function number, PWM value)` -Sets servo outputs of type "function number" to PWM values between 1000 and 2000.
```
WickedShell: The bounds are not strictly 1000 to 2000, it's whatever you want for a PWM value thats in the uint16_t number space, don't imply it has to be 1000 to 2000.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+
+
+Examples
+========
+
+``local current_pos = ahrs:get_position()``  Allocates current_pos as a local variable, uses the AHRS get_position method to assign it the current position as a Location data type if available, or nil if not.
```
WickedShell: It's not necessarily worth introducing the concept of local here. You haven't described what that actually means anywhere.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+
+``global groundspeed = gps::get_velocity_NE(gps:primary_sensor())`` Allocates and sets a global scripting variable, groundspeed, to the ground speed of the vehicle as reported by the primary GPS.
```
WickedShell: Global isn't a keyword. Why do you have 2 colon's? There is no such thing as get_velocity_NE on a GPS object, did you mean AHRS? This clearly hasn't been tested.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+
+``speed = (gps:velocity(1)):length()`` would set speed to the magnitude of the velocity vector of the vehicle including climb/descent rate contributions
+
+``vehicle:set_mode(10)`` Attempts to set the vehicle mode to AUTO.
```
WickedShell: This is wrong. You'd have to say what vehicle you were on, as AUTO is only 10 on plane, not copter.

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+
+``gcs:send_text(6, string.format("LEDs: chan=" .. tostring(chan)))`` Would send a Mavlink meassge of info severity level (6), printing the message ``LEDs: chan= 11`` asssuming chan=11.
```
WickedShell: This makes no sense, you've called string.format on a string literal, there is no reason to use `string.format`

common/source/docs/common-lua-scripts.rst
```diff
@@ -289,6 +366,35 @@ The relay library proivdes access to controlling relay outputs.
...
+Servo Output
+~~~~~~~~~~~~
+
+This is special case that is not generated via the binding methods.
+
+- :code:`servo.set_output (function number, PWM value)` -Sets servo outputs of type "function number" to PWM values between 1000 and 2000.
+
+
+Examples
+========
```
WickedShell: This section is extremely low value. We should link to the examples folder, and include a single well annotated/explained script. This is what we do up in the sample script section, so just delete this whole area, improve that as needed, and link to the examples up there.

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: not familiar with LUA, figured it has a global declaration...it does, but its by NOT using local to declare...dont have time to become a LUA programming expert...brain fart on function call...
and I dont test code...just try to write docs...I rely on experts to correct since the docs were not provided at time of addition of the features..

common/source/docs/common-lua-scripts.rst
Hwurzburg: taken directly from example....might want to discuss with tridge or change the example..

common/source/docs/common-lua-scripts.rst
Hwurzburg: sorry, disagree here....and link to examples was provided lower down

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: is the vehicle ALWAYS disarmed? I thought you could not disarm under certain conditions...if it always disarms, will update...changed re returning false if already disarmed...

common/source/docs/common-lua-scripts.rst
Hwurzburg: taken directly from the examples....but I simplified it....

common/source/docs/common-lua-scripts.rst
Hwurzburg: yep,poorly worded..thanks...fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: is this correct now?

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed, not being a coder, looking (assuming I knew where to look) might not assure that I had it right in any case...

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed...it is used differently as a method...hopefully new wording clarifies this

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

common/source/docs/common-lua-scripts.rst
Hwurzburg: fixed

Hwurzburg:
Michael, thanks for your inputs.
Sorry there were so many errors. I will try my best to assure it will not happen again.
The LUA additions, while not intended to teach the scripting language itself, should be understandable to someone less than expert in LUA or C++  trying to use them as they learn the scripting language. I have no experience in scripting in LUA and little in C++, so I rely on the experts to correct me while I try to make the docs usable to a wide base.

common/source/docs/common-lua-scripts.rst
WickedShell: The sample script is higher quality then the examples here though. It shows much the same thing, and has a description for every line, and includes demos of how you'd actually chain the functions together. The snippets here don't present anything new, and the comments on them make a lot less sense.

common/source/docs/common-lua-scripts.rst
WickedShell: We should fix the example then, because the comment stands, it's illogical and shouldn't be shown to people :)

common/source/docs/common-lua-scripts.rst
Hwurzburg: confirmed that this method always disarms....disarming is prevented in numerous cases using rudder disarm but this method...fixed

common/source/docs/common-lua-scripts.rst
WickedShell: I did fix up the LED example script ( https://github.com/ArduPilot/ardupilot/pull/12992 )

Hwurzburg: thank...incorporated..

common/source/docs/common-lua-scripts.rst
Hwurzburg: removed section....personally it took me a while to understand the  useage, even looking at the examples folder....not being familiar with namespace conventions, etc...
I think that more examples in the examples folders are needed with better commenting in order for this to be useful to the average user...but will let you or others interested in promoting LUA scripting to a wider base do that...

Hwurzburg:
having addressed all reviewers comments, merging

PR #2302, opened 2019-12-05
change plane index page image

Hwurzburg:


Hwurzburg:
may have to resize for mobile...but will try it as is and see...

PR #2327, opened 2019-12-14
Add page on using multiple RC receivers

Hwurzburg:


common/source/docs/common-multiple-rx.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Configuration
+=============
+
+The additional receivers can be attached to any Serial Port's RX input as long as they use a serial protocol.
+These are: SBus, DSM, DSM2, DSM-X, IBus, SUMD, and SRXL.
+
+For whichever Serial Port the receiver is attached, the ``SERIALx_PROTOCOL`` should be set to 23. The baud rate is ignored and auto-detected, as well as the type of the receiver.
+
+Failsafe and Fall Over
```
rmackay9: is it "Fall Over" or "Failover" I wonder..?

common/source/docs/common-optional-hardware.rst
```diff
@@ -47,6 +47,7 @@ information related to Autopilot selection see :ref:`Autopilot Hardware Options
     Landing Gear/ Retractable Camera Mount <common-landing-gear>
 [/site]
     LEDs (external) <common-external-leds>
+    Multiple Radio Control Receivers <common-multiple-rx>
```
rmackay9: I wonder if the title should be "Radio Control Receivers (Multiple)" to keep the "Radio Control" part at the front.  I think it's not important for now but maybe we will eventually decide that we should have a section on RC Receivers.

rmackay9:
Looking good!  I've made some comments but they're not blockers so merged!

common/source/docs/common-multiple-rx.rst
Hwurzburg: will change to "changeover"...thinks its more correct...

PR #2303, opened 2019-12-06
Add battery sag warning to Qplane transition section

Hwurzburg:

PR #2328, opened 2019-12-14
add advanced power monitor setup page

Hwurzburg:


common/source/docs/common-advanced-battery-setup.rst
```diff
@@ -0,0 +1,38 @@
+.. _common-advanced-battery-setup:
+
+=============================================
+Advanced Advanced Power Monitor Configuration
```
rmackay9: "Advanced" twice here..

common/source/docs/common-advanced-configuration.rst
```diff
@@ -10,6 +10,7 @@ tuning options for the vehicle.
 .. toctree::
     :maxdepth: 1
 
+    Advanced Power Monitor Configuration<common-advanced-battery-setup>
```
rmackay9: I'd probably name this one "Power Monitor Advanced Configuration" to keep the "Power " part at the front for alphabetical organisation.

By the way, we use "Module" and "Monitor" quite a bit so we might want to standardise on one or the other.

common/source/docs/common-advanced-battery-setup.rst
```diff
@@ -0,0 +1,38 @@
...
+=============================================
+
+In addition to the normal battery current and voltage setup and scaling parameters addressed in each individual :ref:`power monitor's<common-powermodule-landingpage>` setup page, there are several other parameters available for advanced configuration. Each parameter is well documented by following its link below:
+
+.. note:: While links are provided for the first battery monitor parameter group, there can be up to 10 total monitors in the system depending on the individual autopilot's capabilities and/or use of SMBUS or UAVCAN monitors. These are accessed using ``BATTx_`` parameter prefixes
+
+Parameters for advanced power monitor failsafe setup
+----------------------------------------------------
+
+:ref:`BATT_WATT_MAX<BATT_WATT_MAX>`
```
rmackay9: The BATT_WATT_MAX param is only available on Plane by the way not that it hurts having it too much.

rmackay9:
TBH, I wonder if we need this page.  The content is all included on other pages I think (or should be) - either in the battery failsafe pages or in the pre-arm checks pages... Could we get away with just having links from power module landing page?

For example on the [common-powermodule-landingpage](https://ardupilot.org/copter/docs/common-powermodule-landingpage.html) we could make the "triggering a low battery failsafe" link to the vehicle specific battery failsafe page.  We could add a new line to the list of "benefits" which would link to the [pre-arm checks page](https://ardupilot.org/copter/docs/common-prearm-safety-checks.html).  Happy to discuss of course, I'm just looking for ways to keep duplication low.

common/source/docs/common-advanced-battery-setup.rst
Hwurzburg: its listed in copter and rover params...is this a mistake?

Hwurzburg:
I am considering your last suggestions and agree that it might be better to approach it as additions to the pre-arm page and link to vehicle failsafe pages...but I need to get tridge to approve the PR to the plane failsafes page first to actually have a battery failsafe doc to point to...
as to the naming, I think two names are appropriate, but they must be applied correctly....a Power Module will usually include a Power Monitor, but also a voltage regulator of some kind...its used about 100?  times in the wiki...I will go thru each and try to change as needed...

common/source/docs/common-advanced-battery-setup.rst
Hwurzburg: its really advanced

common/source/docs/common-advanced-battery-setup.rst
rmackay9: I think the script that produces the param wiki page can't handle the vehicle specific tags in the parameters.

> AP_GROUPINFO_FRAME("WATT_MAX", 8, AP_BattMonitor_Params, _watt_max, 0, AP_PARAM_FRAME_PLANE),

So in this example it's not reading that last AP_PARAM_FRAME_PLANE part

Hwurzburg: Sounds like we need a mod to our param build script....will put in an issue....

Hwurzburg:
closing....merged #2345  and #2303 which covered the params I was concerned about...

PR #2334, opened 2019-12-17
Updated LED Meaning List

sakza:
Added ESC calibration light sequence to LED info page.


rmackay9:
Here's the link to the ESC calibration page: https://ardupilot.org/copter/docs/esc-calibration.html

PR #2338, opened 2019-12-17
Update code-overview-adding-a-new-log-message.rst

kd0aij:
link to ArduCopter's LogStructure array  is out of date.
Is there a better way to link to blocks of code that won't go out of date?
Perhaps a branch of source dedicated to the wiki would work.


rmackay9:
Merged, thanks!  I don't know of a better way I'm afraid.

PR #2341, opened 2019-12-18
scripts: added cap_params.py

tridge:


Hwurzburg:
two thumbs up....thanks!

Issue #1209, opened 2018-04-18
Plane: add video of how to do a manual launch


rmackay9:
I think we should add an instructional video (or two!) showing how to do hand launches.  These videos could go on this page: http://ardupilot.org/plane/docs/automatic-takeoff.html#hand-launching

Most of the advice is written on the page but some things people have said to me that I think are not written:

- throw the vehicle mostly flat, i.e. don't point its nose pointing too high
- don't throw it like a football because that leads to rotation at takeoff
- easiest way is to use Auto mode with a takeoff and RTL command only

Naterater:
I would be interested in helping with this.

I'm nearly certain that a takeoff command alone will automatically execute a RTL after.

magicrub:
Kavin Garland has a nice launch right [here](https://youtu.be/Yx1k8VgpHlU?t=4). He'd probably make a take-off only video if asked.

rmackay9:
@Naterater,
This issue was made for you!  It would be really great if you could help us with the videos.  No pressure but it would be greatly appreciated by many I think.

fnoop:
This would be great, especially if it is written for newbies like myself. I've been trying to setup arduplane and get it flying for 6 months (well OK bad weather means rebuilding and the next attempt every couple of months) with no luck so far. 

It's taken a long, long time to read through all the docs and get plane setup, and then work out the easiest way for newbies to get flying is a hand launch and this requires an auto mission to be setup.  Ideally there would be an easier way to hand launch, but at least if the docs talk us through this process it would really help.

So far, ground = 4, flying = 0 :(

Hwurzburg:
#2357 awaiting review and merge

PR #2357, opened 2019-12-23
Add takeoff page to plane

Hwurzburg:
in response to #1209


plane/source/docs/takeoff-landingpage.rst
```diff
@@ -0,0 +1,55 @@
...
+=======
+
+Taking the plane off can be accomplished in many ways: Rolling Takeoff, Hand Launch, Bungee, or VTOL (Verical Takeoff and Landing) if a :ref:`quadplane-support`. And it can be done manually or automatically.
+
+Rolling Takeoff
+===============
+
+In this case, the plane has landing gear with steerable nose gear, usually coupled to the rudder control and the plane is usually launched from a smooth grass or paved surface. Once aligned nose into the wind, throttle is gradually increased to takeoff value (usually full throttle) while the aircraft direction is controlled by rudder. Once sufficient airspeed to lift off is obtained, slight up elevator will rotate the nose to climb attitude.
+
+MANUAL or FBWA modes are most commonly used during takeoffs.
```
Naterater: Add that for FBWA or other automatic modes, the STER parameters may need to be tuned if the nosewheel is configured as a steering servo.

plane/source/docs/takeoff-landingpage.rst
```diff
@@ -0,0 +1,55 @@
...
+===========
+
+Many planes do not have landing gear, or a smooth runway is not available. Hand launching is a very common takeoff method, but also can be very frustrating if not setup and executed properly.
+
+The keys to a successful hand launch are :
+
+- :ref:`Proper CG <guide-center-of-gravity>` 
+- Having flying surfaces centered and neutral at neutral RC stick positions
+- Having the "level" calibration setup such that the level horizon (0 deg pitch) in FBWA mode results in the wing chord line being a few degrees positive, which will result in a positive angle of attack in "level" flight, the normally cruising attitude for a plane.
+- Using FBWA mode or :ref:`automatic-takeoff` . While launching in MANUAL can be done, using a stabilized mode will eliminate a lot of pilot effort to overcome trim issues (see Tip at end of this page). If using FBWA mode, a few degrees of up elevator (~10%) during the launch will result in an automatic climb. Add this via a switch or stick pressure. Do not change TX trim positions used during RC calibration since it affects all other modes of flight.
```
Naterater: Suggest more up elevator; I launch my planes at FULL (100% and 30-degrees pitch) elevator, and that keeps them from diving before climbing... a common problem.

plane/source/docs/takeoff-landingpage.rst
```diff
@@ -0,0 +1,55 @@
...
+- Throttle up and firmly throw the plane level, not nose pointed up. Common mistakes are pointing the nose up, throwing it toward the ground, or rotating your wrist like a football throw. In the first second or so, the plane has low airspeed and the flying surfaces have no authority to correct non-level attitude until it accelerates to flying speed.
```
Naterater: I'd suggest throwing a few degrees nose-up, otherwise it's common to hit the ground.

plane/source/docs/takeoff-landingpage.rst
```diff
@@ -0,0 +1,55 @@
...
+
+This is commonly used for very heavy or very high stall speed planes for launching.
+A tutorial on building and using a bungee launcher is `here <https://www.rcgroups.com/forums/showthread.php?1459906>`__
+
+VTOL
+====
+
+Perhaps the easiest of all takeoffs, it does require that the plane be a VTOL capable :ref:`Quadplane <quadplane-support>` 
+
+.. tip:: While not necessary for a successful launch, enabling :ref:`SERVO_AUTO_TRIM<SERVO_AUTO_TRIM>` will automatically trim the plane during stabilized operations, such that switching to MANUAL or ACRO mode later will have an in-trim plane automatically.
```
Naterater: Add another tip that pitch control after running an AUTOTUNE sequence will greatly improve takeoff performance in pitch-stabilized modes.

plane/source/docs/takeoff-landingpage.rst
```diff
@@ -0,0 +1,55 @@
...
+- Having flying surfaces centered and neutral at neutral RC stick positions
+- Having the "level" calibration setup such that the level horizon (0 deg pitch) in FBWA mode results in the wing chord line being a few degrees positive, which will result in a positive angle of attack in "level" flight, the normally cruising attitude for a plane.
+- Using FBWA mode or :ref:`automatic-takeoff` . While launching in MANUAL can be done, using a stabilized mode will eliminate a lot of pilot effort to overcome trim issues (see Tip at end of this page). If using FBWA mode, a few degrees of up elevator (~10%) during the launch will result in an automatic climb. Add this via a switch or stick pressure. Do not change TX trim positions used during RC calibration since it affects all other modes of flight.
+- Throttle up and firmly throw the plane level, not nose pointed up. Common mistakes are pointing the nose up, throwing it toward the ground, or rotating your wrist like a football throw. In the first second or so, the plane has low airspeed and the flying surfaces have no authority to correct non-level attitude until it accelerates to flying speed.
+
```
Naterater: Add another point about launching into the wind. Hand-launches with a tailwind are often disastrous because the airspeed is negative to start and requires a lot of "delta V" or velocity change to get the control surfaces to do anything.

plane/source/docs/takeoff-landingpage.rst
Hwurzburg: good point will add

plane/source/docs/takeoff-landingpage.rst
Hwurzburg: the problem is that if CG is a bit back, or the throw is weak, this almost always results in a stall....your planes must tend to be a bit nose heavy....since this is aimed at a newbie (experienced users dont need this advice), it is safer to have a grass burner launch (if they are very nose heavy) than a tip rolling stall...the former almost never damages, the latter is a potential rebuild....

plane/source/docs/takeoff-landingpage.rst
Hwurzburg: again, a newbie cant judge a few degrees...we launch as you suggest, but trying for dead level will seldom result in a rebuild if CG is correct..and btw, something like a very heavy MiniTalon wont tolerate even a few degrees..again, the conservative approach here

plane/source/docs/takeoff-landingpage.rst
Hwurzburg: I think this is covered in the tuning section....however, a tip about reducing pitch P from default on small wings I think I should add...btw AUTOTUNE does not always converge...

plane/source/docs/takeoff-landingpage.rst
Hwurzburg: yep...should add nose into the wind...thanks

Hwurzburg:
thanks for the inputs, Nate!

plane/source/docs/takeoff-landingpage.rst
Naterater: The wiki already warns the user about being tail-heavy during setup; I don't think we need to compromise the success of most by thinking a few users will be tail-heavy. Most aircraft today already have very conservative manufacturer CG's. 

Furthermore, CG location actually doesn't have nearly as much of an effect on takeoff stability as one might think. The aerodynamics at play are minimal because we're at an airspeed so low that stability doesn't really come into play. We also often have thrust over an elevator because the throttle is full (unlike a power-off stall/spin).

Saying the wiki is primarily for the novice user isn't really great; I'm a seasoned user and I still get great advice by overlooking the wiki. One of the most frustrating things is being an advanced RC pilot with a plane that is set up very well in Manual, but the first hand launch skims the grass and crashes an otherwise perfect plane. Those people need the most specific and accurate advice, and they are definitely important subjects for this wiki.

I have had a tip rolling stall on takeoff before, but that was because I threw the tail sideways and down. It had nothing to do with any demand from the autopilot; the control surfaces could have been reversed and it would have done the same thing. Only difference is that it would have nearly recovered on its own if it wasn't so close to the ground. Bad throw technique causes lots of problems.

plane/source/docs/takeoff-landingpage.rst
Naterater: I just don't have the same experience at all; I've launched many "overweight" aircraft that barely made it off the ground, and if they hadn't maintained the 30-degrees of pitch; they would be in pieces on the turf. The "level" acceleration that is gained by a lower takeoff AOA is almost never worth it IMO.

plane/source/docs/takeoff-landingpage.rst
Naterater: Maybe just leave a note that after tuning the PTCH2SRV values, the aircraft may achieve and maintain its intended pitch significantly faster. The behavior of the takeoff WILL change after changing the Pitch tune, so do the pitch tune first before experimenting with different takeoff configurations if at all possible.

plane/source/docs/takeoff-landingpage.rst
Naterater: Oh... Running takeoffs are another beef of mine related to tips.

I think we should recommend NOT doing them as the throw technique is compromised so much for so little gain in velocity. Instead, find a location with no obstacles in front of you and do a step forward like tossing (or throwing) a ball. Just ideas.

plane/source/docs/takeoff-landingpage.rst
tridge: I agree with Henry on this

tridge:
@Naterater maybe you could do a page for the believer that explains the launch process for that plane?

rmackay9:
By the way, we have this page that I started on but it was never finished because I don't actually have one. https://ardupilot.org/plane/docs/airframe-believer.html

PR #2373, opened 2019-12-28
Add troubleshooting section for composite port problems in MP

Hwurzburg:
fyi, section has a tag so you can give a user a direct URL  link to it by copying the tag line

@tridge I did not explictily show an image of the driver box with USB Composite pointed to since my computer that driver name was taken over by the "Teensey" board driver...if you have one, send it and I will graphically edit it and replace it either after merge or before...


tridge:
looks great, thanks!

PR #2383, opened 2019-12-30
Expand on tools in MP under CTRL-F page

Hwurzburg:
WIP....am open to suggestions on what other hidden tools might be of use to the average user and will include them also...


khancyr:
I think we already discussed this.
From memory those were intentionally not documented to prevent misuage by novice. It can still be discussed as some tools are really useful

rmackay9:
I wonder if we should consider putting these on the MP wiki instead of the vehicle code wikis?  the problem with this idea is that people may never find them.

For me, the mavlink explorer and proximity buttons are the ones I use the most.

Hwurzburg:
MAVLink inspector is discussed in the rover wiki so its not really hidden, I was asked to document the MAVLink forwarding and updating the bootloader by Tridge, so while I was at it, I thought the the warning manager and the Inspector were really good...I can add a warning not to play with anything else, unless there is something worthwhile to also be added

Hwurzburg:
Randy, it is only on the MP wiki...rather buried at that...

rmackay9:
@Hwurzburg, ah right.  I thought it was appearing on all the wikis.  great!

PR #2387, opened 2019-12-31
Companion Computer Update

stephendade:
A few fixes:
-Better layout of the companion computer section
-Updated instructions for using the Raspberry Pi


rmackay9:
I think this looks OK.  I don't setup companion computers much so I hope that the removal of some information is OK because there are links to other documentation that specify how to do the same thing.

Hwurzburg:
if a platform still is being manufactured and the firmware works on it, why would we archive it? or are they discontinued now?

stephendade:
>I think this looks OK. I don't setup companion computers much so I hope that the removal of some information is OK because there are links to other documentation that specify how to do the same thing.

The idea is to link to the official RPi documentation for the more generic tasks, rather than repeating ourselves (and the maintenance overhead associated with that) here.

>if a platform still is being manufactured and the firmware works on it, why would we archive it? or are they discontinued now?

The Edison was discontinued by Intel a year or two ago. BeaglePilot is covered under the flight controllers section over at https://ardupilot.org/dev/docs/building-for-beaglebone-black-on-linux.html#building-for-beaglebone-black-on-linux (Plus the companion computer page for BeaglePilot is just a set of links to the BeaglePilot  page, rather than any actual setup information.

Hwurzburg:
Stephen, several links were removed in the raspberry-pi-via-mavlink document  that are referenced elsewhere in the wiki, breaking other documents....can you fix these? thanks

stephendade:
@Hwurzburg - sure, I can do that. Is there any tool or command I can use to indicate where the broken links are? Of is it a case of grep though the source code?

Hwurzburg:
here is the error:
/home/henry/Desktop/ardupilot_wiki/dev/source/docs/odroid-via-mavlink.rst:166: WARNING: undefined label: raspberry-pi-via-mavlink_install_the_required_packages_on_the_raspberry_pi (if the link has no caption the label must precede a section header)

/home/henry/Desktop/ardupilot_wiki/dev/source/docs/odroid-via-mavlink.rst:221: WARNING: undefined label: raspberry-pi-via-mavlink_connecting_with_the_mission_planner (if the link has no caption the label must precede a section header)

two labels were removed in raspberry-pi-via mavlink doc that were referenced in the above doc....I suggest that the links be replaced in that doc appropriately...

fyi, when editing, if a link anchor in a document is removed  (anything with _ xxxxxx: ) ..."git grep" the wiki to see if it is referenced by another document and fix appropriately...

stephendade:
@Hwurzburg - anchors fixed in PR: https://github.com/ArduPilot/ardupilot_wiki/pull/2424

Hwurzburg:
see comment in that PR

PR #2392, opened 2020-01-02
Add warning regarding motor interlock for tradheli users

bnsgeyer:
@rmackay9  and @Hwurzburg  I am submitting this PR to address an issue regarding radio failsafe with Copter 3.6 or earlier.  Looking for feedback on the warning placement and clarity.
@JoshWelsh  Any inputs on this PR would be appreciated.


rmackay9:
looking generally good but it would be nice to put TradHeli right near the start of the warning so that multicopter users only need to glance at it to know that it doesn't apply to them.

JoshWelsh:
Agree with Randy in that it should be as specific as possible about applying only to TradHeli users.  Good call on putting this in there, thanks Bill!

rmackay9:
LGTM!

Hwurzburg:
eveeryone agrees...merging

Hwurzburg:
ack! sorry, I should have tested this first....parameter is wrong....will do a patch later today to fix it...

PR #2398, opened 2020-01-03
guide-tilt-rotor:  Q_M_YAW_SV_ANGLE  added

Rolf-G:
I just noticed, that Q_M_YAW_SV_ANGLE ist still missing.


Hwurzburg:
I think we need a bit more here....isnt the lean angle max dependent on how you set the SERVOn_MIN/MAX/TRIM values and mechanical linkage for the output assigned to function 39?
In other words, there must be an assumption of how all that is setup in order to correlate to an actual angle maximum, right?

Rolf-G:
Right, you enter the resulting angle between SERVOn_MID and SERVOn_MAX (or SERVOn_MIN). The firmware assumes symmetrical servo end points.

Hwurzburg:
are you going to add it, or shall I after I merge this?

Rolf-G:
it will be faster and better if you do that ;-)

Hwurzburg:
Will add as part of another fix I am doing to the wiki....merging this now....Thanks!

PR #2403, opened 2020-01-05
Expanded Motor Interlock description in Auxillary Functions page

Hwurzburg:


Hwurzburg:
one question....does this operate as a motor kill in copter ie forces all motors to output min, and is different in Heli since its controlling HeliRSC but has an idle param?

ChristopherOlson:
I'm not exactly sure what it does in Copter for multicopters. For heli the so-called "motor interlock" is throttle hold in any other FBL system, or RC heli setup.

http://www.rcheliwiki.com/Throttle_hold

bnsgeyer:
It operates the same way for multicopters. Users can use it as second line of safety. I think Leonard uses it a lot.

Hwurzburg:
I think its probably a bit different in that it goes to SERVO MIN, not an idle value like Heli....at least Randys video shows it doing that...so that is how I wrote it..think its good to go now...thanks

bnsgeyer:
Yeah. I updated my previous response.  You are right

Hwurzburg:
merged...thanks for the review

common/source/docs/common-auxiliary-functions.rst
```diff
@@ -391,8 +391,7 @@ Stops motors immediately
    <td><strong>Motor Interlock</strong></td>
    <td>
 
-Opposite of Emergency stop (above) in that switch must be ON for motors
-to spin (`video <https://youtu.be/-Db4u8LJE5w?t=51>`__).
+Motor Interlock controls the way the heliRSC (motor throttle control) output is generated in Traditional Helicopters and Multicopters. If > 1200us, it enables the Motor Interlock function, below its is disabled. When <1200us, it is similar to what is sometimes referred to as Throttle Hold in RC Helicopter terminology for Traditional Helicopters and HeliQuads. For Mulit-copters it is used as a motor stop function. (`video <https://youtu.be/-Db4u8LJE5w?t=51>`__).
```
bnsgeyer: This is a little confusing.  You mention heliRSC and then multicopters in the same sentence. Multicopters don’t use the heliRSC servo function

Hwurzburg: I see what you mean...will fix...meant HeliQuads...

PR #2417, opened 2020-01-08
Gazebo sitl update

bromeara:
Removed claims that there is no built-in gazebo support for ardupilot. I have tested out the plugin for gazebo9 however the plugin seemed to be added in gazebo8 according to the source repository. I left the plugin installation instructions and considered moving them to the bottom but decided to leave them where they were and added a note that it could be skipped so it wouldn't throw off the flow since I am unsure which version is standard for Ubuntu 16.04.

### DIFF
--- a/dev/source/docs/using-gazebo-simulator-with-sitl.rst
+++ b/dev/source/docs/using-gazebo-simulator-with-sitl.rst
@@ -11,7 +11,8 @@ Overview
 ========
 
 Gazebo is a well-known and respected robotics simulator, and is also the official DARPA Virtual Robotics Simulator.
-No current release has built-in support for ArduPilot,  however. (Previous PRs for built-in support were not merged as of April-2017).
+Built-in support for ArduPilot is available starting with gazebo8.
+If you are using gazebo8 or above, you should skip the following Plugin installation section.
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


bromeara:
After a short exchange with @khancyr I learned the external plugins are still the preferred way to interact with gazebo so I will be revising my edits to reflect that while still removing the statment there is no built in support for gazebo.

khancyr:
I will lookback into gazebo, but I am pretty sure that only the build in pluggin is outdated and cannot work on all vehicle. That is why I won't recommend it

bromeara:
That's fair I was under the assumption that since it wasn't mentioned it was added after the wiki page was created so if it is the case that it is just out of date it might be worth mentioning on the wiki it is not recommended to use. If you can confirm its outdated I can make the change.

EceChaik:
Guys, this might be a silly question, but I was wondering, how does the SITL know that Gazebo is being used for dynamics simulation and switches to Gazebo instead of using the built-in sitl multicopter simulator?

Hwurzburg:
the command to start simulation specifies the gazebo frame type....https://ardupilot.org/dev/docs/using-gazebo-simulator-with-sitl.html

lets move support questions off a closed issue an into the support forums, please discuss.ardupilot.org

PR #2420, opened 2020-01-09
Uavcan setup

ukmook:
First time PR, UAVCAN re-write.


Hwurzburg:
Malcolm, after some discussion we decided that the CANBUS page should remain as is...showing use of only one driver....we intend to expand instructions for the use of the second driver for redundancy in the future. Also,  it is better to enable only one CAN interface unless absolutely necessary to save memory...also a topic for future expansion since the second can be used for backup also...
I also expanded on some things and changed the formatting....added the F7/H7 page, and squashed the intervening commits so that there is only two ....I am unfamiliar with the FLASH_BOOTLOADER stuff...I am not sure that the parameter has to be changed, but left it untouched for @tridge to review...
thanks for the contributions! hope there are more in the future on the wiki

ukmook:
Henry, happy that it is useful, I did test everything along the way which is why I couldn’t do the F7/H7 page. Changing the Flash bootloader param to 1 just gives the debug message. 
I will keep helping out with the wiki.

Hwurzburg:
Actually, changing that param forces a bootloader update immediately after setting it, if available...if bootloader is current it does nothing....thanks!!

Hwurzburg:
broke out bootloader into its own section

tridge:
a note on using a short file name for uavcan gui tool fw update would be nice
also, SERIAL0_BAUD never matters for chibios boards

Hwurzburg:
@tridge , will update after merge....thanks Malcolm! merging

ukmook:
Thanks for the help with this Henry. @tridge I’ll shorten the terms in future.

PR #2424, opened 2020-01-10
Dev: Clearer instructions for RasPi serial configuration

stephendade:
I missed updating this part in my earlier cleanup of the companion computer documentation.


Hwurzburg:
odroid-via-mavlink.rst:166: WARNING: undefined label: raspberry-pi-via-mavlink_install_the_required_packages_on_the_raspberry_pi (if the link has no caption the label must precede a section header)

odroid-via-mavlink.rst:221: WARNING: undefined label: raspberry-pi-via-mavlink_connecting_with_the_mission_planner (if the link has no caption the label must precede a section header)

fyi, above are producing build errors from first PR merge...which is what I was referencing in my previous comment....thanks

Hwurzburg:
also, if images are replaced, I think we should delete them to prevent clutter growth....also from first PR merge....thanks

stephendade:
@Hwurzburg - The second commit in this PR has fixed both issues -  the labels are now included and I removed the unused images.

Hwurzburg:
thanks! merged

PR #2425, opened 2020-01-10
Update environnement

khancyr:
Little clean up of sphinxsetup
Make use of the script for vagrant
Clean up vagrant setup
remove vimeo plugin source as they belong to root

Tested on windows with vagrant

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -1,16 +1,35 @@
 #!/bin/bash
-    apt-get -y update
-    apt-get install -y unzip git imagemagick mercurial python-pip curl
-    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && rm -f get-pip.py
-    # Install sphinx
-    pip install -U sphinx==1.8.3
-    # Install sphinx theme from ArduPilot repository
-    pip install git+https://github.com/ArduPilot/sphinx_rtd_theme.git -UI
-    # and a youtube plugin:
-    pip install git+https://github.com/sphinx-contrib/youtube.git -UI
-    # and a vimeo plugin:
-    hg clone https://bitbucket.org/jdouglass/sphinxcontrib.vimeo
-    pushd sphinxcontrib.vimeo
-    python setup.py build
-    python setup.py install
-    popd
+set -e
+set -x
+
+if [ "$UID" -gt 0 ]; then
+     echo "Sorry, this script must be run as ROOT!"
+     exit 1
+fi
+
+apt-get -y update
+apt-get install -y unzip git imagemagick mercurial curl wget make
+
+# Get pip through the official website to get the lastest release
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && rm -f get-pip.py
+
+# Install sphinx
+pip install -U sphinx==1.8.3
+
+# Install sphinx theme from ArduPilot repository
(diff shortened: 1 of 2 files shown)


Sphinxsetup.sh
```diff
@@ -1,16 +1,35 @@
...
+hg clone https://bitbucket.org/jdouglass/sphinxcontrib.vimeo
+pushd sphinxcontrib.vimeo
+python setup.py build
+python setup.py install
+popd
+# remove vimeo plugin source as they belong to root
+rm -rf sphinxcontrib.vimeo
+
+# Say that we finish
+echo "Setup complet successfully !"
```
Hwurzburg: completed

Hwurzburg:
Has this been tested on a virgin Ubuntu install with only the wiki repo cloned?

khancyr:
Yep I test it on a fresh VM, in vagrant on windows and on docker.
there is still an error spotted by the docker container, I will fix that. We miss to explicitly install wget !

khancyr:
review will be easier on github if you check the box to remove space changes !

khancyr:
ok all dependencies are installed now ! it works on minimal docker too !

khancyr:
it is working on windows with vagrant, I need to find another computer as I cannot launch vagrant from virtualbox. But there shouldn't be difference between vagrant on windows and vagrant on linux

Vagrantfile
```diff
@@ -1,33 +1,14 @@
...
-    pushd sphinxcontrib.vimeo
-    python setup.py build
-    python setup.py install
-    popd
-
-  SHELL
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  config.vm.box = "ubuntu/bionic64"
+  config.vm.provider "virtualbox" do |vb|
+      vb.name = "ArduPilot_wiki"
```
brunoolivieri: If I have more than one Wiki Repo in different folders on my machine and I evoke Vagrant from each folder, it creates more than one VM. Vagrant creates a VM for each repo with a different sufix.

Would this behavior cease with the VM name being explicit set?

I do not see it as an issue because it is not usual to be useful. I just want to understand if there are some new behaviour.

brunoolivieri: If I have more than one Wiki Repo in different folders on my machine and I evoke Vagrant from each folder, it creates more than one VM. Vagrant creates a VM for each repo with a different sufix.

Would this behavior cease with the VM name being explicit set?

I do not see it as an issue because it is not usual to be useful. I just want to understand if there are some new behaviour.

khancyr: it will use the same machine but change the directory in /vagrant to target the right repo where you launch the vagrant up. 
So yes, that the end of the multiple VM. The VM can still be updated with vagrant provision

khancyr:
ok I have test on a fresh ubuntu 18.04, and there is still an error. Mercurial installation need to enable universe repo. I am fixing this and testing back on vagrant

khancyr:
All good now ! The script is corrected.
It works on Ubuntu 18.04 out of the box, on vagrant, and docker

Vagrantfile
brunoolivieri: No problem I did use it only when changing stuffs on the Sphinx´s theme.  tks!

Hwurzburg:
duh, was setting up VM for Ubuntu install with only 4MB of memory...installs a bit faster with 4GB reserved for it.....I can now test in a VM also here...will do so later this AM

Hwurzburg:
verified on minimal install with no updates...thanks

PR #2427, opened 2020-01-10
common: Change FLOW_ENABLE to FLOW_TYPE

muramura:
I have no variable for FLOW_ENABLE.
I know that setting the device type to FLOW_TYPE disables and enables it.
I change the config parameter name.


rmackay9:
Thanks very much, merged!

brunoolivieri:
Actually, there are both [FLOW_ENABLE](https://ardupilot.org/copter/docs/parameters.html#foll-enable-follow-enable-disable) and [FLOW_TYPE](https://ardupilot.org/copter/docs/parameters.html#flow-type-optical-flow-sensor-type).

rmackay9:
@brunoolivieri, thanks, I've raised a PR to resolve some of the issues with FLOW_TYPE https://github.com/ArduPilot/ardupilot/pull/13228

Hwurzburg:
when merged there will be a lot of wiki build errors....considerable cleanup....also, as it stands now FLOW_ENABLE going away requires these to be fixed....the links will need to be removed and the params highlighted with ticked boxes...but hopefully we have already gotten the migration notes in for all the parms now in 4.0 not in 3.6

brunoolivieri:
I suggest to wait a week, for the parameters versioning be alive, so I could maintain some references to the 3.6.x and the new links to the 4.x.

PR #2435, opened 2020-01-13
Minor link update needed

rishabsingh3003:
The link that "_`Change the delay here from 1ms to 2ms [here in AP_Param::show_all]`_" was forwarding to, needed an update.


Hwurzburg:
merged...thanks

PR #2439, opened 2020-01-14
Fix versioning script

brunoolivieri:
Small fix due some tests on the server.


Hwurzburg:
wish there was a way for the pages to open to either the latest or the current stable instead of the bottom of the list in the drop down boxes

brunoolivieri:
They should open in lastest. Working on that. :-)

PR #2456, opened 2020-01-16
Add new param and correct incorrect param names

Hwurzburg:


Hwurzburg:
corrected in this PR....are other places needed? you said wiki not PR...need pointers if there are since both MC and Trad share this param, correct?

bnsgeyer:
@Hwurzburg i meant this PR.  But this only addresses the Helicopter tuning wiki.  I'm sure there may be other places in the MC wiki's that address these filters like their tuning wiki.  But Leonard may have already caught that.

PR #2458, opened 2020-01-16
Re-organize and expand LEDs page

Hwurzburg:
addressing #2423

PR #2459, opened 2020-01-16
Add NTF config page

Hwurzburg:
adding page for adv config on NTF devices....I have tried to avoid as much as possible, any duplication on the hardware pages for leds and buzzers...an image here is used in #2458


Hwurzburg:
added TOC change per review with @rmackay9 ...merging

PR #2465, opened 2020-01-21
common-external-leds: correct pwm vs gpio note

IamPete1:
This corrects a typo in the output setup instructions.


Hwurzburg:
sorry, this was my understanding as I ramped up on Neopixel....if it needs a PWM output and not GPIO for operation (I am assuming that includes its use as an NTF device as well as a LUA scripted string), then a lot more changes to this page are required...once this is confirmed, I will change all of them on the page...

IamPete1:
Yeah, needs PWM output, its works like BLHeli rather than traditional IO, I was supprised to see the diode advice on there so quick. Though i was dumb not to have checked on the wiki when i first saw it. ;). Very quick work!

IamPete1:
Closing in favor of @Hwurzburg doing a more comprehensive fix. Thanks!

Hwurzburg:
2466 merged....thanks!!

PR #2467, opened 2020-01-21
Dev: Add note about Cygwin not wanting to do ARM builds

stephendade:
Based on recent comments in the Ardupilot Gitter chat, multiple people have reported issues building the ARM-based (ie using gcc-arm) builds under Cygwin.

Adding in this small message to suggest people use WSL instead (which does work), as will cut down on the number of "why is Cygwin not working" posts in the Gitter chat.


rmackay9:
It might be a good idea to add this advice on the landing page which lists cygwin and WSL.. that might be the point where people make the decision as to which way to go

PR #2468, opened 2020-01-22
update output functions list

Hwurzburg:


Hwurzburg:
@tridge, just guessed at rate outputs...let me know how to change

Hwurzburg:
@tridge, thanks for looking at this...however, I really havent found the scaler for the THRUST output...I was expecting a correction ... there isnt a ATC_RAT_THRST_FF as far as I know, but dont know what it is...

PR #2473, opened 2020-01-23
fix wiki build errors due to retired params

Hwurzburg:
@rmackay9 haven't the foggiest idea why these did not appear until today's build check....a build of master an hour ago did not produce them....maybe something to do with the new parameter versioning building process?


rmackay9:
@Hwurzburg,

Hi there, I'm not sure either.  Thanks for sorting it out though in any case!

Issue #1070, opened 2018-01-26
Rover: add list of known working motor drivers


rmackay9:
In the Assembly section of the rover wiki, perhaps on the [motor-and-servo connections page ](http://ardupilot.org/rover/docs/rover-motor-and-servo-connections.html) we should add a list of known working motor drivers to allow the DIYers to pick a motor driver that we know works.

Hwurzburg:
anyone have a list? I will add....

rmackay9:
I know that these ones work:

- [RoboClaw 2x7a](https://www.pololu.com/product/3284) ([from this ArduRoller reference page](https://ardupilot.org/rover/docs/reference-frames-arduroller.html))
- [BlueRobotics Basic ESCs](https://bluerobotics.com/store/thrusters/speed-controllers/besc30-r3/) ([from the boogie board boat reference vehicle](https://ardupilot.org/rover/docs/reference-frames-boogieboard-boat.html))

Hwurzburg:
closing

PR #2483, opened 2020-01-28
update brushed motor drivers list

Hwurzburg:
addresses #1070


rmackay9:
OK, I see, it's a bit odd that it's a brushless motor driver that we talk about in the brushed motor section.  On the otherhand, it's a high end motor driver that I know a user on the forum confirmed works with AP.

don't know.. I'm not sure what to do .. i don't mind keeping it or removing it

Hwurzburg:
okay merging.... @tridge  PWM speed parameter is discussed in the document already....will add diagram for FET drive in the future

PR #2487, opened 2020-01-29

tatsuy:
add the i2c's address description

### DIFF
--- a/common/source/docs/common-benewake-tfmini-lidar.rst
+++ b/common/source/docs/common-benewake-tfmini-lidar.rst
@@ -32,7 +32,7 @@ If the SERIAL4/5 port on a Pixhawk is being used then the following parameters s
 -  :ref:`SERIAL4_BAUD <SERIAL4_BAUD>` = 115 (115200 baud)
 -  :ref:`RNGFND1_TYPE <RNGFND1_TYPE>` = 20 (Benewake TFmini/Plus-Serial)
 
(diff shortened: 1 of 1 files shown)


rmackay9:
Merged, thanks.  Is it actually necessary to set the _ADDR parameter to make it work?

tatsuy:
Yes, we need to set the _ADDR to use the sensor.

rmackay9:
@tatsuy, OK thanks.  That's not good really, we should default to using the correct I2C address in the driver.  For example I just did this recently for the new LeddarVu8 driver.  Here's an issue on the issues list so we don't forget. https://github.com/ArduPilot/ardupilot/issues/13419

tatsuy:
Do you mean it should use default I2C address if setting RNGFND1_ADDR = 0?
Then, we may need to change the following description.
https://ardupilot.org/copter/docs/parameters-Copter-latest-V4.0.2-dev.html#rngfnd1-addr-bus-address-of-sensor
> A value of 0 disables the sensor.

rmackay9:
@tatsuy, yes, I totally agree with you. There's no need to provide a second way for users to disable the sensor.  Unfortunately this means we will need to go through and see if any of the other drivers are actually allowing the sensor to be disabled by setting the ADDR = 0.

tatsuy:
OK. I understand we will edit the wiki after checking whether the default address is used by setting the ADDR = 0 in the other drivers.

PR #2493, opened 2020-01-30
Fix toctree for parameters in versioning build

brunoolivieri:
Fixes the left toctree for the latest parameters page.  (Only and just only for the latest firmware version, the previous stable versions are not supposed to appear)

Obs.: it includes minimal enhances to debug better and present comparations of os.stat() and hash comparations for the cache.


brunoolivieri:
Tested locally, and I will watch the results of the next couple of runs on server.

PR #2499, opened 2020-01-30
Update gsoc-ideas-list.rst

jmachuca77:
Add Swift package


jmachuca77:
Yes, and I'll be willing to mentor and maintain, but would also like for the student to become involved in the project and maintain it. Our overall goal should be to capture more long term contributors, so that should not be a criteria to discard a project.
MAVSDK is already working and would be a matter of just adding ardupilot support, updating the pymavlink generator is more involved but its certainly doable in 3 months.

auturgy:
Before tackling language specific bindings, the mavsdk-server would need to support ardupilot. I’ve had a very brief discussion with Jonas, the mavsdk maintainer, and he’s willing to assist if a GSoC student wants to take that on.

jmachuca77:
Sounds good and better if they are willing to help with the support. I think that would be good for AP.

rmackay9:
I've rebased separately and merged, thanks!

PR #2502, opened 2020-02-02
Re-organize ESC,Dshot, and BLHeli32 into config and hardware pages

Hwurzburg:


rmackay9:
I think this may get a bit confusing with the existing [Peripheral Hardware >> ESCs and Motors](https://ardupilot.org/copter/docs/common-escs-and-motors.html) section..

Hwurzburg:
@rmackay9  The structure logic I have been trying to follow is that descriptions to enable peripheral hardware choices are in the Peripheral Hardware pages along with any required minimal or specific device unique setup necessary for basic function, while Advanced Configuration are things not required for basic functions beyond that which is described as one flows thru First Setup and First Flight. Nothing is perfect in this regard, but long descriptions of BLHeli pasthru and telemety definitely belong in Advnced Config in my opinion.

The brushless escs need no setup to work in the default configuration of themselves nor Ardupilot, hence no configuration info for protocol selection or BLHeli extensions is given in the DShot section but, rather, is pointed back to the Advanced config section in several places in the brushless ESC hardware document.

rmackay9:
Merged, thanks!

PR #2504, opened 2020-02-04
Add docker support for building

khancyr:
Add support for docker
add instruction for docker
Correct Sphinxsetup for debian
add pycharm as ide for visualisation

### DIFF
--- a/common/source/docs/common-wiki-editing-setup.rst
+++ b/common/source/docs/common-wiki-editing-setup.rst
@@ -137,6 +137,7 @@ A combination of two Windows tools can help you previewing your modifications:
   	
 * `Notepad++ plugin for RST files <https://github.com/steenhulthin/reStructuredText_NPP>`__
 * `restview (on-the-fly renderer for RST files) <https://mg.pov.lt/restview/>`__
+* `PyCharm <https://www.jetbrains.com/pycharm/>`__
 
 The Notepad++ plugin helps you with code completion and syntax highlighting during modification.
 Restview renders RST files on-the-fly, i.e. each modification on the RST file can be immediately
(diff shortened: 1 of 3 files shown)


common/source/docs/common-wiki-editing-setup.rst
```diff
@@ -98,6 +99,28 @@ With your favourite web browser, open the locally built wiki which should be nea
 - For Rover look for ``ardupilot_wiki/rover/build/html/index.html``
 - For Developer look for ``ardupilot_wiki/dev/build/html/index.html``
 
+Setup with docker
+-----------------
+
+#. Download and install Docker according to their `official documentation <https://docs.docker.com/install/>`__ .
+
+#. Open a command prompt in the root of the ardupilot_wiki repo and build the docker container for the wiki:
```
OlivierJB: Needs to be a Powershell command prompt on windows. Old DOS command prompt crashes on path in docker run -v option:

docker run --rm -it -v "${PWD}:/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python update.py
docker: Error response from daemon: create ${PWD}: "${PWD}" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.

OlivierJB: Nice, works great!

khancyr:
@OlivierJB Could you work up the powershell command ? I don't have windows computer to try it ...

OlivierJB:
It just works with Powershell given how it parses file paths (as opposed to dos command prompt), so no need to change anything.

khancyr:
Then I will update to add a comment about windows and powershell !

PR #2507, opened 2020-02-05
Update common-wiki-editing-setup.rst

OlivierJB:
Add example wiki URL


OlivierJB:
Re-visiting this and some  conversations, this may add unnecessary complexity and is not really needed, so closing.

PR #2513, opened 2020-02-07
adds tradheli project idea to GSOC list

bnsgeyer:
@rmackay9 here is my idea for GSOC for trad heli.  Let me know what you think.  Certainly not as challenging as autonomous autorotation but is a capability that would be great to have for tradheli.


rmackay9:
Merged, thanks!

PR #2522, opened 2020-02-10
Update building-setup-linux.rst 

vishal-raj15:
a note to avoid firewall problem

### DIFF
--- a/dev/source/docs/building-setup-linux.rst
+++ b/dev/source/docs/building-setup-linux.rst
@@ -25,6 +25,16 @@ Clone ArduPilot repository
     :start-after: inclusion-marker-do-not-remove
     :end-before: Cloning with the GitHub GUI (Windows or MAC)
 
+
+.. note:: in case some firewalls do not allow ssh access which can cause the above submodule updates to fail, in this instance you can tell git to unilaterally use https through the following command:
+
+    ::
+
+         git config --global url."https://" 
+
+    to use https protocols instead of the default git:// prefix.
+
+
 Install some required packages
 ------------------------------
 


dev/source/docs/building-setup-linux.rst
```diff
@@ -24,6 +24,9 @@ Clone ArduPilot repository
 .. include:: git-clone.rst
     :start-after: inclusion-marker-do-not-remove
     :end-before: Cloning with the GitHub GUI (Windows or MAC)
+    
+.. note::in case you get a connection error because of firewalls type command :
```
andyp1per: You could say:
"Some firewalls do not allow ssh access which can cause submodule updates to fail, in this instance you can tell git to unilaterally use https through the following command:"

auturgy:
I thought that the Git:// interface had been deprecated. Is this even an issue?

vishal-raj15:
@auturgy may be it is not an issue but some people may stuck on that , i mean beginners like me

dev/source/docs/building-setup-linux.rst
vishal-raj15: ook , thanks!

Hwurzburg:
fixed induced wiki build errors, formatting, and minor wording clarification.
@peterbarker if you approve, it can be merged...

Hwurzburg:
@khancyr Pierre, changed format as suggested and pushed to his branch....but I have reservations about adding the note globally....
 if you follow our setup guide, all other accesses besides submodule updates will be via remote alias and already be explicit for https, correct? thats why he only ran into it when he did the submodule update with his firewall setup...therefore someone following the guide would not know they needed the change until the submodule update, correct? 
if its 100% innocuous in all configurations we can add it to the git setup, if not, should it not stay where it is?

khancyr:
ha, I didn't see that we are asking to git clone with http already.
The issue is that it could appear on all setup (mac, windows, cygwin). I guess we can keep it here until we get somebody that complain about this error.
There is also a git-submodules page that could be a good candidate to hold this tips and get linked from the linux page

vishal-raj15:
> offset the git config command of 4 space

@khancyr sory but ,i dont understand what it means

Hwurzburg:
@vishal-raj15 dont worry about it...I already pushed a fix for it into your repo and it appears in this PR now...

Hwurzburg:
and I just pushed the correct version back into your branch...the format is now re- corrected...this should be merged early next week during our wiki call ...thanks for helping to improve the wiki!

vishal-raj15:
@Hwurzburg  thanks for help

rmackay9:
Merged, thanks!

PR #2554, opened 2020-02-19
fix list formatting in qacro-mode.rst

kd0aij:
@Hwurzburg I see I had also screwed up the list format. This is an attempt to fix that without losing the param hotlinks.


Hwurzburg:
LGTM...thanks...merged

PR #2560, opened 2020-02-21
Update sitl-with-morse.rst

ashvath100:
Since [#13455](https://github.com/ArduPilot/ardupilot/pull/13628) is merged , morse now has support for Separate Steering and Throttle Rover


rmackay9:
Merged, thanks!

PR #2581, opened 2020-02-26
Add compatibility warning on MiniPix

Hwurzburg:
added note per wiki call...also an expansion that page info is not to be used


rmackay9:
Merged, thanks!

PR #2594, opened 2020-03-03
Added information on using I2C on Wing FC10

alteman:
Found a way to use I2C on this board.


Hwurzburg:
nice...can you take a photo of the mod? I can help edit and add pointers,etc. thanks

alteman:
Here's the pad on the board version from [Banggood](https://www.banggood.com/Wing-FC-10-DOF-Flight-Controller-INAV-OSD-Accelerometer-Barometer-Gyro-Compass-For-RC-Airplane-Drone-p-1318626.html?cur_warehouse=CN) (no silkscreen there, I guess it's only on my version).
Wing FC-10 SDA
Wing FC-10 SCL

You're right, it's worth mentioning that builtin compass needs to be disabled via config parameters.

Hwurzburg:
Just added images and some more explanations.....merging...thanks!

PR #2592, opened 2020-03-03
Fix for antennatracker parameters versioning

brunoolivieri:
Fix an error regarding previous versions.

PR #2598, opened 2020-03-04
Fix tracker parameters

brunoolivieri:
PR #2592 was not fully effective.


brunoolivieri:
Applying and watching on server.

PR #2602, opened 2020-03-07
Description of In-Flight FFT support

andyp1per:


andyp1per:
@Hwurzburg this is ready for review.
@lthall and @bnsgeyer you may want to throw in your thoughts here.

common/source/docs/common-advanced-configuration.rst
```diff
@@ -67,6 +67,7 @@ tuning options for the vehicle.
     Notification Devices (LEDs,Buzzer,etc.) <common-ntf-devices>
 [site wiki="copter,plane"]
     Notch Filter Configuration <common-imu-notch-filtering>
+    In-Flight FFT Configuration <common-imu-fft>
```
Hwurzburg: Alphabetical order

andyp1per: Fixed

Hwurzburg:
LGTM....I will merge once one of the other reviewers approve the content

common/source/docs/common-imu-fft.rst
```diff
@@ -0,0 +1,107 @@
...
+
+======================================
+Managing Gyro Noise with In-Flight FFT
+======================================
+
+*Note that this feature will be released in Copter/Plane 4.1*
+
+As :ref:`discussed<common-imu-notch-filtering>`, the effect of motor vibration on gyros in ArduPilot autopilot installations can be managed effectively through the use of targeted notch filtering. Dynamic notch filtering is most effective when the dynamic notch center frequency can be driven by ESC telemetry, but not all aircraft have access to ESC telemetry or the ESC telemetry is not entirely representative of the dominant noise.
+One alternative to ESC telemetry is to measure the noise peak directly through the use of digital signal processing techniques, most notably the Fast Fourier Transform or FFT.
+
```
bnsgeyer: Add the paragraph below for tradheli
For helicopters, use of this feature depends on how well the rotor speed is being held.  Any helicopter using an ESC governor will not need this feature even if the ESC cannot provide arducopter rotor speed.  ESC governors hold rotor speed very well and the harmonic notch frequency can be set to the rotor speed set in the ESC governor.  If the internal rotor speed controller governor is used, which requires an RPM sensor, then it is recommended that the harmonic notch use the measured RPM to set the notch frequency.  This feature would be most useful for throttle curve method for controlling rotor speed.

andyp1per: Thanks done

tridge:
I love having the maths explanation in the description, but it shouldn't come before the explanation of the practical use of the feature. The maths part should be at the end of the page, with the leading part of the page focussed on:
 - why you would use this
 - how to use it

Hwurzburg:
re-org'd as discussed in wiki call, fixed order on advanced config page, expanded note with disclaimer in case we run out of room by 4.1 on 1MB boards again, can be changed when the time comes...

Hwurzburg:
merging...thanks Andy!...always nice to have docs ready when a feature enters master..

PR #2608, opened 2020-03-11
Common: Added mission rewind on resume instructions

Gone4Dirt:
Added feature explanation and instructions for mission rewind and resume feature that was recently added to master.


Hwurzburg:
LGTM....tested......should be on delay merge until it appears in a beta...or get with Tridge and Randy and figure out what future version, then we can add a note and merge...

Gone4Dirt:
Thanks @Hwurzburg.  Just so I know for future, when should we ideally be adding info to the wiki for new features?  I thought it was when it's merged into Master, but from what you are saying it sounds as though the feature needs to be in a beta version.  Can you clarify for me please.  Thanks.

Hwurzburg:
Most new and many established users will only use STABLE versions of ArduPilot, therefore having information in the Wiki that is not in those versions is misleading and can cause confusion, especially to new users. But we encourage testers of the Beta versions and so its hard to test if you dont have info on what is new in the Betas. That is why we have been holding master-only wiki info until at least a Beta with it appears. Also, things not in STABLE versions are likely to change, and their info with it. 
If the feature can be tagged hard to an upcoming release, then a note can be added and it merged if its deemed necessary. But I personally do not like that, since if a change is needed, its easier to overlook and still wind up with an inaccurate WIKI.
I hope that code devs create the WIKI page at the same time the code is merged into the master, since this prevents the converse case of having a beta or even stable release without documentation. Its easy to put a DELAY MERGE label on the page. It allows time also for the feature to be further tested and/or the WIKI entry be enhanced.....use sometimes highlights things that should be mentioned , that aren't yet...

Gone4Dirt:
OK, makes sense.  Thanks for clarifying 😊

Hwurzburg:
Matt, just realized, this needs to be added to the common mission command list and also the copter docs mission list (it was broken out from the common one as a start to a reorg, I believe, and I will continue that process once I get some help on verifying the validity of the common command list page...something I intend to ask for help with at the conference...thanks

common/source/docs/common-mission-rewind.rst
```diff
@@ -0,0 +1,108 @@
+.. _common-mission-rewind:
+
+========================
+Mission Rewind on Resume
+========================
+
+This page explains what the mission rewind-on-resume feature is and how to use the feature.  This feature 
+is available for Plane, Copter, and Rover.
```
rmackay9, line 8: This will be released with the 4.1 releases of Copter, Plane and Rover

Gone4Dirt:
Apologies, took me a little while to get to this @Hwurzburg.  As requested, I have now added the command to the common mission command list as a part of this PR.

Hwurzburg:
Thanks Matt, if you can add the note that Randy requests, I will merge...

Gone4Dirt:
Note added.

Hwurzburg:
thanks! merged

PR #2609, opened 2020-03-12
Warn users about Spline WP with Dijkstras

rishabsingh3003:
Related to: https://github.com/ArduPilot/ardupilot/issues/12691
I tested it out in SITL. The vehicle just breaches fence and does the fence breached fail-safe option. I feel we should warn the user's of this outcome until we solve this issue.


rmackay9:
@rishabsingh3003, yes, indeed.  thanks very much again!

PR #2612, opened 2020-03-16
Add a note to user about Accel calibration

rishabsingh3003:
This comes as a follow-up for the closed issue:
https://github.com/ArduPilot/ardupilot/issues/13663


Hwurzburg:
merged thanks!

PR #2616, opened 2020-03-18
TIP to make a sync in forked repo

brunoolivieri:
Minimal tip to help users to update their wiki forks.  (Built and tested)


Hwurzburg:
super....I use a script called .sync which has all this but not the push to origin (dont want to waste time logging in since I use this everyday and after every commit I merge on upstream)....I would also add a note at the first rebase on line 33 to do this locally before working on a new change, but will do that after I merge this....thanks!

PR #2617, opened 2020-03-18
Expand info on QuadPlane Transition behaviour

Hwurzburg:
adds info on how throttle/motors behave during transitions....and clarify some confusing language

### DIFF
--- a/plane/source/docs/quadplane-flying.rst
+++ b/plane/source/docs/quadplane-flying.rst
@@ -22,18 +22,27 @@ ground station to command a mode change.
 
 -  If you transition to :ref:`MANUAL <manual-mode>` then the VTOL motors
    will immediately stop. In the case of a tilt-rotor, the motors will immediately rotate to foward flight orientation.
+
+.. warning:: If you do not have sufficient airspeed, an immediate stall will occur! Since MANUAL mode is often setup as a reflex driven "bail-out", some users move, or remove this mode, and substitute QSTABLIZE or QLOITER as an alternative "bail-out" for a QuadPlane
+ 
 -  If you transition to any other fixed wing mode then the VTOL motors will
    continue to supply lift and stability until you have reached the
-   :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed (or airspeed estimate if no airspeed sensor). In the case of tilt-rotors, the motors will tilt to :ref:`Q_TILT_MAX<Q_TILT_MAX>` to begin building forward airspeed for the transistion.
+   :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed (or airspeed estimate if no airspeed sensor). In the case of tilt-rotors, the motors will tilt to :ref:`Q_TILT_MAX<Q_TILT_MAX>` to begin building forward airspeed for the transition. The VTOL motors will behave similar to that in QHOVER and will try to maintain present altitude through the transition. The forward thrust is controlled by the throttle stick in a manner similar to whatever fixed wing mode was entered. In FBWA, it is directly controlled, ie low stick is zero thrust and the QuadPlane will just hover. For tilt-rotors, stick positions below mid-stick will proportionately rotate VTOL motors back towards vertical, since that controls the forward thrust component. In FBWB/CRUISE, throttle stick controls forward thrust as in that mode, as a speed or throttle value, depending on whether or not an airspeed sensor is in use. 
+
+.. note:: Unless the :ref:`Q_OPTIONS<Q_OPTIONS>` bit 0 is set, pulling back on elevator will not only pitch the nose up but also increase the VTOL motor output to assist in climbing during the transition airspeed wait phase.
+ 
 -  Once that airspeed is reached the quad motors will slowly drop in
    power over :ref:`Q_TRANSITION_MS <Q_TRANSITION_MS>` milliseconds (default is 5000, so 5
    seconds) and will switch off after that. And tilt-rotors will slowly rotate to full forward thrust configuration.
+
+.. note:: Usually by this time the VTOL motor contribution is already very low, since the QuadPlane is already flying, providing lift or climbing, and the VTOL contribution is aiding attitude stabilization.
+
 -  If :ref:`Q_TRANS_FAIL<Q_TRANS_FAIL>` is not zero, then exceeding this time before reaching  :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed will cancel the transition and the aircraft will immediately change to QLAND. The default is 0, which disables this timeout.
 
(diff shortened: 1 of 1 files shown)


plane/source/docs/quadplane-flying.rst
```diff
@@ -22,18 +22,27 @@ ground station to command a mode change.
...
+.. note:: Usually by this time the VTOL motor contribution is already very low, since the QuadPlane is already flying, providing lift or climbing, and the VTOL contribution is aiding attitude stabilization.
+
 -  If :ref:`Q_TRANS_FAIL<Q_TRANS_FAIL>` is not zero, then exceeding this time before reaching  :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed will cancel the transition and the aircraft will immediately change to QLAND. The default is 0, which disables this timeout.
 
 If you transition from a fixed wing mode to a QuadPlane mode then the
-forward motor will immediately stop, but the control surfaces will
+forward motor/thrust will immediately stop, but the control surfaces will
 continue to provide stability while the plane slows down. This allows
-for transitions to QuadPlane modes while flying at high speed. Tilt-rotors will immediately move to VTOL position.
+for transitions to QuadPlane modes while flying at high speed. Tilt-rotors will, therefore, immediately move to VTOL position.
```
kd0aij: I don't see what "therefore" is referring to?

plane/source/docs/quadplane-flying.rst
```diff
@@ -203,12 +212,10 @@ What Will Happen?
...
 
-I am hovering in QHOVER and switch to FBWA mode
------------------------------------------------
+I am hovering in QHOVER/QLOITER and switch to FBWA mode
+-------------------------------------------------------
 
-The aircraft will continue to hover, waiting for pilot input. If you
-take your hands off the sticks at zero throttle the aircraft will
-continue to hold the current height and hold itself level. It will drift
+The aircraft will continue to hover, setting forward thrust/throttle at whatever the throttle stick position dictates and gaining speed. If you zero throttle during the transition, the aircraft will continue to hold the current height and hold itself level, slowing to a halt. It will drift
```
kd0aij: If it continues to hover with zero throttle stick input, are you still in VTOL mode with the mode switch in FBWA? Does the GCS indicate FBWA? Is there a timeout for this transition? Does it matter if qassist is enabled?

plane/source/docs/quadplane-flying.rst
```diff
@@ -22,18 +22,27 @@ ground station to command a mode change.
...
    continue to supply lift and stability until you have reached the
-   :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed (or airspeed estimate if no airspeed sensor). In the case of tilt-rotors, the motors will tilt to :ref:`Q_TILT_MAX<Q_TILT_MAX>` to begin building forward airspeed for the transistion.
+   :ref:`ARSPD_FBW_MIN <ARSPD_FBW_MIN>` airspeed (or airspeed estimate if no airspeed sensor). In the case of tilt-rotors, the motors will tilt to :ref:`Q_TILT_MAX<Q_TILT_MAX>` to begin building forward airspeed for the transition. The VTOL motors will behave similar to that in QHOVER and will try to maintain present altitude through the transition. The forward thrust is controlled by the throttle stick in a manner similar to whatever fixed wing mode was entered. In FBWA, it is directly controlled, ie low stick is zero thrust and the QuadPlane will just hover. For tilt-rotors, stick positions below mid-stick will proportionately rotate VTOL motors back towards vertical, since that controls the forward thrust component. In FBWB/CRUISE, throttle stick controls forward thrust as in that mode, as a speed or throttle value, depending on whether or not an airspeed sensor is in use.
```
kd0aij: are there any caveats for operation without an airspeed sensor? 
e.g. QLOITER takeoff in strong headwind then switch to FBWA without flying around first; will the airspeed estimate be too low?

plane/source/docs/quadplane-flying.rst
Hwurzburg: preceding sentence, but it can be removed....

plane/source/docs/quadplane-flying.rst
Hwurzburg: no in transition wait
yes
only if set the overall timeout,covered later
yes, qassist is independent and a fixed wing mode phenomenon

plane/source/docs/quadplane-flying.rst
Hwurzburg: biggest issue I have had is when transition AS is reached....ie a loooong time if flying into a headwind...as soon as you turn crosswind it will attain the aspeed...or if you waggle back and forth to allow the turn math to calculate a better synth airsped..

assist is a bitch without an arsd sensor...because synth as can occasionally radically miscalculate and throw you inot assist with an accompanying transition...I intend to add a note about both issues

plane/source/docs/quadplane-flying.rst
kd0aij: So we're flying at high speed in FW mode and switch to a VTOL mode...
It makes sense that tilting motors would immediately begin slewing toward vertical, but doesn't this also apply to tilt-wings?

plane/source/docs/quadplane-flying.rst
kd0aij: Seems confusing that zero throttle means hover in this case, although I guess you would normally be at hover throttle when beginning the transition.
I realize that I'm talking about design instead of documentation though, and also showing my ignorance of how things actually work.

plane/source/docs/quadplane-flying.rst
kd0aij: Thanks, that makes sense.

plane/source/docs/quadplane-flying.rst
Hwurzburg: Yeah.... I welcome Greg's comments...you don't usually see this since throttle is always at hover position when you flip the switch......but I had  snafu with my quadplane that made me want to determine what really goes on..so I sim'd  it. .Rather than always having Tridge be bothered, I just wanted some other plane guy to review

plane/source/docs/quadplane-flying.rst
Hwurzburg: Aren't tilt wings just tilt rotor? I don't know that it has any difference in Ardupilot?Unless you use Copter bicopter I guess

kd0aij: I think so... which says that the wing will start tilting before airspeed has had any time to slow down.

Hwurzburg: just make the slew up slow...right?

kd0aij: That works; I was thinking the slew rate was hardwired, but there are parameters for both tilt_rate_up and tilt_rate_down.

PR #2621, opened 2020-03-18
Add frame types table to SITL doc

Hwurzburg:
baloon? plane-jet?

### DIFF
--- a/dev/source/docs/using-sitl-for-ardupilot-testing.rst
+++ b/dev/source/docs/using-sitl-for-ardupilot-testing.rst
@@ -40,6 +40,51 @@ The frame type can also be changed with the ``-f`` parameter.
     sim_vehicle.py -v ArduPlane -f quadplane --console --map
     
 
+Frame Types:
+------------
+
+A partial listing of frame types is show below. For a current list, just type:
+
+::
+
+    sim_vehicle.py --help
+
+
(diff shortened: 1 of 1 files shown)


khancyr:
hum, there is a lot more that those. Maybe we can generate it from the help ?

Hwurzburg:
yeah....will do..I had used MP's list...will update,but I think I will restrict it to just vehicles, with a note on other frame types for specialized purposes see the help

dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -40,6 +40,51 @@ The frame type can also be changed with the ``-f`` parameter.
...
++                          + vtail quadplane-cl84   +
++                          + quadplane-tilthvec     +
++                          + quadplane-tilttri      +
++                          + quadplane-tilttrivec   +
++                          + quadplane-tri          +
++--------------------------+------------------------+
++                          + quad (default if -f    +
++                          + is not used)           +
++                          +------------------------+
++                          + coaxcopter djix dodeca-+
```
rmackay9, line 72: an extra space appears in "dodeca-hexa"

Hwurzburg: fixed

PR #2631, opened 2020-03-23
update.py: fetch LogMessages from autotest server

peterbarker:


peterbarker:
Merged after @Hwurzburg tested it

PR #2632, opened 2020-03-23

acxz:
direct readers to the correct file path for sim_vehicle.py


khancyr:
hello,

Those changes are wrong ... If you followed the installation correctly Tools/autotest is in your PATH and you can invoke sim_vehicle.py from vehicle directory and thus don't have to pass the vehicle parameter.
Your changes don't work are invoking sim_vehicle.py from autotest directory need the vehicle parameter.

PR #2648, opened 2020-03-29
Dev: Update AirSim page

rajat2004:


rajat2004:
A new release is also expected to happen fairly soon (https://github.com/microsoft/AirSim/issues/2477), will make another PR for that when it happens

PR #2656, opened 2020-04-03
Fix parameters order in listbox

brunoolivieri:
The option of versions was presented randomly in the Listbox on parameter pages of each vehicle. This PR only sorts that.


brunoolivieri:
Tested locally and applying under supervision.

PR #2658, opened 2020-04-04
add description of tailsitter gainscaling in master

### LINKED ISSUE #2654: Description of tailsitter gain scaling options is wrong
This has apparently been out of date since last April:
https://github.com/ArduPilot/ardupilot/commit/95da094bfec00a36f46b9fc3ba0c0996d1d619ad

I thought I had updated this though, with a table of entries for each value of the gainscaling mask...
Will see if I can find the wiki commit

kd0aij:
closes #2654


Hwurzburg:
Mark, thanks...but I don't think we will put documentation in that has not even been merged into master yet....and the part that IS in master, we usually don't include unless we have a specific release target to mention since things can change often between master, beta testing, and release..and not everything in master gets put in a release...but I will bring it up in the Wiki call this Tuesday...
also, the exact impact of GSCMIN needs some expansion....

kd0aij:
something went wrong :(

Hwurzburg:
@kd0aij Mark, did this get resubmitted after rebase , cant find it....think you checked that stable and beta are same as wiki notes, right? and this would be a delay merge for a future release?

kd0aij:
I think I just created a new delayMerge PR instead of trying to fix this one, and
it looks like this got merged to the wiki, but the associated commit never made it into beta or stable
wiki: https://github.com/ArduPilot/ardupilot_wiki/commit/10f89a584a01d991a69ae23efd863281296bc17c
commit in ArduPilot master: https://github.com/ArduPilot/ardupilot/commit/95da094bfec00a36f46b9fc3ba0c0996d1d619adda094b

PR #2661, opened 2020-04-08
Common: Give details of minimal firmware for some autopilots

stephendade:
This comes from the my research into ``HAL_MINIMIZE_FEATURES`` define in ArduPilot.

This will assist users in determining if their autopilot is missing any features that they require.

### DIFF
--- a/common/source/docs/common-autopilots.rst
+++ b/common/source/docs/common-autopilots.rst
@@ -29,9 +29,11 @@ Open hardware
     mRo X2.1-777 <https://store.mrobotics.io/mRo-X2-1-777-p/mro-x2.1-777-mr.htm>
     OpenPilot Revolution <common-openpilot-revo-mini>
     PocketPilot* (Linux) <https://github.com/PocketPilot/PocketPilot>
-    TauLabs Sparky2 <common-taulabs-sparky2>
+    TauLabs Sparky2** <common-taulabs-sparky2>
 
 \* these devices are sensor add-on boards for a Beagle Bone microcomputer. See board links for details 
+
+\** due to flash memory limitations, these boards do not include all ArduPilot features. See :ref:`Firmware Limitations <common-autopilots_limited_firmware>` for details.
 
 Closed hardware
 ---------------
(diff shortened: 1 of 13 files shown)


common/source/docs/common-autopilots.rst
```diff
@@ -64,6 +66,8 @@ Closed hardware
     VR Brain 5 <http://www.virtualrobotix.it/index.php/en/shop/autopilot/vrbrain5-detail>
     VR uBrain 5.1 <http://www.virtualrobotix.it/index.php/en/shop/autopilot/vrbrainmicro51-detail>
 
+\^ due to flash memory limitations, these boards do not include all ArduPilot features. See :ref:`Firmware Limitations <common-autopilots_limited_firmware>` for details.
```
Hwurzburg, line 69: I would prefer using the ** notation for the second note instead of ^....looks a bit awkward to me...personal perference

common/source/docs/common-autopilots.rst
```diff
@@ -88,4 +92,101 @@ available if you're still working on those platforms:
...
 
+.. _common-autopilots_limited_firmware:
+
+Firmware Limitations on Selected Hardware
+-----------------------------------------
+
+The ArduPilot firmware in some configurations exceeds 1 MB in size. Some autopilots may not have enough
+flash memory to store the full firmware.
+
+For the affected autopilots, a minimal firmware is generated. This firmware omits less-commonly used features
```
Hwurzburg, line 103: instead of "minimal" perhaps "reduced", since it really isnt the minimal set that could fly or operate

common/source/docs/common-autopilots.rst
```diff
@@ -88,4 +92,101 @@ available if you're still working on those platforms:
...
+in order to reduce the firmware size to less than 1 MB.
+
+Miminal firmware is required for the KakuteF7, KakuteF7 Mini, OmnibusF7V2, sparky2 and older versions of
```
Hwurzburg, line 106: same comment

Hwurzburg: @tridge why is it that some 1MB boards have this greatly reduced feature set, and others much less reduction?
Also, as a maintenance issue, how do we flag when a new feature is added to the MINIMIZE_FEATURES list or a specific board in order to keep the Wiki up to date?

Hwurzburg: This section should and the following section for other boards should be in the same format style.... with the boards bulleted and feature list indented...perhaps for these first ones, first indent is vehicle, then feature...

common/source/docs/common-autopilots.rst
```diff
@@ -88,4 +92,101 @@ available if you're still working on those platforms:
...
+
+   -  N/A
+
+
+-  **Sub Only**
+
+   -  N/A
+
+
+The following autopilots have specific minimal feature sets with the listed features not present.
```
Hwurzburg, line 162: perhaps "have different reduced feature sets"

@tridge (again raising the question, why others have so many more removed?)

common/source/docs/common-autopilots.rst
stephendade: Fixed to **

common/source/docs/common-autopilots.rst
stephendade: Fixed

common/source/docs/common-autopilots.rst
stephendade: Fixed the wording

common/source/docs/common-autopilots.rst
stephendade: Fixed

Hwurzburg:
Stephen, had a discussion with Tridge....since many F4 boards have recently run out of space, we are going to eliminate EKF2 from all F4 builds not using HAL_MINIMIZE_FEATURES currently. see #14066, so we can remove that EKF2 comment from the MatekF405-Wing board. It will be transparent switching to EKF3 upon upgrade.
Also, we discussed ways of automating this in the wiki, and Tridge may discuss with you regarding a  python script development...

peterbarker:
On Sun, 12 Apr 2020, Henry Wurzburg wrote:

> HAL_MINIMIZE_FEATURES currently. see #14066, so we can remove that EKF2 comment from the MatekF405-Wing board. It will be transparent switching to EKF3
> upon upgrade.

No, it won't be transparent.  While we do switch people from EKF2 to EKF3 
without notification, we won't (and can't) do parameter conversion.

A warning that if people have fiddled with significant EKF2 parameters 
that they should ensure those are reflected in their EKF3 parameters would 
be appropriate IMO.

Peter

Hwurzburg:
Wouldn't that be part of the release notes in 4.1 when it becomes part of a release?
Like the change in quadplane tailsitter types that was recently merged, it should have a cautionary note in the release notes for 4.1 also, I would think...

stephendade:
@Hwurzburg - In that I'll:
-Remove the EKF2 comment from the MatekF405-Wing
-Move the "Firmware Limitations" to a separate page, so make future automation easier

I will leave the automation for a separate PR.

Not sure whether to include the lack of EKF2 for all F4 boards. It would make the list much longer (and include most boards) and I suppose it's more of an upgrade (to EKF3) rather than a missing feature.

stephendade:
I've committed the updates - removed EKF2 for the MatekF405-Wing and put the firmware limitations section into a separate page.
Should be good to go.

Hwurzburg:
thanks....merging

PR #2662, opened 2020-04-08
Saving build time of wiki and its subtasks

brunoolivieri:
The idea is to save the time that tasks from the build script consume. This will help to identify the pre-built caching issues.


brunoolivieri:
Tested locally. I will apply it and watch the run on the server to check possible issues.

PR #2663, opened 2020-04-09
Fixing PR #2662

brunoolivieri:
Fixing the build script


brunoolivieri:
Applying the fix and watching the next build.

PR #2665, opened 2020-04-10
Add Kraus Aerospace as an official partner

magicrub:
Add Kraus Aerospace as an official partner


rmackay9:
Merged, thanks!  Let's see how it looks on the real page..

PR #2667, opened 2020-04-11
Dev: Add Rpanion-server and Pi-Connect as options

stephendade:
Adding in references to the Rpanion-server software and the Pi-Connect hardware.


brunoolivieri:
LGTM, thanks!

Issue #2675, opened 2020-04-15
Rover parameters are not found


brunoolivieri:
The link https://ardupilot.org/rover/docs/parameters.html  is not available.

The rename of "APMrover2" to Rover broke part of the building script. I already working on that.

brunoolivieri:
PR #2676 does some working around to provide latest parameter version.

PR #2676, opened 2020-04-15
Fixing parameters broken for Rover due APMRover2 renaming

brunoolivieri:
Addresses issue #2675. The parameters will appear but only from current to future versions. 

Old versions still need correction.


brunoolivieri:
Merge and first run under supervision.

PR #2678, opened 2020-04-15
Add note about RCOut banner indicating output types to GPIOs page

Hwurzburg:
need to add this to DShot page also...


IamPete1:
Yeah, you get the message along with the firmware version eacct eh time you get params

PR #2688, opened 2020-04-22
Add more uart description

andyp1per:


dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -431,9 +431,19 @@ To use a real serial device you can use a command like this:
 
 what that does it pass the --uartB argument to the ardupilot code,
 telling it to use /dev/ttyUSB0 instead of the normal internal simulated
-GPS for the 2nd UART.
+GPS for the 2nd UART. You can find the serial port mappings :ref:`here <common-pixhawk-serial-names>`
```
Hwurzburg, line 434: Probably should be more generalized for other bds...serial->uart is usually in the boards description in common_autopilots

andyp1per: I don't think this is right? The link I'm looking for is this one https://ardupilot.org/copter/docs/common-pixhawk-serial-names.html and the PX4_HAL column. They are synthetic values. I think the page is misnamed

tridge: this is not relevant for SITL

andyp1per: How about we add the uartA names etc to common-serial-options ?

Hwurzburg:
Andy, from the discussions above, I think we can close this one....agree?

andyp1per:
No, do not agree! How do I map from the non-obvious uart letters to serial numbers. That is what this is trying to address. Need a different table I guess.

peterbarker:
Need a common-sitl-serial-names document?  I can't imagine it would be particularly complicated - there's not *that* many names we give serial ports in SITL.  As opposed to PixHawk/NuttX which had 6 IIRC!

andyp1per:
@peterbarker exactly

Hwurzburg:
if you would like for me to add it, I would need to know where to find the information to do so...the sim_vehicle help dialog does not seem to even mention the use of real serial devices, format for adding, or mapping of sim UARTx to sim SERIALx_ ports...

andyp1per:
Ok I have cherry-picked @peterbarker 's SITL mappings and updated.

Hwurzburg:
Made some changes, added to TOC, fixed some formats....merging..thanks....

PR #2690, opened 2020-04-24
common: all-vehicle-types: add gyrocopter video

peterbarker:


Hwurzburg:
better than nothing....its really too high to tell anything other than it does fly...merging

PR #2697, opened 2020-04-29
Reorg Optical Flow, common setup page, archived Mouse sensor

Hwurzburg:
no content change other than making sure it applies to Quadplane for all flow sensors


rmackay9:
Merged, thanks!

PR #2701, opened 2020-04-30
common: tweak SiK radio information

peterbarker:
Most notably removing the suggestion that ECC is always a good thing.
It most often is not.


common/source/docs/common-3dr-radio-advanced-configuration-and-technical-information.rst
```diff
@@ -633,8 +631,8 @@ If that approach doesn't work do the following:
 
 #. To verify wireless telemetry:
 
-   -  Remove the FTDI to USB cable from APM
-   -  Connect the air radio to APM & LIPO
+   -  Remove the FTDI to USB cable from the autopilot
+   -  Connect the air radio to the autopilot & LIPO
```
TunaLobster, line 635: Would it be better to say connect the radio to a power source as some radios have particular voltages that they operate at?

PR #2705, opened 2020-05-01
dev: add floating-point-annotation and mul-vs-div to style guide

peterbarker:


dev/source/docs/style-guide.rst
```diff
@@ -467,3 +467,36 @@ The user field helps to categorize and hide advanced parameters from being adjus
 
 - Standard - Available to anyone
 - Advanced - Available to advanced users
+
+Floating Point Annotation
+=========================
+
+ArduPilot is compiled with ``-fsingle-precision-constant``.
+
+That means it is currently allowable to leave off the float specifier from constants.  It is also permissable to have them.
```
khancyr, line 476: I would rather say : please use them ... otherwise when we will actually need double we will have issues !

dev/source/docs/style-guide.rst
```diff
@@ -467,3 +467,36 @@ The user field helps to categorize and hide advanced parameters from being adjus
...
+Multiplication vs Division
+==========================
+
+Use multiplication rather than division where possible:
+
+**Right**
+
+::
+
+   const float foo_m = foo_cm * 0.01;
```
khancyr, line 494: 0.01f ;-P

dev/source/docs/style-guide.rst
tridge: I disagree, it makes the code harder to read

PR #2707, opened 2020-05-02
Update bd docs for alternate bd configs

Hwurzburg:


Hwurzburg:
Since I have pretty high confidence in the accuracy, I will go ahead and merge and get it off Tridges plate

PR #2712, opened 2020-05-03
Update copter stabilize mode page

Hwurzburg:


common/source/docs/common-when-problems-arise.rst
```diff
@@ -16,3 +16,38 @@ What to do if you have an issue
...
+-  copter wobbles when descending quickly.  This is caused by the copter
+   falling through it's own props wash and is nearly impossible to  tune
+   out although raising the Rate Roll/Pitch P values may help.
+-  copter yaw right or left 15degrees on take-off.  Some motors may not
+   be straight or the :ref:`ESCs have not been calibrated <esc-calibration>`.
+-  copter always tends to fly in one direction even in a windless
+   environment.  Try :ref:`SaveTrim or AutoTrim <autotrim>` to level the
+   copter.
+-  copter does not maintain altitude or does not stay perfectly still in
+   the air.  As mentioned above this is a manual flight mode and
```
rmackay9, line 42: Ah, this bit is incorrect now because it's not on the stabilize page anymore so we should remove it, "As mentioned above this is a manual flight mode and requires constant control of the sticks to maintain altitude and position".

The most common reason that the vehicle cannot maintain altitude is because of high vibrations.  So we could add an item, "copter climbs rapidly even if the pilot pulls the throttle down.  This is likely caused by high vibrations. See https://ardupilot.org/copter/docs/common-vibration-damping.html for methods to improve vibration isolation".

rmackay9:
One small issue and then this can be merged, thanks!

common/source/docs/common-when-problems-arise.rst
Hwurzburg: done...thanks...merging

PR #2714, opened 2020-05-04
Add warning note on BN-220 re time required to pass pre-arm checks

Hwurzburg:
move to EKF3 highlited this issue on several of my planes using mini-gps...


rmackay9:
Looks OK, thanks!

PR #2715, opened 2020-05-04
all-vehicle-types: add thrust-vectored-coaxcopter

peterbarker:


peterbarker:
Should wait until we have at least a link to a branch before merging this.

PR #2716, opened 2020-05-05
Replace APMrover2 with Rover

peterbarker:


Hwurzburg:
fixed table format error....merging...
@rmackay9  I assume that you are going to continue to use the APMrover2-Stable tag on releases...if not, I will need to change reference in Lua scripting when you do change to Rover-Stable or whatever....let me know if you change....thanks

PR #2736, opened 2020-05-10
Update common-zubax-gnss-positioning-module-gps-compass-and-barometer…

auturgy:
….rst


auturgy:
Image was of the original Zubax v1, but page referred to v2.  Minor but this adds a v2 image and switches the reference.

PR #2756, opened 2020-05-17
Update RFD900 Warning

NickNunno:
@tridge per our conversation

### DIFF
--- a/common/source/docs/common-rfd900.rst
+++ b/common/source/docs/common-rfd900.rst
@@ -9,9 +9,8 @@ designed for long range serial communication.  Ranges of `15km and much higher h
 
 .. warning::
 
-   These modems are high-power RF devices. It is suggested to
-   provide separate +5 voltage to the modems. It is NOT suggested to power
-   these modems directly from Autopilots telemetry port.
+   Many common flight controllers have telemetry port current limits that prevent directly powering RFD900 radios at their full transmit power. You should power high power radios externally or limit their maximum transmit power to an acceptable level.All commonly used flight controllers should be able to handle transmit powers of 20dBm.  See the manual of your specific flight controller for the detailed power specifications of each telemetry port. To set the transmit (TX) power see :ref:`Configuring a Telemetry Radio using Mission Planner <common-configuring-a-telemetry-radio-using-mission-planner>`.
+
 
 .. image:: ../../../images/RFD900_telemetry_radio.jpg
     :target: ../_images/RFD900_telemetry_radio.jpg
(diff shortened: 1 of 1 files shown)


proficnc:
No, this is not acceptable.

Under NO circumstances should you power an RFD900 via the telemetry ports. 

This is already covered in the Ardupilot wiki.

And already covered in the CubePilot docs, and it applies to all cubes and FMUv2/3 boards.

NickNunno:
@proficnc I've changed it to be more generic as each manufacturer will have different hardware capabilities and recommendations.

NickNunno:
The purchase links on the wiki page provide RFD900 radios with cables to directly power them from telem. 

The RFD900 manual provides graphs showing voltage/current draw over the full transmit range of the radio. IE- @25dBi, it draws aproximately .7A. 

With the information given in the RFD900 manual and proper hardware ratings/recommendation from the autopilot manufacturer, they can easily calculate whats acceptable. 

http://files.rfdesign.com.au/Files/documents/RFD900%20DataSheet.pdf

Hwurzburg:
First, I would add a link the the current vs dbm graph....and since many autopilots do not specify max current for Telem ports or UART ports; and since Sik radios have a max current spec of 100ma and we currently do not warn of attaching two of those,  you can assume 200ma is commonly supported on TELEM ports....so I would warn RFD900 users that if they cannot determine the available power, use 10 dbm to get ~200ma as a safe course

NickNunno:
Happy to add the graph as its specific to the RFD900 we're referencing. 

The point of this PR was to abstract the autopilot/carrier component since we do not know what they are using with the RFD900. 

With that in mind, I do not think we should make explicit recommendations/assumptions beyond the realm of the RFD900.

Hwurzburg:
This page only refers to the RFD900...it also needs rebasing to be able to be merged

common/source/docs/common-rfd900.rst
```diff
@@ -9,9 +9,8 @@ designed for long range serial communication.  Ranges of `15km and much higher h
 
 .. warning::
 
-   These modems are high-power RF devices. It is suggested to
-   provide separate +5 voltage to the modems. It is NOT suggested to power
-   these modems directly from Autopilots telemetry port.
+   Many common flight controllers have telemetry port current limits that prevent directly powering RFD900 radios at their full transmit power. You should power high power radios externally or limit their maximum transmit power to an acceptable level. See the manual of your specific flight controller for the detailed power specifications.
```
tridge, line 12: All commonly used flight controllers should be able to handle transmit powers of 20dBm, and most will cope with 25dBm.

rmackay9: I think it's best to be more specific and specify the 20dBm or 25dBm (or whatever).  Although most users don't know what a "dBm" is the setup screen has a drop-down for this setting.  Extra points if we include a picture with some red boxes showing how to get to the setting.  I tried to capture a screen shot but my MP is acting up.

Hwurzburg: no need for screen  shot ...easy reference to how to do that already exists....will include in warning...

Hwurzburg:
pushed the changes as discussed in wiki call and merging

PR #2758, opened 2020-05-17
Update parameters in rangefinder example

sslaughter:
New parameter set has numbers in RNGFND parameters. Simply updating the parameter names to the current set.


Hwurzburg:
thanks! merging

PR #2760, opened 2020-05-17
Add info about GCS messages while autotuning

Hwurzburg:


copter/source/docs/autotune.rst
```diff
@@ -126,6 +159,38 @@ ATUN (auto tune overview) and ATDE (auto tune details) messages are
...
+Ground Control Station Messages
+===============================
+
+For each axis there are several phases to the tune. Rate PIDs are adjusted first, then ANGLE parameters. Progress messages during these phases are sent to the GCS (and recorded in the Dataflash logs).
+
+Typical sequence during tuning might be:
+
+::
+
+ 09:09:33	AutoTune: Twitch
```
rmackay9, line 171: we can probably remove the time part of this message I guess.. I think that's probably the ground station printing that..

copter/source/docs/autotune.rst
```diff
@@ -107,6 +139,7 @@ Additional Notes
 -   AutoTune is sometimes unable to find a good tune for frames with very soft vibration dampening of the autopilot or very flexible arms.
```
rmackay9, line 139: we can probably remove this line, "AutoTune is sometimes unable to find a good tune for frames with very soft vibration dampening of the autopilot or very flexible arms." now that we've got the failure reason section at the top.

copter/source/docs/autotune.rst
```diff
@@ -107,6 +139,7 @@ Additional Notes
...
 -   For best results the copter shouldn't be allowed to build up too much horizontal speed. This can be prevented by applying a quick correction between tests (twitches) to stop the vehicle from flying too fast.
 -   Be advised that AutoTune will engage from Stabilize, so don't accidentally flip your AutoTune switch until you are in AltHold and ready to begin the procedure.
+-   As a general rule, for Pitch and Roll, P and I should be equal, and D should be 1/10th P. For Yaw, I should be 1/10th P and D = 0, in most cases.
```
rmackay9, line 142: I'm not sure we should add this new line, "As a general rule, for Pitch and Roll, P and I should be equal, and D should be 1/10th P" because although it's a good rule for manual tuning, I don't think AutoTune will necessarily come up with this ratio of P/I to D.

rmackay9:
Thanks so much for this!

copter/source/docs/autotune.rst
Hwurzburg: copied from DataFlash log directly

copter/source/docs/autotune.rst
Hwurzburg: will do

copter/source/docs/autotune.rst
rmackay9: Ah, ok.. i just thought it was a bit distracting to have the time there but I'm not too fussed.

copter/source/docs/autotune.rst
Hwurzburg: it does...I think...
#define AUTOTUNE_PI_RATIO_FINAL            1.0f     // I is set 1x P after testing
#define AUTOTUNE_YAW_PI_RATIO_FINAL        0.1f     // I is set 1x P after testing
 and during the test
attitude_control->get_rate_yaw_pid().kD(0.0f);

rmackay9: Ah, right.  Just the "D should be 1/10th P" is not correct I'm pretty sure.

Hwurzburg: ah, right...that's for manual tuning..will remove...autotune actually has a D phase for R/P...

rmackay9:
Merged, thanks!

PR #2762, opened 2020-05-18
Add RTL_CLIMB_MIN to RTL mode page

Hwurzburg:


Hwurzburg:
if this is going into a 4.0.x release, I should change the limiter in it....

PR #2775, opened 2020-05-23
Update building-setup-linux.rst

auturgy:
NuttX and make are no longer used by ArduPilot


khancyr:
Waf distclean instead of rm -rf no?

peterbarker:
@khancyr nuke the site from orbit; it's the only way to be sure.

khancyr:
?
Waf distclean does the same things than rm -rf but also clean waf stuff.
Why do you consider it as nuke ? ( git clean -d -x -f)

Hwurzburg:
approvals in ...merging

PR #2780, opened 2020-05-24
Add info on plane level calibration

Hwurzburg:
appears only in plane site


common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -55,6 +53,14 @@ Calibration steps
          -  The level position is the most important to get right as this will
             be the attitude that your controller considers level while flying.
 
+[site wiki="plane"]
+    For planes, often the optimum "level" position is actually done with the nose raised a few degrees (but the wings perfectly level). This is explained in the diagrams below. Most planes fly with a few degrees of Angle of Attack (AOA)and while many planes will have some Angle of Incidence built in, some do not, and some need a little higher AOA to fly at lower cruise speeds. If the level step of calibration is done with the plane's fuselage line level, then first flights will be safe, but cruise may be a bit fast. You can either set a few extra degrees nose up when doing the level step of calibration, or  decide after first flights if AOA needs to be increased for your desired cruise speed/throttle. If so, then you may redo the level only step using the  ``Calibrate Level`` button on the page.
+
```
tridge, line 58: the mission planner screen captures don't look right. They are from copter, and I think they are actually from the 1 axis accel cal, which we definitely don't want to use.
The one we want is the accel trim, which changes AHRS_TRIM_X and AHRX_TRIM_Y only. The 1-axis accel cal changes the INS_ACCOFF* parameters, and zeros the AHRS_TRIM_ values. That is not something we should recommend using

Hwurzburg: going to merge, then fix the historical bad screens right after

PR #2788, opened 2020-05-28
Input Interface of the Navio2 is PPM not PWM

q-wertz:
e.g. see https://docs.emlid.com/navio2/ardupilot/hardware-setup/


rmackay9:
Surely this is correct, thanks!

PR #2790, opened 2020-05-28
Updates to Copter/Rover fences pages

Hwurzburg:
Add rover to the pages....fix outdated params....add note that geo fence setup tab not included in MP for Rover....
this is just phase 1
phase 2 is to update them to show inclusion (which can be used for a gui tin can I believe) and exclusion zones and polygons as well as updating MP screen shots....


common/source/docs/common-ac2_simple_geofence.rst
```diff
@@ -8,29 +8,42 @@ Cylindrical GeoFence
 Overview
 ========
 
-AC 3.0.1 (and higher) includes a simple "tin can" shaped fence centered
+Copter and Rover include a simple "tin can" shaped fence centered
 on home that will attempt to stop your copter from flying too far away
-by stopping at the fence (if in Loiter mode and using Copter-3.4 or higher) or initiating an RTL.
-The maximum circular distance and altitude and the vehicle behavior when the fence is reached can be configured using
+by stopping at the fence (BRAKE or HOLD), LANDing, or initiating an RTL.
```
rmackay9, line 13: By the way, Copter will stop at the horizontal fences (cylindrical, circular and polygon inclusion/exclusion fences) in Loiter and Guided mode.  It will also stop at the FENCE_ALT_MAX in Loiter, PosHold and AltHold.  It doesn't stop at the fence by switching to Brake mode though... it simply stops without changing modes and the pilot can then backup the vehicle to move away from the fence and then keep flying.

Rover will stop at the horizontal fences in Acro and Steering modes but no others I think.

Hwurzburg: hopefully this is now more complete and correct....

Hwurzburg: @rmackay9 I am merging this one so that other changes wont conflict...we can correct any minor errors when you review #2830 which will be over the top of this one

PR #2791, opened 2020-05-29
Cube Autopilot Updates and DJI FPV

MadRC:
I have created proper pages for the Cube Orange, Yellow and Purple as currently there is just link that are going to the Profi site that was not giving much info at all.  

Updated Cube black page with the website and support links as well. 

Also updated the DJI FPV on the Digital page as there is more info now and moved the new FPV system up as that's more relevant. 

I have checked things but I can be 100% there will be mistakes, I think I have played things out but I have not compressed the commits and I hate GitHub and I can't work out how too. 

The links would need to be updated to the new pages but I have not looked at that in this. 

Please let me know your thoughts.


Kigurai:
Hi,
DJI FPV isn't designed for drone racing as the latency is too high.  It is
however awesome for 'general' fpv'ing.

tridge:
this should not have been merged in this form. It didn't build properly and should have had proper commit names, and not used 40 commits to change 12 files

Issue #2776, opened 2020-05-23
dev/source/docs/building-setup-linux.rst


auturgy:
We should re-order this, so that users run the install script as the first step.
Also need to add meat to the Docker instructions - they're very light.

Hwurzburg:
#2799 but I cant help with Docker

Hwurzburg:
closing until someone complains of error in Docker instructions, since I cant expand them and no complaints so far...help there is alwasys welcome

PR #2799, opened 2020-05-31
reodering and re-org of linux build setup page

Hwurzburg:
addresses #2776


dev/source/docs/building-setup-linux.rst
```diff
@@ -10,6 +10,28 @@ This article shows how to setup your build environment on Linux/Ubuntu machines.
 Setup on Ubuntu
 ===============
 
+Install some required packages
+------------------------------
+
+If you are on a debian based system (such as Ubuntu or Mint), we provide `a script <https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-ubuntu.sh>`__ that will do it for you. From ardupilot directory :
```
khancyr, line 16: I am not sure of this change as we need git to get the install script ...

Hwurzburg: Absolutely! will fix...but still needs a little re-org...will redo and ask for a re-review...thanks!

Hwurzburg: Think I have the flow correct now....Unbuntu with git/clone first, other Linux, MacOS,Docker

Hwurzburg:
@khancyr good now?

dev/source/docs/building-setup-linux.rst
```diff
@@ -57,6 +58,37 @@ Now you should be able to build with waf as described in `BUILD.md <https://gith
 .. youtube:: 4B8BVskH0vc
 
 
+ccache for faster builds
```
khancyr, line 61: this is already part of the install script, so we don't need to do it manually

dev/source/docs/building-setup-linux.rst
```diff
@@ -79,14 +111,12 @@ Then reload your PATH by using the "dot" command in a terminal
 ::
 
     . ~/.bashrc
-
-Setup for other Distributions
-=============================
-
+
```
khancyr, line 114: maybe a title here like :
STM toolchain

khancyr: the right toolchain tar is gcc-arm-none-eabi-6-2017-q2-update-linux.tar.bz2 not the 4.9 one (L152) and L157 the name is also wrong.
We should put Compiler section (L144) before Permissions ones (L133)
And put back the CCache install at end

dev/source/docs/building-setup-linux.rst
Hwurzburg: moved back under other dists

dev/source/docs/building-setup-linux.rst
Hwurzburg: please check that I guessed correctly at the right path to the compiler dir...

Hwurzburg: @khancyr Pierre?

PR #2806, opened 2020-06-02
Build Wiki with sphinx-build's -W

peterbarker:
This turns all warnings into errors.

This will allow us to keep the Wiki clean as travis will enforce everything.

This does include some cleanups to get things to build cleanly, but I was astonished at how few there were.

### DIFF
--- a/antennatracker/Makefile
+++ b/antennatracker/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+# SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
--- a/ardupilot/Makefile
+++ b/ardupilot/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+# SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
--- a/copter/Makefile
+++ b/copter/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+# SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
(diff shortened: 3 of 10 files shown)


peterbarker:
This won't pass CI until we've got a fresh set of parameters generated on the autotest server.

Hwurzburg:
if you stop the copying, it will break the concrete links to those pages....tried that when I cleaned up the 700+ build errors (with Bruno's help)  in the beginning of my involvement with the wiki...the concrete would have to be changed I was told....and not worth it...

peterbarker:
@Hwurzburg I've removed the patch which removed the copying.  But I've added another one to retain the rename of that function.

Hwurzburg:
Sorry, having trouble following the patches....I thought the copying into Ardupilot site for partners,donations, and rtf were being retained to prevent the concrete wrapper links from breaking...it seems that they are still not copied to Ardupilot site? 

the erroneous copies into mavproxy just highlite the need for error free builds...I check a chron every morning of the error outputs  of a master wiki build...it just reports the std error path to a file...I had gotten so used to seeing the handful of ardupilot site errors that I did not "see" the mavproxy errors ...apparently when we were merging the mavproxy site, there WERE errors due to their non-inclusion and I had added mavproxy to those two documents to get clean site build...but it was because of the sphinx sidebar links...removing copying them breaks them in the mavproxy site....I will work on that...

peterbarker:
@Hwurzburg so did you want me to rebase this on top of fixes you're going to do?

If you grab this PR and kill my "fixes" to those rst files you could chase things down pretty quickly...

Hwurzburg:
actually, the mavproxy errors arent errors, they also need to be there....the mavproxy errors only appeared yesterday after the other merges...I believe....I dont think that I had missed them before...they are new...so both mavproxy and ardupilot copies are correct...think we need to understand why the build is flagging them first

peterbarker:
>  Yipee! finally figured out how to eliminate the last wiki build errors....it build entirely clean now for the first time since I began working on it a year ago...  @Peter Barker can now merge your test fails for the wiki on warnings!

Removing DelayMerge on that basis.

I've removed all modifications to the .rst files in this PR now - let's see if it passes :-)

peterbarker:
We'll see how this goes; it might prove too painful, but it will save Henry time if we stop silly errors creeping in.

PR #2813, opened 2020-06-03
Add copter flight features page

Hwurzburg:
makes it easier to find advance features for copters, like plane wiki

### DIFF
--- a/copter/source/docs/copter-flight-features.rst
+++ b/copter/source/docs/copter-flight-features.rst
@@ -1,0 +1,23 @@
+.. _copter-flight-features:
+
+===============
+Flight Features
+===============
+
+Copter has a lot of advanced features which can be accessed either via
+configuration parameters or via mission commands. The list below will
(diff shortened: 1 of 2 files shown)


Hwurzburg:
Advanced config page is hard to find major features in my opinion...I like the plane's short quick page to major things most people would be interested in doing with plane...I just picked what seemed to be major operational modes vs thins like DShot setup..

rmackay9:
It's up to you.. I just don't know how we decide what goes on that page.. lots of people ask about lots of different things.  for example, I probably wouldn't put "boat mode" on the page.. "terrain following" definitely yes.. lots of people ask about object avoidance.. not many ask about motor thrust scaling.. 

It's just a tough choice and so I'll not so sure what to put there.. anyway.. i leave it in your capable hands.

Hwurzburg:
okay, going to merge as is...we can always tweak it later

PR #2814, opened 2020-06-03
Split geofencing fs out from under object avoidance

Hwurzburg:
treating OA, Geofences, and Terrain Following/Surface tracking as the independent features they are....this will allow easy expansion of Copter/Rover Geofences for new inclusion/exclusion zone docs and allow easy inclusion of Plane in the future if it ever gets the new fences...


common/source/docs/common-object-avoidance-landing-page.rst
```diff
@@ -4,8 +4,9 @@
 Object Avoidance
 ================
 
-Ardupilot supports many kinds of object avoidance in vehicles. Airborne Vehicles (ADSB), Object Avoidance (Object/Ground/Ceiling), and Geo-Fences (Beacons, Exclusion zones, Fences).
-Supported types vary with vehicle. Some kinds of avoidance require external hardware, such as ADSB receivers or  Rangefinders.
+Ardupilot supports several kinds of object avoidance in vehicles. Avoidance of Airborne Vehicles (ADSB) and Object Avoidance (Object/Ground/Ceiling).
```
rmackay9, line 7: a small thing but we can probably remove "in vehicles".  We don't actually support avoiding the ground although I see this is an existing issue with the docs.

common/source/docs/common-object-avoidance-landing-page.rst
```diff
@@ -24,20 +25,18 @@ Avoidance Features
     Simple Object Avoidance <common-simple-object-avoidance>
     Autonomous Mission Re-routing via Bendy Ruler <common-oa-bendyruler>
```
rmackay9, line 26: An existing issue so not a blocker for this PR but at some point I'd like to change these to "BendyRuler path planning around obstacles" and "Dijkstra's path planning around obstacles"

common/source/docs/common-object-avoidance-landing-page.rst
```diff
@@ -24,20 +25,18 @@ Avoidance Features
...
+
+
 Hardware
 ========
 
 -  :ref:`ADS-B Receiver <common-ads-b-receiver>`
-[site wiki="copter,rover"]
+
 -  :ref:`Rangefinders <common-rangefinder-landingpage>`
 -  :ref:`Fence Beacons (Non GPS navigation) <common-non-gps-navigation-landing-page>`
```
rmackay9, line 39: another pre-existing issue so not a blocker but I don't think we should call our Beacons, "Fence Beacons".  Instead we should call the Pozyx Beacons like we do on this page: https://ardupilot.org/copter/docs/common-pozyx.html.  The link should also go to this page I think.

common/source/docs/common-geofencing-landing-page.rst
```diff
@@ -0,0 +1,18 @@
+.. _common-geofencing-landing-page:
+
+==========
+GeoFencing
```
rmackay9, line 4: It's up to you, but personally I like the simpler word "Fence" rather than "GeoFence".  I think the "Geo" part just makes it slightly less understandable to people.

Hwurzburg: I dont care myself...but to be consistent, we need Tridge to buy in for Plane because I will change to it everywhere...

common/source/docs/common-geofencing-landing-page.rst
```diff
@@ -0,0 +1,18 @@
...
+==========
+
+ArduPilot supports several types of GeoFencing failsafes to prevent the vehicle from traveling higher or further than desired or into unwanted areas. Types of GeoFences supported varies by vehicle.
```
rmackay9, line 7: Is it really several fence failsafes or is it really just a single failsafe that works with the various types of fences AP supports?  Anyway, I'm being a bit pedantic and I don't feel too strongly about it.

common/source/docs/common-object-avoidance-landing-page.rst
Hwurzburg: will do

common/source/docs/common-object-avoidance-landing-page.rst
Hwurzburg: might as well do it now...

common/source/docs/common-object-avoidance-landing-page.rst
Hwurzburg: will do

common/source/docs/common-geofencing-landing-page.rst
Hwurzburg: I can change...no problem...."supports failsafes generated by several types of Fences"

common/source/docs/common-geofencing-landing-page.rst
Hwurzburg: Thinking about it, shouldn't we differentiate a geo-fence from a beacon fence?

Hwurzburg: Since you leave it up to me.....GeoFence

common/source/docs/common-object-avoidance-landing-page.rst
Hwurzburg: that should be moved...its not object avoidance per se, its non gps navigation

Hwurzburg: should be moved to non-gps nav stuff

Hwurzburg:
changes made...merging

PR #2815, opened 2020-06-03
Update mission-command-list.rst : Note about LAND

EosBandi:
Land changes wpnav_speed to 2m/s. If you continue mission after land (with a takeoff) you have to change it back with a CHANGE_SPEED

### DIFF
--- a/copter/source/docs/mission-command-list.rst
+++ b/copter/source/docs/mission-command-list.rst
@@ -163,6 +163,10 @@ provided.  This is the mission equivalent of the :ref:`LAND flight mode <land-mo
 **Lat, Lon** - the latitude and longitude targets. If left as zero it
 will land at the current location.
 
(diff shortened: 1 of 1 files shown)


rmackay9:
I don't think this is correct, I just tested it in SITL and after the takeoff the vehicle's speed climbed to WPNAV_SPEED.  Perhaps there were do-change-speed commands earlier in the mission?

EosBandi:
I dig a little deeper, the speed limit was reproable in  in 4.0.3. However this change https://github.com/ArduPilot/ardupilot/commit/7b1e34d219eaa559ef85c015a8f453cfa4f8a07d had a side effect which fixed the behavior described above, So in 4.0.4rc it is working just fine.

Hwurzburg:
okay so closing

PR #2816, opened 2020-06-04
dev: add a gcs-resources document

peterbarker:


peterbarker:
@Hwurzburg ideas on where we should slot this into the wiki?

Hwurzburg:
I'd put it in the Dev index.rst TOC so that show as Info for GCS Developers in the left side bar for the wiki....

Hwurzburg:
added to TOC and merged...thanks!

PR #2817, opened 2020-06-04
Add HobbyPorter to Partners Page

auturgy:
HobbyPorter is the parent of two new Partners, Skydroid and RadioMaster.
Tested, uneventfully, with a local Vagrant build (Rover only for convenience/speed).


Hwurzburg:
thanks...merging

PR #2821, opened 2020-06-05
Increasing setrecursionlimit

brunoolivieri:
Setting limit to 15000 (tested locally)


brunoolivieri:
Tested locally and applying on the server under supervision.

PR #2826, opened 2020-06-05
Added Sabertooth 2x32A Motor Controller

MadRC:
I have used this extensively with Rover and its fine.

I have not tested this change on a build as its small.


rmackay9:
awesome, thanks!

PR #2828, opened 2020-06-06
Update advanced compass setup page

Hwurzburg:


common/source/docs/common-compass-setup-advanced.rst
```diff
@@ -19,183 +19,68 @@ will not move in the correct direction in autopilot modes (i.e. AUTO,
 LOITER, PosHold, RTL, etc). This can lead to circling (aka
 "toiletbowling") or fly-aways.
 
-ArduPilot currently allows up to three compasses to be connected. Only
-one compass (specified using the ``COMPASS_PRIMARY`` parameter) is used
-for navigation. While many autopilots have an internal compass or compasses, many
-will instead use an external compass. This provides more reliable data
+ArduPilot multiple compasses to be connected, but only three are used at one time. And usually, only the primary compass is used, unless there are consistency problems with its readings versus other compasses and/or other sensors. In this case, the autopilot will automatically determine which of the first three compasses to use.
```
rmackay9, line 22: I think we are missing "supports" or "allows" here.

rmackay9: Is this true, I'm actually not sure if the EKF ever fails over to a 2nd compass.. perhaps it does..

common/source/docs/common-compass-setup-advanced.rst
```diff
@@ -19,183 +19,68 @@ will not move in the correct direction in autopilot modes (i.e. AUTO,
...
 than an internal compass because of the separation from other
 electronics. See :ref:`common-autopilots` for details about the specific autopilot to determine how many built-in compasses the autopilot may have, if any.
 
-Most users will only need to select compass configuration and perform the :ref:`Basic Compass Calibration<common-compass-calibration-in-mission-planner>` but details are also given on the less-used  :ref:`CompassMot <common-compass-setup-advanced_compassmot_compensation_for_interference_from_the_power_wires_escs_and_motors>` and Manual Declination. Most of this configuration can be performed from the *Mission Planner*'s **Initial Setup \| Mandatory Hardware \| Compass** screen.  
-Other ground stations may have similar features.
+Most users will only need to  perform the :ref:`Basic Compass Calibration<common-compass-calibration-in-mission-planner>` but details are also given on the less-used  :ref:`CompassMot <common-compass-setup-advanced_compassmot_compensation_for_interference_from_the_power_wires_escs_and_motors>`
```
rmackay9, line 28: I think we should replace "Basic Compass Calibration" with "Onboard Compass Calibration" although it's an existing issue not made worse by this PR.

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: thats what I understood from Tridge....

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: can do

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: sometimes I have to backspace so often because I type dyslexically, my thoughts outpace the words! will add "allows"

rmackay9: Ok, if Tridge says it is so then great

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: will do this globally in another commit

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: added...thanks

common/source/docs/common-compass-setup-advanced.rst
Hwurzburg: I have been looking into this...Compass Calibration is used everywhere....this is the only Basic Compass Calibration (which I added in this PR)...and I think Compass Calibration is better than changing it everywhere to something like Onboard Compass Calibration, which would lead to "what other kinds of compass calibration do I need to look at?" thought processes....I should have used "will only need to perform the normal (then ref tag) Compass Calibration, but details...."....I will change it....

PR #2829, opened 2020-06-06
Update compass wiring page

Hwurzburg:


common/source/docs/common-installing-3dr-ublox-gps-compass-module.rst
```diff
@@ -80,8 +55,7 @@ Connecting to Erle-Brain2
 An external compass is required (drivers are not yet available for the
 internal compass).
 
-If using Copter 3.3 or higher or recent versions of Plane/Rover then
-this parameter should also be set:
+This parameter should also be set:
```
rmackay9, line 58: I'm happy with this change but as a follow-up we can remove all mention of Erle Brain from the wiki now (or archive it) because they're not available anymore.

PR #2830, opened 2020-06-07
Change poly geofence to inclusion and exclusion zones

Hwurzburg:
probably should modify some for Rover....tomorrow


common/source/docs/common-geofencing-landing-page.rst
```diff
@@ -6,12 +6,18 @@ GeoFencing
...
+Inclusion and Exclusion Zones are defined easily in Mission Planner under the PLAN screen and loaded to the autopilot, using the  FENCES item in the drop down box, just like Missions or Rally Points are planned.
+[/site]
+
 .. toctree::
     :maxdepth: 1
 
 [site wiki="copter,rover"]
     Cylindrical Geo-Fencing Failsafe <common-ac2_simple_geofence>
-    Polygon Geo-Fencing Failsafe <common-polygon_fence>
+    Inclusion/Exclusion Zone Failsafe <common-polygon_fence>
```
rmackay9, line 20: I think we should stick to fewer words if possible.  I prefer "Fence" to "Zone".  I also prefer "Fence" to "Geo-Fence". 

I sometimes use the word "Stay-in" and "Stay-out" although I'm not sure we use these on the wiki.  I don't think we should anymore because "inclusion/exclusion" is what MP (and perhaps other GCSs) use to to describe these fences.

So.. I think maybe we've got the "Cylindrical fence" and "Inclusion/Exclusion Circular fences" and "Inclusion/Exclusion polygon fences".

Hwurzburg: Changed for copter/rover.....left plane as GeoFence......
MP is not aligned with this....still uses GeoFence on the Config/GeoFence screen...

rmackay9:
Merged, thanks!

PR #2703, opened 2020-05-01
common: add SERIAL_CONTROL column to PixRacer serial table

peterbarker:
Also remove NuttX/PX4 information

PR #2833, opened 2020-06-08
common: add SERIAL_CONTROL column to PixRacer serial table

Hwurzburg:
Also remove NuttX/PX4 information

fixes to #2703 with permission from PeterB


Hwurzburg:
okay now?

peterbarker:
You removed the observation that bootloader-serial is available?

I guess it's probably fair if all our vehicles have it available on SERIAL1 - but....

PR #2834, opened 2020-06-10
Fix: Rover parameters versioning

brunoolivieri:
It does some workaround to re-enable versioning for Rover parameters files. Tested locally.

It also inserts the ":orphan:" tag in the files in order to stop warnings for parameters files for old versions.

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -55,7 +55,7 @@ BASEPATH = ""
 
 ## Dicts for name replacing
 vehicle_new_to_old_name = { # Used because "param_parse.py" args expect old names
-    "Rover": "Rover",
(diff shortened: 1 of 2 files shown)


Hwurzburg:
so why the delaymerge?

brunoolivieri:
I tested it locally. However, I prefer to apply on the server in a moment that I can watch this and check if the result is the same. It is just an extra caution that I use to do when I am updating anything on the scripts.  I will apply asap.

brunoolivieri:
Applying under supervision.

brunoolivieri:
Built well done in the server and pages are updating with mutiple parameters in all vehicles again.

PR #2835, opened 2020-06-10
Update common-partners.rst

auturgy:
when I updated the Air Supply logo a little while back it seems I had a typo in the image filename.  Rather than fix that, I'm just pointing to the name of the new logo


rmackay9:
Merged, thanks!  let's see how it looks on the real page.

PR #2850, opened 2020-06-16
Backup engine for wiki building.

brunoolivieri:
This PR enables the "--enablebackups" flag in the build script. When used, this keeps up to N_BACKUPS_RETAIM previous wiki builds as backups.

The default used folder is supposed to be another http site.  It will enable navigation between those versions to check for errors or make it easier to get a built copy online due to some issues.

### DIFF
--- a/update.py
+++ b/update.py
@@ -38,6 +38,9 @@ import filecmp
 import time
 import sys
 import hashlib
+from datetime import datetime
+import distutils
+from distutils import dir_util
 
 DEFAULT_COPY_WIKIS =['copter', 'plane', 'rover']
 ALL_WIKIS =['copter', 'plane', 'rover','antennatracker','dev','planner','planner2','ardupilot', 'mavproxy']
@@ -53,6 +56,8 @@ parser.add_argument('--clean', action='store_true', help="Does a very clean buil
 parser.add_argument('--cached-parameter-files', action='store_true', help="Do not re-download parameter files")
 parser.add_argument('--parallel', type=int, help="limit parallel builds, -1 for unlimited", default=1)
 parser.add_argument('--destdir', default="/var/sites/wiki/web", help="Destination directory for compiled docs")
+parser.add_argument('--enablebackups', action='store_true', default=False, help="Enable several backups up to const N_BACKUPS_RETAIN in --backupdestdir folder")
+parser.add_argument('--backupdestdir', default="/var/sites/wiki-backup/web", help="Destination directory for compiled docs")
 parser.add_argument('--paramversioning', action='store_true', default=False, help="Build multiple parameters pages for each vehicle based on its firmware repo.")
 parser.add_argument('--verbose', dest='verbose', action='store_false', default=True, help="show debugging output")
 args = parser.parse_args()
@@ -62,6 +67,8 @@ args = parser.parse_args()
 PARAMETER_SITE={'rover':'APMrover2', 'copter':'ArduCopter','plane':'ArduPlane','antennatracker':'AntennaTracker' }
 LOGMESSAGE_SITE={'rover':'Rover', 'copter':'Copter','plane':'Plane','antennatracker':'Tracker' }
 error_count = 0
+N_BACKUPS_RETAIN = 10
+
 
 def debug(str_to_print):
     """Debug output if verbose is set."""
@@ -136,6 +143,7 @@ def build_one(wiki):
     subprocess.check_call(["nice", "make", "clean"], cwd=wiki)
     subprocess.check_call(["nice", "make", "html"], cwd=wiki)
 
+
 def sphinx_make(site):
     """
     Calls 'make html' to build each site
@@ -172,6 +180,7 @@ def sphinx_make(site):
                 if p.exitcode != 0:
                     error_count += 1
         time.sleep(0.1)
+
 
 def copy_build(site):
     """
@@ -217,6 +226,64 @@ def copy_build(site):
         # delete the old directory
         print('DEBUG: removing %s' % olddir )
         shutil.rmtree(olddir)
+
+
+def copy_and_keep_build(site):
+    """
+    Copies each site into the target location and keep last "n" builds as backups
+    """
+    global error_count
+    for wiki in ALL_WIKIS:
+        if site=='common':
+            continue
+        if site is not None and site != wiki:
+            continue
+        debug('coping: %s' % wiki)
+        targetdir = os.path.join(args.destdir, wiki)
+        distutils.dir_util.mkpath(targetdir)
+
+        if os.path.exists(targetdir):
+            olddir = os.path.join(args.backupdestdir, str(building_time + '-wiki-bkp'), str(wiki) )
+            debug('Checking %s' % olddir )
+            distutils.dir_util.mkpath(olddir)
+            debug('Moving %s into %s' % (targetdir,olddir))
+            shutil.move(targetdir, olddir)
+
+            sourcedir = os.path.join(os.path.abspath(os.getcwd()), str(wiki), 'build', 'html')
+            html_moved_dir = os.path.join(args.destdir, 'html')
+            try:
+                #subprocess.check_call(['mv', sourcedir, html_moved_dir])
+                shutil.move(sourcedir, html_moved_dir)
+                # Rename move! (single move to html/* failed)
+                #subprocess.check_call(['mv', html_moved_dir ,targetdir])
+                shutil.move(html_moved_dir, targetdir)
+                debug("Moved to %s" % targetdir)
+            except:
+                error("FAIL moving output to %s" % targetdir)
+            finally:
+                debug("Creating a backup in %s" % olddir)
+                # subprocess.check_call(['cp', '-r', targetdir ,olddir])
+                distutils.dir_util.copy_tree(targetdir, olddir, preserve_symlinks=0)
+        else:
+            error("FAIL when looking for folder %s" % targetdir)
+
+
+def delete_old_wiki_backups(folder, n_to_keep):
+    try:
+        debug('Checking number of number of backups in folder %s' % folder)
+        backup_folders = glob.glob(folder  + "/*-wiki-bkp/")
+        backup_folders.sort()
+        if len(backup_folders) > n_to_keep:
+            for i in range(0, len(backup_folders) - n_to_keep):
+                if '-wiki-bkp' in str(backup_folders[i]):
+                    debug('Deleting folder %s' % str(backup_folders[i]))
+                    shutil.rmtree(str(backup_folders[i]))
+                else:
+                    debug('Ignoring folder %s because it does not look like a auto generated wiki backup folder' % str(backup_folders[i]))
+        else:
+            debug('No old backups to delete in %s' % folder)
+    except Exception as e:
+        error('Error on deleting some previous wiki backup folders: %s' % e)
 
 
 def generate_copy_dict(start_dir=COMMON_DIR):
@@ -383,6 +450,7 @@ def is_the_same_file(file1, file2):
 
     return(digests[0] == digests[1])
 
+
 def fetch_versioned_parameters(site=args.site):
     """
     It relies on "build_parameters.py" be executed before the "update.py"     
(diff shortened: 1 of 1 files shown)


brunoolivieri:
Delayed in orther to create the vhost for the new server folder.

brunoolivieri:
ping @peterbarker

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
...
 
+def copy_and_keep_build(site):
+    """
+    Copies each site into the target location and keep last "n" builts as backups
+    """
+    global error_count
+    for wiki in ALL_WIKIS:
+        if site=='common':
+            continue
+        if not site==None and not site==wiki:
```
peterbarker, line 237:
```suggestion
        if site is not None and site != wiki:
```

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
...
+            continue
+        debug('copy: %s' % wiki)
+        targetdir = os.path.join(args.destdir, wiki)
+
+        debug("Creating  temporary folders")
```
peterbarker, line 242:
```suggestion
        debug("Creating temporary folders")
```

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
...
+
+        if os.path.exists(targetdir):
+            debug('Moving %s into %s' % (targetdir,olddir) )
+            shutil.move(targetdir, olddir)
+
+        sourcedir='./%s/build/html/' % wiki
+
+        html_moved_dir = os.path.join(args.destdir, 'html')
+        try:
+            subprocess.check_call(['mv', sourcedir, html_moved_dir])
```
peterbarker, line 261: Any chance of using `os.rename` here?

Shelling out is awful - and yeah, I know we already do a lot of it in here.

```

pbarker@bluebottle:~$ mkdir pop
pbarker@bluebottle:~$ python
Python 2.7.17 (default, Apr 15 2020, 17:20:14) 
[GCC 7.5.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.rename("pop", "/tmp/pop")
>>> 
pbarker@bluebottle:~$ ls /tmp/pop
pbarker@bluebottle:~$ 
```

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
...
+
+
+def delete_old_wiki_backups(folder, n_to_keep):
+    try:
+        backup_folders = glob.glob(folder  + "/*/")
+        backup_folders.sort()
+        if len(backup_folders) > n_to_keep:
+            for i in range(0, len(backup_folders) - n_to_keep):
+                debug('Deleting folder %s' % str(backup_folders[i]))
+                subprocess.check_call(['rm', '-rf', str(backup_folders[i])])
```
peterbarker, line 280: `shutil.rmtree`

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
...
+        finally:
+            debug("Creating a backup in %s" % olddir)
+            subprocess.check_call(['cp', '-r', targetdir ,olddir])
+            delete_old_wiki_backups(previousdir, N_BACKUPS_RETAIM)
+
+
+def delete_old_wiki_backups(folder, n_to_keep):
+    try:
+        backup_folders = glob.glob(folder  + "/*/")
+        backup_folders.sort()
```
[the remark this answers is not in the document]
brunoolivieri: Actually, it not breaks. The .sort() does it in situ.

update.py
```diff
@@ -136,6 +141,7 @@ def build_one(wiki):
     subprocess.check_call(["nice", "make", "clean"], cwd=wiki)
     subprocess.check_call(["nice", "make", "html"], cwd=wiki)
 
+
```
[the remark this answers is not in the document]
brunoolivieri: Short answer: Flake8 ignores that.  ||  Long answer: I was using the Pylint, and it is simpler than flake8. When I ran flake8, and it raised more than 100 warnings in this file. I suggest leaving these spaces, and I will sit down to check pep8 warnings in the building scripts. What do you think?

update.py
```diff
@@ -62,6 +65,8 @@
 PARAMETER_SITE={'rover':'APMrover2', 'copter':'ArduCopter','plane':'ArduPlane','antennatracker':'AntennaTracker' }
 LOGMESSAGE_SITE={'rover':'Rover', 'copter':'Copter','plane':'Plane','antennatracker':'Tracker' }
 error_count = 0
+N_BACKUPS_RETAIM = 10
```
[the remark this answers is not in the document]
brunoolivieri: Wopss. Corrected.

update.py
```diff
@@ -219,6 +226,64 @@ def copy_build(site):
         shutil.rmtree(olddir)
 
 
+def copy_and_keep_build(site):
+    """
+    Copies each site into the target location and keep last "n" builts as backups
```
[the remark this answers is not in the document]
brunoolivieri: Wopss. Corrected.

update.py
brunoolivieri: Wopss. Corrected.

update.py
brunoolivieri: Updated here, but not in the original.

update.py
brunoolivieri: I do agree that using the shell is not a good idea. As the wiki scripts are being here for a while and me kinda new around, I make some efforts to maintain it as the same as possible (i.e.: maintain the original methods and insert new methods and enable them IFF an argument is used).

For this specific case, I would like to maintain this set of possible exceptions as the original. Do you mind? Otherwise, I would need to check the same collection of possible errors, and I am not aware of them.

update.py
brunoolivieri: This one is easy to update. I will change it.

brunoolivieri:
@peterbarker  Thanks! I guess it is good now. 

I am intending to run I in the wiki-beta environment and them merge. What you think?

brunoolivieri:
Indeed it worked fine in the wiki-beta environment.

update.py
```diff
@@ -53,6 +54,8 @@
 parser.add_argument('--cached-parameter-files', action='store_true', help="Do not re-download parameter files")
 parser.add_argument('--parallel', type=int, help="limit parallel builds, -1 for unlimited", default=1)
 parser.add_argument('--destdir', default="/var/sites/wiki/web", help="Destination directory for compiled docs")
+parser.add_argument('--enablebackups', action='store_true', default=False, help="Enable several backups up to const N_BACKUPS_RETAIN in PREVIOUS folder")
```
peterbarker, line 57:
```suggestion
parser.add_argument('--enablebackups', action='store_true', default=False, help="Enable several backups up to const N_BACKUPS_RETAIN in --backupdestdir folder")
```

update.py
```diff
@@ -219,6 +226,65 @@ def copy_build(site):
...
+        debug("Creating temporary folders")
+        previousdir = os.path.join(args.backupdestdir)
+        debug('Recreating %s' % previousdir )
+        if not os.path.exists(previousdir):
+            os.mkdir(previousdir)
+    
+        olddir = os.path.join(previousdir, str(building_time) )
+        debug('Recreating %s' % olddir )
+        if not os.path.exists(olddir):
+            os.mkdir(olddir)
```
peterbarker, line 251: Consider using
```suggestion
        olddir = os.path.join(args.backupdestdir, str(building_time) )
        debug('Recreating %s' % olddir )
        distutils.dir_util.mkpath(olddir);
```

update.py
```diff
@@ -219,6 +226,65 @@ def copy_build(site):
...
+
+        if os.path.exists(targetdir):
```
peterbarker, line 253: Could consider only creating `olddir` if `targetdir` exists.

update.py
```diff
@@ -219,6 +226,65 @@ def copy_build(site):
...
+            debug('Moving %s into %s' % (targetdir,olddir) )
+            shutil.move(targetdir, olddir)
+
+        sourcedir='./%s/build/html/' % wiki
```
peterbarker, line 257: whitespace

peterbarker: Should probably also be `os.path.join`

update.py
```diff
@@ -219,6 +226,65 @@ def copy_build(site):
...
+        try:
+            subprocess.check_call(['mv', sourcedir, html_moved_dir])
+            # Rename move! (single move to html/* failed)
+            subprocess.check_call(['mv', html_moved_dir ,targetdir])
+            debug("Moved to %s" % targetdir)
+        except:
+            error("FAIL moving output to %s" % targetdir)
+        finally:
+            debug("Creating a backup in %s" % olddir)
+            subprocess.check_call(['cp', '-r', targetdir ,olddir])
```
peterbarker, line 269: This `check_call` does seem to be whole-cloth new.  Please use a Python library to do this one.  Probably `shutil.copytree`

update.py
```diff
@@ -219,6 +226,65 @@ def copy_build(site):
...
+        backup_folders = glob.glob(folder  + "/*/")
+        backup_folders.sort()
+        if len(backup_folders) > n_to_keep:
+            for i in range(0, len(backup_folders) - n_to_keep):
+                debug('Deleting folder %s' % str(backup_folders[i]))
+                shutil.rmtree(str(backup_folders[i]))
+        else:
+            debug('No old backups to delete in %s' % folder)
+    except Exception as e:
+        error('Error on deleting some previous wiki backup folders: %s' % e)
```
peterbarker, line 285: Should we be incrementing `error_count` here?

Similarly in some other exception blocks, actually.

update.py
brunoolivieri: Yes, it is better. Updated.

update.py
brunoolivieri: Sure. Updated ln 264 due to args.backupdestdir

update.py
brunoolivieri: shutil.copytree() is slightly different from python 2.7 to 3.8. So, I used the previous indication distutils.dir_util lib.

update.py
brunoolivieri: ok

update.py
brunoolivieri: This, specifically, is a good sign that directories that have not been deleted could fill the disk.

I took as a standard accounting in all, just in case.

Anyway, given this review, I see that I will have to go back with some previous PR to adjust some things. Maybe evaluate everything concerning python 3 and also with the same libraries.

update.py
brunoolivieri: done

brunoolivieri:
@peterbarker , ok now?

update.py
```diff
@@ -219,6 +228,58 @@ def copy_build(site):
...
+        finally:
+            debug("Creating a backup in %s" % olddir)
+            # subprocess.check_call(['cp', '-r', targetdir ,olddir])
+            distutils.dir_util.copy_tree(targetdir, olddir, preserve_symlinks=0)
+
+
+def delete_old_wiki_backups(folder, n_to_keep):
+    try:
+        debug('Checking number of number of backups in folder %s' % folder)
+        backup_folders = glob.glob(folder  + "/*/")
```
peterbarker, line 271: Could we tighten the glob here somehow?  As it stands someone can pass in a bad archive folder and nuke stuff they didn't want to nuke.  I'm usually an "enough rope" sort of person, but this sort of construct makes me a bit wary.

Something like `os.path.join(folder, "*-*-*-*-*/")` might work, for example.

brunoolivieri:
> 
> 
> @brunoolivieri I'm guessing you might like to merge this yourself and monitor the server, rather than me merging it?

Sure, I ususally do that with the scripts, np. I took some of your time this time to get an idea of how I am doing these changes. Thank you :-)

update.py
brunoolivieri: I do agree because I don't trust users, especially when I'm one of them :-)

I forced a nomenclature in the folders and when it was time to delete them.

brunoolivieri:
Script done. I will rebase it and merge when I can watch the first execution in the server.

brunoolivieri:
Merging and watching first build.

PR #2854, opened 2020-06-17
Add UAV LogViewer wiki information

Hwurzburg:
only one tbd...in this, how to remove an added preset...but I will merge and fix later if Willian approves everything else...

Willian once merged you can use the url of the page on your viewer as a help file


common/source/docs/common-uavlogviewer.rst
```diff
@@ -0,0 +1,90 @@
...
+Plot Features
+-------------
+
+#. When each field is added, a Y axis number is assigned, and the data autoscaled and centered. This can be changed by hovering over the axis and scrolling to change scale, or clicking and moving the origin. If you wish to automatically align and scale several fields to one another, just set the y axis number to be the same. They will remain together from that point on.
+
+#. You can change the color code by clicking the color swatch.
+
+#. You can remove a data plot by clicking the trash can icon.
+
+#. You can even create and plot expressions that use the fields by "Add Expression" button. See ` Graph Expressions for MavExplorer <https://ardupilot.org/dev/docs/using-mavexplorer-for-log-analysis.html#graph-expressions>`__ for examples of using expressions in a plot analysis.
```
Williangalvani, line 61: I would add a note that currently only a subset of the "functions" available in mavproxy is available. I ported a bunch but there are still some missing

Hwurzburg: if you have a list, I could include

Williangalvani: unfortunately I don't have it right now, I'll have to compare https://github.com/ArduPilot/pymavlink/blob/master/mavextra.py
and
https://github.com/ArduPilot/UAVLogViewer/blob/master/src/mavextra/mavextra.js
and also think of built-ins

Williangalvani: Can we skip this for now?

Hwurzburg: sure....

PR #2859, opened 2020-06-18
Frontend proposal for C5 shutdown

brunoolivieri:
This is an early and naive PR to discuss a very clean and plain new frontend to substitute C5 for ArduPilot main page. Here is a single page sample:  www.olivieri.com.br/ardupilot/bootstrap/

The whole idea is to use plain HTML that could be updated inside the Wiki environment like any other Wiki's  PR.

To proceed, I need to address some points:

- [x] The header and footer of the page are good enough?  IFF it is close to being good enough, I can recreate the C5 pages statically. ~~or try to create a new sphinx theme for those.~~ (updated)

- [x] Presenting Discourse links and images alert users about cross-site items (SSL issue). I am not sure, but I guess this is not good for being found from google and other SEO stuff. Could we just present the text? XOR process it not dynamically on the client-side, but import it when the wiki is being built?


brunoolivieri:
This new commit updates the main page. It inserts ArduPilot wiki items: www.olivieri.com.br/ardupilot/bootstrap/

Any thoughts about header and footer? If not I will recreate the C5 other pages:

Partners
Wiki landing page
User cases
Team members

Hwurzburg:
How do I test this?

brunoolivieri:
> 
> 
> How do I test this?

@Hwurzburg 
Not necessarily building. What do you think about the overall appearance and content?

Hwurzburg:
How do I view i

rmackay9:
Looking forward to this!

brunoolivieri:
Notes from weekly-devcall from 2020-07-20:

1) Looks a good way to proceed.
2)Logs section will stay by creating something in Python to run on the server-side.
3) Separate repo or a page inside the Wiki repo?
4) We’ll go ahead once we’ve got the blog posts happening.

brunoolivieri:
After some absence, I would like to return to this.

So, the last conversation was about to insert blog posts on that. So, I did a python script that creates a JSON in the server, and the pages load that.

There is a test version on http://wiki-beta.ardupilot.org:83/frontpage/

@rmackay9 :  Is that what do you were talking about for as the blog presentation on the main page? The number of posts can vary easily, and it is kind responsive to the screen.

rmackay9:
@brunoolivieri,

I was hoping that it would look a bit more like ardupilot.org does at the moment with the blogs including a picture and the top few lines of text would appear.  I think the blog posts should be one of the first things that people see when they come to ardupilot.org.

brunoolivieri:
Get text from the Discourse is a little tricking because users' texts may have some self-made HTML. 

I got the text and removed the standard HTML and shorted it for 140 characters for a couple of trials:

|          |                                                                    |
|----------|--------------------------------------------------------------------|
| [Layout 1](http://wiki-beta.ardupilot.org:83/frontpage/index_layout_01.html) | Basic blog images.                                                 |
| [Layout 2](http://wiki-beta.ardupilot.org:83/frontpage/index_layout_02.html)  | The community almost on the page top and title of the blog and lastest posts. |
| [Layout 3](http://wiki-beta.ardupilot.org:83/frontpage/index_layout_03.html)  | Layout 2 was added by the first 140 characters of the post itself.         |

@rmackay9  Is there something that you think is a good path to follow?

brunoolivieri:
ping @rmackay9

brunoolivieri:
We are restarting this initiative. 

Please @Hwurzburg, check the 3 proposed laytouts and choose one. The main diferrence is how the page gets the last blog posts from the forum.

Then, I will do a script to build it during wiki building and we can test that.

Hwurzburg:
I like the third one.....not exactly the same but close enough....I think having it easily editable greatly outweighs any small issues with it not being the exact same format as the old....if we could measure it, I bet that  jumps from this page are to documentation vastly outweighs any other

brunoolivieri:
So, to put it in place, I guess at least three things need to be put in place:

[] - insert it on the build script (doing)
[] - enhance the blog posts importer script (I will do it after the build one)
[] - update plain pages and decide what to put on empty pages* 


*so far, plain or empty pages are:
-> /partners.html
-> /team.html
-> /cases-lading-page.html
-> /documentation-lading-page.html
-> /common-stores.html
-> /commercial-support.html
-> /gsoc-lading-page.html

tridge:
>     * enhance the blog posts importer script (doing)

I did some recent edits to the php script that imports blog posts. I could convert it to python if you like, or it could be used as-it.

Note that simplest initial solution is to create a HTML frame with this URL:

  https://ardupilot.org/blogfeed-static.html

that URL is auto-updated from the php script and this is exactly how the C5 code works (HTML frame with that URL)

brunoolivieri:
Thank you @tridge !  I wish I saw that php before. So, I did a python that get data from discourse JSONs.

brunoolivieri:
I think we finished the last set of proposals: http://wiki-beta.ardupilot.org:83/frontend/

@khancyr and @Hwurzburg : What you think about put it on next dev meeting?

Hwurzburg:
Looks great to me!

brunoolivieri:
Merging as discussed and watching the building.

PR #2863, opened 2020-06-20
Fix link errors from previous commit

Hwurzburg:
@stephendade rpi changes deleted links in other documents...unfortunately local build will not catch these since they are cross site links in the .inv files on the server....I think...

@hamishwillee please check that the Odroid changes look right, if you can..

I need to be more careful regarding deleted cross-site-refs since they cannot be caught until merged... @peterbarker   is something we can add to the build check for local builds?


peterbarker:
FWIU (and this is only from watching messages scroll by!), we slurp the .inv files from the actual documentation server, rather than generating them locally.

I actually remarked on this last time I played in update.py...

@brunoolivieri said you'd need to do a pair of builds initially to have this work.  I don't know the details.

PR #2864, opened 2020-06-20
Fixing some dead links to hardkernel.com

adamskt:
I was browsing the docs and came across these dead links.


Hwurzburg:
thanks....merged

PR #2865, opened 2020-06-20
copter: Change from GSC to GCS

muramura:
I would change the GCS to indicate the ground control station.


Hwurzburg:
thanks...merged

PR #2867, opened 2020-06-21
Update for LGR_OPTIONS bit function

Hwurzburg:


rmackay9:
Thanks for this.  It's a bit wordy perhaps but looks OK.

Hwurzburg:
reworded....as compact as I can make it and still have all the info buried...

rmackay9:
Great, thanks!!

Hwurzburg:
@rmackay9 what about the question? question: if you enter and exit Copter LAND before actually landing is completed, will the gear automatically retract?

rmackay9:
@Hwurzburg, no I don't think it will automatically retract.  It will only retract with the takeoff is complete I think.

PR #2878, opened 2020-06-26
Fix typo in Matek F743 page

samuelctabor:
Fix mis-title as F765.


Hwurzburg:
Damn, its hard to get everything right....took a third set of eyes to notice the copy and paste error ....but even this is wrong....needs to be 'H743'....can you changeand I will merge?  thanks

samuelctabor:
Thanks Henry. Didn't know this FC existed, and I just bought two F765s!

Hwurzburg:
thanks!

PR #2885, opened 2020-06-30
dev: add document with information about serial mapping in SITL

peterbarker:


Hwurzburg:
this is actually alread there : https://ardupilot.org/dev/docs/learning-ardupilot-uarts-and-the-console.html but this is a better format....it is also necessary in https://ardupilot.org/dev/docs/using-sitl-for-ardupilot-testing.html
   
suggestion: since this is likely to never be changed, it might be easier just to copy and past into those documents, eliminating the necessity to jump to the page, AND to find a place for it in a TOC...

peterbarker:
@Hwurzburg A TOC entry might not be a bad idea, and a page titled this way will help people find it easily.  Currently a web search doesn't find the existing information - where it easily does for the Pixhawk equivalent.

Hwurzburg:
incorporated in #2688...thanks

PR #2888, opened 2020-07-01
Update common-rally-points.rst

tatsuy:


peterbarker:
A better commit message would be nice for next time - "Correct spelling in common-rally-points", perhaps.

But thanks for the fix!

tatsuy:
I understand.
thanks for the checking and merging.

PR #2901, opened 2020-07-05
Add more detailed FFT tuning information, including recently added options

andyp1per:


Hwurzburg:
Since this is an update to an already marked futures page, I am merging....thanks!

PR #2902, opened 2020-07-05
ICM20689 does not have compass

Daniel-1276:
Updated line 25, as line 123 already stated correctly, ICM20689 does not include any compass.


Hwurzburg:
LGTM..thanks

PR #2903, opened 2020-07-06
Add CUAV Nora and X7 autopilots

Hwurzburg:


cuhome:
@Hwurzburg Thank you, it looks great. I think it should add an item: Power module: Both X7 and nora packages include CAN PMU SE power module, please connect it to the Power C interface and use it without any configuration (no need to configure battery monitor)

Hwurzburg:
Added...merging

PR #2904, opened 2020-07-06
common: GY-US42v2 Serial TYPE tbd to 31

muramura:
GY-US42v2 serial was merged into the master.
The TYPE is 31.
I change TBD to 31.


Hwurzburg:
thanks...merged

PR #2915, opened 2020-07-15
Fix pix32v5 file charset

brunoolivieri:
The wiki building did break due to file common-holybro-pix32v5.rst charset was not set to utf8.

This PR converts the file to ut8 and remove bad characters.

### DIFF
--- a/common/source/docs/common-holybro-pix32v5.rst
+++ b/common/source/docs/common-holybro-pix32v5.rst
@@ -24,8 +24,8 @@ Specifications
 ==============
 
 -  **Processors**
-     - 32 Bit Arm è¢Ó Cortexè¢Ó -M7, 216MHz, 2MB memory, 512KB RAM
-     - 32 Bit Arm è¢Ó Cortexè¢Ó -M3 IO co-processor, 24MHz, 8KB SRAM
+     - 32 Bit Arm Cortex -M7, 216MHz, 2MB memory, 512KB RAM
+     - 32 Bit Arm Cortex -M3 IO co-processor, 24MHz, 8KB SRAM
 
 -  **Sensors**
      - Accel/Gyro: ICM-20689


brunoolivieri:
Tested locally. Applying.

Hwurzburg:
@brunoolivieri  
1. it does not break the wiki or produce a build error
2. It displays fine on my computer in the webpage, it displays fine in RETEXT

so what is the proper symbol for that registered trademark symbol, as  it needs to be replaced in the document for legal purposes?

brunoolivieri:
Hi Henry, 

(1) I could not build in my box and checked inside the server, and the builds were being aborted. Travis also had problems. Now I can build it and the server built as well. Did you test in windows? Because the file was not in utf8, and I test only in Linux. 

(2) I could not see the correct character in my box and GitHub either.
Can you see it on these links? 

https://github.com/ArduPilot/ardupilot_wiki/blob/1368d1be059df9a912fdced80ff886b609bbcfd1/common/source/docs/common-holybro-pix32v5.rst 

https://raw.githubusercontent.com/ArduPilot/ardupilot_wiki/1368d1be059df9a912fdced80ff886b609bbcfd1/common/source/docs/common-holybro-pix32v5.rst

Because I can not...

So, the server is building fine now. I just need to check how to insert the Copyright symbol. Am I correct?

brunoolivieri:
Guess PR #2916 covers the gap of missing symbol.

Hwurzburg:
thanks...but that is the wrong symbol....I was building locally and viewing with browser and nothing was wrong that I could see on my Linux box...

brunoolivieri:
If you saw in linux I cannot explain. I noticed that was something wrong by checking the server build log. The server was not building and I only changed the encode of the file and removed the symbol and worked fine. 

Sorry if I missed something.

PR #2925, opened 2020-07-19
common: Remove Under Development

muramura:
I know that GY-US42v2 has finished its development.
I remove under development.


Hwurzburg:
thanks...merging

PR #2927, opened 2020-07-23
frsky: updated yaapu images and text

yaapu:
minimal changes with updated screenshots and supported radio list


Hwurzburg:
merged ...thanks!

PR #2937, opened 2020-08-03
removing apparently superfluous <th>value</th>

arikrupnik:
It appears that `<th>value</th>` appears twice in the table of Tilt Servo functions & values (https://ardupilot.org/plane/docs/guide-tilt-rotor.html#tilt-servos)

### DIFF
--- a/plane/source/docs/guide-tilt-rotor.rst
+++ b/plane/source/docs/guide-tilt-rotor.rst
@@ -133,7 +133,7 @@ You control that with 3 possible servo function values.
 .. raw:: html
 
    <table border="1" class="docutils">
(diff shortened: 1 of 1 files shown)


Hwurzburg:
funny how you can look at something repeatedly and never see something like this! thanks! merged

arikrupnik:
Thank you for merging this in, @Hwurzburg! As I'm reading the docs, these little typographical issues come up. if you want to keep merging them, I'm happy to keep making PRs as I find them.

Hwurzburg:
please do!

kd0aij:
the table headers are a bit inconsisent though:
image

Hwurzburg:
Value in the last could be Function...and columns swapped....I'll do that now

PR #2943, opened 2020-08-07
Sphinx: remove mercurial and change vimeo plugin location

khancyr:
This is RFC.
It removed mercurial dependency to save some space and bandwidth on setup. It change the target to a clone of the original vimeo plugin on github. The plugin hasn't been updated from long time. Ideally we should put the plugin on ArduPilot github

It works as previously no issue spotted. We gain 40Mo of space and some second on environment setting. We also remove dependency to mercurial and bitbucket to have only git and github.

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -12,7 +12,7 @@ if [ ${DISTRIBUTION_CODENAME} == 'Ubuntu' ]; then
   add-apt-repository universe
 fi
 apt-get -y update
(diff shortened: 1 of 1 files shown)


stephendade:
+1 for merging this in, as the current wiki build is broken (has been for the last couple of days). Looking at the build log, Bitbucket has removed mercurial access, which means the Sphinx vimeo plugin is no longer accessible

PR #2945, opened 2020-08-08
fix broken link due to param name change

Hwurzburg:


rmackay9:
Aha, so I just merged a PR which corrected the parameter name.  So it was the flight code that had the odd name with two underscores.  The winch code has been around a while but has always been defined out of the firmware but it will go out with Copter-4.1.

PR #2950, opened 2020-08-09
Dev: Add info about instance number when using MAVExplorer

rajat2004:


Hwurzburg:
LGTM..thanks...merged

PR #2954, opened 2020-08-11
Update MAVProxy-in-Windows

stephendade:
A few updates for Windows users:
-Mission Editor module now works
-Developer instructions checked and updated


Hwurzburg:
LGTM....thanks...merged

PR #2956, opened 2020-08-12
Add flight options page in Plane wiki

Hwurzburg:


plane/source/docs/flight-options.rst
```diff
@@ -0,0 +1,20 @@
...
+==============
+
+
+:ref:`FLIGHT_OPTIONS<FLIGHT_OPTIONS>` is a bitmask that allows configuring several alterations to Plane's behavior.
+
+
+=====================================   ======================
+:ref:`FLIGHT_OPTIONS<FLIGHT_OPTIONS>`   Function
+=====================================   ======================
+0                                       Allows rudder input to be used in only MANUAL,STABILIZE,and ACRO modes. In other modes, rudder input mixing is controlled by the :ref:`STICK_MIXING<STICK_MIXING>` bitmask parameter.
```
tridge, line 14: STICK_MIXING isn't a bitmask

Hwurzburg: thanks...fixed....did not read closely enough...thought it was obvious that you could want both VTOL yaw option AND FBWA mixing in QuadPlane landings (I would),  so assumed it was bitmask instead of reading closely...

PR #2966, opened 2020-08-16
Add Airmode upcoming feature for Plane

Hwurzburg:


plane/source/docs/airmode.rst
```diff
@@ -0,0 +1,28 @@
...
+This is not an actual flight mode, but rather, an important feature of QACRO and QSTABILIZE modes. Other flight control software refer to it as a flight mode, so for consistency, its included in the QuadPlane flight modes documentation.
+
+What it is
+==========
+
+Normally, in QuadPlane, if throttle is lowered to idle in QACRO or QSTABILIZE modes, the stabilization is removed. While this prevents movement due to vibration while on the ground at idle throttle, or when using rudder arming, it prevents stabilization in flight for aerobatic movements or rapid descents.
+
+Instead, stabilization even at zero throttle can be enabled/disabled for these modes using an RC channel switch via ``RCx_OPTION`` = 84 (Airmode). 
+
+.. note:: if :ref:`ARMING_RUDDER<ARMING_RUDDER>` is set to 2 (arm and disarm using rudder stick), while in Airmode, you cannot disarm using rudder. This allows full rudder stick throws without the chance of a disarm occurring.
```
kd0aij, line 18: rudder disarm is disabled when is_flying(), but there isn't a check for airmode
I wonder if this has been tested adequately?

Hwurzburg: I did not assume this....I discovered it when running thru the variations in SITL in order to document....disarm by rudder is prevented on the ground....set ARMING_RUDDER=2.....then evoke AIRMODE in any manner...AIRMODE switch or arm by switch...it will not disarm on the ground with rudder....must think that since the stabilizing is running the motors over spin min, its flying....or there is something else preventing disarm...

kd0aij: you're right, I thought I was in airmode when I ran my sitl test, but was not

is_flying returns true when (is_vtol_man_throttle() && air_mode == AirMode::ON), so the method name is misleading in this case.

kd0aij: could add that to autotest....

Hwurzburg: so everything look okay? if so, approve and I will merge...

PR #2968, opened 2020-08-17
Scripts: Ensure js files are picked up

stephendade:
This is a follow-on to https://github.com/ArduPilot/ardupilot_wiki/pull/2926. A required js file wasn't picked up by the build system. This patch fixes this.

Tested locally and works fine.


Hwurzburg:
seems to fix it....thanks!

PR #2973, opened 2020-08-20
Add note about joystick not working in WSL

TunaLobster:


Hwurzburg:
does it work in WSL2?

TunaLobster:
I don't believe so because libusb support hasn't been added yet. I'll have to set up a WSL2 to verify though.

PR #2981, opened 2020-08-24
Fix build issues from User Alerts

stephendade:
Fixing up the issues @Hwurzburg noted in https://github.com/ArduPilot/ardupilot_wiki/pull/2926:

-Moved custom css to common
-Ensure useralerts.js is only copied for plane,copter,rover,antennatracker


Hwurzburg:
thanks! merged

PR #2987, opened 2020-08-27
Update common-rfd900.rst

kentmartin99:
Change link to RFD900/868x modem instead of RFD900/868+ legacy modems


Hwurzburg:
thanjks....while we are at it...added another partner source...I like to have multiple sources listed if available...merged

PR #2990, opened 2020-08-28
Add caution on prop strike on TVBS and solution

Hwurzburg:


Hwurzburg:
yeah, but they dont prop strike....it is smoother going into auto from qstab....perhaps that should be the recommendation for auto takeoffs...only from Qstab?what do you think?

kd0aij:
OK, it's certainly better than the alternative.

Try this fix: https://github.com/ArduPilot/ardupilot/pull/15195
It looks good to me, but I may be missing something important...

Hwurzburg:
I went ahead and recommended starting from QSTAB for AUTO takeoffs...its the safest until your PR makes it to stable...merging....thanks for fixing this @kd0aij

PR #3004, opened 2020-09-06
Update :ref: link and photo size

vincentpoont2:
-Not sure why :ref:'common-holybro-rtk-f9p' wasn't working before, does this fix it?
-Update photo size

image


Hwurzburg:
we should not change the width...some browsers will not display properly the larger format...they are hot linked so clicking on them will bring a maximum sized image for the particular browser...

I will fix the other link...

Hwurzburg:
closing as explained above...

PR #3005, opened 2020-09-06
Update Photo Size

vincentpoont2:
Update Photo Size, previously the photo appeared to be too small on website.
image


Hwurzburg:
we should leave as is, as explained on other PR

Hwurzburg:
going to close this as explained above

PR #3008, opened 2020-09-08
Fixed incorrect indentation

abingham:
The ``init`` function should be at module scope, but it was nested in the ``TestModule`` class.


stephendade (approved):

Hwurzburg:
thanks...merging

PR #3014, opened 2020-09-09
amount -> number

arikrupnik:
"Meters" is a count noun. You can climb this *number* meters or this *amount* of mass noun, e.g., "this amount of altitude." I propose changing this to "will climb that many meters," but it could equally be "will climb that number of meters."


peterbarker (approved):
I'd prefer "that number", but this is clearly better.

arikrupnik:
I'm OK with either "that number of meters" or "that many meters."

Hwurzburg:
thanks...merged

PR #3025, opened 2020-09-13
Update tailsitter info for BiCopter

Hwurzburg:
Pete, another question...does BiCopter use the Q_TILT_YAW_ANGLE param? seems like it would have to ....but I assume the other transition parameters are exactly like the tailsitter transition stuff and not the tilt rotor transition stuff?

### DIFF
--- a/plane/source/docs/guide-tilt-rotor.rst
+++ b/plane/source/docs/guide-tilt-rotor.rst
@@ -14,9 +14,8 @@ where transition between hover and forward flight is accomplished by
 tilting one or more rotors so that it provides forward thrust instead
 of upward thrust.
 
-This is distinct from :ref:`tailsitters <guide-tailsitter>` where the
-autopilot and main fuselage change orientation when moving between
-hover and forward flight.
+.. note:: This is distinct from :ref:`tailsitters <guide-tailsitter>` where the autopilot and main fuselage change orientation when moving between hover and forward flight. Do  not use the information below for a Tailsitter, some parameters are shared, but use the  instructions in the :ref:`guide-tailsitter`.
+
 
 Types of Tilt-Rotors
 ====================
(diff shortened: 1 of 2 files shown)


IamPete1:
Servo min is the forward flight position, trim is up, and max is back. So there should be 90 deg of moment between min (forward) and trim (up). Q_TILT_YAW_ANGLE should be set to the angle between the up (trim) and back (max).

Hwurzburg:
I dont think trim is involved in normal tilt rotor setup.....set the tilt yaw angle, then adjust max for up (its really not that PWM..you are adjusting up+yaw angle with max, but QSTAB drives the up) and min for forward (manual)....dont think trim value is involved or has any effect....so this is different

kd0aij:
I'm confused; are we talking about tiltrotors or copters here? 
The only SITL vehicle I have is a copter; is there an RF8 model configured as a bicopter for quadplane?

kd0aij (commented):
the yaw torque comment looks good to me
But the bicopter/tailsitter thing seems confusing, especially since the V22 is not a tailsitter.

Hwurzburg:
the yaw torque comment is wrong you just told me....stable does not have it (no yaw torque...it has it)....or maybe, it needs a stable number as to when it was added to stable

kd0aij:
What's confusing is that a bicopter tailsitter is not a tailsitter, but it uses frame class 10.  It might be better to define a new frame class for the V22, just to avoid confusion.

Hwurzburg:
does it use Quadplane transitions or Tailsitter transitions and params?if it uses thr-lft/rgt I expect it uses tailsitter transitions...very confusing....hence the need for a clear wiki

IamPete1:
It does standard quad-plane transmissions. I think the confusion is simply that the 'bicopter' frame type is called tailsitter because that is what it was original added for. 

So you can have a 'bi-copter tailsitter' or a 'bi-copter tilt-rottor' or in ArduCopter  just a 'bi-copter'. 

We probably should look at separating the tailsitter and tilt rotor types from frame types, now you can have more combinations of each.

The tilt and tailsitter mask are the current tirggers, but the bicopter frame type does not use the same masks so needs a special case for each.

Hwurzburg:
After talking with Pete, and finally understanding how BiCopter is setup, I have added a BiCopter section and modified previous entries....

kd0aij (approved):
the bicopter tiltrotor stuff sounds good to me

plane/source/docs/guide-tilt-rotor.rst
```diff
@@ -66,9 +65,11 @@ hovering. Currently supported tilt-rotor frame classes are:
    <tr><td>Tricopter</td><td>7</td></tr>
    </table>
 
+.. note: the BiCopter tilt-rotor (either motors tilting ,or entire wing with fixed motors tilting) QuadPlane is a special case. It requires that :ref:`Q_FRAME_CLASS<Q_FRAME_CLASS>` be set to 10 which is reserved for Tailsitter configurations only, normally. See :ref:`BiCopter <bicopter>`  below.
```
IamPete1, line 68:
```suggestion
.. note: the BiCopter tilt-rotor QuadPlane is a special case. It requires that :ref:`Q_FRAME_CLASS<Q_FRAME_CLASS>` be set to 10 which is reserved for Tailsitter configurations only, normally. See :ref:`BiCopter <bicopter>`  below.
```

Bi-copter needs its two motors to be able to tilt interdependently.

Hwurzburg: fixed

IamPete1 (approved):

Hwurzburg:
thanks! merging

PR #3026, opened 2020-09-14
vicon: fixed instructions

tridge:
key change is use of code tag so that "--" displays correctly. We need
to watch out for this in future PRs


Hwurzburg:
fixed a few link display errors while we're at it....grepped for other instances...none found...thanks, merging

PR #3033, opened 2020-09-15
Omnibus F4 Nano V6.x: fix 4in1 socket pinout image

vierfuffzig:
and add a note to double check socket pinout against solder pads, as early batches of V6.1 revision seem to use a non-standard socket pinout.


Hwurzburg:
thanks...merging

PR #3034, opened 2020-09-16
Update common-omnibusnanov6.rst

CJ-Davies:
Presumably there was no specific OmnibusNanoV6_bl file back when this doc was first written? I have just tried flashing both omnibusf4pro_bl.bin & OmnibusNanoV6_bl.bin to a new V6.1 board & only the latter produces the expected behaviour (solid power LED & rapidly blinking blue LED). I then successfully flashed Plane 4.0.6 via Mission Planner.

### DIFF
--- a/common/source/docs/common-omnibusnanov6.rst
+++ b/common/source/docs/common-omnibusnanov6.rst
@@ -158,73 +158,29 @@ Set :ref:`BATT_MONITOR <BATT_MONITOR>` to 4 (= analog voltage and current) and r
 
 Flashing Firmware
 =================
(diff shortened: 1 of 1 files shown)


vierfuffzig:
nice find! there have been some updates to this file recebtly, but this obviously has slipped my attention...

Hwurzburg:
I think we may want to change to the more current way of flashing onto a BF/iNav preloaded board and using the one step STM32CubeProgrammer with the _bl.hex file...
it will prevent problems with those not having the DFU driver installed already (current instructions tacitly assume this) and is a single step...

just remove all the instructions and link to : https://ardupilot.org/plane/docs/common-loading-firmware-onto-chibios-only-boards.html and 
https://ardupilot.org/plane/docs/common-loading-firmware-onto-pixhawk.html for future firmware changes

Hwurzburg:
thanks....squashing it.....merging

PR #3035, opened 2020-09-17
Add start date for User Alerts

stephendade:
This PR adds a small note to the User Alert pages explaining that the User Alerts don't 100% cover the time period before 16/09/2020.


Hwurzburg:
thanks... merged

PR #3037, opened 2020-09-18
common: remove duplicate link

tatsuy:
unify to "GeoFencing Failsafe"
image


Hwurzburg:
Unfortunately, Randy wants the Copter/Rover titles to be FENCE FAILSAFES in those sections....
instead, isolate the lower GEOFENCING entry to be plane only

Hwurzburg:
Randy wants them called Fences in Copter/Rover....its a GEOFENCE in plane and shows different info....that link should just be isolated for PLANE site and not removed....

tatsuy:
Thank you for your review.
I changed that GEOFENCE is displayed in plane only.

Hwurzburg:
thanks...merging

PR #3041, opened 2020-09-18
Add SITL parameter list page

Hwurzburg:
got tired of tracking them all down in the code....just did the most general ones...hopefully , someone will add descriptions for the others over time...


rmackay9:
I've struggled with this as well.  I wonder if we could add them into the code like we do for regular parameters so they appear in all the ground stations.

Hwurzburg:
had to fix that title....typing dyslexia strikes again

peterbarker:
There's no problem adding them to the code.  The problem is them all mixing in with the other parameters.  I suggest we go down that path anyway as it's the right thing to do and we don't want to have to maintain them separately.

Hwurzburg:
merging...its a start

PR #3042, opened 2020-09-19
Add Matek M8Q UAVCAN GPS/MAG/BARO

Hwurzburg:


Hwurzburg:
thanks Sampson, merging

PR #3046, opened 2020-09-22
Clarify note about FS continue modes

Hwurzburg:


copter/source/docs/gcs-failsafe.rst
```diff
@@ -40,7 +40,7 @@ In Mission Planner's  full parameter list or full parameter tree, set the :ref:`
...
+-  **Enabled Continue with Mission in Auto Mode (this bit has no effect in 4.0 and later with :ref:`FS_OPTIONS<FS_OPTIONS>` parameter replacing function, see below )** (Value 2) will ignore the failsafe in an Auto Mode mission. Otherwise, it will behave the same as *Enabled Always RTL*. This option no longer exists in Copter 4.0 and later. Instead, see the :ref:`FS_OPTIONS<FS_OPTIONS>` parameter for this function. Setting this value in Copter 4.0 and later version will automatically be converted and set to Value 1 (Always RTL) and the :ref:`FS_OPTIONS<FS_OPTIONS>` bitmask will be modified to include bit (1) for "Continue auto mode mission on GCS failsafe".
```
rmackay9, line 43: Thanks for this.

maybe we should say, "(this **value** has no effect in 4.0 and later with the FS_OPTIONS parameter replacing this function, see below)"

FYI I checked the code and in fact, in 4.0.x it will be treated the same as "1" (Always RTL) but we probably don't need to mention this.

Hwurzburg: fixed....

Hwurzburg:
merging....

Issue #3038, opened 2020-09-18
Error in ArduPPM Module PPM Encoder Fuse Settings


code0303456:
It says:
"Fuse setting for internal MCU

All ATMEGA 328P based PPM Encoders have strict fuse settings.

Following fuse settings should be used when programming with AVR Studio:

    LOW: 0xFF
    High: 0xDA
    Ext: 0xFD
"

However External Fuse can't even take that value

it has value Ext:0x05 on the hex code to be downloaded

and that makes CH1 or CH2 (TX on AtMega328p) not to work propperly

it should be evaluated to use:

  LOW: 0xFF
    High: 0xDA
    Ext: 0x07

for no brownout detection, thus enabling the pin on arduPPM to receive data

i hope this issue will soon be fixed!!

https://ardupilot.org/rover/docs/common-ppm-encoder.html

link to the page with the typo

PR #3049, opened 2020-09-24
Rover: Updated Ext Fuse Settings for common-ppm-encoder

Jovian-Dsouza:
Changed Ext Fuse Settings from 0xFD to 0x07 

Resolves : #3038

### DIFF
--- a/common/source/docs/common-ppm-encoder.rst
+++ b/common/source/docs/common-ppm-encoder.rst
@@ -117,7 +117,7 @@ Following fuse settings should be used when programming with AVR Studio:
 
 - LOW: 0xFF
 - High: 0xDA
-- Ext: 0xFD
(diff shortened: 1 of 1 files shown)


Hwurzburg:
we ought tocompletely eliminate the DYI encoder stuff since its not maintained....plenty of commercial solutions....and the need itself is outdated with modern receivers....but will merge this anyway....thanks

code0303456:
Despite the fact that having to recognise that no-one is perfect, old parts in the code help many students in universities to further comprehend some theory directly applied in practise, and shoud be really really proud of that... be more positive, and dont discard the past from where we all came. Best Regards!!

Hwurzburg:
when we obsolete something, its is archived, not removed...check the appendix

PR #3052, opened 2020-09-27
Add notes about RAM limits

Hwurzburg:
Please add any other suggestions for temporary RAM savings and/or symptoms on the bench....I did not put a note about examining MEMINFO because there is no definite number to tell them if thats the problem, I believe, due to varying setups and whats not starting...

### DIFF
--- a/common/source/docs/common-limited-firmware.rst
+++ b/common/source/docs/common-limited-firmware.rst
@@ -76,3 +76,21 @@ The missing features are listed below.
 
    -  Parachute
    -  Sprayer
+
+RAM Limitations
+===============
+
+There may be insufficient RAM available in some flight controllers to support all enabled firmware features. Some possible symptoms are:
+
+- MAVFTP does not work. Parameter downloads to GCS are delayed starting while the GCS tries to establish a MAVFTP link which cannot be setup, and then proceed slowly to download parameters using the normal download method.
+- Compass Calibration will not start
+- Logging will not start
+- Terrain downloading from GCS will not start
+
+If this occurs, several possible options are available to allow temporary use of MAVFTP, download terrain tiles, and/or Compass Calibration. All require a reboot to take effect:
+
+- Make sure IMU Batch Sampling (used for FFT analysis) is not running by setting :ref:`INS_LOG_BAT_MASK<INS_LOG_BAT_MASK>` = 0.
+- Try temporarily disabling logging by setting :ref:`LOG_BACKEND_TYPE<LOG_BACKEND_TYPE>` to 0, then returning to 1 (default) after calibrating.
+- Reduce the size of :ref:`LOG_FILE_BUFSIZE<LOG_FILE_BUFSIZE>` . However, lowering below 16KB can introduce small gaps in the log. This may be used temporarily to download terrain or calibrate compass.
+- Disable Terrain Following temporarily by setting :ref:`TERRAIN_ENABLE<TERRAIN_ENABLE>` to 0.
+- Disable SmartRTL on Copter by setting :ref:`SRTL_POINTS<SRTL_POINTS>` = 0.


peterbarker:
Flash and RAM are orthogonal.  I would guess that several full-features boards will also suffer from the lack-of-RAM problem.

I'd guess people have noticed it on these boards as they're more likely to have the fft and OSD stuff enabled....

One other option to avoid this is to calibrate the compasses sequentially rather than in parallel.  Obviously that's a bunch more tumbling the user would have to do.

Do we only get a one-short message if terrain allocation fails?  I thought we got a prearm for that?

Hwurzburg:
the f4s have less ram I am told

my failure happened with only one compass, one gps...nothing else

I got NO message about terrain not downloading from the GCS, just remained pending after GPS lock, and MAVftp did not work....

got compass to cal by disabling logging temporarily....

ultimately, I found that I had IMU batch sampling active which consumes about 6K....had copied a previous TVBS with it running (Pete said he keeps it always enabled...but that is obviously not good  now)

peterbarker:
On Sun, 27 Sep 2020, Henry Wurzburg wrote:

> the f4s have less ram I am told

Sure, but several F4s are full-features.

> I got NO message about terrain not downloading from the GCS, just remained pending after GPS lock, and MAVftp did not work....

I wonder if we should just go into the poorly-named "config error loop" if 
we get these allocation errors on startup.  We *could* add prearm checks, 
but that would be a lot more code.

> ultimately, I found that I had IMU batch sampling active which consumes about 6K....had copied a previous TVBS with it running (Pete said he keeps it
> always enabled...but that is obviously not good now)

Dumping the dynamic allocation tables from ChibiOS into a ftp'able file 
would be cool.  Of course, if FTP fails to allocate.... :-)

Hwurzburg:
@peterbarker any problems merging this as is?

common/source/docs/common-limited-firmware.rst
```diff
@@ -76,3 +76,19 @@ The missing features are listed below.
 
    -  Parachute
    -  Sprayer
+
+RAM Limitations
+===============
+
+Occasionally, in the above boards, RAM will become limited with enabled options in the firmware also. This is manifested by various symptoms:
```
kd0aij, line 83:
```suggestion
There may be insufficient RAM available in some flight controllers to support all enabled firmware features. This is manifested by various symptoms:
```

common/source/docs/common-limited-firmware.rst
```diff
@@ -76,3 +76,19 @@ The missing features are listed below.
...
+
+- MAVFtp does not work for fast connections to GCS
```
kd0aij, line 85:
```suggestion
- MAVFtp does not work: parameter fetches to GCS will be slow
```

kd0aij:
If MEMINFO shows less than 16K RAM free at bootup, there's not enough to run mavftp.
It seems that mavproxy doesn't automatically do a param fetch on connection, so that's one way to measure it. Otherwise, enabling log_disarmed should provide the necessary info. 
Ideally though, when something fails to start or run due to insufficient RAM it should emit a warning message.

common/source/docs/common-limited-firmware.rst
peterbarker: Should probably be simply, `Some symptoms are`.

(the symptoms would be of the manifestation, but that's implied).

common/source/docs/common-limited-firmware.rst
peterbarker: Should be `MAVFTP`  (FTP and MAV both being acronyms).

peterbarker (commented):
SRTL on Copter (set points to zero)
Scripting (disable it)
Having both EK2 and EK3 (set (e.g.) EK2_ENABLE to zero)

kd0aij:
Related issue and start of a PR: https://github.com/ArduPilot/ardupilot/pull/15468

common/source/docs/common-limited-firmware.rst
Hwurzburg: done

common/source/docs/common-limited-firmware.rst
Hwurzburg: done

Hwurzburg:
@peterbarker did not mention scripting nor EKF...the 1Mb boards are most likely to have the issue now...2MB boards have more ram, in addition to the additional flash

Hwurzburg:
getting more people wit this issue....going to merge this....any further issues, we can modify

PR #3057, opened 2020-10-02
add multi_veh.png

Satyamtripathi1996:


Hwurzburg:
this does not look like an add! more like a delete...

stephendade (changes requested):
Why is this file being deleted? It's still referenced by ``mavproxy/source/docs/getting_started/multi.rst``, so will break the page.

Hwurzburg:
looks like a mistake...closing

PR #3059, opened 2020-10-02
Update altholdmode.rst

GiliardiSantos:
Updating parameters name - https://discuss.ardupilot.org/t/alt-hold-page-old-pilot-velz-max-parameter/62475


Hwurzburg:
you can drop the comment about earlier versions...if a param is two releases or older, we drop it from the wiki...only 3.6 or later at this point stays...otherwise LGTM...thanks

Hwurzburg:
I will just fix it since you probably are using the easy editor and not a local repo....

Hwurzburg:
merging thanks...

PR #3060, opened 2020-10-03
Add images for tilt rotor setup

Hwurzburg:
still have to add Bicopter QuadPlane setup image


Hwurzburg:
added bicopter...ready for review...

Hwurzburg:
thanks for the input @IamPete1 .....fixed and merging...

PR #3064, opened 2020-10-04
Add criticality field to user alerts

stephendade:
Documentation to reflect the addition of the "criticality" field to user alerts.

Ping @auturgy


Hwurzburg:
LGTM...merging...thanks

PR #3068, opened 2020-10-06
common ADSB: add Sagetech

magicrub:
This adds Sagetech directions.

DO NOT MERGE YET


Hwurzburg:
since this will not be in a release, we will need to put the notes in the future-wiki-changes.rst rather than directly on the ADSB page......dont worry about it...when its ready I will fix and push....thanks

Hwurzburg:
@magicrub any updates?

magicrub:
The ADSB module has changed in other ways. I'll close this in favor of a larger update

PR #3109, opened 2020-10-30
Add C-RTK 9p to estimate yaw

cuhome:
Add C-RTK 9p to estimate yaw


rmackay9:
I'll leave this to Henry to merge but I definitely think it's good to have the video of a copter using GPS-for-Yaw.

Hwurzburg:
builds fine...merging as is

PR #3112, opened 2020-10-30
Fix typo in using-python-scripts-in-mission-planner.rst

holadepo:
Fix a minor typo in using-python-scripts-in-mission-planner.rst


Hwurzburg:
this will cause a build error....the end marker for the column is now miss-aligned

Hwurzburg:
thanks, LGTM...merging

PR #3116, opened 2020-10-31
Extensive FrSky reorg and update

Hwurzburg:
demoted (not removed) non-passthru protocols
removed outdated hardware bloat and software references
reorganized the Frsky main page to allow quick setup
generalized connection diagrams to allow use for both SPort and FPort upcoming feature
added best supported mav to passthru converter to that page

next step is to add bi-directional to upcoming features..


Hwurzburg:
extensively tested all links and build..mostly content reduction and reorg...a few expansions and clarifications

PR #3128, opened 2020-11-06
Changed to a better tutorial link for Jetson users

cnpcshangbo:
I updated the Jetson tutorial link, because the old link can only get the librealsense.so. Jetson can not install pyrealsense2 with pip. In order to get the Python wrapper, compiling from source is a must. The updated link contains the compiling tutorials that leads to both ``librealsense`` and ``pyrealsense2``.


rmackay9:
It's probably best for @thien94 or @patrickpoirier51 to decide on this one.

Hwurzburg:
@thien94  @patrickpoirier51 comments?

patrickpoirier51:
Fine for me

Thanks

thien94:
Sorry for the late reply. The changes are ok.

Many thanks @cnpcshangbo.

Hwurzburg:
Thanks.. merging

PR #3129, opened 2020-11-07
HereFlow: fix mounting orientation

tajisoft:
The Wiki mounting orientation photos always confuse me. I think the photo should be in the right direction.


rmackay9 (approved):
LGTM. I agree fixing the image is the right thing to do, txs!

PR #3136, opened 2020-11-09
common-dshot: fix SERVOx_TRIM pwm mix-up

vierfuffzig:


rmackay9 (approved):
This seems like it's correct to me.  I've never used Dshot but it seems like a good change.

Hwurzburg:
thanks! merging

PR #3140, opened 2020-11-13
Add github actions test

khancyr:


brunoolivieri:
Hi there!

Would this intend to check a PR before a merge as CI does? IFF yes, I highly suggest use the build command "python update.py --clean --paramversioning --parallel 4 --enablebackups" as it runs on the server.

davidbuzz:
lgtm

tridge:
needs to also run the github actions

Hwurzburg:
@khancyr can you address Tridge's comment?

khancyr:
@brunoolivieri I don't think we want the `clean` flag, as it set to use master banch. 
`paramversioning`, need more complex build as it need to generate all the parameter. I will open an issue an do it later
`enablebackups` : what the use for this one ? just testing that we can do backup ?

@Hwurzburg nope, I don't have enough priviledge on wiki to enable actions. The action result could still be seen on my fork : https://github.com/khancyr/ardupilot_wiki/actions/runs/395823375

PR #3150, opened 2020-11-16
updated replay docs for new implementation

tridge:


rmackay9:
This looks like a good improvement so I've gone ahead and merged it.  I may add some additional notes as I play with replay

PR #3154, opened 2020-11-19
Update airmode.rst

EosBandi:
Add a note, that you have to use the switch to arm for engaging "AirMode"

### DIFF
--- a/copter/source/docs/airmode.rst
+++ b/copter/source/docs/airmode.rst
@@ -11,7 +11,7 @@ What it is
 
 Normally, in Copter, if arming is setup via the rudder stick, if throttle is lowered to idle in ACRO or STABILIZE modes, the stabilization is removed. While this prevents movement due to vibration while on the ground at idle throttle, it prevents stabilization in flight for aerobatic movements or rapid descents.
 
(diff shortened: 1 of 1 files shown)


copter/source/docs/airmode.rst
```diff
@@ -14,3 +14,5 @@ Normally, in Copter, if arming is setup via the rudder stick, if throttle is low
 If an ``RCx_OPTION`` is enabled for switch ARM/DISARM (41) then stabilization at idle throttle is still fully active, which is often referred to as AIRMODE.
```
Hwurzburg, line 14:
```suggestion
If an ``RCx_OPTION`` is enabled for switch ARM/DISARM (41) nad used to arm, then stabilization at idle throttle is still fully active, which is often referred to as AIRMODE.
```

copter/source/docs/airmode.rst
```diff
@@ -14,3 +14,5 @@ Normally, in Copter, if arming is setup via the rudder stick, if throttle is low
...
 
 .. note:: AIRMODE has no effect in Traditional Helicopter
+
+.. note:: It isn't sufficient to just enable arming on an RCn_OPTION switch, you *also* have to actually use it for arming. So if you have it on a switch but then use rudder-arming then you won't get "airmode".
```
Hwurzburg, line 18: then this is not necessary

Hwurzburg:
thanks...merging

PR #3156, opened 2020-11-21
Add updating instructions for MAVProxy

Hwurzburg:


mavproxy/source/docs/getting_started/download_and_installation.rst
```diff
@@ -1,100 +1,127 @@
...
+The system will need to be restarted to apply the above change.
+
+Updating
+--------
+
+To update an existing installation with the current release on Python 3 based systems:
+
+.. code:: bash
+
+    sudo pip3 install mavproxy --user --upgrade
```
tridge, line 80: it makes no sense to use --user and use sudo. Using sudo is so it can write to system files (eg. /usr/local/bin). Using --user means write to ~/.local
Note that when you do mix them like this you end up with root owning part of ~/.local. That means if you now leave off the sudo it won't work. To fix this you'll need to do:
sudo chown -R $USER.$USER ~/.local
I don't think we should put that in the docs, I'm just letting you know as you probably want to fix your test machine

tridge (changes requested):

mavproxy/source/docs/getting_started/download_and_installation.rst
Hwurzburg: thanks...still a linux newbie....removed sudo ....okay now?

peterbarker:
Merged, thanks!

PR #3158, opened 2020-11-23
Sphinxsetup.sh: install Python

peterbarker:
Bionic doesn't have python


khancyr (commented):
I think we should enforce to python3 as we will have issue like on ardupilot with the different version of ubuntu

tridge (approved):

PR #3160, opened 2020-11-23
Update last APM2 firmware version in common-autopilots.rst

holadepo:
Update last APM2 firmware version in common-autopilots.rst based on firmware available from https://firmware.ardupilot.org/Plane/


brunoolivieri:
I am pretty sure he is right. Could you confirm @tridge?

I have a couple of APMs here running AC 3.2.1 and AP 3.4.

rmackay9:
This this seems correct. https://firmware.ardupilot.org/Plane/stable/apm2/git-version.txt.  Thanks!

PR #3179, opened 2020-11-27
Add baud rate notes to Serial Passthru

Hwurzburg:


common/source/docs/common-serial-passthrough.rst
```diff
@@ -19,6 +19,7 @@ Setup
 - Connect the autopilot to a PC using a USB cable and connect with a Ground Station (i.e. Mission Planner, QGC, etc)
 - Set :ref:`SERIAL_PASSTIMO <SERIAL_PASSTIMO>` to a lenght of time (in seconds) that gives you enough time to connect with the sensor's configuration software.  30 to 60 seconds is a good choice
 - Set :ref:`SERIAL_PASS2 <SERIAL_PASS2>` to the number of the serial port connected to the sensor.  I.e. "2" if the sensor is connected to Telem2/Serial2
+- Be sure to set each port's baud rate appropriately using the ``SERIALx_BAUD`` parameter. The rates may be different for each port. ArduPilot will do the buffering. However, flow control would need to be used, either with the UART's CTS/RTS on the higher baud rate UART's port or externally managed.
```
tridge, line 22: flow control would only be needed if you were streaming quite a lot of data. The common uses for passthru don't involve a large amount of data, and easily the most common case it one end of the link is a USB port where flow control isn't configurable.

tridge:
I think it is worth mentioning that on F7 and H7 boards you can use the 2nd USB endpoint (which will be the last SERIALn_PROTOCOL). That allows you to keep using the GCS with mavlink on main USB while doing pass-thru on the 2nd USB endpoint to a device. It makes for a really useful way of configuring a device

common/source/docs/common-serial-passthrough.rst
Hwurzburg: removed flow control

Hwurzburg:
added ntoes about using second virtual USB port

tridge (approved):

PR #3181, opened 2020-11-29
Changing the URI of dronekit-python

muramura:
When I click on the title of dronekit-python, it will take me to dronekit-android.
I change the URI.


Hwurzburg:
LGTM...merging...thanks

PR #3183, opened 2020-11-29
update install-prereqs-ubuntu.sh -y directory

Kametor:
install-prereqs-ubuntu.sh -y script directory was changed but not edited in wiki

### DIFF
--- a/dev/source/docs/building-setup-linux.rst
+++ b/dev/source/docs/building-setup-linux.rst
@@ -39,10 +39,10 @@ Clone ArduPilot repository
 Install some required packages
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
this is not correct.....  this link is still correct https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-ubuntu.sh

Kametor:
> this is not correct..... this link is still correct https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-ubuntu.sh

yes you are right, my PR was conflicted.

PR #3186, opened 2020-12-02
Update common-durandal-overview.rst

vincentpoont2:
- Removing "Beta Hardware", there is no beta hardware, only beta Ardupilot firmware. Since Durandal is running Stable firmware now,  removing this will clear some confusion. 
- Updating the "Order from" link to the direct product purchase link.


Hwurzburg:
thanks...merging

PR #3191, opened 2020-12-04
Add notes to the Replay set up and use

Hwurzburg:


dev/source/docs/testing-with-replay.rst
```diff
@@ -8,24 +8,22 @@ Introduction
...
 Replay is a program that takes a dataflash log file and replays it
-through the latest master code allowing for state estimation issues to
-be analysed with new code.
+through any branch's code allowing for state estimation issues to
+be analyzed using that branch's code instead of the code used while generating the dataflash log.
 
-It is recommended that if a problem is reproducible that the log be
-generated with both the `LOG_REPLAY` and `LOG_DISARMED` parameters set
-to 1.
+.. note:: In order to use Replay the log must be generated with both the `LOG_REPLAY` and `LOG_DISARMED` parameters set to 1.
```
peterbarker, line 14: This is not true; it is just recommended.

Logs do not Replay as well if those parameters are set, but tridge did try to make that work.

dev/source/docs/testing-with-replay.rst
```diff
@@ -8,24 +8,22 @@ Introduction
 ============
 
 Replay is a program that takes a dataflash log file and replays it
-through the latest master code allowing for state estimation issues to
-be analysed with new code.
+through any branch's code allowing for state estimation issues to
+be analyzed using that branch's code instead of the code used while generating the dataflash log.
```
peterbarker, line 12: A note here saying that the logging structures of the two branches must be the same belongs here....

peterbarker (changes requested):

dev/source/docs/testing-with-replay.rst
Hwurzburg: if LOG_REPLAY is not enabled, you will not get the replay messages, correct? then what purpose does replaying a log serve?
and what is the impact of not logging prior to arm?
Tridge and Paul had told me I had to have those set to do a proper analysis of an EKF issue...

dev/source/docs/testing-with-replay.rst
Hwurzburg: will add

peterbarker:
On Fri, 4 Dec 2020, Henry Wurzburg wrote:

> -generated with both the `LOG_REPLAY` and `LOG_DISARMED` parameters set
> -to 1.
> +.. note:: In order to use Replay the log must be generated with both the `LOG_REPLAY` and `LOG_DISARMED` parameters set to 1.
> 
> if LOG_REPLAY is not enabled, you will not get the replay messages, correct? then what purpose does replaying a log serve?

Sorry.  LOG_REPLAY must be set LOG_DISARMED does not - we just get a much, 
much better log if LOG_DISARMED is set.

dev/source/docs/testing-with-replay.rst
Hwurzburg: changed to require replay, preferred to also have disarmed set also

Hwurzburg:
should be good to go now..

peterbarker (approved):
Thanks Henry.

PR #3192, opened 2020-12-04
Update learning-ardupilot-storage-and-eeprom-management.rst

Kametor:
This link gives 404 error, the directory was changed but not edited in wiki i think

### DIFF
--- a/dev/source/docs/learning-ardupilot-storage-and-eeprom-management.rst
+++ b/dev/source/docs/learning-ardupilot-storage-and-eeprom-management.rst
@@ -164,7 +164,7 @@ set of FMT messages which have a well known format and which describe
 the format of the messages that follow.
 
 Go and have a look at
-`libraries/DataFlash/examples/DataFlash_test/DataFlash_test.cpp <https://github.com/ArduPilot/ardupilot/blob/master/libraries/DataFlash/examples/DataFlash_test/DataFlash_test.cpp>`__.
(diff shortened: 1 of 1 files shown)


Hwurzburg:
@peterbarker Peter, could you look at this? while the link is wrong, the entire description I believe is now inaccurate also...does not seem to match the code example...who should revise this?

peterbarker:
@Kametor @Hwurzburg I'm afraid the new link is wrong, too.

DataFlash was the old name for `AP_Logger` - we retained the name `DataFlash` well past the time it was correct as for a long time we only supported writing files to POSIX-style filesystems or over mavlink.  Now we *can* write to dataflash again - but that's relatively new....

So, sadly, it's probably my job to fix any innaccuracies.

The correct link would be to `libraries/AP_Logger/examples/AP_Logger_test/AP_Logger_test.cpp`


However, the entire page probably needs splitting up; the logging is no longer intrinsically linked with the Storage layer (which is basically why "eeprom" in there is talking about....

Kametor:
@peterbarker  so should I close this PR? or close after changing the link with you suggested?

Hwurzburg:
@Kametor leave open... I will occasionally ~~annoy~~ remind Peter to update the page based on this PR being open

Hwurzburg:
@peterbarker consider yourself ping'd ;-)

peterbarker:
@Kametor @Hwurzburg Let's just correct the link and commit it....

Hwurzburg:
okay...merging...thanks

PR #3202, opened 2020-12-11
Insert REGISTRED MARK symbol for ARM Cortex in pix32v5

MasterBoBo:
aaa

### DIFF
--- a/common/source/docs/common-holybro-pix32v5.rst
+++ b/common/source/docs/common-holybro-pix32v5.rst
@@ -24,8 +24,8 @@ Specifications
 ==============
 
 -  **Processors**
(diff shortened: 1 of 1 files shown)


Hwurzburg:
This is incorrect and will not display properly....and unfortunately its only the tip of the iceberg....there are dozens of pages in the wiki without the symbol...help in updating them would be appreciated...
see here for proper way to add the symbol: https://docs.anaconda.com/restructuredtext/detailed/#smart-quotes-dashes-and-unicode

Hwurzburg:
Sphnix update allowed the html insert, and Bruno did it and I merged it, so this is not necessary any longer..closing

PR #3204, opened 2020-12-12
Clarify F.Port setup steps required on OpenTX side

alteman:
I spent a lot of time confused why the sensors don't show up in OpenTX even though they worked in s.port mode.. Added a blurb to hopefully clarify the point that Yaapu script has to be enabled for the sensors to appear.

### DIFF
--- a/common/source/docs/common-FPort-receivers.rst
+++ b/common/source/docs/common-FPort-receivers.rst
@@ -18,7 +18,7 @@ Most F7 and H7 autopilots, on the other hand, have internal inverters, and can b
 Telemetry protocol
 ==================
 
-Only the :ref:`Passthrough telemetry <common-frsky-passthrough>` protocol is supported. To get sensors to show up in OpenTX you have to install and enable :ref:`Yaapu FrSky Telemetry Script for OpenTX<common-frsky-yaapu>`. Otherwise the only sensor discovered would be GPS.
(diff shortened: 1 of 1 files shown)


common/source/docs/common-FPort-receivers.rst
```diff
@@ -15,6 +15,11 @@ Most F7 and H7 autopilots, on the other hand, have internal inverters, and can b
 
 .. note:: ArduPilot supports both 16 and 24 channels FPort2 but telemetry is enabled only for 16 channels mode.
 
+Telemetry protocol
+==================
+
+Only the :ref:`Passthrough telemetry <common-frsky-passthrough>` protocol is supported. To get sensors to show up in OpenTX you have to install and enable :ref:`Yaapu FrSky Telemetry Script for OpenTX<common-frsky-yaapu>`. Otherwise the only sensor discovered would be GPS.
```
Hwurzburg, line 21: Not quite accurate
```suggestion
ArduPilot will send telemetry data to the receiver over Fport using the :ref:`Passthrough telemetry <common-frsky-passthrough>` protocol. To get sensors to show up in OpenTX you have to install and enabled a script which can display Passthrough telemetry, such as  :ref:`Yaapu FrSky Telemetry Script for OpenTX<common-frsky-yaapu>`. Otherwise the only sensor that will be discovered would be GPS.
```

Hwurzburg (changes requested):

alteman:
Thanks for clarification!

Hwurzburg:
LGTM...thanks!

PR #3208, opened 2020-12-15
dev: design decisions: add example for bitfields, options for std:: f…

peterbarker:
…unctions

### DIFF
--- a/dev/source/docs/style-guide.rst
+++ b/dev/source/docs/style-guide.rst
@@ -521,6 +521,27 @@ Bit fields are generally frowned upon
 
 Using bit fields reduces RAM usage but can considerably increase flash usage, as to extract a boolean truth value from a bit field requires more machine instructions.  If the variable is frequently accessed then this can be a LOT of flash.
 
+**Not preferred:**
+
+::
+
+   class Foo() {
+   private:
+       bool should_fly  : 1;
+       bool should_grow : 1;
+   };
+
+**Preferred:**
+
+::
+
+   class Foo() {
+   private:
+       bool should_fly;
+       bool should_grow;
+   };
+
+
 Initialise member variables in header files rather than in constructors
 -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dev/source/docs/style-guide.rst
```diff
@@ -521,6 +521,27 @@ Bit fields are generally frowned upon
...
 Using bit fields reduces RAM usage but can considerably increase flash usage, as to extract a boolean truth value from a bit field requires more machine instructions.  If the variable is frequently accessed then this can be a LOT of flash.
 
+**Not preferred:**
+
+::
+
+   class Foo() {
+   private:
+       uint8_t should_fly  : 1;
+       uint8_t should_grow : 1;
```
khancyr, line 531: humm, I not sure that is clearer. In the text you are speaking of bool and here you are using uint8_t . Isn't using bitfields on uint8_t for packing a structure a right thing ?

dev/source/docs/style-guide.rst
```diff
@@ -558,3 +579,23 @@ For efficiency reasons, ArduPilot doesn't use the C standard library (``std::``)
 This means no ``std::vector`` and no ``std::string``, for example.
 
 If you really want to go that way, including the header to get these isn't sufficient - you will need to fiddle with the build system to link ``std`` in.
+
+Alternatives to ``std::vector``
```
khancyr, line 583: maybe we should say that we try to avoid dynamic allocation like on vector and string

dev/source/docs/style-guide.rst
peterbarker: Changed, thanks.

dev/source/docs/style-guide.rst
peterbarker: Added some more words.

Hwurzburg:
tested..merging since review comments addressed

PR #3218, opened 2020-12-17
Notes about submodule init and VS version

EosBandi:


meee1:
@Hwurzburg looks good to me

Hwurzburg:
merging

PR #3223, opened 2020-12-18
Add note re F9P constellation setup

Hwurzburg:


khancyr (approved):

rmackay9:
txs for this.. it's a bit out of place in some ways because we don't really have notes on specific GPSs on the landing page.  Moving it down below the "U-Blox F9P Firmware Update Procedure" link might be better.. but.. I don't mind too much if this is merged.

Hwurzburg:
I think that it might be missed easily if lower....alternatively, I could add to the F9P gps individually, unfortunately since the mRo device is by reference to their page, it would not be seen for that GPS...I can do it either way...

rmackay9:
I think on the individual F9 pages might be better.. as you say we don't have our own pages for some but where we do have them, that might be best

Hwurzburg:
moved note...merging

PR #3228, opened 2020-12-23
Plane: Falisafe TYPE=X REASON=X enum references

Naterater:
We can point users here to help with their diagnosis. Random numbers mean nothing to users, and I can't find this table anywhere else. I hope I've pulled it correctly from ModeResason.h and defines.h

### DIFF
--- a/plane/source/docs/apms-failsafe-function.rst
+++ b/plane/source/docs/apms-failsafe-function.rst
@@ -302,6 +302,95 @@ There are three possible enabled settings. Seeing :ref:`FS_GCS_ENABL<FS_GCS_ENAB
    </tbody>
    </table>
 
+Failsafe Diagnosis in Logs or GCS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


plane/source/docs/apms-failsafe-function.rst
```diff
@@ -302,6 +302,95 @@ There are three possible enabled settings. Seeing :ref:`FS_GCS_ENABL<FS_GCS_ENAB
    </tbody>
    </table>
 
+Failsafe Diagnosis in Logs or GCS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GCSs will often display text indicating the type of failsafe encountered. Refer to this table if you failsafe displays as a TYPE=X, REASON=X
```
Hwurzburg, line 308:
```suggestion
GCSs will often display text indicating the type of failsafe encountered, such as "Failsafe Short event on: type=1/reason=3". Type and Reason can be determined using the table below:
```

Hwurzburg (changes requested):

Naterater:
@Hwurzburg, please do anything you want to it so long that the general idea gets into the wiki somewhere.

PR #3229, opened 2020-12-23
telemetry-xbee:Add parameter list

konishi-junichi:
Added setting parameters that have been used in xbee(S2C).


tridge (approved):

tridge:
i can't really validate this as I don't have any xbees, but seems reasonable

PR #3231, opened 2020-12-25
Update quadplane-tips.rst

kd0aij:
add note on Q_OPTIONS bit for Disarmed Yaw Tilt


Hwurzburg:
thanks...moved to future wiki page since this is master only and not stable....merging

PR #3242, opened 2020-12-31
Add a link to the LightWare SF45

muramura:
I was interested in SF45.
I didn't see a link to it on the rangefinder list.
I will add the SF45 to the list.


Hwurzburg:
that is because its in the master branch only and its not supposed to be in the main wiki yet....see https://ardupilot.org/plane/docs/common-master-features.html
if you put SF45 in as a search term in the wiki it will be found....

closing this

PR #3243, opened 2020-12-31
Add a submarine link

muramura:
There was no link to the submarine.
I will add the submarine firmware link.


Hwurzburg:
thanks..merging

PR #3252, opened 2021-01-04
Add Generator and Intelligent Energy fuel cell pages

Gone4Dirt:
This PR adds the setup instructions for the intelligent energy fuel cells.

I have also added a generator page and moved Richenpower under that page.  This organises the wiki as per the generator library in AP.


rmackay9:
@Gone4Dirt, 

This looks OK but it would be good to include:

- where to buy
- diagram of how to physically connect the generator to the autopilot

common/source/docs/common-generators.rst
```diff
@@ -0,0 +1,23 @@
+.. _common-generators:
+
+==========
+Generators
+==========
+
+.. figure:: ../../../images/generators.png
+   :target: ../_images/generators.png
+
+Support for generators is currently available in master.  Within ArduPilot the term generator is used to broadly group all devices that convert some form or fuel into electrical power.  Hence, the list of currently supported generators includes hydrogen fuel cells.
```
rmackay9, line 10: "or" should be "of" I think.

No big deal but it is good to be concise so we could drop, "Hence, the list of currently supported generators includes hydrogen fuel cells."

I'd probably replace, "Support for generators is currently available in master" with "ArduPilot 4.1 (and higher) support generators which can be used to extend the flight time".

common/source/docs/common-ie24-fuelcell.rst
```diff
@@ -0,0 +1,204 @@
...
+- Fuel cell state.
+- Error codes.
+- Total power out (log only).
+- Stack Power Module (SPM) power (log only).
+
+
+Setup
+-----
+
+Enable the fuel cell driver by setting :ref:`GEN_TYPE <GEN_TYPE>` to 2.
```
rmackay9, line 26: I think it would be better to just provide a bullet point list of parameter and what they should be set to like we've done on the richen power page. https://ardupilot.org/copter/docs/common-richenpower-generator.html

Sorry to be annoying but the reader just wants to get in and out as quickly as possible so let's not hide the parameters that need changing within long sentences.

common/source/docs/common-ie24-fuelcell.rst
```diff
@@ -0,0 +1,204 @@
...
+
+As previously mentioned, normal voltage failsafes can be used on this fuel cell model.  Hence, as a final belt-and-braces approach a critical voltage failsafe has been set on the electrical generator monitor instance to initiate an immediate landing if voltage gets to 3.3 V per cell:
+
+- :ref:`BATT_CRT_VOLT <BATT_CRT_VOLT>` = 39.6
+
+
+Driver Not Healthy
+------------------
+
+If you see the `Generator: Driver Not Healthy` message in the GCS then ArduPilot has not received any data packets from the fuel cell for at least 5 seconds.  Check the following for common causes to this issue:
```
rmackay9, line 183: as a side note, I think we should drop "Driver" from the error message.  For other sensors (like optical flow, etc) we just say the sensor is unhealthy not the driver that consumes the sensor's data.

rmackay9:
Great to see this wiki page!

common/source/docs/common-ie24-fuelcell.rst
Gone4Dirt: I disagree.  Why give users less information?  I acknowledge a balance has to be struck so as to not make the page overly verbose.  However, if sufficient information is provided then users will be better equipped to work out the specifics of how to configure their setup, if it is different from the example given.  Given the way that the fuel cells interact with the battery monitor library is a little odd/unconventional, I think the information provided is needed.  Plus, all parameter names are highlighted in blue so can be easily picked out if users don't want to read all of the info.

common/source/docs/common-generators.rst
Gone4Dirt: Fixed.

common/source/docs/common-ie24-fuelcell.rst
Gone4Dirt: Fixed here.  Will PR AP code shortly.

Gone4Dirt:
Thanks for the review @rmackay9.  I have made most of the suggested changes.  I just disagree on the reducing the instructions from what's there.

rmackay9:
Merged, thanks!  Re removing info, I was more suggesting that we put the parameter names at the beginning of the sentence instead of in the middle.  it's not a big deal though and it's your baby.

Hwurzburg:
Since this is not in a stable release, it should be reformatted and moved into Upcoming Features...@rmakay9 let's discuss tomorrow AM

Gone4Dirt:
@Hwurzburg, I don't understand.  It is in the master features section.  Is there somewhere else it should have been added?

Hwurzburg:
@Gone4Dirt @rmackay9 Sorry, long night...3AM brain glitch...the unnecessary 4.1 reference got me going down a rabbit hole....will fix that tomorrow..need sleep

PR #3258, opened 2021-01-06
Cleanup of out of date firmware references

Hwurzburg:
Start of wiki cleanup of outdated firmware reference. Per previous agreements only references to last two releases in a vehicle will be retained. Archived topics are exempt. And where large pieces of information would be removed, they will be archived in some manner instead of deleted.

### DIFF
--- a/copter/source/docs/gps-failsafe-glitch-protection.rst
+++ b/copter/source/docs/gps-failsafe-glitch-protection.rst
@@ -20,7 +20,7 @@ A *GPS failsafe* event will occur if GPS 3D
 lock or the position "Glitches" for at least 5 seconds while Copter is
 in a mode that requires the GPS (RTL, Auto, Loiter, Circle, Position,
 Guided or Drift).  The GPS failsafe response can be set to Land or
-switch to AltHold mode (in AC3.1 and higher) so that you can retake
+switch to AltHold mode so that you can retake
 manual control.
 
 Without GPS updates, the inertial sensors allow approximately 10 seconds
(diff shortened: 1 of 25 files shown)


copter/source/docs/gps-failsafe-glitch-protection.rst
```diff
@@ -20,7 +20,7 @@ A *GPS failsafe* event will occur if GPS 3D
 lock or the position "Glitches" for at least 5 seconds while Copter is
 in a mode that requires the GPS (RTL, Auto, Loiter, Circle, Position,
 Guided or Drift).  The GPS failsafe response can be set to Land or
-switch to AltHold mode (in AC3.1 and higher) so that you can retake
+switch to AltHold mode so that you can retake
```
rmackay9, line 23: I think this page is perhaps not linked from anywhere anymore.  Not a problem to update it but I don't see how users can get to it.  It should probably be archived if it's not already.

rmackay9:
Merged, thanks!

copter/source/docs/gps-failsafe-glitch-protection.rst
Hwurzburg: it is....from EKF failsafe....and an unlinked page generates a build error...
are you sure that the DCM and EKF GPS glitch protection talked about on this page is no longer used?

rmackay9: This page (https://ardupilot.org/copter/docs/gps-failsafe-glitch-protection.html) is linked from the bottom of the EKF failsafe page but it was for the gps glitch protection that built into copter and the EKF1 (which we don't use anymore). It's possible the EKF1 section could be moved to the EKF failsafe page and updated to apply to the EKF2/EKF3.  It seems that:

- EKF_GLITCH_RAD has become EK2_GLITCH_RAD and EK3_GLITCH_RAD
- EKF_POS_GATE has become EK2_POS_I_GATE, EK3_POS_I_GATE
- GPSGLITCH_ACCEL doesn't exist

PR #3259, opened 2021-01-06
removed outdated firmware revs from tricopter

Hwurzburg:
there was info that needed to be archived with this pages cleanup so created an archive page for it


rmackay9:
Merged, thanks!

PR #3260, opened 2021-01-06
removed references to old firmware revs

Hwurzburg:
focusing on Plane now

### DIFF
--- a/common/source/docs/common-reach-rtk-receiver.rst
+++ b/common/source/docs/common-reach-rtk-receiver.rst
@@ -9,9 +9,6 @@ This article provides a brief overview of Reach RTK GPS receiver.
 .. image:: ../../../images/Reach.jpg
 	:target: ../_images/Reach.jpg
 
-.. note::
-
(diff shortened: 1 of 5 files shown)


brunoolivieri:
I do understand that makes the text easier to maintain and read. However, the old versions will be still available those versions on the firmware? Maybe it would be a good idea to set the floor version as a modern version as v4.x.x.

Hwurzburg:
> I do understand that makes the text easier to maintain and read. However, the old versions will be still available those versions on the firmware? Maybe it would be a good idea to set the floor version as a modern version as v4.x.x.

ALL versions of older released firmware are available (via MP)....but we decided last year to only carry warnings on when new features/operations occur for the last two firmware stable revs.. and I am archiving any significant info blocks for older firmware ..otherwise clutter accumulates and makes reading difficult...

rmackay9 (approved):
LGTM at least for the non-plane pages

tridge (approved):

tridge:
good idea, thanks Henry!

PR #3013, opened 2020-09-09
Soaring updates

samuelctabor:
Update for new soaring behaviours described [here](https://discuss.ardupilot.org/t/soaring-gliding-for-planes/15323/250?u=samuel_tabor) based on top of [existing PR for soaring RC option](https://github.com/ArduPilot/ardupilot_wiki/pull/2914/files).

PR #3262, opened 2021-01-07
Soaring: Improve MAX_RADIUS doc and remove unnecessary warning.

samuelctabor:
@Hwurzburg I haven't been able to reproduce the behaviour described in the warning - this removes it and fixes the description to match the behaviour.

### DIFF
--- a/plane/source/docs/soaring.rst
+++ b/plane/source/docs/soaring.rst
@@ -203,10 +203,6 @@ Limit maximum distance from home
 
 If using FBWB or CRUISE mode, the parameter :ref:`SOAR_MAX_RADIUS<SOAR_MAX_RADIUS>` can be used to trigger RTL if the aircraft is more than this distance from home when thermalling while in LOITER mode. Note that this parameter won't stop the aircraft from exceeding this distance before it enters LOITER mode.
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
Is this with master or Stable? and if different, something changed right?

samuelctabor:
> Is this with master or Stable? and if different, something changed right?

Master - SOAR_MAX_RADIUS doesn't exist in stable.

Hwurzburg:
okay, so lets put this in #3013 and close this one,okay?

samuelctabor:
I'd rather just merge this now - no point having information that isn't correct for any version.

A separate discussion is whether we should move all the features available only in master out of the live wiki and into https://github.com/ArduPilot/ardupilot_wiki/pull/3013 - interested in your thoughts on that too.

Hwurzburg:
Merging.. Sorry,I had forgotten that SOARING was actually in a Stable release since there have been so many improvements....Since it will be months, apparently, before a 4.1 plane beta, its probably better to not have mixed messages...its not life or death, but if its not too complicated to move the master only stuff into 3013 that would be more consistent...if you are willing

PR #3267, opened 2021-01-09
Update path for CMAC-circuit.txt

nkruzan:
path is no longer valid


Hwurzburg:
LGTM...thanks! merging

PR #3269, opened 2021-01-10

vincentpoont2:
Add Holybro Nano Ublox M8 5883 GPS Module


Hwurzburg:
very cool Vincent! I have a small TVBS project that can really use something like this instead of a BN-220 and external compass to avoid the weight of a BN-880....merged

PR #3270, opened 2021-01-10
rover: sailboat: add simulator demo vid

IamPete1:
This adds a link to this excellent demo vid on the sailboat landing page.

https://youtu.be/kzGciG7cCBc

### DIFF
--- a/rover/source/docs/sailboat-home.rst
+++ b/rover/source/docs/sailboat-home.rst
@@ -4,15 +4,27 @@
 Sailing Vehicles Home
 ======================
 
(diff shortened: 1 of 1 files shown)


rmackay9:
Merged, thanks!

By the way, the spelling change of familiarised to familiarized is just an American vs British spelling difference and since the number of dev team members from Canada, Australia, Wales, etc outnumbers those from the US we normally still with the British spelling.

IamPete1:
> By the way, the spelling change of familiarised to familiarized is just a British vs American spelling difference and since the number of dev team members from Canada, Australia, Wales, etc outnumbers those from the US we normally stick with the British spelling.

I should probably work out how to set my VSCode spell check to British English.....

PR #3282, opened 2021-01-20
added IMU temperature calibration documentation

tridge:
the corresponding PR should be merged soon:
https://github.com/ArduPilot/ardupilot/pull/16262


common/source/docs/common-imutempcal.rst
```diff
@@ -0,0 +1,219 @@
+.. _common-magfit:
```
Hwurzburg, line 1: needs to match the file name

Hwurzburg (changes requested):

Hwurzburg:
suggest a title Online Temperature Calibration since you have an Offline one...

Hwurzburg:
builds okay but params are unlinked since the code has not merged.....will need to add this to the Upcoming features page when we are ready to merge to get the TOC entry and it belongs there since not in a release yet....

common/source/docs/common-imutempcal.rst
tridge: fixed

tridge:
> suggest a title Online Temperature Calibration since you have an Offline one...

the doc covers both

tridge:
> builds okay but params are unlinked since the code has not merged.....will need to add this to the Upcoming features page when we are ready to merge to get the TOC entry and it belongs there since not in a release yet....

yep, I didn't link it at all yet. We can do that when ardupilot PR merged

Hwurzburg:
> > suggest a title Online Temperature Calibration since you have an Offline one...
> 
> the doc covers both

so you can start right at the section titled Offline.... without the predecessor? may be just editorial fancy, but normally I would expect a flow like (not to suggest titles, just content partitioning): 

-common setup
-Online cal
-Offline cal
-Factory cal

we start by stating two methods.... by only one is subtitled...

tridge:
> we start by stating two methods.... but only one is subtitled...

that is because the process for doing an offline cal starts by doing an online cal. Happy to change it if you think it will work better of course

tridge:
PR is now merged, so we can add this into the TOC

Hwurzburg:
merged vai another branch

PR #3288, opened 2021-01-22
Update common-understanding-altitude.rst

emin1461:
- ASL: Altitude above (mean) Sea Level - the altitude of the vehicle expressed as distance above the mean seal level of the world.
seal level changed . sea level


Hwurzburg:
thanks...merged

PR #3296, opened 2021-01-24
RTF Update

auturgy:
Hopefully there'll be a few of these over the coming days.


Hwurzburg:
tested, merged..thanks

PR #3307, opened 2021-01-29
Add a page on how to setup SITL on Mac OSX

spark404:
I managed to get SITL running on my Mac. The instructions might be useful for other folks try to do the same.


Hwurzburg:
@andyp1per are you still the one to review this?

Hwurzburg:
@andyp1per bump!
@stephendade are you in a position to review this?

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+::
+
+   cd ardupilot/ArduCopter
+
+The sim_vehicle script checks several environment variables to determine the OS it's running on. For Mac OSX it
+requires that DISPLAY is set.
+
+::
+
+  export DISPLAY=:0.0
```
andyp1per, line 87: Don't you need the macOS X environment installed for this to work?

stephendade:
@Hwurzburg - I haven't got a Mac, so unable to review.

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+Preparing Python
+----------------
+
+Install the required python 3 version using homebrew and activate it. If you are familiar with pyenv or virtualenv it might be better to use one of those to setup the environment. This limits potential problems with other tools that require specific versions of python. How to do that is beyond the scope of this document.
+
+The following steps install python version 3.9.x for system wide usage.
+
+::
+
+	brew install python@3.9
```
timtuxworth, line 47: I found that in order to get ArduPilot to compile using waf on an M1Pro Mac running OSX 12.1 Monterey I had to install Python 3.10 from python.org.

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+  export PATH="$(brew --prefix)/opt/python/libexec/bin":$PATH
+	
+	pip install wheel
```
timtuxworth, line 50: This command may require sudo. Best practice to install for the local user is to use the --user option to install for the user only.

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+	pip install future pymavlink mavproxy
```
timtuxworth, line 51: Using Python 3.10 from Python.org for Mac M1 ARM version, this will fail with the following errors:
wx/svg/_nanosvg.c:20946:19: error: implicit declaration of function '_PyGen_Send' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                ret = _PyGen_Send((PyGenObject*)yf, value == Py_None ? NULL : value);
                      ^
    wx/svg/_nanosvg.c:20946:17: warning: incompatible integer to pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from 'int' [-Wint-conversion]
                ret = _PyGen_Send((PyGenObject*)yf, value == Py_None ? NULL : value);
                    ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    wx/svg/_nanosvg.c:20951:19: error: implicit declaration of function '_PyGen_Send' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                ret = _PyGen_Send((PyGenObject*)yf, value == Py_None ? NULL : value);
                      ^
    wx/svg/_nanosvg.c:20951:17: warning: incompatible integer to pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from 'int' [-Wint-conversion]
                ret = _PyGen_Send((PyGenObject*)yf, value == Py_None ? NULL : value);
                    ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    wx/svg/_nanosvg.c:21035:19: error: implicit declaration of function '_PyGen_Send' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                ret = _PyGen_Send((PyGenObject*)yf, NULL);
                      ^
    wx/svg/_nanosvg.c:21035:17: warning: incompatible integer to pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from 'int' [-Wint-conversion]
                ret = _PyGen_Send((PyGenObject*)yf, NULL);

timtuxworth: I'm not totally sure what these lines do, but it seems that there is some code in _nanosvg.c that needs to be fixed.

timtuxworth (commented):
I tried following these instructions on my new 2021 MacPro M1Pro running 12.1 Monterey. This is what I found.

dev/source/docs/setting-up-sitl-on-macosx.rst
srmainwaring: Wondering if we need two sub-guides for macOS - one for Intel based machines and one for M1 versions. The default install locations for brew are different for each architecture and I suspect we'll keep seeing more differences as more functionality is documented. For example the instructions for adding other simulators like Ignition will be specialised for each architecture.

dev/source/docs/setting-up-sitl-on-macosx.rst
srmainwaring: I've not needed to set the DISPLAY when running SITL (no need for an X server that I'm aware of?)

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Optional: Add the following to the either .bashrc or .zshrc to enable this version of python be default
+
+::
+
+  export PATH="$(brew --prefix)/opt/python/libexec/bin":$PATH
+
+
```
srmainwaring, line 59: It's also useful to add 

```bash
################################################################################
# ARDUPILOT
#
export PATH=$PATH:$HOME/Code/ardupilot/ardupilot/Tools/autotest
export ARDUPILOT_ROOT=$HOME/Code/ardupilot/ardupilot
```

to the shell initialisation scripts while you're at it. I put customisations including the PATH modification for Python in `~/.zprofile`.

dev/source/docs/setting-up-sitl-on-macosx.rst
```diff
@@ -0,0 +1,137 @@
...
+  brew install sfml
+
+Note that not all sfml features work on Mac OSX currently. The following error can appear `Cannot create a window from a worker thread. (OS X limitation)`, this is a result of Mac OSX specifics that haven't been catered for in the sitl code yet.
+
+Start SITL simulator
+--------------------
+
+.. warning::
+
+	Due to ideosyncracies with Python and using screens there are limitations when Python needs to use the screen to display a graphical UI. Python 3.9 solves part of the problem by no longer requiring a separate pythonw binary. However the requirement to run on the main display still stands. When observing the error "This program needs access to the screen. Please run with a Framework build of python, and only when you are logged in on the main display of your Mac.", disconnect external displays during startup. When SITL is started the windows can be moved around as normal.
```
srmainwaring, line 73: sp.

```diff
- ideosyncracies
+ idiosyncrasies
```

timtuxworth: I haven't been able to get this to work at all on my M1 Mac. So I just run SiTL with no UI. It works fine and I connect to with with QGroundControl which shows the map and all the messages.
Maybe the message should say the UI is optional?

srmainwaring (commented):
@spark404 nice one, thanks for expanding on the docs for macOS. My main thoughts (also commented in the review details) are that we may need to have custom install instructions for the newer M1 based machines, as Apple have changed the permitted locations for where user libraries can be placed. This combined with the sanitising of `DYLD_LIBRARY_PATH` means that software developed on linux that should nominally run fine on macOS can be tricky to set up unless care is taken to ensure everything winds up in the right place - you can't fix it with library path remapping now unless you disable SIP (which I prefer not to).

dev/source/docs/setting-up-sitl-on-macosx.rst
srmainwaring: @timtuxworth unfortunately I don't have a M1 to test at present - still Intel based machines for me. When I get round to upgrading my macbook I'll take another look at the UI modules in MAVProxy which seem to be causing the problems. As you suggest it might be worth emphasising the difference between a bare bones SITL install (with alternative GCS) and full MAVProxy support for macOS which should have all the UI elements functioning correctly.

dev/source/docs/setting-up-sitl-on-macosx.rst
timtuxworth: This mayncause problems if you run more than one clone in different directories. I often do that.

dev/source/docs/setting-up-sitl-on-macosx.rst
ntamas: I can compile ArduPilot just fine with Python 3.9 on an M1-based MacBook Pro so I don't think you'll need Python 3.10. I'm using Python 3.9 from Homebrew.

dev/source/docs/setting-up-sitl-on-macosx.rst
ntamas: Not sure how much we want to dive into this in the instructions, but in my experience the best is to isolate ArduPilot-related Python dependencies in a virtualenv; what I do is as follows:

```shell
python3 -m venv .venv
.venv/bin/pip install -U pip wheel future pymavlink mavproxy
```

and then I run everything from the virtualenv. This way I don't mess up the Homebrew-installed Python with random dependencies that are needed for ArduPilot only.

dev/source/docs/setting-up-sitl-on-macosx.rst
ntamas: Indeed; it seems like most of the difficulties that arise in installing SITL stuff on an M1 Mac are due to the GUI stuff. I run the SITL without a GUI and I haven't encountered any significant issues.

dev/source/docs/setting-up-sitl-on-macosx.rst
timtuxworth: I'm pretty sure it would help to put this in the instructions. Setting up a workable M1 Mac dev environment should be easy and without clear instructions it's definitely not. When you say "run everything from the virtualenv" - you mean mavproxy.py or waf or both? 

What I find now is that waf needs a different python environment to mavproxy, but sim_vehicle calls both waf and mavproxy so I can't get sim_vehicle->waf to use a different python from sim_vehicle->maxproxy.

For now in order to get sim_vehicle to work, I just run it 'headless', which seems like what @srmainwaring  and others are doing but I shouldn't have to do that.

ntamas: I have to admit I don't use `mavproxy` in the virtualenv because I'm using my own tools to talk to the simulator. `mavproxy` is problematic because it insists on bringing the entire wxPython distribution with it and I don't need that. wxPython currently provides pre-compiled wheels for Intel Macs only on PyPI, and only for Python 3.9, so until wxPython publishes newer wheels one would have to compile wxPython from source.

So, ultimately my goal was to get the build process running and to be able to run most of the non-GUI scripts from the `Tools` folder (e.g., the one that builds the bootloaders). For that, a virtualenv is fine, the only catch is that some of the scripts invoke other Python scripts or `waf` in a way that it does not respect the current Python interpreter setting (like [this one](https://github.com/ArduPilot/ardupilot/pull/20420) ) and just fall back to whatever Python is at the front of the system PATH. Activating the virtualenv from the shell with `source .venv/bin/activate` does the trick in 99% of the cases, for instance it ensures that `waf` picks up the correct (virtualenv-based) Python instead of the system Python.

_Edit_: I tried the wxPython snapshot builds from [here](https://wxpython.org/Phoenix/snapshot-builds/); it seems like they have `universal2` wheels for Python 3.9 there so there would be no need to compile wxPython from source, but it did not work for me so I just gave up on GUI stuff.

PR #3310, opened 2021-01-31
Update common-rtf.rst with Holybro Multicopter Kits

vincentpoont2:
Add Holybro S500 V2 kits and Holybro X500 kits


rmackay9:
@vincentpoont2,

Thanks very much for this PR.  My understanding though is that this frame doesn't ship with ArduPilot installed on the autopilot instead it is some other firmware.  If Holybro sells a version that ships with AP we are very happy to include it on the RTF page.  If you need help getting a tuned up version of the frame, if you send me one I'll make up a parameter file and make it easy to select it from MP.  I did something similar for another manufacturer recently.  https://github.com/ArduPilot/ardupilot/pull/16263.

Feel free to email me at rmackay9 at yahoo.com, PM me on discord or skype (id: rmackay9).

Thanks again and sorry that we can't immediately merge this.

PR #3313, opened 2021-02-02
Add RC_PROTOCOL param explanation

Hwurzburg:


common/source/docs/common-future-wiki-changes.rst
```diff
@@ -348,6 +348,12 @@ Add to table:
 7                                       Allow Aux Switches to honor the ``RCx_REVERSED`` parameter
 =================================       =========
 
+
+Add section at end:
+
+
+There is also an :ref:`RC_PROTOCOLS<RC_PROTOCOLS>` bitmask that can be used to restrict which RC protocols are detected and used. This is useful in cases where operation at the fringes of RC reception is desired but a radio control failsafe (see individual vehicle setup sections) will not be used. Such as when a long range ground control station is being employed and it is desired that loss of RC signal not interrupt operation. In these situations, it is possible for the autopilot to lose RC signal and then incorrectly re-detect it as an incorrect protocol during partial signal reception, delivering erroneous control inputs. Since one usually knows the correct protocol being used in a particular setup, this bit mask can be used to restrict detection to that protocol only and avoid this problem.
```
tridge, line 355: i think this explanation is a bit too obtuse.
How about this:
 This is useful in cases where the RC protocols autodetection fails and an incorrect RC protocol handler is chosen. This is rare, but if you do find it happens then you can lock in a single RC protocol to detect using this parameter.

Hwurzburg: will change...thanks

PR #3328, opened 2021-02-13
Update common-rc-options.rst

mattdholmes00:
I believe there was a simple numerical error.  "setting bit 2" is a value of 4, not 8.  It might be even better to make the example with bit 3, then 2^3=8 has three un-repeated numbers that makes it utterly clear.


Hwurzburg:
good catch..thanks...merged

PR #3334, opened 2021-02-16
Introduce OneWire

amilcarlucas:


common/source/docs/common-dshot.rst
```diff
@@ -80,3 +81,9 @@ Ardupilot is compatible with BLHeli32 Passthrough and ESC Telemetry Functions:
 
     BLHeli32 Passthrough <common-blheli32-passthru>
     BLHeli32 ESC Telemetry <common-dshot-blheli32-telemetry>
+    
+OneWire
```
khancyr, line 85: this shouldn't be here as this isn't related to dshot method that are dma based ... Onewire is serial ..

amilcarlucas: Where should we put it then?

amilcarlucas: Can we talk on discord?

Hwurzburg: I agree, but not for that reason....since this page is about ESC configuration not only DSHOT (the file name is historical)....but really should be restricted to ESCs attached to the SERVO/MOTOR outputs...I think the Serial ESCs deserve a separate listing in the common-esc-and-motors page....a new page titled common-serial-escs   then the fettec one wire page can be linked from that page...I will re-title the Brushless ESC configuration page to SERVO/MOTOR Output ESC Configuration and Brushless ESCs page to ESCs connected to SERVO/MOTOR outputs...

PR #3343, opened 2021-02-21
Add note about GPS connection ordering

Hwurzburg:
Perhaps we should change the code to try each SERIAL port configured for GPS before giving up during initialization?

### DIFF
--- a/common/source/docs/common-flight-controller-wiring.rst
+++ b/common/source/docs/common-flight-controller-wiring.rst
@@ -21,6 +21,9 @@ GPS is mandatory in all vehicles. Compass is mandatory for Copter and Rover and 
 
 .. note:: Usually GPS is attached to the logical Serial Port 3 by default in ArduPilot. However, which  physical UART is assigned to ArduPilot's Serial Port 3 on the autopilot is documented in the autopilots :ref:`documentation <common-autopilots>`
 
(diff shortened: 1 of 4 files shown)


common/source/docs/common-flight-controller-wiring.rst
```diff
@@ -21,6 +21,9 @@ GPS is mandatory in all vehicles. Compass is mandatory for Copter and Rover and
 
 .. note:: Usually GPS is attached to the logical Serial Port 3 by default in ArduPilot. However, which  physical UART is assigned to ArduPilot's Serial Port 3 on the autopilot is documented in the autopilots :ref:`documentation <common-autopilots>`
 
+.. note:: While it is possible to use multiple GPSes in a system (See :ref:`<ek3-affinity-lane-switching>`), most use only one. It is important that it be connected to the first SERIALx port that has its ``SERIALx_PROTOCOL`` parameter set to "5" (GPS) since it will stop searching for GPS during bootup if not found on the first port configured for GPS protocol.
```
rmackay9, line 24: I think we should change the first part to be something like:

note:: If only one GPS is used it is important that it be connected to the first SERIALx port that has its ``SERIALx_PROTOCOL`` parameter set to "5" (GPS) since it will stop searching for GPS during bootup if not found on the first port configured for GPS protocol

Hwurzburg: thanks..done

PR #3349, opened 2021-02-23
Wiki: Updates to fences to support fence mission item protocol

joshanne:
_**This is WIP**_

Updates some of the wiki to start pointing to the AC_Fence documentation

These changes are to support: https://github.com/ArduPilot/ardupilot/pull/15288


amilcarlucas (approved):
Looks OK to me, just the 20m push-back sounds a bit strange. It is plane specific right? It does not happen on copter. Or does it?

joshanne:
20m push back occurs on copter and rover, up to a maximum of 100m, at which point it lands or holds respectively.

Plane/geofence doesn't have the concept of pushing the fence back, and an ultimate override if the vehicle is more than 5x (100m) the pushback value. Best be addressed in discussion/reviews.

amilcarlucas:
Can you update it with the latest changes and fixes from the PR review?

joshanne:
> Can you update it with the latest changes and fixes from the PR review?

Done

amilcarlucas (approved):

Hwurzburg:
will review soon...

joshanne:
Ta. Keep in mind, the changes to plane mean that there is now a lot of commonality between Copter/Rover and Plane (they're all using the same AC_Fence Library).

I've tried to simplify a lot of the fence documentation - a lot of it can be confusing and fragmented IMO.

Hwurzburg:
@joshanne  how is the fence return point for plane now set for inclusioin/exclusion fences since its not a copter/rover concept? is there a fence item for this in the new mission style lists? now that missionplanner apprently no longer has a click menu to allow the setting of the return point, how is this done?

joshanne:
> 
> 
> @joshanne how is the fence return point for plane now set for inclusioin/exclusion fences since its not a copter/rover concept? is there a fence item for this in the new mission style lists? now that missionplanner apprently no longer has a click menu to allow the setting of the return point, how is this done?

I'm not aware of a way to do it in a context menu (eg. Right Clicking, or selecting an icon).

In the top right hand corner of mission planner's _**PLAN VIEW**_ is a drop down box, which has `Mission`, `Fence` and `Rally`. Select Fence. Add a waypoint. In the bottom, change the command to `FENCE_RETURN_POINT`.

I'm not sure if there's a simpler way, but that's how it's done now.

FWIW: QGroundControl has a simple "Add Breach Return Point" (or something similar) button.

image

Hwurzburg:
> > @joshanne how is the fence return point for plane now set for inclusioin/exclusion fences since its not a copter/rover concept? is there a fence item for this in the new mission style lists? now that missionplanner apprently no longer has a click menu to allow the setting of the return point, how is this done?
> 
> I'm not aware of a way to do it in a context menu (eg. Right Clicking, or selecting an icon).
> 
> In the top right hand corner of mission planner's _**PLAN VIEW**_ is a drop down box, which has `Mission`, `Fence` and `Rally`. Select Fence. Add a waypoint. In the bottom view (which lists the points in the polygon group), change the command of this waypoint to `FENCE_RETURN_POINT`.
> 
> I'm not sure if there's a simpler way, but that's how it's done now.
> 
> FWIW: QGroundControl has a simple "Add Breach Return Point" (or something similar) button.
> 
> image

I assume that Copter/Rover ignore the fence item if its FENCE_RETURN_POINT or is an error message returned?

joshanne:
> 
> I assume that Copter/Rover ignore the fence item if its FENCE_RETURN_POINT or is an error message returned?

Plane explicitly uses the fence return point ([ArduPlane/fence.cpp:91](https://github.com/ArduPilot/ardupilot/blob/ffd8f3678cd7df5a377339c9d6482cc03555cba9/ArduPlane/fence.cpp#L91)).

It appears Copter/Rover do not use the `get_return_point` function, thus would most likely ignore the point. Interesting that copter/rover still allows the addition of a fence breach return point in both QGC and MP.

Hwurzburg:
another question...Copter does some special things when reaching FENCE_MAX_ALT in some modes....like just refusing to breach it....does Plane now or is the action triggered for max and min alt breachs, independent of mode?

joshanne:
> another question...Copter does some special things when reaching FENCE_MAX_ALT in some modes....like just refusing to breach it....does Plane now or is the action triggered for max and min alt breachs, independent of mode?

I believe copter has some ability to do avoidance with the fence, on both the walls and ceiling - if avoidance is turned on.

Plane however maintains the triggering of `FENCE_ACTION` for max and min alt breaches, independent of mode.

Hwurzburg:
@joshanne James, we can close this now, right?

joshanne:
Certainly can, thanks for your work on updating the fence docs

PR #3352, opened 2021-02-24
correct DO_CHANGE_SPEED descriptions

Hwurzburg:
@rmackay9 in addition to reviewing this, why do we have https://ardupilot.org/copter/docs/mission-command-list.html and this page? I see you have mission planner shots which are nice but outdated and some incorrect now, but this is duplication and a maintenance issue...

I think it might be better to expand the copter section of the common page with any missing information (if any) for commands, put a couple of mission planner shots showing a few waypoint type shots for commands at the top of the common page to give someone the info on where/how to change in mission planner and eliminate that page....lets discuss when you get time soon...

### DIFF
--- a/common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
+++ b/common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
@@ -2718,17 +2718,17 @@ speed-type and throttle settings are ignored.
    <tr style="color: #c0c0c0">
    <td><strong>param1</strong></td>
    <td>speed m/s</td>
-   <td>Speed type (0=Airspeed, 1=Ground Speed).</td>
+   <td>Speed type (0,1=Ground Speed,  2=Climb Speed, 3=Descent Speed).</td>
    </tr>
    <tr>
    <td><strong>param2</strong></td>
-   <td>speed m/s</td>
+   <td>speed in m/s</td>
    <td>Target speed (m/s).</td>
    </tr>
    <tr style="color: #c0c0c0">
    <td>param3</td>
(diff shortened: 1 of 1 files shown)


common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
```diff
@@ -2848,7 +2848,7 @@ set to 0).
    <tr style="color: #c0c0c0">
    <td><strong>param1</strong></td>
    <td>Type (0=as 1=gs)</td>
-   <td>Speed type (0=Airspeed, 1=Ground Speed). Set to 0 before v2.50, otherwise ignored.</td>
+   <td>Speed type . Set to 0 before v2.50, otherwise ignored.</td>
```
rmackay9, line 2851: we probably don't need to mention v2.50 anymore.

rmackay9:
@Hwurzburg I'm happy to have just a single page for all vehicles but I find the copter page is still better than the combined page.  

In particular the Copter page gets to the commands very high up on the page while the combined page requires the user to scroll down through many pages of detailed discussion before getting to the commands.

common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
Hwurzburg: fixed

Hwurzburg:
merging....as to single page...will work on that once @Gone4Dirt finished reviewing our mavlink command docs for accuracy...I will then update, then reorg in a manner similar to the standalone copter page for better navigation....finally archiving that page

PR #3357, opened 2021-03-01
adding documentation for Mamba F405 MK2

arikrupnik:
Support for this board is now in master. I mostly copied Sebastian's readme.md, but reformatted it to RST to fit the rest of the wiki.


common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+     :target: ../_images/mamba_f405mk2_board.jpg
+
+https://www.diatone.us/products/mamba-f405-flight-controller-mk2
+
+The Mamba F405 MK2 is a flight controller produced by
+`Diatone <https://www.diatone.us>`__.
+
+.. note::
+
+   Support for this board is available with Plane 4.1.0, and higher
```
Hwurzburg, line 15: remove note please....trying to avoid any revs references if possible....created Upcoming features section for things not in stable but in master, however, the boards have been an exception and  we have just put it in since there will not be firmware downloadable except for appropriate revs (in this case latest)

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+SERIAL0 COMPUTER         USB
+SERIAL1 PPM/RX1/SBUS/TX1 UART1 (RC Input)
+SERIAL3 TX3/RX3          UART3 (Telem1)
+SERIAL6 TX6/RX6          UART6 (GPS)
+======= ================ ================
+
+RC Input
+========
+
+RC input is configured on the PPM (UART1_RX) pin. It supports all RC
```
Hwurzburg, line 49: I have doubts that this actually works with the hwdef as it stands...it appears that the inverter from SBUS to PPM is a controlled inverter and therefore would need to be actively tri-stated in order to attach anything to the PPM input...I have asked the author, but he never responded....so I dont want to post incorrect information until that is resolved...

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+   Mamba F405 MK2 Board
+
+UART Mapping
+============
+
+======= ================ ================
+Name    Pin              Function
+======= ================ ================
+SERIAL0 COMPUTER         USB
+SERIAL1 PPM/RX1/SBUS/TX1 UART1 (RC Input)
```
Hwurzburg, line 41: Using RX1 as UART is not possible unless alt config is selected....see how this is done here: https://ardupilot.org/copter/docs/common-omnibusf4pro.html

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+The Mamba F405 MK2 supports OSD using OSD_TYPE 1 (MAX7456 driver).
+
+PWM Output
+==========
+
+The Mamba F405 MK2 supports up to 4 PWM outputs. The pads for motor
+output ESC1 to ESC4 on the above diagram are for the 4 outputs. All 4
+outputs support DShot as well as all PWM types.
+
+The PWM are in one group.
```
Hwurzburg, line 64:
```suggestion
All the PWM outputs are in one group.
```

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+an external compass using I2C on the SDA and SCL pads.
+
+Alternate settings
+==================
+
+It is possible to set alternate configurations with the BRD_ALT_CONFIG
+parameter.
+
+============== ============
+BRD_ALT_CONFIG RX1 function
```
Hwurzburg, line 97: since there is no RX1 shown in the diagrams, it should be called PPM pin function in the heading.. and instead of referring to TX1 in the table , put a note that TX1 is usable in both configurations below it.

Also, we need to verify that ALT1 works, again due to controlled inverter output on PPM (RX1 pin)

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
...
+============== ============
+ALT 0(default) RCININT
+ALT 1          RX1/TX1
+============== ============
+
+Loading Firmware
+================
+
+Initial firmware load can be done with DFU by plugging in USB with the
```
Hwurzburg, line 106: use the flashing firmware section from other board docs as in here https://ardupilot.org/copter/docs/common-omnibusf4pro.html

Hwurzburg (changes requested):

common/source/docs/common-autopilots.rst
```diff
@@ -73,6 +73,7 @@ Closed hardware
     SpeedyBee F4 (this board currently is non-verified) <common-speedybeef4>
     VR Brain 5 <http://www.virtualrobotix.it/index.php/en/shop/autopilot/vrbrain5-detail>
     VR uBrain 5.1 <http://www.virtualrobotix.it/index.php/en/shop/autopilot/vrbrainmicro51-detail>
+    Mamba F405 MK2* <common-mamba405-mk2>
```
Hwurzburg, line 76: Should be in alphabetical order

Hwurzburg (changes requested):

common/source/docs/common-autopilots.rst
arikrupnik: I should have figured this out. Done.

common/source/docs/common-mamba405-mk2.rst
arikrupnik: Aren't we using software serial for RC input?

common/source/docs/common-mamba405-mk2.rst
arikrupnik: done

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,112 @@
+================================
+Mamba F405 MK2 Flight Controller
+================================
+
+.. image:: ../../../images/mamba_f405mk2_board.jpg
+     :target: ../_images/mamba_f405mk2_board.jpg
+
+https://www.diatone.us/products/mamba-f405-flight-controller-mk2
```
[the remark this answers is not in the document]
arikrupnik: done

common/source/docs/common-mamba405-mk2.rst
arikrupnik: Done

common/source/docs/common-mamba405-mk2.rst
arikrupnik: Looking at this closer, I'm unclear on the author's assertions of SERIALx usage. Do you know if there are compelling considerations to list GPS on `SERIAL6` instead of the more common `SERIAL3`, for instance?

common/source/docs/common-mamba405-mk2.rst
arikrupnik: I pushed a commit that may address your feedback. Let me know how you like it.

arikrupnik:
I pushed a couple of commits that should address most of the feedback. @Hwurzburg let me know what you think, and if there are additional improvements I can make. An open question is serial port function assignments that disagree with defaults.

Hwurzburg:
Ari, perhaps we best chat on this....I have a real problem with this board joining the wiki for the following:
1. I must have a baro added externally.....while theoretically we should be able to setup EKF3 to use GPS alt, Tridge and I have not gotten it to actually fly on a board yet....having EKF stability problems and its not on the top of his list....that means the average user will have to add an external baro...non trivial for many users..just asking for more support calls....SpeedyBeeF4 is in the same boat...I really want to move it out of the wiki
2. I have been told that the hwdef is incorrectly setup to even FIND an external baro...so its reported as non-functional as of now
3. The PPM in pin is probably non-functional and the ALT config also due to the inverter driving it from SBUS....so we would need to only allow SBUS in and no ALT config
the only benefit of this is its $19 price, but adding an external BARO  and current sense defeats that when for $20 you can get an OmnibusF4Pro....

arikrupnik:
@Hwurzburg these are all valid concerns. I would invite @tridge to share his thoughts.

common/source/docs/common-mamba405-mk2.rst
Hwurzburg: The way this hwdef is configured you need to use the SBUS input for all RC protocols , PPM input is unuseable since its driven by the SBUS inverter all the time and not disabled easily (can by forcing the GPIO)...and I am pretty sure that ALT config 1 does not work...it would need to also disable the inverter to use the UART1 RX input (PPM pin)

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,172 @@
...
+======= ========== ===================
+Name    Pin labels Function
+======= ========== ===================
+SERIAL0            USB
+SERIAL1 PPM / SBUS RC Input / UART1 RX
+SERIAL3 TX3 / RX3  UART3 (Telem1)
+SERIAL6 TX6 / RX6  UART6 (GPS)
+======= ========== ===================
+
+It is possible to use PPM pin as SERIAL1 TX instead of RC input by setting :ref:`BRD_ALT_CONFIG<BRD_ALT_CONFIG>` to 1.
```
Hwurzburg, line 47: incorrect....and its the RX not TX pin....I would just not mention ALT config until its gets fixed to also disable the inverter output driving the PPM pin from the SBUS pin

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,172 @@
...
+Mamba F405 MK2 Flight Controller
+================================
+
+.. image:: ../../../images/mamba_f405mk2_board.jpg
+     :target: ../_images/mamba_f405mk2_board.jpg
+
+The Mamba F405 MK2 is a flight controller produced by
+`Diatone <https://www.diatone.us>`__.
+
+
```
Hwurzburg, line 11: We need to add a warning here that this autopilot does not have a BARO and currently cannot be used without an external BARO addition. I intend to move SpeedyBee, this, and the other autopilot to a special section of the autopilot landing page since they need external baros

common/source/docs/common-mamba405-mk2.rst
Hwurzburg: and the hwdef has UART1 defaulted to TELEM1(which does not matter) and UART3 defaulted to GPS1, UART6 is USER2, so the pinout diagram needs changing also...its for a BETAFLIGHT setup, not AP

common/source/docs/common-mamba405-mk2.rst
```diff
@@ -0,0 +1,172 @@
...
+Where to buy
+============
+
+Available from multiple retailers and directly from the vendor at `diatone.us <https://www.diatone.us/products/mamba-f405-flight-controller-mk2>`__
+
+
+Pinout
+======
+
+.. figure:: ../../../images/mamba_f405mk2_pinout.png
```
Hwurzburg, line 30: we need a new pinout/wiring diagram for ArduPilot....UART3 is GPS by default in the hwdef...none is shown in the BF diagram...also we would need to show an external I2C BARO on the I2C bus along with an optional compass...

Hwurzburg:
Made some changes to text and figures....per above....merging....thanks!

PR #3358, opened 2021-03-01
Copter: add Hexsoon TD-650 reference frame

rishabsingh3003:
This adds the Hexsoon TD 650 as a new reference frame:
http://www.proficnc.com/all-products/262-td-650.html

The link to the param file won't work till https://github.com/ArduPilot/ardupilot/pull/16777 isn't merged. So best to merge this after that is in Master.
Also, I have included a short flight video: 
https://www.youtube.com/watch?v=FbzXvi3beDI

I am not sure how good it is, but its not doing any harm either. If it isn't relevant, please feel free to comment down below. I'll remove it.


rmackay9:
Looking good, thanks!

PR #3363, opened 2021-03-03
Add Xterm to Cygwin's setup parameters

muramura:
I add XTERM in the setup list as a parameter.


rmackay9:
Thanks.  I'm not actually totally sure this is correct but I think XTerm is probably required to show the console.

PR #3367, opened 2021-03-05
dev: add warning to eclipse page that it is out-of-date

peterbarker:
See https://github.com/ArduPilot/ardupilot/issues/16681#issuecomment-790975227


Hwurzburg:
thanks...merging

PR #3368, opened 2021-03-05
Add OEM firmware customization page

Hwurzburg:


common/source/docs/common-oem-customizations.rst
```diff
@@ -0,0 +1,54 @@
...
+
+.. warning:: The ``defaults.parm`` file should be as small as possible. Some boards only allow 1024 bytes total for this file. Every ASCII byte in the file counts against this limit (except for comment lines). Use integer values where possible. Below is a simple example.
+
+::
+
+       # setup for NTF LEDs on output5
+       SERVO5_FUNCTION 120
+       NTF_LED_TYPES 256
+
+5. You can also embed LUA scripts in the ROM of the chip that will automatically be run. Since these currenly only run on autopilots with a lot of flash space, they are only restricted in total aggregate size to under 8192 bytes total. Put the scripts in a sub-directory called ``scripts``, ie
```
tridge, line 47: the lua scripts don't have a 8192 byte limit. The limit is based on how much flash and memory you have. For F7 and H7 boards with 2M flash the limit will be quite high (certainly much higher than 8k)

tridge (changes requested):
very nice! just small fix for limit of lua scripts

common/source/docs/common-oem-customizations.rst
Hwurzburg: thanks...was just guessing, figured it would be corrected on review if wrong...

PR #3370, opened 2021-03-06
Update "How to Add New Bindings" section in Lua Scripting docs

rishabsingh3003:
Thanks to @IamPete1 for helping me figure this out.
I have updated the docs to help others.

You can't do "make run" in master


Hwurzburg:
rather than having different instructions for different code bases, just use the distclean and recompile approach for everything?

rishabsingh3003:
@Hwurzburg you can't do the distclean in 4.0.x . The only option is to do "make run".
This PR brought the change: https://github.com/ArduPilot/ardupilot/pull/14329 in master

Hwurzburg:
thanks...merging

PR #3376, opened 2021-03-10
Update common-downloading-and-analyzing-data-logs-in-mission-planner.rst

### LINKED ISSUE #3375: Wrong gyro units in "Downloading and Analyzing Data Logs in Mission Planner"
[Downloading and Analyzing Data Logs in Mission Planner](https://ardupilot.org/copter/docs/common-downloading-and-analyzing-data-logs-in-mission-planner.html)

I am working with some logs from ArduPilot. The GyrX units are definitely not degree/s. When I calculate the gyro rates by taking the derivative of the tilt angles, the gyro rates seem to have a unit of rad/(Pi/2)/s which sounds weird.

I tried to find units in the mavlink message from the files but nothing is there.

It is stated in this [old discussion](https://discuss.ardupilot.org/t/imu-units-of-the-pixhalk-log/19458/5) that the units are rad/s.

I'm not sure what the units are, they just don't seem to be degree/s so the page can confuse people.

hendjoshsr71:
The rotation rates are logged as radians/seconds. Closes #3375


Hwurzburg:
sorry...already made the change when I came across #3375 before I reviewed this one....

PR #3378, opened 2021-03-11
Autotune: Check maximum angular rate parameters

samuelctabor:
Linked to issue https://github.com/ArduPilot/ardupilot/issues/16856 , autotune can fail if the maximum angular rates are larger than the aircraft can achieve. Add a note to the wiki page to check the rates are appropriate.

### DIFF
--- a/plane/source/docs/automatic-tuning-with-autotune.rst
+++ b/plane/source/docs/automatic-tuning-with-autotune.rst
@@ -41,7 +41,7 @@ important to set a level appropriate for the rates that the aircraft can physica
 This is because excessive target rates will prevent the autotune from adjusting the gains.
 In particular, the level 6 roll rates can be too large for slow aircraft with large wing spans,
 like gliders. These often have a maximum roll rate of 30 - 40 degrees per second and should use
-level 2 or 3. See :ref:`automatic-tuning-with-aututune_level-settings` for the rates.
+level 2 or 3. See :ref:`automatic-tuning-with-autotune_level-settings` for the rates.
 
 You also need to make sure that all of the basic settings for your
 airframe are correct. In particular, ensure that all surface reversals
@@ -231,7 +231,7 @@ If you adjust the I gain manually, then this also changes the value for
 P that is required to maintain the correct response, so adjusting the I
 gain is only recommended for advanced users.
 
-.. _automatic-tuning-with-aututune_level-settings:
+.. _automatic-tuning-with-autotune_level-settings:
 
 AUTOTUNE_LEVEL settings
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/plane/source/docs/automatic-tuning-with-autotune.rst
+++ b/plane/source/docs/automatic-tuning-with-autotune.rst
@@ -59,6 +59,9 @@ Other things to check:
    your transmitter with your plane on the ground (and propeller removed
    or made safe) and check how the plane reacts
 -  setup a rally point for a safe place to RTL if needed
+-  make sure that the maximum roll and pitch rates can be achieved by your aircraft at 
+   normal flight speeds. In particular, the default roll rates can be too large for
+   aircraft with large wing spans. These often have a maximum roll rate of 30 - 40 degrees per second.
 
 Flying in AUTOTUNE
 ~~~~~~~~~~~~~~~~~~
--- a/plane/source/docs/automatic-tuning-with-autotune.rst
+++ b/plane/source/docs/automatic-tuning-with-autotune.rst
@@ -36,6 +36,13 @@ then you could choose level 7, which will result in a bit sharper tune
 have done an initial tune with a lower level. Levels above 8 should only
 be used by very experienced pilots.
 
+:ref:`AUTOTUNE_LEVEL<AUTOTUNE_LEVEL>` also sets the target pitch and roll rates. It is
+important to set a level appropriate for the rates that the aircraft can physically achieve.
+This is because excessive target rates will prevent the autotune from adjusting the gains.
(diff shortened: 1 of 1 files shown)


plane/source/docs/automatic-tuning-with-autotune.rst
```diff
@@ -59,6 +59,9 @@ Other things to check:
    your transmitter with your plane on the ground (and propeller removed
    or made safe) and check how the plane reacts
 -  setup a rally point for a safe place to RTL if needed
+-  make sure that the maximum roll and pitch rates can be achieved by your aircraft at
```
Hwurzburg, line 62: need to point to specific params for this

Hwurzburg (changes requested):

samuelctabor:
After some more investigation the right way to deal with this is to set AUTOTUNE_LEVEL appropriately as this automatically sets the rates. Updated and added table of level vs rate.

plane/source/docs/automatic-tuning-with-autotune.rst
```diff
@@ -223,3 +230,35 @@ lower throttle settings and airspeeds.
 If you adjust the I gain manually, then this also changes the value for
 P that is required to maintain the correct response, so adjusting the I
 gain is only recommended for advanced users.
+
+.. _automatic-tuning-with-aututune_level-settings:
```
Hwurzburg, line 234: spelling?

plane/source/docs/automatic-tuning-with-autotune.rst
```diff
@@ -36,6 +36,13 @@ then you could choose level 7, which will result in a bit sharper tune
 have done an initial tune with a lower level. Levels above 8 should only
 be used by very experienced pilots.
 
+:ref:`AUTOTUNE_LEVEL<AUTOTUNE_LEVEL>` also sets the target pitch and roll rates. It is
+important to set a level appropriate for the rates that the aircraft can physically achieve.
+This is because excessive target rates will prevent the autotune from adjusting the gains.
+In particular, the level 6 roll rates can be too large for slow aircraft with large wing spans,
+like gliders. These often have a maximum roll rate of 30 - 40 degrees per second and should use
+level 2 or 3. See :ref:`automatic-tuning-with-aututune_level-settings` for the rates.
```
Hwurzburg, line 44: spelling?

plane/source/docs/automatic-tuning-with-autotune.rst
samuelctabor: fixed

plane/source/docs/automatic-tuning-with-autotune.rst
samuelctabor: fixed.

plane/source/docs/automatic-tuning-with-autotune.rst
```diff
@@ -36,6 +36,13 @@ then you could choose level 7, which will result in a bit sharper tune
 have done an initial tune with a lower level. Levels above 8 should only
 be used by very experienced pilots.
 
+:ref:`AUTOTUNE_LEVEL<AUTOTUNE_LEVEL>` also sets the target pitch and roll rates. It is
```
Hwurzburg, line 39: Samuel, just trying to understand a bit better....
the angle limit determines angle at full stick deflection set by LIM_X_CD
a step input on angle error generates an angle error which is multiplied by a gain (P? since I have been told Plane is really not a rate but rather an angle controller historically) and ramped by the TCONST to produce a demanded rate which is limited by the max rates below and generates the control deflection...the max rate is the target of the autotune process? and if so, then having it too high would seem to make it drive gains too high trying to meet it instead of giving up..obviously I don't understand....it this is not the case, then maybe having the rates in a table is TMI and we should just put the warning for sailplanes in and skip the rate table, since this seems counter-intuitive to someone with a dangerously small (ie some but not enough...like me with an MS in it) amount of control theory knowledge

samuelctabor: Hi Henry,

Close, actually the angle error is multiplied by 1/TCONST to get the target rate. This is then limited by RMAX.

The actual rate is subtracted from target rate to get rate error. The output to control surface is based on;
 - P + FF multiplied by target rate
 - D multiplied by rate error
 - I multiplied by integrated rate error. 

So yes it is not a PID controller on angular rate like the copter controllers, but more like FF + PI.

If the target rates are too high the control outputs will be maxxed out trying achieve the excessive target. The autotune logic detects this and won't adjust gains while this is the case, to avoid over-gaining like you describe.

IMO the default of 75 deg/s is very aggressive but obviously my experience is mostly from gliders. I think the numerical values do need to be documented somewhere.

Hwurzburg: thanks...going to merge as is...but I am going to add some info like what TCONST and RMAX does here or elsewhere and maybe expand a little here to make it clearer what the failure to tune mechanism is as you have explained...thanks

PS I also have question about the I term that gets set during AT...I always reset it back higher, otherwise SERVO_AUTOTRIM may never actually trim...but it warns that it impacts P, which I have found to be a weak interaction....perhaps its out of date...the defaults for I are high in case CG is off so it can save a plane...and many never autotune and use it as is...

Hwurzburg:
merging...thanks

plane/source/docs/automatic-tuning-with-autotune.rst
samuelctabor: Thanks. There is an interaction. I don't fully understand the reasoning for the logic, but the interaction term is scaled by TCONST squared, making it quite small, and it is subtracted from P, so larger I should be on the safe side.

PR #3418, opened 2021-04-02
Dev: autotest-framework fix text off page

hendjoshsr71:
This fixes a previous change of mine that had the build test steps going off the edge of the page. 
If there is a better way to fix it let me know?


Hwurzburg:
LGTM...merging

PR #3424, opened 2021-04-06
Update loiter-mode.rst to correct LOIT_ANG_MAX

JoshWelsh:
Correcting description of LOIT_ANG_MAX to reflect real-world and also full-param-list documentation


SUI-Eric:
There's an additional mismatch between the Loiter documentation and the full parameter list documentation as well. The loiter page states that ANGLE_MAX will be used if LOIT_ANG_MAX is set to zero, but the full parameter list page states that 2/3 of PSC_ANGLE_MAX or ANGLE_MAX will be used.

Hwurzburg:
merging this,then fixing the remaining issue in description

PR #3431, opened 2021-04-08
Corrects link for EV ID definitions

ccackam:
The previous link ([https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/defines.h#L281](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/defines.h#L281)) No longer contains this information. It is now found at the corrected link ([https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Logger/AP_Logger.h#L94](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Logger/AP_Logger.h#L94)).


Hwurzburg:
thanks...merging

PR #3443, opened 2021-04-11
update.py: flake8 correctness

peterbarker:
nothing but make it flake8-clean

### DIFF
--- a/update.py
+++ b/update.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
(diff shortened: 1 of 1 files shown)


Hwurzburg:
not sure what this does, but if you are sure its correct, merge it!

khancyr (approved):

khancyr:
@Hwurzburg flake8 is a small software to check python script against the python norm. So this allow to  check for python correctness, and general allow to catch simple syntax mistakes

PR #3444, opened 2021-04-11
Generate ap periph paramaeters

peterbarker:
Based on previous PR.  This is actually a trivial change.

I'm not sure this is actually useful because only a very small subset of these parameters will be available on many devices - not all AP_Periph nodes support all of the libraries we document in here.


Hwurzburg:
@peterbarker I tired this locally and it does not seem to generate the document...

PR #3446, opened 2021-04-11
Add OMC converted docs in temp location:ardupilot for review

Hwurzburg:
Just a temporary placement until an omc actual site is created....this will allow review and mods in the meantime....just dumped it in the ardupilot site so it would appear in a side menu....also did not separate the images out into an upper level dir, just to avoid changes to all the links at the present time...once a site is created, the proper hierarchy can be implemented....
I have tested builds in a local "site" and update.py will construct it properly locally...just need a real https site for correct links...

if you build locally, you can see the page as a "fake" site on the left hand nav menu on the first docs page...


tridge (approved):

tridge:
thanks Henry!

tridge:
i think we'd need some activity from OMC first

tridge:
@Hwurzburg there was a suggestion on the call to put it straight into archived?

PR #3433, opened 2021-04-08
Dev: Update Setup of Windows WSL Environment

hendjoshsr71:
There are number of requests on discord and discuss regarding how to setup the WSL environment. I added a few things that are required for setting it up to use with SITL. I also added a few more notes on useful tools and integrations with WSL. 

I think new developers should be pointed to use WSL and not Cygwin as I have done here. 

The note I have about why one may need to compile using cygwin may need to be fixed up a bit?

I'm not sure if the note at the end about "Ground Control Station software on wsl" is true anymore?

PR #3453, opened 2021-04-14
Add section for building SITL binaries for Windows

TunaLobster:
Dependent on ArduPilot/ardupilot#17083

### DIFF
--- a/dev/source/docs/building-setup-windows-cygwin.rst
+++ b/dev/source/docs/building-setup-windows-cygwin.rst
@@ -16,19 +16,19 @@ These setup instructions describe how to setup `Cygwin <http://www.cygwin.com/>`
 Install Cygwin
 --------------
 
-#. Open a web browser to `www.cygwin.com/install.html <https://www.cygwin.com/install.html>`__ and download  "setup-x86_64.exe"
+#. Open a web browser to `www.cygwin.com/install.html <https://www.cygwin.com/install.html>`__ and download  "setup-x86.exe"
 
-#. Paste the  following command at a DOS prompt (press window and R key then enter cmd) in the directory where you downloaded "Run setup-x86_64.exe"
+#. Paste the  following command at a DOS prompt (press window and R key then enter cmd) in the directory where you downloaded "Run setup-x86.exe"
 
 ::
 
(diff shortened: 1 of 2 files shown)


dev/source/docs/building-setup-windows-cygwin.rst
```diff
@@ -162,4 +164,13 @@ You can run SITL just as you can with any other build environment. In order to s
 
 ::
 
-      export DISPLAY=:0.0 
+      export DISPLAY=:0.0
+      
+Build SITL binaries for Windows
+===============================
+
+You can build 32-bit binaries to be run outside of Cygwin will a small set of companion files. This can be useful when used in conjunction with Mission Planner to run SITL with a stable release build that is not from the master developement branch of ArduPilot.
```
Hwurzburg, line 172:
```suggestion
You can build 32-bit binaries to be run outside of Cygwin with a small set of companion files. This can be useful when used in conjunction with Mission Planner to run SITL with a stable release build that is not from the master development branch of ArduPilot.
```

Hwurzburg (changes requested):

hendjoshsr71:
You might want to add a note to this page about why a developer would need to use Cygwin. As it is the first page 
 in Win10 installs. https://ardupilot.org/dev/docs/building-setup-windows.html

TunaLobster:
@hendjoshsr71 Has updated the linked page in #3433. This information was added in the first note on the wiki page.

hendjoshsr71:
Sorry what I wrote was confusing. 

I just wanted to make sure you thought what I just added to that page made sense? Or to add more details about why someone would use Cygwin as I was a bit vague.

TunaLobster:
Updated other lines to try and weave everything together and include more information. Also took a quick pass at the grammar on these files.

stephendade:
A few notes from me:
-The ``cygwin32-gcc-g++`` package isn't required, since we're using 32-bit Cywin here
-``./waf --color yes --toolchain i686-pc-cygwin --board sitl configure bin`` can be simplified to ``./waf --board sitl configure``

TunaLobster:
Updated the commit to fix those items.

stephendade:
Looks good from here. LGTM

Hwurzburg:
squashed, rebased,tested...LGTM...merging..thanks

PR #3454, opened 2021-04-15
Wiki conversion to 4.1

Hwurzburg:
Moving Copter and Rover and common items out of upcoming into main wiki


common/source/docs/common-advanced-configuration.rst
```diff
@@ -62,13 +62,13 @@ tuning options for the vehicle.
     Ground Control Station Only Operation <common-gcs-only-operation>
 [site wiki="copter"]
     Ground Effect Compensation <ground-effect-compensation>
-    In-Flight FFT Configuration <common-imu-fft>
```
rmackay9, line 63: Hi, is it "FFT" or "FTT"?

rmackay9:
This looks OK to me.  I guess we should merge this now because we've started beta testing?  If we do it will perhaps cause some confusion because users of stable will find some new features don't work... but beta testers are more likely to find these changes.

I guess another option would be to more explicitly point beta testers to the "New Features" section of the wiki.

Anyway, I'm happy with this being merged any time.

common/source/docs/common-advanced-configuration.rst
Hwurzburg: Fast Fourier Transform

Hwurzburg:
Still a lot more work to do, but am going to merge this now and continue working on existing PRs and issue list

PR #3458, opened 2021-04-17
Change the second pin of the Pixhawk power connector to be red for po…

ardugitter:
The first two pins of the Pixhawk power connector pinout should be red to indicate power. The previous version had the first pin red and second black leading to people assuming the second pin was ground - a dangerous situation.

Signed-off-by: roger <73884126+ardugitter@users.noreply.github.com>


Hwurzburg:
LGTM...thanks

PR #3475, opened 2021-04-21
Fix typo

vergissberlin:
Discovered in this video https://youtu.be/x_L4RlZegXg?t=55


khancyr (commented):
I don't think the change is right, as  we are speaking of an area. My english my be weak, but I think that a 7 by 8km square is right

Hwurzburg:
original wording is correct..."an area x by y" describes the dimensions where x,y are lengths...."an area x to y " describes the variation in the area and x,y would be in area dimensions like square km.... .closing

PR #3476, opened 2021-04-22
Add AP_Periph updates in checks to update it on the Wiki

brunoolivieri:
Wiki is build when something is updated.

Inserting AP_Periph check as in the others vehicles.


brunoolivieri:
Waiting for the check and I will apply it and observing in the server.

brunoolivieri:
Applying and watching in the server.

PR #3481, opened 2021-04-23
Add Iflight Beastf7 AIO

Hwurzburg:


common/source/docs/common-iflight-beastf7AIO.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The board has a built-in voltage and current sensors.
+
+The correct battery monitor parameters are:
+
+-    :ref:`BATT_MONITOR<BATT_MONITOR>` =  4
+-    :ref:`BATT_VOLT_PIN<BATT_VOLT_PIN>` = 12
+-    :ref:`BATT_VOLT_MULT<BATT_VOLT_MULT>` ~ 10.9
+-    :ref:`BATT_CURR_PIN<BATT_CURR_PIN>` = 13
+-    :ref:`BATT_AMP_PERVLT<BATT_AMP_PERVLT>` ~ 28.5
```
andyp1per, line 100: This was wrong in my original definition it should be 100

andyp1per (changes requested):

common/source/docs/common-iflight-beastf7AIO.rst
Hwurzburg: you should set the log backend to 4 also in the defaults

Hwurzburg: so its a 300A sensor?

Hwurzburg: its specs say 200mv/10a which would be 50 for the amp/volt

andyp1per: > you should set the log backend to 4 also in the defaults

Not needed - this is done automatically by specifying flash logging

andyp1per: > so its a 300A sensor?

The docs say the "current scale: 200", but this value is what gives me believable results

Hwurzburg: okay, if you have measured that 100 is the correct value...will change in both docs...that implies a 300A current sensor range..which I will change also...

andyp1per (approved):

PR #3484, opened 2021-04-23
4.1 update FENCE_ACTIONS for Copter and Rover

Hwurzburg:
please check the existing entries for Rover...they do not agree with its param metadata...let me know what to change...thanks


rmackay9:
looking good, thanks!

PR #3488, opened 2021-04-25
Add TradHeli param list at a glance page

Hwurzburg:
first addition to the wiki as I learn and build...


bnsgeyer:
@Hwurzburg  I modified the page.  I purposely don't reference transmitter throttle curve.  I think this is confusing enough to RC heli users.  you say throttle curve and they think channel 3 and channel 3 is what we use for collective input from the transmitter.  I am not sure there is a way to use the throttle curve from the transmitter to be the input for the RSC.  You would probably have to use the pitch channel (channel 6) as the RSC passthrough channel.  It would not be very clean plus it adds the potential that they mess up the failsafe settings.  

Please review what I've changed and see if you have any other suggestions.

Thanks,
Bill

Hwurzburg:
thanks....added a few missing params for second custom swash....merging

PR #3491, opened 2021-04-26
Update Obstacle Avoidance docs for Copter/Rover 4.1

rishabsingh3003:
This updates Path Planning libraries documentation to Copter/Rover 4.1. This also sets up a new "proximity" hardware page, so that it can be removed from the rangefinder setup and have its own identity. I have also done some rearrangement to the main OA landing page.. hope it is fine. 

The only thing remaining from OA point of view is simple avoidance which is next on my list..  
I am really sorry @Hwurzburg for the huge delay despite the constant reminders from you. I had a branch ready, but needed videos or something else to complete it and hence did not push..

### DIFF
--- a/common/source/docs/common-ads-b-receiver.rst
+++ b/common/source/docs/common-ads-b-receiver.rst
@@ -1,9 +1,9 @@
 .. _common-ads-b-receiver:
+[copywiki destination="plane,copter"]
 
-
-==============
+=====
 ADS-B
-==============
+=====
 
 This article describes how to attach and configure an ADS-B module so that your aircraft can be aware of, and/or transmit to, other aircraft and air-traffic control nearby. This also allows the pilot on the ground to be aware of nearby manned aircraft and optionally to allow the vehicle to automatically avoid them.
 
--- a/common/source/docs/common-oa-bendyruler.rst
+++ b/common/source/docs/common-oa-bendyruler.rst
@@ -9,19 +9,44 @@ Object Avoidance with Bendy Ruler
 
 Copter and Rover 4.0 (and higher) support "BendyRuler" for path planning around obstacles and fences.  The BendyRuler algorithm probes around the vehicle in many directions looking for open spaces and then tries to pick the direction that is sufficiently open while also moving the vehicle towards the final destination.
 
-.. note:: 
+.. note::
 
     This is only applicable in AUTO, GUIDED, and RTL flight modes.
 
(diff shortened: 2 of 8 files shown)


rishabsingh3003:
This PR introduces a couple of errors I can't get rid of: "/home/rishabh/ardupilot_wiki/copter/source/docs/common-proximity-landingpage.rst: WARNING: document isn't included in any toctree"
and  a few others of these similar "WARNING: document isn't included in any toctree"

Can someone help me on how to fix this, and explain what this really means?

Hwurzburg:
I will review and fix any issues in the build this week

rishabsingh3003:
In the latest commit, this now contains Simple Avoidance updates too.
I want to update other small things and add more videos but I think I will do that in a separate PR.

Now majorly developer documentation is left.

Hwurzburg:
getting a lot of build errors in other wiki sites due to shifting or eliminating some docs and references...need to make sure this reorg makes sense and then fix the issues its generates....going out of town tomorrow for 5 days...will  work on fixing these when I return....

rishabsingh3003:
@Hwurzburg I wanted to make a distinction between rangefinders and proximity sensors. That's why I removed the 360 lidars from rangefinder section and created a new proximity page, where rangefinders are referenced as a possible proximity sensor .
Proximity sensors like RPlidar A2, Lightware SF45 etc are only supported by Rover and Copter and are used for horizontal avoidance. At the moment they cannot be used as a replacement for downward-facing rangefinder.
AFAIK, plane only supports downward facing range finders which belong to the RangeFinder lib..

Hwurzburg:
we will need to make a separate page under peripheral hardware for those, if moved and added to the main landing page for peripherals....I can do that when I get back in town next week

rmackay9:
Sorry if I've gone back and forth on this but I wonder if it is really helpful for users to have proximity sensors separated from lidar.  The distinction in the code is large but for users it may not be.  It might be better instead of separate them into different sections on the rangefinder landing page and maybe even have a more general proximity sensor setup page (linked from the landing page) but having a completely different section under Peripherals may not be the best choice.

Hwurzburg (approved):
I re-org'd per Randy's comments, a few minor edits, fixed all build errors....merging
thanks for all the work on this!

rishabsingh3003:
Thanks @Hwurzburg . Sorry for troubling you with this :)

PR #3497, opened 2021-04-28
minor re-formatting and tweaks

Hwurzburg:
just a few tweaks and an addition re acro setup

### DIFF
--- a/copter/source/docs/traditional-helicopter-swashplate-setup.rst
+++ b/copter/source/docs/traditional-helicopter-swashplate-setup.rst
@@ -9,8 +9,9 @@ pitch, roll, and collective commands and translate them to individual blade pitc
 inputs. There are many different servo arrangements for both three and four servo 
 swashplates. The swashplate library covers all of the popular configurations for 
 both three and four servo swashplates. If you have a unique swashplate, you can 
-also use the generic three servo swashplate. Linearize servo is a feature that you 
-can use for both three and four servo swashplates however it is optional for the 
+also use the generic three servo swashplate. 
+
+Linearize Servo is a feature that you can use for both three and four servo swashplates. However, it is optional for the 
 three servo swashplates as there is no chance for binding. It will most likely 
 always be required for four servo swashplates because of the potential for binding 
 in a four point attachment. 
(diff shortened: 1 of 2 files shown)


bnsgeyer (commented):
I would like to get your thoughts on the outputs vs servo.  Look at the heli page in MP or QGC.

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -26,30 +27,39 @@ by using shims underneath the landing gear as shown in the picture below.
...
+For single heli, the servo function assigned to Servo 1
+is motor 33, Servo 2 is motor 34, and Servo 3 is motor 35.  These assignments are the same 
 for swashplate 1 for a dual heli frame.  Swashplate 2 for a dual heli defaults to 
-servos 4, 5, and 6 with motors 36, 37 and 38 assigned respectively.  For four servo swashplates,
-the fourth servo on the single heli frame defaults to servo 5 and is assigned motor 37. 
-For the dual heli frame, the fourth servo on swashplate 1 defaults to servo 7 and is assigned 
-motor 39. The fourth servo on swashplate 2 defaults to servo 8 and is assigned motor 40. 
-In the case of dual heli frame with four servo swashplates, the RSCHeli function (motor 31) will 
-need to be moved to one of the aux servo outputs (9-16).
+outputs 4, 5, and 6 with motors 36, 37 and 38 assigned respectively.
```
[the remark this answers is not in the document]
Hwurzburg: changed back to servo outputs

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -26,30 +27,39 @@ by using shims underneath the landing gear as shown in the picture below.
...
+
+For four servo swashplates,
+the fourth servo (Servo 5) on the single heli frame defaults to output 5 and is assigned motor 37. 
+For the dual heli frame, the fourth servo (Servo 7) on swashplate 1 defaults to output 7 and is assigned motor 39. The fourth servo (Servo 8) on swashplate 2 defaults to output 8 and is assigned motor 40 function.
```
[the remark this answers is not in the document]
Hwurzburg: changed back to servo outputs

Hwurzburg:
@bnsgeyer this should be ready to merge...

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -188,13 +199,16 @@ Center              +0
...
 
-which may be sensitive for some pilots. Therefore, it is recommended to change the values from default to something like in this case:
+which may be too sensitive for some pilots. Therefore, for those setups, it is recommended to change the values from default values to something like this:
 
 - :ref:`IM_STB_COL_1<IM_STB_COL_1>` = 42
 - :ref:`IM_STB_COL_2<IM_STB_COL_2>` = 62
 - :ref:`IM_STB_COL_3<IM_STB_COL_3>` = 80
 - :ref:`IM_STB_COL_4<IM_STB_COL_4>` = 92
 
+to provide a similar curve in STABILIZE to that of the narrower collective range. At least until the pilot becomes accustomed to the collective feel.
```
bnsgeyer, line 209: Why would a user ever want to set up the Stabilize collective range to include negative collectives down to -12.   Your statement here implies that a user might want to set up the stabilize mode this way so that it is the same for a aerobatic acro curve.  I would recommend that they always set up the Stabilize collective curve for non-aerobatic collective range.

Hwurzburg: a user wouldn't...and that is not what it said...those values set -2 to +10.... to provide a collective in STABLIZE like the narrower collective range (-2 to 10....not the wider one -12 to +12)

Hwurzburg: I am not sure that is true....on my 3D machines, I had non idle up curves that followed the idle up curves except for the lower stick range where they  kept -2 max and increased throttle from 0 at low stick to maintain the correct head speed at by mid stick.... I never had my upright hover point different (3/4 stick)  and 0 deg always mid stick.....to do otherwise confuses your thumb on where upright hover is and where 0 deg is....just wanted the non idle up to be spool controlled on the lower half of the stick...

bnsgeyer: What you said about how you did it for aerobatic setups makes sense however the stabilize collective curve does not support the curve you are suggesting where it stays at minus 2 deg for the collective stick on the lower half. And the built in throttle curve would remain at the value that corresponds to -2deg collective if it was able to achieve the collective curve you are suggesting.  I don’t think the code supports aerobatic heli’s very well.

Hwurzburg: it does not remain at -2...low stick is -2, 25% stick is -1 mid is 0...

Hwurzburg: https://attituderc.com/rc-helicopter-curves/

bnsgeyer: Currently the code doesn’t support a 5 point collective curve in stabilize mode. So I’m not sure that the setup you are suggesting works. The code is set up to provide a transition from a -2 to 12 collective in Stabilize mode to a -12 to 12 deg collective in acro. Like you said this would cause the user to have to transition from a mid stick hover to a 3/4 stick hover.  So it was not designed for 3D users.

Hwurzburg: it does not need 5 points, although I did make an error in the math for the 25%/75% points of the STB_COL param...will fix
Screenshot from 2021-06-28 07-13-13

Hwurzburg: I believe someone desiring to ultimately do inverted flight would setup a symmetric min/max collective with 10 or 12 deg pitch extremes and have hover points at 25% and 75% in ACRO....but first would need to get everything tuned with a STABILIZE curve matching your intended defualt with hover at mid stick, matching normal throttle stick position in the autopilot collective controlled modes....then graduate to ACRO with hover points at 25/75....they may or may not decide to change the STABILIZE curve back by changing the STB_COL values to 0/25/75/100... I would , just have to raise throttle to midstick after interlock engage for STABILIZE/ACRO mode to avoid frame stress..
Screenshot from 2021-06-28 07-32-23

Hwurzburg:
@bnsgeyer broke out aerobatic setup....should be okay now...especially since I got my mind unscrewed up

Hwurzburg:
@bnsgeyer Bill, fairly certain everything is addressed...merging

PR #3502, opened 2021-05-03
MAVProxy: clarify direction in MAVLink over IP

arikrupnik:
MAVLink over IP is confusing in how it sets up the client and server. GCS in effect acts as a (network) server: it listens on a specific port; AP acts as a client, sending packets to a known address/port combination. This is the opposite of how it sets up a serial port, where anyone who can establish a serial connection can start listening to MAVLink from AP. For a reader who expects GCS to be the client that connects to an AP server on a known address, the text I'm editing would deepen the confusion.

My copy isn't perfect, and if @Hwurzburg or anyone else has a better way to express this idea, I support that. I'm simply trying to clarify the assumptions about which side is listening and which side need to know the other's IP address.


mavproxy/source/docs/getting_started/quickstart.rst
```diff
@@ -86,7 +86,7 @@ the local IP address. The type of stream (tcp or udp) should also be specified.
     mavproxy.py --master=udp:127.0.0.1:14550
     mavproxy.py --master=tcp:0.0.0.0:14550
     
-If connecting to a remote IP address, the udpout or tcpout arguments should be used:
+If connecting to a remote instance of MAVProxy that distributes telemetry, the udpout or tcpout arguments should be used:
```
khancyr, line 89: hello, this is wrong, you could connect to any remote MAVLink stream, MAVProxy isn't necessary

arikrupnik: Thank you for noting this, @khancyr . If you can enumerate other practical use cases of nodes distributing MAVLink over IP, we should add them here.

Hwurzburg: I routinely use Mission Planner to forward mavlink from autopilot on my windows box to a mavproxy instance on my Linux box...can also be a companion computer remotely...can be an autopilot connected over the cell network

arikrupnik: Thank you, @Hwurzburg and @khancyr . The MAVProxy in my copy id clearly incorrect. I'm removing that.

What I'm unclear on, and what the copy isn't addressing for me, is in what case would the GCS need to know the AP's IP instead of the other way around. Every case I've seen, the sender (AP or proxy) has to know the receiver's IP ahead of time, and the receiver just listens. What's the use case where the receiver needs to specify the sender's IP? Is it for cases where the sender is doing a UDP broadcast, and the receiver wants to listen only to that specific sender and ignore all other senders who are also broadcasting in the same broadcast domain?

arikrupnik: I still have no idea how I would use `--master=udpout:` Have you used it?

Hwurzburg: yes many times....example...on my windows box (ip:196.168.0.41 on my local network), I attach an autopilot....I then forward MAVLink as a UDP host from Mission Planner...then open Mavproxy on my Linux box on the same network using --master=udpout:192.169.0.41:14550  to connect...I can use either gcs to talk to the autopilot...

arikrupnik: Let me see if I'm following this correctly (a PR is a weird place to do this, but I'm learning a lot, so thank you for staying with me).
On your windows machine, you tell MP to forward MAVlink, and specify the IP of the Linux host. On the Linux host, you tell MAVProxy the IP of the windows host to listen to. Each host needs to know the other's IP address. Is that right?

Hwurzburg: no, only the MAVProxy side needs the  ip...the forwarder just listens for anything (heartbeats) coming in on port 14550 (assuming that is the port setup in Mission Planner )...it know then who to send back requested data to (IP packets have destination and source IPs)

Hwurzburg:
overall, I believe the existing wording is correct....I understand that you had misunderstandings with directions....but hopefully that is clearer now....closing

PR #3508, opened 2021-05-07
Scripting: clarifying which parameters apply where

arikrupnik:
I couldn't find any `SRTL_*` parmeters on my vehicle until I realized they didn't apply to planes.


Hwurzburg:
LGTM...thanks....

PR #3509, opened 2021-05-07
Update common-holybro-rtk-f9p.rst

Ionstream:
Added clarification on wire re-crimping instructions for use with Pixhawk 2.1 board.


Hwurzburg:
thanks...merging

PR #3521, opened 2021-05-11
Update Sphinxsetup.sh

khancyr:
Fix python3 and docker and CI


peterbarker:
@khancyr as @brunoolivieri this moves dev envs which *could* be close to the server to be further from it.

Was anything actually broken?  Can we fix Python3 without switching to use it exclusively?

PR #3526, opened 2021-05-12
Docs about ESC based setup and dynamic harmonics

andyp1per:


andyp1per:
Still working on this ...

common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -98,6 +98,39 @@ The harmonic notch is designed to match the motor noise frequency as it changes
 
 .. _common-imu-notch-filtering-helicopter-setup:
 
+ESC based setup
+===============
+
+If ESCs with telemetry are used (either via a separate telemety wire or bi-directional dshot) then the harmonic notch reference frequency can be set dynamically using ESC telemetry.  The harmonic notch reference frequency parameter, :ref:`INS_HNTCH_FREQ <INS_HNTCH_FREQ>`, is used to indicate the lowest motor speed for which the ESC telemetry should be used to dynamically set the harmonic notch reference frequency.  It is recommended that this be set to below the hover frequency but above the control bandwidth.
```
Hwurzburg, line 104: what does control bandwidth mean? you need to refer to a param value, I think....sched freq?

common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -98,6 +98,39 @@ The harmonic notch is designed to match the motor noise frequency as it changes
...
+
+To configure this option set :ref:`INS_HNTCH_OPTS <INS_HNTCH_OPTS>` to 2
+
+Double-notch
+------------
+
+The software notch filters used are very "spikey" being relatively narrow but good at attenuation at their center. On larger copters the noise profile of the motors is quite dirty covering a broader range of frequencies than can be covered by a single notch filter. In order to address this situation it is possible to configure the harmonic notches as double notches that giver a wider spread of significant attenuation. To utilize this feature set :ref:`INS_HNTCH_OPTS <INS_HNTCH_OPTS>` to 1.
+
+Note that each notch has some CPU cost so if you configure both dynamic harmonics and double notches (:ref:`INS_HNTCH_OPTS <INS_HNTCH_OPTS>` set to 3) you will end up with 8 notches on your aircraft per IMU. On flight controllers with 3 IMUs this totals 24 notches which is computationally significant.
```
Hwurzburg, line 125: probably should give some guidance on when NOT to do this or how to tell if computation load is too much...

common/source/docs/common-imu-notch-filtering.rst
andyp1per: Control bandwidth is the frequency at which the controller needs to operate and reflects the size and agility of the copter. Thus for a big, heavy copter the control bandwidth might be 30Hz, for small racer it might be 80Hz. Maybe we could say that a good rule of thumb is that this should be above the configured value for INS_GYRO_FILTER. Basically we don't want the notch filtering either stick or control inputs.

common/source/docs/common-imu-notch-filtering.rst
andyp1per: It also relates to INS_GYRO_RATE. Not sure what we can say here other than give some examples.

Single IMU with dynamic harmonics at INS_GYRO_RATE 0 on an F4 is safe (4 notches @1Khz)
Three IMUs with dynamic harmonics and double notches at INS_GYRO_RATE 1 on an H7 is safe (24 notches @2Khz)

Everything else is in-between. Given time I could probably do some comparisons with --enable-stats turned on.

Hwurzburg: okay, we need to then inform a user as to the bench symptoms of cpu overload, so that they can disable double notching if this occurs...

andyp1per: Only symptom I know of is glitchy motor control

Hwurzburg:
@andyp1per pushed a couple changes with respect to above conversation...if okay, will merge...thanks

andyp1per (commented):
Looks good!

Hwurzburg:
except for the typo...got cocky and pushed before build test completed on my pc...fixing now and will merge

PR #3527, opened 2021-05-12
Fix Sphinxsetup.sh script for Ubuntu 18.04 and 20.04

TunaLobster:
Hit an issue while trying to use the script on bionic WSL 1. I've narrowed the scope of when the script can be used.

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -7,27 +7,42 @@ if [ "$UID" -gt 0 ]; then
      exit 1
 fi
 
-DISTRIBUTION_CODENAME=$(lsb_release -i -s)
-if [ ${DISTRIBUTION_CODENAME} == 'Ubuntu' ]; then
-  add-apt-repository universe
+DISTRIBUTION_ID=$(lsb_release -i -s)
+if [ ${DISTRIBUTION_ID} == 'Ubuntu' ]; then
+  DISTRIBUTION_CODENAME=$(lsb_release -c -s)
+  if [ ${DISTRIBUTION_CODENAME} == 'focal' ] || [ ${DISTRIBUTION_CODENAME} == 'bionic' ]; then
+    add-apt-repository universe
+  fi
 fi
+
 apt-get -y update
-apt-get install -y unzip git imagemagick curl wget make python3 python-is-python3
+apt-get install -y unzip git imagemagick curl wget make python3
+
+# Install packages release specific
+if [ ${DISTRIBUTION_CODENAME} == 'bionic' ]; then
+  apt-get install -y python3-distutils
+elif [ ${DISTRIBUTION_CODENAME} == 'focal' ]; then
+  apt-get install -y python-is-python3
+else
+  apt-get install -y python-is-python3
+fi
 
 # Get pip through the official website to get the lastest release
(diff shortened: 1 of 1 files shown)


TunaLobster:
Something doesn't seem correct with how I've used git on this one. Any helpful hints?

Hwurzburg:
not sure why you are trying to add an existing document...need to rebase perhaps?

TunaLobster:
I think it has to do with the file being in the .gitignore. I have the file locally when I rebased this morning.

TunaLobster:
Somehow my rebase didn't move the files from /dev to /common. Let's try this again.

TunaLobster:
@peterbarker Could you review this change?

peterbarker (commented):
Have those Sphinxsetup.sh changes actually been tested in their current form?

Note that you can bring up a VM to test this with minor modifications to the Vagrantfie.

TunaLobster:
Whoops! That was not great. Tested on WLS 1 with bionic and Vagrant with focal64. Got the expected results. Tried to test in WSL 2 with focal, but could not get do-release-upgrade to behave.

Sphinxsetup.sh
```diff
@@ -7,27 +7,41 @@ if [ "$UID" -gt 0 ]; then
      exit 1
 fi
 
-DISTRIBUTION_CODENAME=$(lsb_release -i -s)
-if [ ${DISTRIBUTION_CODENAME} == 'Ubuntu' ]; then
-  add-apt-repository universe
+DISTRIBUTION_ID=$(lsb_release -i -s)
+if [ ${DISTRIBUTION_ID} == 'Ubuntu' ]; then
```
khancyr, line 11: This isn't good as it is working perfectly on debian

Sphinxsetup.sh
```diff
@@ -7,27 +7,41 @@ if [ "$UID" -gt 0 ]; then
...
 # Get pip through the official website to get the lastest release
 curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py && rm -f get-pip.py
 
 # Install sphinx
-pip install -U sphinx==1.8.3 || pip3 install -U sphinx==1.8.3
+pip install --upgrade sphinx==1.8.3 || pip3 install --upgrade sphinx==1.8.3
 
 # Install sphinx theme from ArduPilot repository
-pip install git+https://github.com/ArduPilot/sphinx_rtd_theme.git -UI ||  pip3 install git+https://github.com/ArduPilot/sphinx_rtd_theme.git -UI
+pip install git+https://github.com/ArduPilot/sphinx_rtd_theme.git --upgrade ||  pip3 install git+https://github.com/ArduPilot/sphinx_rtd_theme.git --upgrade
```
khancyr, line 38: This won't do it, it got a pr that correct that

khancyr (changes requested):

Sphinxsetup.sh
peterbarker: This certainly fixes the problem in front of us - the `-I` was causing Sphinx to be updated to a version that isn't compatible with our configuration files.

Sphinxsetup.sh
peterbarker: Fixed.

PR #3532, opened 2021-05-13
Common: Clarify motor controllers

stephendade:
Based on the discussion at https://discuss.ardupilot.org/t/is-my-skid-steer-configuration-correct-solved/70854, we should clarify which interface to use on motor controllers that have multiple interface options.


rmackay9 (approved):
Thanks for this!

khancyr (approved):

PR #3533, opened 2021-05-13
DShot documentation

andyp1per:


common/source/docs/common-dshot.rst
```diff
@@ -40,17 +43,29 @@ The DShot protocol can run at different speeds. ArduPilot supports four speeds:
...
+
 Reversible DShot ESCs
 ---------------------
 
-Currently, only BLHeli32 style reversible DShot ESCs are supported. In order to use one, the output which drives it must be designated with the appropriate bit in the :ref:`SERVO_BLH_RVMASK<SERVO_BLH_RVMASK>` bitmask parameter. This will map the outputs 1000-1500-2000 values to the correct digital values for the ESC to provide FullReverse-Idle-FullForward range operation, respectively.
+Currently, only BLHeli32 style reversible DShot ESCs are supported. In order to use one, the output which drives it must be designated with the appropriate bit in the :ref:`SERVO_BLH_3DMASK<SERVO_BLH_3DMASK>` bitmask parameter. This will map the outputs 1000-1500-2000 values to the correct digital values for the ESC to provide FullReverse-Idle-FullForward range operation, respectively.
+
+If the craft has been setup for DShot commands then ArduPilot will supply the correct command at startup in order to set the ESCs in reversible mode. In a similar fashion outputs can be reversed by setting :ref:`SERVO_BLH_RVMASK<SERVO_BLH_RVMASK>` without any changes needing to be made through ESC setup software (e.g. BLHeliSuite).
```
Hwurzburg, line 68: Sorry, I am confused by this...when do you use SERVO_BLH_3DMASK vs SERVO_BLH_RVMASK? for reversing escs

common/source/docs/common-dshot.rst
```diff
@@ -40,17 +43,29 @@ The DShot protocol can run at different speeds. ArduPilot supports four speeds:
...
+
+DShot Commands
+--------------
+
+On certain ESCs DShot commands are supported. These allow functions such as ESC LEDs, beeps and motor direction to be manipulated by the flight controller. In order to use DShot commands :ref:`SERVO_DSHOT_ESC<SERVO_DSHOT_ESC>` must be set to the type of ESC that is in use. Notify functions (e.g. LEDs and Buzzer) can then be configure to include DShot as an output type.
```
Hwurzburg, line 61: what happens if you set this and the ESC does not support the commands or not all escs support them? nothing?

common/source/docs/common-dshot.rst
andyp1per: 3DMASK is for reversible ESCs i.e they can go backwards as well as forwards. RVMASK is for ESCs you want to physically run in the opposite direction (e.g. because you wired them up wrong)

common/source/docs/common-dshot.rst
andyp1per: The command set for other ESCs is different so undefined - so its important to set the type correctly

Hwurzburg: but right now there is only one type of command set....if set to 1 for an ESC wihtout Dshot command capability, what happens? and if set to 1 for an ESC that does not have the command set ("BLHeli32/BLHeli_S/Kiss" ) what happens? smoke? explosions? nothing?

common/source/docs/common-dshot.rst
Hwurzburg: aha! is this clearer?
```suggestion
If the craft has been setup for DShot commands then ArduPilot will supply the correct command at startup in order to set the ESCs in reversible mode.

 In a similar fashion, normal output rotation direction can be reversed by setting :ref:`SERVO_BLH_RVMASK<SERVO_BLH_RVMASK>` without any changes needing to be made through ESC setup software (e.g. BLHeliSuite). This can also be used on ESCs with forward and reversed active operation, ie reversible ESCs to set the "forward" directions rotation.
```

common/source/docs/common-dshot.rst
andyp1per: It depends on the ESC - on some nothing, on others maybe a beep instead of an LED flash. This only came up because the ESCs that tridge is working with have a slightly different command set. I suppose the worse that could happen is that 3D is selected - or not selected - unintentionally.

Hwurzburg: sorry to keep on this, but are there any features , besides notify, that require the SERVO_DSHOT_ESC be set to 1? you imply reversible(3D)....or can this be set using passthru?if so it should be noted also...
perhaps something like this should be added: " Currently, ArduPilot supports the command set (SERVO_DSHOT_ESC=1) that is commonly used, however, others are appearing and may not be compatible, resulting in undefined operation. Use caution (remove blades!) until correct operation using type=1 is verified for your ESC."

andyp1per: Seems reasonable. Currently supported features:
- Dshot LEDs via NTF_LED_TYPES
- Dshot Buzzer via NTZ_BUZZ_TYPES
- Reverse motor direction via RVMASK
- Enable 3D operation via 3DMASK

Both of the latter can be set via passthru but obviously requires a different tool be fired up so more convenient this way

Hwurzburg:
@andyp1per pushed a few changes, corrected some params incorrectly named, squashed....if okay, will merge...let me know

common/source/docs/common-dshot.rst
```diff
@@ -15,16 +15,19 @@ Ardupilot supports several communication protocols with brushless motor ESCs:
 - Oneshot125 (PWM pulse widths divided by 8)
 - DShot150 at 150kbaud (recommended)
 - DShot300 at 300kbaud
-- DShot600 at 600kbaud
+- DShot600 at 600kbaud (recommended)
 - DShot1200 at 1200kbaud
+- Bi-directional DShot at 150, 300, 600 and 1200kbaud on supported firmware
 
-These are selected by the :ref:`MOT_PWM_TYPE <MOT_PWM_TYPE>` parameter on Copter or :ref:`Q_M_PWM_TYPE <Q_M_PWM_TYPE>` on QuadPlanes for copter function motors. On Plane, other motors use Normal (PWM) protocol.
+These are selected by the :ref:`MOT_PWM_TYPE <MOT_PWM_TYPE>` parameter on Copter or :ref:`Q_M_PWM_TYPE <Q_M_PWM_TYPE>` on QuadPlanes for copter function motors. On Plane, other motors use Normal (PWM) protocol. Bi-directional DShot is selected by selecting DShot in :ref:`MOT_PWM_TYPE <MOT_PWM_TYPE>` and then selecting the required channels through :ref:`SERVO_BLH_BDSMASK <SERVO_BLH_BDMASK>`.
```
andyp1per, line 22: should be SERVO_BLH_BDMASK

andyp1per (commented):
One small error in naming

common/source/docs/common-dshot.rst
Hwurzburg: right....build did not catch that typo since its the label not link....

Hwurzburg:
thanks Andy! merging

PR #3557, opened 2021-05-22
Expand TradHeli info to include tail and esc setup

Hwurzburg:


copter/source/docs/traditional-helicopter-connecting-apm.rst
```diff
@@ -84,3 +95,13 @@ Accelerometer Calibration
...
 If the accelerometers were not calibrated on the bench prior to installation it must be calibrated before proceeding.(It is usually easier to calibrate on the bench and then re-calibrate only the level position, if required, once installed.)
 :ref:`See this topic <common-accelerometer-calibration>`.
+
+ESC Calibration
+---------------
+
+
+.. warning:: be sure to remove all blades when doing ESC calibration.
+
+If an ESC is used, it must be calibrated to the throttle range. In addition, it is often required to change settings, such as enabling the governor mode and/or setting voltage protection levels.
```
bnsgeyer, line 105: I disagree with this.  If an esc is used it does not need to be calibrated if you are using its internal governor.  I suppose it would need to be calibrated if you are using the throttle curve or throttle curve/ardupilot gov.

Hwurzburg: Since the output low values of ESCs vary, a value of 1000 (default) may or may not be within its actuation and/or lockout value (some escs lock out until power cycle if the value is out of range)....

bnsgeyer: You are talking about the value that actually starts the motor to spin.  The ESC regardless of when the motor starts to spin reads PWM.  My Castle creation ESC says that I need to set it at 90% for one governor speed and 60 for another and 30 for the 3rd speed.  I have never had a problem with the initialization of the ESC.  

So you are telling me that regardless of whether I am using the internal governor or a throttle curve, I should calibrate my ESC.  What am I doing in calibrating?  I'm setting the limits of my servo output to when the motors starts (min) and reaches max speed (max)?  Or I am I telling the ESC the limits that I have set for my servo output? i.e I want it to start turning the motor at 1000 PWM and reach max at 2000 PWM.

Hwurzburg: when you calibrate the ESC, it sets the input PWM range for motor off (it ill start a few tens of PWM above that) and max output , that range will also be used by the ESC to set valid PWM input range. Depending on the ESC it may after boot if PWM is outside that range (and the tolerance varies by ESC) it may lock up to prevent operation until power is cycled. Also we have escs that just plain fail with ArduPilot boot because we take too long to start pulses and you can never calibrate or use them...I dont know for sure, but I think that the Align internal governor uses percentage of established input range instead of absolute PWM for speed setpoint....never used or programmed a Castel ESC....I have alwasy calibrated ESCs on any vehicle.

Hwurzburg: @bnsgeyer made changes we discussed and merged...thanks

PR #3558, opened 2021-05-22
GPS: add paragraph on GPS_AUTO_SWITCH

IamPete1:


common/source/docs/common-positioning-landing-page.rst
```diff
@@ -40,6 +40,22 @@ positioning technologies:
...
+    Always use the primary GPS, this can be either the first or second as set with :ref:`GPS_PRIMARY <GPS_PRIMARY>`
+
+- 1: Use best
+    Automatically select the best GPS, this is done based on GPS fix status (2D / 3D / rtk) if both GPSs have the same fix status the one with the larger number of satellites is used.
+
+- 2: Blend
+    Blend is best suited for use with two identical GPS units, see :ref:`GPS blending<common-gps-blending>`
+
+- 4: Use primary if 3D fix or better
+    Use primary GPS as set with :ref:`GPS_PRIMARY <GPS_PRIMARY>` if it has a 3D fix or better. This should be used with two dissimilar GPS units are used, one high quality primary unit and a secondary less accurate unit. In this case the
```
Hwurzburg, line 55:
```suggestion
    Use primary GPS as set with :ref:`GPS_PRIMARY <GPS_PRIMARY>` if it has a 3D fix or better. This should be used when two dissimilar GPS units are used, one high quality primary unit, and a secondary less accurate unit. In this case the 
```

common/source/docs/common-positioning-landing-page.rst
```diff
@@ -40,6 +40,22 @@ positioning technologies:
...
+
+- 1: Use best
+    Automatically select the best GPS, this is done based on GPS fix status (2D / 3D / rtk) if both GPSs have the same fix status the one with the larger number of satellites is used.
+
+- 2: Blend
+    Blend is best suited for use with two identical GPS units, see :ref:`GPS blending<common-gps-blending>`
+
+- 4: Use primary if 3D fix or better
+    Use primary GPS as set with :ref:`GPS_PRIMARY <GPS_PRIMARY>` if it has a 3D fix or better. This should be used with two dissimilar GPS units are used, one high quality primary unit and a secondary less accurate unit. In this case the 
+    high quality GPS will often have a better quality fix even if it has fewer satellites. Using 'Use best' in this case would result in using the poor quality GPS and resulting in more switching between GPS units. Unlike 'Use Primary' this option
```
Hwurzburg, line 56:
```suggestion
    high quality GPS will often have a better quality fix even if it has fewer satellites. Using 'Use best' in this case would result in using the poorer quality GPS and result in more switching between GPS units. Unlike 'Use Primary' this option 
```

Hwurzburg (changes requested):
How does EKF3 affinity work with this? since you can assign the two gps to  whatever IMU lane, what happens when you have them independently assigned? if one is the primary and fails can it switch to the second, even though its assigned to another lane?can you assign the primary to all lanes and they fall back to secondary?

IamPete1:
I have added a line on EKF affinity, but not gone into depth of all the possible combinations. But in answer to your questions yes I think you can do all of that if you setup affinity appropriately.

Hwurzburg:
> I have added a line on EKF affinity, but not gone into depth of all the possible combinations. But in answer to your questions yes I think you can do all of that if you setup affinity appropriately.

Based on Tridge's response, the last addition seems to need work/expansion since both can be made to work together it seems...

common/source/docs/common-positioning-landing-page.rst
```diff
@@ -40,6 +40,24 @@ positioning technologies:
 
 .. note:: It is important that a GPS be connected to the first SERIALx port that has its ``SERIALx_PROTOCOL`` parameter set to "5" (GPS) since it will stop searching for GPS during bootup if not found on the first port configured for GPS protocol.
 
+When using two GPS unit there are a number of switching options that can be selected with :ref:`GPS_AUTO_SWITCH <GPS_AUTO_SWITCH>`.
```
nkruzan, line 43: Units

common/source/docs/common-positioning-landing-page.rst
```diff
@@ -40,6 +40,24 @@ positioning technologies:
...
+
+- 2: Blend
+    Blend is best suited for use with two identical GPS units, see :ref:`GPS blending<common-gps-blending>`
+
+- 4: Use primary if 3D fix or better
+    Use primary GPS as set with :ref:`GPS_PRIMARY <GPS_PRIMARY>` if it has a 3D fix or better. This should be used when two dissimilar GPS units are used, one high quality primary unit, and a secondary less accurate unit. In this case the 
+    high quality GPS will often have a better quality fix even if it has fewer satellites. Using 'Use best' in this case would result in using the poorer quality GPS and result in more switching between GPS units. Unlike 'Use Primary' this option 
+    still allows falling back to the secondary GPS if 3D fix is lost on the primary.
+
+An alternative option to GPS switching is :ref:`EKF3 affinity and lane switching <common-ek3-affinity-lane-switching>`, a EKF lane can be setup to use each GPS the whole EKF lane is then switched based on its health rather than judging the GPS in isolation.
```
Hwurzburg, line 59: perhaps this:?
```suggestion
An additional variation with GPS switching is :ref:`EKF3 affinity and lane switching <common-ek3-affinity-lane-switching>`. An EKF lane can be setup to use either GPS and the whole EKF lane is then switched based on its health. If the :ref:`GPS_PRIMARY <GPS_PRIMARY>` is used for a lane, and :ref:`GPS_AUTO_SWITCH <GPS_AUTO_SWITCH>` is enabled, then the lane switch will occur only if both GPS are unhealthy.
```

Hwurzburg (changes requested):

common/source/docs/common-positioning-landing-page.rst
nkruzan, line 43:
```suggestion
When using two GPS units there are a number of switching options that can be selected with :ref:`GPS_AUTO_SWITCH <GPS_AUTO_SWITCH>`.
```

PR #3559, opened 2021-05-22
Airspeed: remove APM2 reference and add paragraph on WIND_MAX and failure.

IamPete1:


Hwurzburg:
thanks! added a few tweaks...

PR #3581, opened 2021-05-29
feat: Adds link to flashcards

darigovresearch:
Thought it would be helpful to learn key topics to get a head start with ArduPilot so thought it would be helpful for other people as well to link it from the docs

### DIFF
--- a/common/source/docs/common-glossary.rst
+++ b/common/source/docs/common-glossary.rst
@@ -9,6 +9,14 @@ Glossary
    The glossary is not indexed. It is in alphabetical order so just
    scroll down to find the topic of interest.
 
+.. note::
+
+   A `set of flashcards <https://www.darigovresearch.com/ardupilot-glossary-flashcards>`_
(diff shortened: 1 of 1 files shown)


khancyr (commented):
Thanks for the idea but I don't think this is usefull as it rely on other website and need a special program... So that isn't worth it

darigovresearch:
No worries, thanks for the feedback @khancyr. For context Anki is incredibly popular in various industries including robotics, engineering, mathematics and physics for it's feature rich ability such as images, HTML, LaTeX and more. It's a free & open source app so can be used by anyone including a free online website as well. An example of a similar project being done for another open source project which was merged (and also coincidentally also uses read the docs) can be found here https://huggingface.co/transformers/community.html

When you say `as it rely on other website` which other website do you mean?

For clarification this is our website so it is as assured as any website can be, but in principle can link to the GitHub page instead however this has less context than the website currently linked in the pull request.

However totally would understand if even given all the givens you and other maintainers wish to not merge this pull request

PR #3583, opened 2021-05-30
copter: Change to the URL of the maintenance company

muramura:
Japanese drone manufacturer enRoute has transferred this vehicle maintenance to NTT e-Drone Technology.

### DIFF
--- a/copter/source/docs/sprayer.rst
+++ b/copter/source/docs/sprayer.rst
@@ -17,9 +17,9 @@ Required Hardware
 =================
 
    .. image:: ../images/sprayer_EnRoute_AC940D.jpg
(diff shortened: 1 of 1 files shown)


nkruzan:
This seems to be a poor destination to send someone there is no information showing just a picture? I am using translate to view page perhaps that is the reason. Is there a different product page that could be used instead ?

muramura:
This aircraft has been discontinued.  There is no description page for the aircraft.  Therefore, if the content of pr is ng, I think it is better to delete the link.

rmackay9:
I think @muramura is the owner of this really so merged!

PR #3584, opened 2021-05-30
Add stab collective curve info to swash setup

Hwurzburg:


copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -155,9 +155,45 @@ that produces zero thrust. In Copter 3.6 and ealier, the minimum collective pitc
 could command was done with the parameter ``H_COL_LAND_MIN``. 
 
 Set :ref:`H_SV_MAN <H_SV_MAN>` to 3 to have the swashplate move to the mid position.  Adjust :ref:`H_COL_MID <H_COL_MID>` until the blade
-pitch angle measures the desired collective blade pitch that corresponds to zero thrust.  
+pitch angle measures the desired collective blade pitch that corresponds to zero thrust.
 
+.. note:: If the minimum or maximum collective is measured in ACRO (see below for impact of ACRO on collective pitch curve vs stabilized modes), the value will be less than those since some servo movement is reserved for cyclic. Of course, any swashplate tilt due to pilot or autopilot input will interefere with the measurements also.
```
bnsgeyer, line 160: What is this supposed to mean? “The value will be less than those”. The value of what?  I believe in the SV_MAN, cyclic is locked out in min, mid and max positions. Why does cyclic factor into this discussion?

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -155,9 +155,45 @@ that produces zero thrust. In Copter 3.6 and ealier, the minimum collective pitc
...
 
+ACRO versus STABILIZED Collective Curves
+========================================
+
+ArduPilot provides a programmable curve for stabilized modes (all except ACRO). This curve is set by the ``IM_STB_COL_x`` parameters as shown in the graph below. Only ACRO mode uses the collective extremes that were setup above.
```
bnsgeyer, line 165: The collective curve is for stabilized mode only. It is not used for any other mode

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -155,9 +155,45 @@ that produces zero thrust. In Copter 3.6 and ealier, the minimum collective pitc
...
+.. image:: ../images/collective_curves.png
+   :target: ../_images/collective_curves.png
+
+By default, these settings mirror the ACRO mode curve, and are satisfactory if the -2° to +10° range is setup. This will give the following values:
+
+==============      =====
+Throttle Stick      Pitch
+==============      =====
+Low                 -2
+Mid                 +4 (close to hover)
```
bnsgeyer, line 176: To reduce confusion with the collective mid (H_COL_MID), I would suggest you replace mid with center in your tables

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -155,9 +155,45 @@ that produces zero thrust. In Copter 3.6 and ealier, the minimum collective pitc
...
+==============      =====
+
+
+However, if the extremes were setup for full aerobatic, the defaults for ``IM_STB_COL_x`` would yield:
+
+==============      =====
+Throttle Stick      Pitch
+==============      =====
+Low                 -10
+Mid                 +0
```
bnsgeyer, line 187: Same here

copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -155,9 +155,45 @@ that produces zero thrust. In Copter 3.6 and ealier, the minimum collective pitc
...
+High                +10
+==============      =====
+
+which is fine for ACRO, but the auto-collective controlled modes, like AUTO and HOVER, would be overly sensitive to collective. Therefore it is recommended to change the values from default to something like:
```
bnsgeyer, line 191: Hover?  This is not a mode. Also the stabilize collective curve is not used for auto-collective modes. The use the full collective range. They are set to the reduced range for an aerobatic setup because stabilize is a 2D mode and doesn’t need that much negative collective.  It is really that the full range is too sensitive for the pilot

bnsgeyer (changes requested):

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: if H_SRV_MAN =0 then you will not get max and min collective at the extremes of stick movement  in ACRO, the only mode that has full collective range, irrespective of the IM_STB_COL params...will make it clearer

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: then it is not used in the other stabilized modes, which is every mode besides ACRO? I thought they would inherit from stabilized mode, otherwise, you could be running different curves depending on the history of how and when ACRO was entered..if this is the case, I think it is incorrect behavior....

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: will do

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: sorry ALTHOLD (plane guy)....and I am confused by the contradiction in "They are set to the reduced range for an aerobatic setup because stabilize is a 2D mode and doesn’t need that much negative collective"...

copter/source/docs/traditional-helicopter-swashplate-setup.rst
bnsgeyer: Well they shouldn’t be measuring collective blade pitch with SV_MAN = 0. I think you should just state that measurement of collective blade pitch will not be accurate with SV_MAN disabled because the autopilot is making swashplate inputs according to control laws or something to that end

copter/source/docs/traditional-helicopter-swashplate-setup.rst
bnsgeyer: Yes all the other modes use the stabilize mode for pitch roll and yaw but the non manual throttle modes or auto collective modes do not use the collective control range of stabilize. They are given the full range that is given to acro.

copter/source/docs/traditional-helicopter-swashplate-setup.rst
bnsgeyer: The stabilize collective curve is provided so the pilot can set a non aerobatic collective range in stabilize mode. That all it is there for.

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: so the curve selection is determined by STABILIZE or ACRO mode entry, and it use the Input handler for non-auto-colllective modes? so DRIFT mode (the other piilot controlled collective mode)  curve is determined by whatever was last used ACRO or STABILIZE?

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: okay so no problem using the full range for aerobatic setups for all the non pilot controlled collective modes....assumed that the softer (in aero setups) stabilize range would be used....will redo the explanations...but does DRIFT (which is like STABILIZE with extras) work correctly ? ie use STABILIZE collective curve?

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: will do...

copter/source/docs/traditional-helicopter-swashplate-setup.rst
bnsgeyer: It reverts to the full collective range from min to max. There is no special curve or handler

copter/source/docs/traditional-helicopter-swashplate-setup.rst
Hwurzburg: fixed

PR #3590, opened 2021-06-02
Nitpicking

arBalasquide:
Only proper concern is that the wiki states to set the drone to 5ft in altitude but that could be too low in some cases. Everything else is just my opinion on what reads better.

### DIFF
--- a/copter/source/docs/ac2_followme.rst
+++ b/copter/source/docs/ac2_followme.rst
@@ -28,45 +28,39 @@ Instructions for Mission Planner
 ================================
 
 #. Set one of your flight modes to "Loiter"
-#. Set up your Copter at the field and establish a MAVLink connection
+#. Set up your Copter and establish a MAVLink connection
    over wireless telemetry
 #. Ensure that your GPS USB dongle or Bluetooth device is plugged into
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):

copter/source/docs/ac2_followme.rst
```diff
@@ -28,45 +28,39 @@ Instructions for Mission Planner
...
-#. In the Mission Planner, enter Control-F, which will open the
+#. In Mission Planner, enter Control-F to open the
    following window. Click on "Follow Me"
 
 .. image:: ../images/followme1.jpg
     :target: ../_images/followme1.jpg
 
--  This will bring up this window. Select the serial port that is
-   assigned to your GPS device and whatever baud rate it uses.
+-  This will bring up this window. Select the serial port that's assigned to your GPS device and the baud rate it uses.
```
Hwurzburg, line 48: we autobaud...so
```suggestion
-  This will bring up this window. Select the serial port that's assigned to your GPS device.
```

Hwurzburg (changes requested):

Hwurzburg (approved):
we need to update the images....will put it on my list...

PR #3597, opened 2021-06-08
Add note regarding airspeed sensor use to Soaring page

Hwurzburg:


samuelctabor:
Looks good - add the same to `plane/source/docs/soaring.rst `?

PR #3602, opened 2021-06-10
plane: added a note on CRUISE mode

tridge:
I added this note after reviewing a flight log where a user flew a large quadplane too close to the ground in CRUISE mode


Hwurzburg:
LGTM..thanks...merging

PR #3604, opened 2021-06-15
wiki: fix link

fagoth:


fagoth:
May be should add a new link to 16.9 version? https://github.com/JazzMaverick/BLHeli/tree/JazzMaverick-patch-1/BLHeli_S%20SiLabs/Hex_files_16.9

Hwurzburg:
thanks

fagoth:
Found on forums that "At the moment, the 16.73 version is the only recommended version that properly supports RPM filtering.", newer versions are very experimental.

Hwurzburg:
okay...merging as is

PR #3620, opened 2021-06-21
add line break in common-video-landingpage

NhanNgo2306:
add line break in common-video-landingpage


Hwurzburg:
thanks....I decided that all the sections on this page really needed clear separators....so I added them....so this PR is no longer needed, but thanks for bringing this to our attention..

PR #3628, opened 2021-06-27
Re-org heli tuning page

Hwurzburg:
The tuning page presently is huge and too long a read for a single page. This re-orgs it a bit, separating the control system theory description to a separate page for users not already knowledgeable and removing it from the tuning page for PID and PIDFF knowledgeable users . The advanced tuning setup items are also moved to a separate page.
also:
- unified the area for flybarred heads
- added my PID loop diagram used elsewhere, modified appropriately and a post tune graph for YAW (intentionally slightly imperfect but flyable)...I need a good ATT.DesP vs ATT.P graph, post tune also, as well as Bill's preferred FF calculation method from logs (we match a PID totals expression vs FF*PID.Act for Quadplane)
- substituted "demanded" or "requested" rate for "predicted" since this is consistent with the rest of the wiki and we really dont predict attitude or rate in the PID loops....actual/sensed vs demanded/desired/requested drives the control loop outputs
- moved some things to make the flow the same for pitch/roll and yaw, ie table position, etc.
- update the  ILMI for tune start in the tables
- general formatting


bnsgeyer:
@Hwurzburg i am away this week and won’t have much time to review this. There is some terminology that you and I will need to discuss.  I would agree that maybe predicted  is not the right word but requested is different from target or demanded. I will have to look over copter terminology.

Hwurzburg:
minor updates of  pid diagrams and description based on conversations with Leonard as to proper terminology (changed actual to measured) and added some more post tune diagrams...still need to decide what to use for FF calculations from logs....too bad RATE.xOut does not include FF terms, could use FF times it vs RATE.x instead of Sum of individual PIDs vs FF*PID.Act (my understanding from Tridge), would be useful in Plane also

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -175,19 +103,44 @@ problem. Correct the problem before proceeding with roll and pitch tuning.
...
 roll. Land and pull the log from the microSD card and look at the signals in
 your ground station software. If the actual rate is more than the desired rate
-then you'll want to decrease FF. If it is less, increase FF. If the desired
-and actual rates are offset by some amount it means that your swash was not
-properly leveled in the setup or the CG is not right.  In this case, just make
-sure the change in rate is similar between desired and actual.  If you get the
+then you'll want to decrease FF. If it is less, increase FF.
+
+**(to be added...FF calculation directly from log data)**
+
```
bnsgeyer, line 140: See the [tuning guide](https://drive.google.com/file/d/1BBnF2jPyAN-AXjm39TO_NW3uQFiDw-sZ/view?usp=sharing) that I put together but just haven't put it into this wiki.  If you want to try it and modify the wiki from it, that's fine.  What I wanted you to look at in particular is the VFF tuning on slides 7-9.  You can show the mavexplorer method once they perform this maneuver but the maneuver is what I really wanted you to see.  Tell the user to set the max roll or pitch rate to 50 deg/s and then go fly and perform a maneuver that goes from 30 deg in one direction to 30 deg  in the opposite using full stick in Stabilize mode.  The command model will limit the rate to 50 deg/s and the mavexplorer can be used afterwards to determine the best value of VFF.

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -315,86 +260,28 @@ discrepancy between desired and actual pitch and roll attitude when the
...
+**Note on Tuning Flybar Helicopers** - for flybar heads the flybar is a
+mechanical version of the rate PID loop. So flybar is tuned using only FF in
+pitch and roll. The rate D and P gains mentioned in the next topic below are
+left set to zero for flybar. I-gain, IMAX, and ILMI are tuned just like FBL.
+Below is a plot of beginning a tune on a flybar helicopter using the starting
+setting of :ref:`ATC_RAT_RLL_VFF<ATC_RAT_RLL_VFF>` = 0.22. The graph shows the aircraft's response to
+the rate request of the attitude controller is low, meaning the FF value must
+be increased to achieve proper rate response. For all flybar helicopters be sure
+to set :ref:`H_FLYBAR_MODE<H_FLYBAR_MODE>` = 1
```
bnsgeyer, line 284: Flybar mode is a bit misleading.  It only provides direct passthrough for acro mode.  In stabilize and all other modes, it uses the normal command model and PID archetecture.

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -7,78 +7,13 @@ Traditional Helicopter – Tuning
...
-its side.  So the integrator leak along with the leak minimum parameter keep 
-enough of the integrator to make it effective in keeping the attitudes matching
-but not so powerful to cause the aircraft to roll over.
+.. warning:: For making setting changes to traditional helicopters, users are reminded to use only the Full or Complete Parameter List, or the Heli SETUP page in your ground station software. **Do not use the Basic, Extended or Advanced Tuning pages that are designed for multi-rotor aircraft in Mission Planner.** These pages will make unwanted setting changes to traditional helicopters. And remember to write the changes to the flight controller after making them or they won't be saved!
+
+.. note:: If the reader is unfamiliar with PID control systems then reading :ref:`traditional-helicopter-control-system` is recommended.
 
 Initial Setup of Pitch and Roll Tuning Parameters
 =================================================
+
```
bnsgeyer, line 16: In the Table below.  We should break out the ATC_RAT_XXX_FLTT and ATC_RAT_XXX_FLTD filter cut off frequencies as well.   Recommend that ATC_RAT_XXX_FLTD be set to zero and ATC_RAT_XXX_FLTT be set to 20 hz.  This is  defaulted to 10 hz currently but I think it is too low and adds to the time delay in the response.

bnsgeyer (changes requested):
@Hwurzburg thanks for the push on this.  I have some suggested changes.  Let me know if you want to sit down and talk about them.

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -129,28 +64,21 @@ from the defaults.
...
 It is recommended to make sure the tail functions properly before proceeding
 with tuning pitch and roll.
 
 **Important Note** - UAV helicopters, as opposed to sport helicopters, will
 usually be running low headspeed and higher disc loading. With a mechanically
 driven tail this also means lower than normal tail speed and reduced tail
 authority. If your helicopter meets this description, it is recommended to set
-:ref:`ATC_RAT_YAW_VFF<ATC_RAT_YAW_VFF>` to 0.05 before the first test hover.
+:ref:`ATC_RAT_YAW_VFF<ATC_RAT_YAW_VFF>` to 0.024 before the first test hover.
```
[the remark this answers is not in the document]
Hwurzburg: thought it was a typo since the table below says use 0.024.....but now I see that the note should be placed AFTER the table as an exception....will move below and keep the 0.05...

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -175,19 +103,44 @@ problem. Correct the problem before proceeding with roll and pitch tuning.
 
 * This param name changed in ArduPilot 4.0 and later.
 
+Spool up the heli and hover it
+no more than .25 meters above ground in Stabilize flight mode and test the
+starting tail settings. If the tail seems "loose" and doesn't want to hold
+increase the :ref:`ATC_RAT_YAW_P<ATC_RAT_YAW_P__AC_AttitudeControl_Heli>`. If the tail slowly wags back and forth, the :ref:`ATC_RAT_YAW_P<ATC_RAT_YAW_P__AC_AttitudeControl_Heli>` may be too high. If the tail rapidly shakes side to side reduce the
+:ref:`ATC_ANG_YAW_P<ATC_ANG_YAW_P>`.
```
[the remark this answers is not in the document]
Hwurzburg: seemed to work for me.....I thought in general P oscillations are slow as opposed to D oscillations (not applicable here) and the original had ANG_P as a factor in fast shakes, which makes sense to me in that its directly a gain on angle error to rate demand.....but I never got fast shakes only wags...I never messed with the ANG_P, only got the wags out with RAT_YAW_P....thats why I added the slow wag statement

copter/source/docs/traditional-helicopter-tuning.rst
Hwurzburg: its exactly the same math as what we do for figuring out pitch ff in VTOL.....but it can be done without holding the full stick for any length of time....you dont have to hold it, just demand it and get it close....the FF is the same at any rate point, its a linear expression..otherwise limiting the max rate at full stick like you suggest would mess with the outcome

....then we take the PID total output (ROut...but we cant use that since ours does not have FF in it, so we graph the sum of the PIDS) * proposed FF and try to match to actual RATE graphs....getting the amplitudes matched by trail of FF values...only take a couple of guesses..that way rate doesnt have to be maxd out and held..I dont think you need to limit the rates, you can just rock it back and forth as hard as you are comfortable with, but its a safe to limit them....but fear limits it pretty well, haha...I dont have any objection to either method, limited rate or no, graphical match or math....my only issue with the math is that it requires eyeballing for a value, whereas overlaying the two graphs could be a bit more precise since the eye is better at declaring a match rather than estimating an extended line to an axis... but am happy to put either method in...perhaps do the math and then re-graph with the FF*ROut*57.3  vs R and graphs should ovelay as a double check on the math and eyeballing of R and ROut values

copter/source/docs/traditional-helicopter-tuning.rst
Hwurzburg: will do

copter/source/docs/traditional-helicopter-tuning.rst
bnsgeyer: > .the FF is the same at any rate point, its a linear expression..otherwise limiting the max rate at full stick like you suggest would mess with the outcome

Its not a linear response.  We want to match the steady state rate.   Now there is probably some don't let perfection get in the way of good enough factor here.  I prefer they do the maneuver to evaluate the amount of VFF required for steady state response.  and they should probably look at it in both directions.  If they didn't trim the swashplate correctly, then this will show that pretty distinctly.  so either they take the time to trim it properly or use an average of the VFF they found in each direction.

copter/source/docs/traditional-helicopter-tuning.rst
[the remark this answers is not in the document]
bnsgeyer: yep that would be good.  Thanks.

copter/source/docs/traditional-helicopter-tuning.rst
```diff
@@ -315,86 +260,28 @@ discrepancy between desired and actual pitch and roll attitude when the
...
+With a flybar head, where the linkage rate is normally lower, it is recommended
+to start with 0.22 FF for both pitch and roll and you will likely have to go
+higher with FF. But for a flybarless head, FF shouldn't be more than 0.22 
+unless you have really really slow servos or slow linkage rate. With all 
+helicopters, the FF gain compensates for differences in servo and linkage
+speed. 
+
+**Note on Tuning Flybar Helicopers** - for flybar heads the flybar is a
+mechanical version of the rate PID loop. So flybar is tuned using only FF in
+pitch and roll. The rate D and P gains mentioned in the next topic below are
```
bnsgeyer, line 278: It references rate P and D gain sections which are not after this section.  I would just remove the reference and just tell the reader to set the Rate P and D gains to zero.

copter/source/docs/traditional-helicopter-tuning.rst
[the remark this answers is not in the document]
bnsgeyer: I am working off of your branch.  I fixed this.

copter/source/docs/traditional-helicopter-tuning.rst
bnsgeyer: I will add the additional params and my current recommendations

copter/source/docs/traditional-helicopter-tuning.rst
bnsgeyer: fixed.

copter/source/docs/traditional-helicopter-tuning.rst
bnsgeyer: reworked this

copter/source/docs/traditional-helicopter-tuning.rst
bnsgeyer: done

bnsgeyer:
@Hwurzburg I rebased and added a commit with the discussed changes above.  I force pushed to your branch

Hwurzburg:
merging

PR #3633, opened 2021-06-30
Plane: Change verbage for FBWA attitude holding

TunaLobster:
I'm not sure how much this section of the docs is actually useful. FBWA should be for maintaining attitude and FBWB should for be height. This change at least makes it clearer that FBWA is for attitude.

### DIFF
--- a/plane/source/docs/starting-up-and-calibrating-arduplane.rst
+++ b/plane/source/docs/starting-up-and-calibrating-arduplane.rst
@@ -149,7 +149,7 @@ List** and adjusting the parameters as shown in the screenshot below.
 
    These parameters are in radians (every 0.01 is about 0.6 of a
    degree) so adjust in increments of 0.01 initially. If the plane turns to
-   the left, :ref:`AHRS_TRIM_X<AHRS_TRIM_X>` should be increased. If the nose of the plane
+   the left, :ref:`AHRS_TRIM_X<AHRS_TRIM_X>` should be increased. If the nose of the plane is too low for a given throttle setting (ie flying too fast while holding altitude)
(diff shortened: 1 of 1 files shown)


Hwurzburg:
To be truthful I have never carefully read that page after the bootup stuff, it needs a lot of work....not how I would recommend the flight sequences....checking navigation should be the last thing done....getting cruise set correctly and trimmed should be done first....and not in the manner suggested.....I had reworked the acc cal instructions previously to explain how to get cruise setup with and without airspeed....needs to be referred to here....should not be messing with AHRS trims.....will work on this after I return from holiday

plane/source/docs/starting-up-and-calibrating-arduplane.rst
```diff
@@ -152,7 +149,7 @@ List** and adjusting the parameters as shown in the screenshot below.
 
    These parameters are in radians (every 0.01 is about 0.6 of a
    degree) so adjust in increments of 0.01 initially. If the plane turns to
-   the left, :ref:`AHRS_TRIM_X<AHRS_TRIM_X>` should be increased. If the plane loses height
-   with mid throttle, :ref:`AHRS_TRIM_Y<AHRS_TRIM_Y>` should be increased.
+   the left, :ref:`AHRS_TRIM_X<AHRS_TRIM_X>` should be increased. If the nose of the plane
+   dips down, :ref:`AHRS_TRIM_Y<AHRS_TRIM_Y>` should be increased.
```
Hwurzburg, line 153:
```suggestion
  , :ref:`AHRS_TRIM_Y<AHRS_TRIM_Y>` should be increased.
```

Hwurzburg (changes requested):
since it will be a while for me to get to redoing all this with a cruise setup page, lets merge this with the minor changes suggested...

PR #3636, opened 2021-07-05
Build: Use more recent version of Sphinx

stephendade:
This appears to fix the CI build errors


khancyr (commented):
Why not the lastest version ? It is working too

Hwurzburg:
closing since Peter merged his fix

PR #3643, opened 2021-07-07
Update airspeed sensor page

Hwurzburg:
SInce analog sensors are almost totally unavailable, moved to separate page (to allow future archiving) and added UAVCAN sensors.
also added links to partner's offerings for airspeed sensors page


magicrub:
I looked through about half of it and it all seems like reasonable improvements.

magicrub (approved):

PR #3646, opened 2021-07-08
scripts: add update.cron

peterbarker:


Hwurzburg:
merging since it needs no review being a script example

PR #3647, opened 2021-07-14

FPVZaphod:
CRSF Telemetry easily reaches 25km, why is this under "Short Range"?


Hwurzburg:
as I understand it, originally Telemetry was limited in range....with higher power and improvements its not now....LGTM...merging...thanks

PR #3649, opened 2021-07-15
Add instructions for using GPS for Altitude

Hwurzburg:


common/source/docs/common-advanced-configuration.rst
```diff
@@ -120,6 +120,7 @@ tuning options for the vehicle.
     Transmitter Based Tuning <common-transmitter-tuning>
 [/site]
     UAVCAN Setup <common-uavcan-setup-advanced>
+    Using GPS for Altitude Sensor <common-gps-for-alt>
```
rmackay9, line 123: I think we should call this page, "GPS for Altitude" and group it with "GPS for Yaw"

rmackay9:
Besides the comment above it looks pretty good to me.

common/source/docs/common-advanced-configuration.rst
Hwurzburg: Will do

Hwurzburg: done

Hwurzburg:
@rmackay9 I had not pushed the actual GPS for Alt page...could you review? thanks

common/source/docs/common-gps-for-alt.rst
```diff
@@ -0,0 +1,26 @@
...
+Until version 4.1, ArduPilot required a barometer altitude sensor. Now, it is possible to boot an autopilot without barometer present, and use GPS (outdoor use) or other altitude sensor (rangefinder, range beacon) for indoor use. 
+
+.. warning:: Without a barometer as a backup sensor, loss of the primary altitude sensor can result in a crash if in an altitude holding mode or autonomous mode. Be prepared to switch to a non altitude holding mode in case of loss of primary sensor.
+
+Setup for using a non Baro Source for Altitude Measurement
+==========================================================
+
+If using EKF2:
+
+- Set the :ref:`EK2_ALT_SOURCE<EK2_ALT_SOURCE>` parameter to GPS.
```
rmackay9, line 16: I think we should specify the actual number so people don't need to look it up... although it is probably displayed in the GCS

common/source/docs/common-gps-for-alt.rst
```diff
@@ -0,0 +1,26 @@
+.. _common-gps-for-alt:
+
+==============================
+Using GPS for Altitude Sensing
+==============================
+
+Until version 4.1, ArduPilot required a barometer altitude sensor. Now, it is possible to boot an autopilot without barometer present, and use GPS (outdoor use) or other altitude sensor (rangefinder, range beacon) for indoor use.
```
rmackay9, line 7: I suspect some will be confused by this line and think that we only supported alt from baro in versions before 4.1.. maybe we should just remove the mention of the version and the past situation.

"This page explains how to setup GPS for use as an altitude source."

rmackay9 (approved):
I've added a comment or two but feel free to merge when you're happy, txs!

common/source/docs/common-gps-for-alt.rst
Hwurzburg: done

common/source/docs/common-gps-for-alt.rst
Hwurzburg: modified it a bit

PR #3652, opened 2021-07-17
Dev: Static Analysis: add clang tools to Ubuntu install instructions

IamPete1:
I also had to install clang tools, set both to version 7 to match CI

### DIFF
--- a/dev/source/docs/static-analysis.rst
+++ b/dev/source/docs/static-analysis.rst
@@ -20,7 +20,7 @@ You can do your own static analysis using autotest:
 
    You may need to install clang for this to work.
 
-   On Ubuntu, "sudo apt-get install clang"
+   On Ubuntu, `sudo apt-get install clang clang-tools`
 
    On Windows see https://clang.llvm.org/get_started.html
 


dev/source/docs/static-analysis.rst
```diff
@@ -20,7 +20,7 @@ You can do your own static analysis using autotest:
 
    You may need to install clang for this to work.
 
-   On Ubuntu, "sudo apt-get install clang"
+   On Ubuntu, `sudo apt-get install clang-7`, `sudo apt-get install clang-tools-7`
```
khancyr, line 23: Why not put them on the same cmd ?

IamPete1:
Turns out using edit in github I made a new branch in the AP repo, not mine as it used to. Oops, we should delete this branch once it gets merged.

dev/source/docs/static-analysis.rst
```diff
@@ -20,7 +20,7 @@ You can do your own static analysis using autotest:
 
    You may need to install clang for this to work.
 
-   On Ubuntu, "sudo apt-get install clang"
+   On Ubuntu, `sudo apt-get install clang-7 clang-tools-7`
```
peterbarker, line 23: Is this going to be wrong on older versions of Ubuntu?

IamPete1: I have no idea, i'm on 18.04 and it worked for me. We could just mention the need for it and not provide the command.....

khancyr: yep, doesn't work on 20.04 ...
so `sudo apt-get install clang clang-tools` will get you the latest available on ubuntu

IamPete1: removed the version numbers, should work for everyone ?

peterbarker (approved):

Hwurzburg:
merging

PR #3657, opened 2021-07-19
prearm: Match the distance to the source

muramura:
The judgment value of the source is 1m. However, in the WIKI, it is 2m.
I think it is better to match the source value.

judgment
https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/AP_Arming.cpp#L84


Hwurzburg:
LGTM...thanks

PR #3665, opened 2021-07-26
Update Note Sensor Performances Outdoors

JLAsensio:
I would like to change the note with the actual note which is written in the Specification sheet of the sensor. 
Also I would like to include the Test Report which hopefully will give better understanding of the sensor's performances outdoors.


Hwurzburg:
LGTM.thanks

PR #3668, opened 2021-07-27
Plane: tailsitter: add note about new enable param

IamPete1:
adds a note for the new tailsitter enable param that is now in master.  https://github.com/ArduPilot/ardupilot/pull/18015

image


Hwurzburg:
is this being back ported to 4.1 release? if not probably should start re updating the Upcoming features stuff....I know that that the BD_PWM_CNT stuff will not be in the first stable releases....will this?

IamPete1:
@Hwurzburg No plans for this to make it to 4.1. Will just be for those on dev for the time being.

Hwurzburg:
I added a more expansive note including enable=2 in the upcoming features section....closing this one...thanks

PR #3696, opened 2021-08-10
Update common-underwater-sonar-analog.rst

MaxtheMaker414:
Greetings, could you please add this change 
Sonarrange in the Preflight Check, Great tip for beginners. 
Sincerely Max

### DIFF
--- a/common/source/docs/common-underwater-sonar-analog.rst
+++ b/common/source/docs/common-underwater-sonar-analog.rst
@@ -60,3 +60,12 @@ Data screen's Status tab. Look closely for "sonarrange".
 .. image:: ../../../images/mp_rangefinder_lidarlite_testing.jpg
     :target: ../_images/mp_rangefinder_lidarlite_testing.jpg
 
+Preflight Check
+==================
+For an easy depth check add the following to your preflight CheckListEditor:
+
+Field: rangefinder1
+Comparison: NONE
+Trigger: 0,5 (in my case)
+Description/Value: Sonarrange
+                   {value}m


common/source/docs/common-underwater-sonar-analog.rst
```diff
@@ -60,3 +60,12 @@ Data screen's Status tab. Look closely for "sonarrange".
 .. image:: ../../../images/mp_rangefinder_lidarlite_testing.jpg
     :target: ../_images/mp_rangefinder_lidarlite_testing.jpg
 
+Preflight Check
+==================
+For an easy depth check add the following to your preflight CheckListEditor:
```
Hwurzburg, line 65: this is going to need a lot more information about exactly how to add using that editor....its not mentioned anywhere in the wiki....would need a section in the Mission Planner wiki and then this cross linked to that section....otherwise it will just raise a lot of questions to support....can you add a section here: https://ardupilot.org/planner/docs/mission-planner-features.html?

Hwurzburg (changes requested):

common/source/docs/common-underwater-sonar-analog.rst
MaxtheMaker414: Hi I added something to the mission planner wiki, maybe you can use this as a suggestion. I think my formatting and writing is not good enough to use it 1:1.

Hwurzburg: sorry I do not see any PRs for that

MaxtheMaker414: Found it in a personal folder, but I'm complety helpless, so here is my text.

====================================
PreFlight Checks
====================================

In the DATA Panel Mission Planner is able to provide useful information about your vehicle.

One of them is the PreFlight screen.
It can help checking GPS, battery and telemetry health with colour coded items. 

Green (everything is fine), red (input is below the minimal value)

When you select edit, you can open the PreFlight CheckListEditor.
Now you can see the inputs and the values needed to trigger an event.
With "+" you can add your own inputs to preflight check like a second battery or a sonar.

Field
====================================

Here you can fill in the needed input, like Rangefinder. 


Comparison (Hope this is right, I'm not sure)
====================================

In which way your trigger value shall be used.

NONE = none
LT = smaller than value
LTEQ = smaller than value or equal?
EQ = equal to value
GT = bigger than value
GTEQ = bigger than value or equal?
NEQ = not equal to value

In my case I want to have a green value for a sonarrange above 0,5 Meter -> GT


Trigger
====================================

Value describes, when the trigger changes the colour of the field


Description/Value
====================================

1. line: Fill in the name of your Preflight Check, like Sonar-down
2. line: Fill in the value and the unit of the value, like {value}m, for distance in Meter


Colour True Colour False
====================================

Option to switch the colours


"+" in the right corner
====================================

can be used to add an sub-group to an input

PR #3697, opened 2021-08-10
Update common-display-onboard.rst

MaxtheMaker414:
Joy-it Display, bought and tested, is working properly,
all the other links show products, which are no more available


Hwurzburg:
I think it will be better just to change the links to general searches in Amazon, and Ebay for SSD1306 and SH1106 displays....I just did this...thanks for pointing the broken links out...

PR #3699, opened 2021-08-10
Fix backslash escape charaters

TunaLobster:
Saw this one on the site. Quick fix to have a Windows path show up correctly with escaping the escape character.


Hwurzburg:
LGTM

PR #3700, opened 2021-08-14
Fix install script bash syntax

TunaLobster:
The install script was warning about unexpected...something. No warnings with this change because bash expects single square bracket operations to have single equal signs. Makes it easier to determine if an actual install error has happened.


peterbarker (approved):
bash actually accepts both, and we do `#!/bin/bash`

But this isn't a bad change

PR #3701, opened 2021-08-15
Update RPM

Gone4Dirt:
Few commits to update the RPM sensor page:
- Added updates to reflect the new param changes in this PR: https://github.com/ArduPilot/ardupilot/pull/18044
- Added setup of average ESC rpm.  @andyp1per would appriciate it if you could have a quick glance over that to make sure I have added it correctly.
- A few minor corrections and formating.

### DIFF
--- a/common/source/docs/common-rpm.rst
+++ b/common/source/docs/common-rpm.rst
@@ -82,20 +82,21 @@ First the board needs to be configured to allow PWM pins to be set for GPIO.  Th
 be used for GPIO.  On non-Pixhawk boards the PWM count will include all PWM outputs.  On Pixhawk boards this parameter only affects AUX pins.  Write the parameter 
 and reboot the autopilot.
 
+Now the RPM library must be enabled.  Set the parameter :ref:`RPMx_TYPE<RPM1_TYPE>` to 1 for a GPIO pin based sensor.  Write the parameter to ArduPilot then refresh/fetch the 
+parameters.  You will now find that the instance of RPM (e.g. RPM1) has a number of other parameters available for editing, allowing you to complete your setup.
+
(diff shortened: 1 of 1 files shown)


andyp1per:
Looks ok to me, although I think this is @yaapu 's code

Gone4Dirt:
Ah, sorry @andyp1per I assumed it was yours.  Should have looked at Git blame.  Thanks for looking at it though.

@yaapu if you could double check the instructions I have done for setting up the ESC telem in the RPM library that would be much appriciated.

yaapu:
@Gone4Dirt  looks good to me too, thanks for documenting the average ESC RPM option

Hwurzburg (changes requested):
Matt, this has a few build errors...there is are no params named RPM1_xxx.....only RPM and RPM2 and RPM3....in addition, I try to avoid using links with RPMx_yy <RPM2_yy>...I'd rather we use something like "in the case of the second RPM instance:" RPM2_MIN<RPM2_MIN> so that the linked param matches the display name they clicked to avoid confusion..ie why did it go there?

Gone4Dirt:
Thanks for taking a look Henry.  Noted about the preferred way to do links.  Will update those.  Regarding RPM1_xxx params.  I have a PR (https://github.com/ArduPilot/ardupilot/pull/18044) which will be changing the RPM_ to RPM1_ hence this pr to the wiki in prep for it going in.

Hwurzburg:
@Gone4Dirt Matt, this will need a rebase since I had to fix build errors in several files caused by the name change....it appears that this is a NFC, as such, when rebased, I can merge rather than waiting for 4.2....because I must fix the build errors when they occur and the param generator is always shows dev, and hence the build errors

Hwurzburg:
closing in lieu of #3771 and #3772

PR #3705, opened 2021-08-16
common: remove links to shop.ardupilot.org

peterbarker:


peterbarker:
shop.ardupilot.org no longer exists in the state it was.

PR #3708, opened 2021-08-19
Update common-px4flow-overview.rst

vincentpoont2:
Add Holybro PX4Flow shop link, remove broken links.


vincentpoont2:
The current links seems to be broken or the website no longer exist. Please help adjust accordingly.

rmackay9:
Thanks for the fix.

@Hwurzburg 
We should fix the links but it looks like CUAV also has the product available so we should provide links to both CUAV and the holybro version. https://store.cuav.net/shop/flow-sensor/

rmackay9:
Thanks again for this. I've updated the page including the links to holybro.

Issue #3397, opened 2021-03-23
Add Peripherals/OptFlow page for the UPixels UPFlow sensor


rmackay9:
Support for this sensor will be released with Copter-4.1
Issue is here: https://github.com/ArduPilot/ardupilot/issues/11328
PR is here: https://github.com/ArduPilot/ardupilot/pull/16883

Hwurzburg:
Closing as already addressed. `common-upixels-upflow.rst` exists and is linked from `common-optical-flow-sensors-landingpage.rst`.

PR #3709, opened 2021-08-19
Add Peripherals/OptFlow page for the UPixels UPFlow sensor

zhouzhiwen2000:
Solves this issue: [Add Peripherals/OptFlow page for the UPixels UPFlow sensor #3397](https://github.com/ArduPilot/ardupilot_wiki/issues/3397)


rmackay9 (approved):
Looks good to me thanks!

As a side note we probably need to update our optical flow pages for 4.1 because the default becomes the EKF3 and we have the GPS/Non-GPS transitions too.  This can be done separately though

zhouzhiwen2000:
> Looks good to me thanks!
> 
> As a side note we probably need to update our optical flow pages for 4.1 because the default becomes the EKF3 and we have the GPS/Non-GPS transitions too. This can be done separately though

Maybe in a separate PR when I get some time. Can this PR be merged for now?

Hwurzburg:
I am a little concerned that since this is only supported in 4.1 and 4.1 uses EKF3 that the setup will be confusing, especially since the 1MB boards do not even have EKF2 as an alternative...shouldn't we give the EKF3 basic setup instead of EKF2?

rmackay9:
@Hwurzburg, I'll add the changes so that it talks about EKF3 instead.

zhouzhiwen2000:
> I am a little concerned that since this is only supported in 4.1 and 4.1 uses EKF3, that the setup will be confusing, especially since the 1MB boards do not even have EKF2 as an alternative...shouldn't we give the EKF3 basic setup instead of EKF2?

I will update ekf3 setup tutorials for all opticalflow sensors soon.

PR #3710, opened 2021-08-20
Grammar pass on loading firmware page

TunaLobster:
Not perfect, but better.

### DIFF
--- a/common/source/docs/common-loading-firmware-onto-pixhawk.rst
+++ b/common/source/docs/common-loading-firmware-onto-pixhawk.rst
@@ -4,14 +4,14 @@
 Loading Firmware
 ================
 
-These instructions will show you how to download the latest firmware onto the autopilot using the Mission Planner ground station,which already has ArduPilot firmware installed . See :ref:`common-loading-firmware-onto-chibios-only-boards` .
+These instructions will show you how to download the latest firmware onto the autopilot hardware that already has ArduPilot firmware installed. This process will use the Mission Planner ground control station. See :ref:`common-loading-firmware-onto-chibios-only-boards` .
 
 [copywiki destination="copter,plane,rover,planner"]
 
 Connect autopilot to computer
 =============================
 
-Once you've :ref:`installed a ground station <common-install-gcs>` on your computer, connect
+Once you have :ref:`installed a ground station <common-install-gcs>` on your computer, connect
 the autopilot using the micro USB cable as shown
 below. Use a direct USB port on your computer (not a USB hub).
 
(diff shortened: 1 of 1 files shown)


common/source/docs/common-loading-firmware-onto-pixhawk.rst
```diff
@@ -47,36 +44,34 @@ Answer **Yes** when it asks you "Are you sure?".
...
 
 It usually takes a few seconds for the bootloader to exit and enter the main code after programming or a power-up. Wait to press CONNECT until this occurs.
 
-.. note:: Updating the firmware to a newer version does not alter existing parameters unless the firmware is for a different vehicle, in which case parameters are reset to their default values for that vehicle. However, it is always a good idea to save your parameters to a file using the "Save to File" button on the Mission Planner's **CONFIG/Full Parameter Tree** tab before any firmware updates, just in case of any issues while updating.
+.. note:: Updating the firmware to a newer version does not alter existing parameters unless the firmware is for a different vehicle, in which case parameters are reset to their default values for that vehicle. However, it is always a good idea to save your parameters to a file using the "Save to File" button on the Mission Planner's **CONFIG/Full Parameter Tree** tab before any firmware updates, just in case of any issues while updating. Do not apply all parameters after upgrading to a new version as some parameters might have a different meaning.
```
Hwurzburg, line 59: not actually certain this is true...we convert params that have change to another name, true but loading old ones wont actually load in that case....I do not think we change the meaning of any existing param on something that is released...perhaps rarely in DEV if we think no one really has used it yet...like on some quadplane frames I recall...so I do not think this statement should be included...you can upgrade firmware to any release version and re-apply old param files without unexpected consequences, to my knowledge

Hwurzburg (changes requested):

common/source/docs/common-loading-firmware-onto-pixhawk.rst
TunaLobster: Currently moving to Plane 4.1 and applying the old PID gains will cause issues because the parameter names are the same but the values have different meaning. There have been a few users on the forums that had this issue so documenting it would be useful before 4.1 leaves beta.

Hwurzburg: old pid gains are different names...only the RLL2SRV_TCONST and PTCH2SRV_TCONST remained unchanged..which ones are you talking about?

TunaLobster: My bad on the PID parameters. I'm not sure where that came from in my head.

https://discuss.ardupilot.org/t/plane-4-1-0-beta/72434/67

Mainly sparked from this post by tridge and noticing that we do not have this recorded anywhere. It is important to tell users not to upload old parameters because of the possibility of similar parameters names even when the auto upgrade scripts works.

Hwurzburg: something I do often...because I change firmware often when experimenting with code on the bench and reset defaults and then reload old param files...never have had an issue...will discuss more with Tridge...I am still not convinced that this is warranted...

PR #3717, opened 2021-08-24
SiTL: Fix SIM_WIND_DIR explanation error

Maarrk:
Parameter wind direction sets **from where** the wind is blowing

Tested by doing ALT_HOLD on 3m altitude, the copter is drifting north for these params:

- SIM_WIND_DIR 180
- SIM_WIND_SPD 3
- SIM_WIND_T_ALT 2

Just in case: drifts west for SIM_WIND_DIR 90


Hwurzburg:
Thanks,merging

PR #3719, opened 2021-08-25
common: remove reference to compass calibration stick gestures

peterbarker:


rmackay9:
It might be better to just add a note that this has been removed in 4.1

Hwurzburg:
and is it removed in 4.1? Tridge told me master changes do not go into beta/stable automatically

peterbarker:
On Tue, 24 Aug 2021, Randy Mackay wrote:

> It might be better to just add a note that this has been removed in 4.1

Why?

If we wanted people to know about it *we shouldn't've removed it!* :-)

peterbarker:
Changed this to just add a `.. note::`

Hwurzburg:
rather than ask you to redo and add to futures page, I just did it....thanks closing

PR #3721, opened 2021-08-25
Update sitl-with-JSON.rst

TunaLobster:
Add detail for starting the JSON interface with the vehicle executable.


Hwurzburg:
LGTM...sorry for the delay

PR #3731, opened 2021-08-31
common: add watchdog-debugging document

peterbarker:


common/source/docs/common-watchdog-debugging.rst
```diff
@@ -0,0 +1,60 @@
...
+   MC           MAVLink Command:          246: ?????
+  IEH       Internal Error Mask:          0x0: ?????
+  IEC      Internal Error Count:            0: 0
+   TN               Thread name:        :Rove: ?????
+  pbarker@bluebottle:~/rc/ardupilot(pr/floating-point-radiuses)$ 
+
+Converting the LR field into a line of code
+===========================================
+
+Certain faults will fille the Function Level Reset (FLR) register with the address of an instruction which caused the processor to fault.  That address can be turned into a line of code.
```
peterbarker, line 51:
```suggestion
Certain faults will fill the Function Level Reset (FLR) register with the address of an instruction which caused the processor to fault.  That address can be turned into a line of code.
```

rmackay9:
I think this should either go into the dev wiki (instead of common) or it should be combined with the existing "independent watchdog" page which is in common. https://ardupilot.org/copter/docs/common-watchdog.html

common/source/docs/common-watchdog-debugging.rst
peterbarker: Fixed

peterbarker:
> I think this should either go into the dev wiki (instead of common) or it should be combined with the existing "independent watchdog" page which is in common. https://ardupilot.org/copter/docs/common-watchdog.html

I've moved it into dev - only devs are going to be interested in it - I expect it will get a link from that other page, 'though.

Hwurzburg:
when its ready, I can check formatting and add a link from the other page...

rmackay9:
It's a bit odd to have the page called "common-" when it's in the dev section... I think the file itself should also be renamed to remove the common-.

peterbarker:
> It's a bit odd to have the page called "common-" when it's in the dev section... I think the file itself should also be renamed to remove the common-.

Done.

What it really  needs is *much* better content....

nkruzan:
I think this should go in as-is and can be expanded on more in future

Hwurzburg:
fixed errors, cleaned up a little, added to debugging page TOC....thanks...merging

PR #3733, opened 2021-09-01
Convert RIFF image encodings and add CI check

TunaLobster:
Had some issues with these images while working on the latexpdf builder. Converted them to a different encoding and added a CI check so that it won't happen again. Also made some of the images html builder only as they were externally referenced.


Hwurzburg:
LGTM

PR #3735, opened 2021-09-01
Remove sphinxcontrib.vimeo plugin

TunaLobster:
The sphinxcontrib.youtube plugin supports embedding Vimeo videos now. I did some spot checking to make sure that the video still worked as expected. The issue with using the current Vimeo plugin from our forked repo is that it does not support the latexpdf builder. The upstream repo is gone now.

One other option would be to add an `.. only:: html` block around each Vimeo video, but that creates an issue for when new Vimeo videos are added to the wiki.

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -47,6 +47,11 @@ python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinx_
 python3 -m pip install --user --upgrade git+https://github.com/sphinx-contrib/youtube.git
 
 # and a vimeo plugin:
-python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
+# python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
+# we previously used this plugin, but the features are now included in the youtube plugin
+# remove if installed (will warn if not installed)
+if python3 -m pip show sphinxcontrib.vimeo > /dev/null 2>&1; then
+  python3 -m pip uninstall -y sphinxcontrib.vimeo
+fi
 
 echo "Setup completed successfully!"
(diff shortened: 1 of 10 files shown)


TunaLobster:
@auturgy @peterbarker Thought you might want to take a look at this before it goes in.

Sphinxsetup.sh
```diff
@@ -47,6 +47,9 @@ python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinx_
 python3 -m pip install --user --upgrade git+https://github.com/sphinx-contrib/youtube.git
 
 # and a vimeo plugin:
-python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
+# python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
+# we previously used this plugin, but it is now included in the youtube plugin
+# remove if installed (will warn if not installed)
+python3 -m pip uninstall sphinxcontrib.vimeo
```
khancyr, line 53: maybe a test here with python -m pip show to see if it is installed or not first

TunaLobster: Fixed. Should not give any output if the package is not present.

peterbarker:
@TunaLobster which browsers did you test with?

TunaLobster:
Firefox 90. Is there an issue with Chromium?

peterbarker:
> Firefox 90. Is there an issue with Chromium?

That's a very good question, isn't it?  Considering 80%+ of the planet uses it, we should probably know before merging this :-)

TunaLobster:
Tested in Firefox 90 (the whole 4% of us) and Chrome 93 (70% of the world). I am unable to test for the 20% that use Safari.

peterbarker:
> Tested in Firefox 90 (the whole 4% of us) and Chrome 93 (70% of the world). I am unable to test for the 20% that use Safari.

Eh.  MaxOSX users deserve what they get ;-P

peterbarker (approved):

PR #3736, opened 2021-09-02
Update download_and_installation.rst to install pathlib before MAVProxy to MacOS

DaijiroYamazaki:
I got following error when install MAVProxy to Mac OS (Big Sur 11.5.2):

--
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
conda-repo-cli 1.0.4 requires pathlib, which is not installed.
--

This error is cleared with installing pathlib.


stephendade:
``pathlib`` is included as standard in Python 3.4+

What version of Python are you using?

DaijiroYamazaki:
@stephendade thanks for review.
I found 2 python version: anaconda 3.8.8 and official 3.9.6, default is anaconda 3.8.8.
I'll try to removing anaconda and reinstall newest official version.
I think this request is caused by only my environment, so I can close.

PR #3739, opened 2021-09-03
Add Ainstein to the list of ArduPilot stores

CamronMyers:
Ainstein offers the US-D1 radar altimeter for a variety of drone products running on ArduPilot


Hwurzburg:
I dont believe that Ainstein is an ArduPilot partner....current policy is the stores list shows Partner's stores...

Hwurzburg:
Updated links...awaiting feedback from @tridge on what, if anything, to add regarding CAN device and setup...closing this however

PR #3743, opened 2021-09-04
Copter: common-pixhawk-camera-setup AUX6 is Pin 55

hendjoshsr71:
From ArduPilot Facebook report

Aux6 should be Pin 55 not Pin 56 see picture above here,
https://ardupilot.org/copter/docs/common-pixhawk-overview.html


Hwurzburg:
thanks...merging

PR #3746, opened 2021-09-05
Add exception message for vimeo module

TunaLobster:
Should cause an issue if the module is present and not just build without Vimeo videos included.

@peterbarker The issue should be easier to spot now.

### DIFF
--- a/update.py
+++ b/update.py
@@ -51,6 +51,12 @@ from datetime import datetime
 # without the following import on old versions of Python:
 from distutils import dir_util  # noqa
 
+try:
+    import sphinxcontrib.vimeo
+    raise Exception("Module sphinxcontrib.vimeo was found. Please run python3 -m pip uninstall -y sphinxcontrib.vimeo")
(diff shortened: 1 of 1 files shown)


TunaLobster:
Tested with python 3.8 with and without the sphinxcontrib.vimeo module installed.

TunaLobster:
This should not go in yet. Need to wait on a few other PRs to get the Vimeo part correct and to get the YouTube plugin working.

Hwurzburg:
@TunaLobster  @peterbarker ready now?

TunaLobster:
No. This was meant to be a follow up PR for #35755. But that has been reverted. Then I had to fix something in the plug-in repo so that it would work with python 2.7 and sphinx 1.8. Now I'm wait for the new venv things to get settled before trying to remove this dependency. I might also have to wait until we update the sphinx version which will also be dependent on updating the theme. Now I need to reopen a PR that includes several other things, including this one, but I won't do that until everything is ready because I don't want something to be merged into master without you testing it and have it upset the build.

PR #3752, opened 2021-09-07
Revert "Remove sphinxcontrib.vimeo plugin"

peterbarker:
This reverts commit d82bd556627ef368346c0934fb96fa161fa9ecc1.

We can't install a recent enough version on Python2 for the youtube
plugin to cover this


peterbarker:
@Hwurzburg hopefully this gets those videos back!

TunaLobster:
Since python 2 doesn't support f strings, this isn't exactly trivial. I'll try making a VM that matches the build server so things like this can be caught before going live.

PR #3756, opened 2021-09-09
Update common-flight-controller-wiring.rst

auturgy:
Tidy up.  Some of the buzzer/switch stuff was inaccurate/misleading.  Needs a bigger scrub through, as for copter the vast majority of users should be using FMU outputs.


Hwurzburg:
thanks

PR #3757, opened 2021-09-10
Add KakuteF4 Mini

Hwurzburg:


common/source/docs/common-holybro-kakutef4-mini.rst
```diff
@@ -0,0 +1,145 @@
+.. _common-holybro-kakutef4-mini:
+
```
andyp1per, line 2: You should probably add that this board support bdshot by default

andyp1per (approved):

common/source/docs/common-holybro-kakutef4-mini.rst
Hwurzburg: done

PR #3763, opened 2021-09-12
Add script to create install of actual wiki server

peterbarker:
The normal add-environment scripts get you most things - but this one will create users, install apache and configure it, that sort of thing.


Hwurzburg:
does everyone want to have a wiki server on their PC? am confused a bit what this is for and who should have it....since its part of anyone who wants to locally build a wiki for checks...

TunaLobster:
This is to match the exact environment that the wiki is built with on another machine. I've been having issues with that because the current method installs the latest versions of all the modules rather than the versions than what is used to build the wiki. That's part of the problem why I was unable to find the Vimeo plugin issue before it went live.

Another benefit of this is that we can have whatever python package versions we want without disrupting the user environment. If something happens upstream with say the docutils package, we can roll back to working version in our environment and let the user environment have some other version of docutils. That is especially helpful when working with multiple projects that need different versions of packages to function.

khancyr (approved):
LGTM

brunoolivieri:
I guess that the mentioned apache/nginx still a WIP.

Hwurzburg:
Since Peter and Pierre approve...merging

peterbarker:
@TunaLobster I goofed when I merged your PR into mine here (which then got merged).

The idea behind my changes here were an install script for the *new* server, not the old.  So done the right way - new packages, `--user` and whatnot.

I really don't understand vitrualenvs - but I also can't find this "activate" script which is referenced in here.

TunaLobster:
The activate script is created when the venv is created. There are issues with the theme that need to be addressed before rolling forward with sphinx version.

brunoolivieri:
I confess that I don't understand venvs. However, we have script for docker, vagrant and venv. Do we need all of them? (the server does not use them yet). I understand that they can all coexist, but it might be interesting to put/update some notes about this in the wiki build documentation.

brunoolivieri:
There really is some inconsistency in the python environment right now. Between python2.7 and python 3.5. There is no pip3 but when calling pip, it is calling pip3.

PR #3765, opened 2021-09-14
corrected FS "reason" 5 to GCS Failsafe

pompecukor:
Bat FS reason was missing from the list and made the numbering incorrect from 3 up.

### DIFF
--- a/plane/source/docs/apms-failsafe-function.rst
+++ b/plane/source/docs/apms-failsafe-function.rst
@@ -360,22 +360,26 @@ GCSs will often display text indicating the type of failsafe encountered, such a
    </tr>
    <tr>
    <td>4</td>
-   <td>GCS Failsafe</td>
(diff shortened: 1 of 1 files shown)


plane/source/docs/apms-failsafe-function.rst
```diff
@@ -358,24 +358,28 @@ GCSs will often display text indicating the type of failsafe encountered, such a
    <td>3</td>
    <td>Radio Failsafe</td>
    </tr>
-   <tr>
+    tr>
```
Hwurzburg, line 361: this is in an error in format

Hwurzburg (changes requested):
Needs format fixed....also, an existing issue, reasons above GSP glitch are incorrect and out of date....but that not your issue...

Hwurzburg:
updated table and fixed existing format error....squashed...thanks merging

PR #3769, opened 2021-09-18
Wrong information about Loiter_turns radius setting

pompecukor:
With regards to COPTER, there was a false statement that the circle radius cannot be set as part of a NAV command. It is possible to set number of turn with param3. I don't know if this was different before, but in current firmware this is possible and tested. 
PS: Mission planner may not display the name of the param3 correctly (i did not check), but in QGC it is correct and works too.

image


Hwurzburg:
1. does 0 use circle_radius param? if so probably should be mentioned
2. does this apply to loiter unlimited and/or loiter time? and only to copter, not plane or rover?

Hwurzburg:
@pompecukor Alex....needs answer

tridge (approved):

PR #3782, opened 2021-09-23
Add multi screen for MSP OSD to upcoming

Hwurzburg:
Co-authored-by:  Alessandro Apostoli <30294218+yaapu@users.noreply.github.com>

Alex, when in a release, will just drop the 4.2 from the page name and it will become the main page in wiki

will do the same for Display port....also will update the existing pages with the spelling, phrasing, and new telem tables in yet another PR...


yaapu:
Henry some images are missing or obsolete. In the msp osd emulation section I had updated the screenshot with a more recent one and also added a stats screen example, those images are both in my original PR

Hwurzburg:
yep, forgot to git add them on the commit...will need to update the table also with your new comments....I will do that....

yaapu:
thanks Henry!

PR #3786, opened 2021-09-23
Update learning-ardupilot-rc-input-output.rst

TunaLobster:
Remove text about RCIN and SERVO_OUT being the same layer.

There are probably more things on this page to update, but I'm a bit lost on how to describe the new layers and decoding correctly. This kind of info would be helpful for the few murmurings of adding support for Jeti EX.Bus.


Hwurzburg:
@peterbarker ping

PR #3790, opened 2021-09-25
fix broken verbatim

shiv-tyagi:
Fixed a broken verbatim in testing the vehicle section.


Hwurzburg:
LGTM...merging..thanks

PR #3804, opened 2021-09-29
dev: add note on being able to specify mavlink message telem rates in…

peterbarker:
… file

### DIFF
--- a/dev/source/docs/mavlink-requesting-data.rst
+++ b/dev/source/docs/mavlink-requesting-data.rst
@@ -10,6 +10,7 @@ The ground station or companion computer can request the data it wants (and the 
    - Send `REQUEST_DATA_STREAM <https://mavlink.io/en/messages/common.html#REQUEST_DATA_STREAM>`__ messages to set the rate for groups of messages
    - Send a `SET_MESSAGE_INTERVAL <https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL>`__ command (within a `COMMAND_LONG <https://mavlink.io/en/messages/common.html#COMMAND_LONG>`__ message) to precisely control the rate of an individual message.  Note this is only supported on ArduPilot 4.0 and higher
    - Send a `REQUEST_MESSAGE <https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE>`__ command (within a `COMMAND_LONG <https://mavlink.io/en/messages/common.html#COMMAND_LONG>`__ message) to request a single instance of a message.  Note this is only supported on ArduPilot 4.0 and higher
+   - create configuration files stored on SD card or in ROMFS specifying message stream rates on a per-channel basis
 
 More details of these methods can be found below.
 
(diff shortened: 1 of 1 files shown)


dev/source/docs/mavlink-requesting-data.rst
```diff
@@ -146,6 +146,27 @@ Send a `COMMAND_LONG <https://mavlink.io/en/messages/common.html#COMMAND_LONG>`_
 - param2: depends on message requested; see that message's definition for details.
 - param3 to param7: 0 (not used)
 
+Specifying Message Rates in a File
+----------------------------------
+
+At boot ArduPilot will populate the initial message intervals from files found in either ROMFS or in the filesystem (e.g. SD card).
```
Hwurzburg, line 152: at SD root? at APM dir root?

this is really cool....did not know it was in the code....is this master only or in stable? I assume it can be overwritten by the GCS....its a common problem on Dragonlink...I think we dismissed the option of stopping GCS from overwriting, hoping systems they would include it in DL docs, but still is a re-occurring problem for that system

also, if we could create an example for Dragonlink and include it, it would help a lot of people using that system

Hwurzburg (changes requested):

dev/source/docs/mavlink-requesting-data.rst
peterbarker: @Hwurzburg this does not stop the GCSs from reconfiguring the telemetry rates using `SET_STREAM_RATE` or whatever.  It just permits absolute customisation of the stream rates for streaming messages, so you can prune out everything that doesn't matter to you.

GCSs should have options to turn off that auto-configuration of stream rates, as @khancyr mentions.  One extreme fix if a GCS doesn't behave is to turn on mavlink signing - can't reset streamrates if you don't know the key!

peterbarker: Also, I haven't specified "APM directory" because that's not correct in the Linux HAL.  If you feel like adding more words, I'm happy to think about them :-)

Hwurzburg: what dir in Linux systems?

peterbarker: > what dir in Linux systems?

Varies based on board type, but typically the directory in which the binary is started.

Hwurzburg: still needs specific instructions where to place the files for ROMFS or SD...by system if needed...otherwise will just cause support questions...

peterbarker: I've added instructions on SD and ROMFS.

dev/source/docs/mavlink-requesting-data.rst
```diff
@@ -146,6 +146,37 @@ Send a `COMMAND_LONG <https://mavlink.io/en/messages/common.html#COMMAND_LONG>`_
...
+
+At boot ArduPilot will populate the initial message intervals from files found in either ROMFS or in the filesystem.
+
+On ChibiOS-based boards the SD card will be searched for specially-named files in the APM subdirectory.
+
+Configuration files can be included in ROMFS (i.e. compiled into the image) by specifying their path in the relevant board's hwdef file.
+
+.. code-block::
+
+   ROMFS message-intervals-chan0.txt message-intervals-chan0.txt
```
Hwurzburg, line 160: so, the first param is the file name that will be added into ROMFS and must be "message-intervals-chanX.txt" format for each channel, and the second param is the path to a text file with the date to be used in that file?   having the same name without some more explanation (assuming I have figured it out correctly) is confusing..
perhaps ROMFS message-intervals-chan0.txt <path to text file with data to load into the ROMFS file> ?

Hwurzburg: also, need to add at very top as 5th method to set rates....but I can add after merge if you want

peterbarker:
On Thu, 30 Dec 2021, Henry Wurzburg wrote:

> so, the first param is the file name that will be added into ROMFS and must be "message-intervals-chanX.txt" format for each channel, and the second param is the path to a text file with
> the date to be used in that file? having the same name without some more explanation (assuming I have figured it out correctly) is confusing..
> perhaps ROMFS message-intervals-chan0.txt ?

That's correct on all counts.  I've added some path components and some 
better(?) text around this.  I've also moved it to below the explanation 
of the file names.

peterbarker:
On Thu, 30 Dec 2021, Henry Wurzburg wrote:

> also, need to add at very top as 5th method to set rates....but I can add after merge if you want

Added it.

Hwurzburg:
thanks...I hate the markup used here....a couple of expressions I wrote turned into just a ?

Hwurzburg:
okay has a couple of build errors...but easiest is to merge and I will fix immediately after....thanks

PR #3810, opened 2021-10-01
.gitignore: Add */_build/*

TunaLobster:
I think sphinx started making these directories at some point. Not sure when. Might be related to the Latex testing.

@khancyr


Hwurzburg:
does not occur in local builds

PR #3815, opened 2021-10-03
Dev: update SITL Binaries Install Instructions

hendjoshsr71:
This adds the toolchain option required to build custom binaries for mission planner's sitl.

From discord, @stephendade,

"Mission Planner needs a 32bit build of ArduPilot. Issue is that cygwin64 gcc builds a 64bit by default. Changing the toolchain to i686-pc-cygwin tells Cygwin to use the 32bit toolchain. See https://stackoverflow.com/questions/27344540/what-do-the-c-compilers-on-cygwin-generate for details of the Cygwin64 toolchains.
To be precise, the supplied dlls in the Mission Planner sitl folder are all 32bit, so the sitl exe needs to be 32bit as well"

### DIFF
--- a/dev/source/docs/building-setup-windows-cygwin.rst
+++ b/dev/source/docs/building-setup-windows-cygwin.rst
@@ -166,8 +166,9 @@ You can run SITL just as you can with any other build environment. To see the te
 Build SITL binaries for Windows
 ===============================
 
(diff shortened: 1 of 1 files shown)


stephendade (approved):
All good. LGTM

PR #3816, opened 2021-10-03
Revert "Add script to create install of actual wiki server"

Hwurzburg:
Reverts ArduPilot/ardupilot_wiki#3763 entirely...this continues to cause issues in my local builds, even after re-running the new Spinixsetup.sh, and apparently also gives the venu warning on the build server as reported by @brunoolivieri


Hwurzburg:
Closing since I found the command to activate the environment...not liking it, but found it...I need to update the wiki build instructions

TunaLobster:
I opened a PR on Peter's branch for the original PR to update the build instructions. Please don't merge this. There are more things on the infrastructure side that need to be changed outside of these PRs.

PR #3821, opened 2021-10-06
Test auto build.

brunoolivieri:
I'm testing the fix for the wiki auto build. This is a poor change to test if the server will be able to build without intervention.


brunoolivieri:
I'm testing autobuild on the server and looking at the effect.

brunoolivieri:
Did not work yet.

PR #3822, opened 2021-10-06
Testing auto build

brunoolivieri:
I'm testing the fix for the wiki auto build. This is a poor change to test if the server will be able to build without intervention.


brunoolivieri:
Applying and observing the server.

(Running manually by the shell, wiki builds, but by cron job no. There is still and PIP error)

PR #3831, opened 2021-10-09
Dev: Eclipse setup instructions remove waf in command

hendjoshsr71:
I was creating a new project folder and the extra `waf` here leads to errors when creating new projects. If you look at the photo here you will see the working command. 
https://ardupilot.org/dev/_images/eclipse-install3.png


hendjoshsr71:
I think there are issues between different versions of Eclipse.  For example a user posted to Discord about Eclipse issues, but after discussion it turns out he was using STM's STM32CubeIDE. It is Eclipse based and slightly behind the official Eclipse release. 

I would be tempted to recommend STM's IDE instead for Eclipse since it integrates onboard hardware debugging. But first I'd think we need to demonstrate that capability?

rmackay9:
@hendjoshsr71, thanks for this.   I think that extra "waf" was just a typo.  I actually don't think we should add the extra note about the STM32CubeIDE because that's a different editor and we will just add confusion.

hendjoshsr71:
I've removed the note for now. If a few more folks hit the issue then we can reconsider. 

STM32CubeIDE is Eclipse just with wrappers around it to add in additional tools for STM hardware debugging built-in. It could be a good a fit for us given all of our hardware in STM based. I haven't yet figured out out how to make the hardware debugging part functional yet though.

PR #3834, opened 2021-10-15
On GPIOS page include note re: SERVOx_FUNCTION, -1

mattdholmes00:
Added comment that SERVOx_FUNCTION needs to be -1 for GPIO use.


Hwurzburg:
this is already documented in the upcoming features section...its not in stable, so does not go in yet...thanks anyway

PR #3836, opened 2021-10-15
dev: add note about adding parameters in G2 class.

shiv-tyagi:
Adding a note about ParametersG2 that the global parameters can also be added to this class and can be accessed in main code just like g parameters.

### DIFF
--- a/dev/source/docs/code-overview-adding-a-new-parameter.rst
+++ b/dev/source/docs/code-overview-adding-a-new-parameter.rst
@@ -127,6 +127,8 @@ characters.
 You're done!  The new parameter can be access from the main code (not
 the libraries) as g.my_new_parameter.
 
+.. note:: ArduPilot vehicle's global parameter classes are crowded. To avoid exceeding the maximum number of parameters in a vehicle's Parameter class, another class for global parameters (named 'ParametersG2') has been introduced. Adding parameters to this class is similar to adding parameters in libraries. The parameters added to this class can similarly be accessed from within the vehicle's main code.
+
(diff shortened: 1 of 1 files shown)


dev/source/docs/code-overview-adding-a-new-parameter.rst
```diff
@@ -127,6 +127,8 @@ characters.
 You're done!  The new parameter can be access from the main code (not
 the libraries) as g.my_new_parameter.
 
+.. note:: As the ardupilot is improving, the global parameter class is filling up. To avoid getting past the maximum number of parameters in Parameter class, another class for global paramters (named 'ParametersG2') has been introduced. Adding parameters to this class is similar to adding paramters in libraries. The paramters added to this class can also be accessed from the main code.
```
Hwurzburg, line 130:
```suggestion
.. note:: ArduPilot vehicle's global parameter classes are crowded. To avoid exceeding the maximum number of parameters in a vehicle's Parameter class, another class for global parameters (named 'ParametersG2') has been introduced. Adding parameters to this class is similar to adding parameters in libraries. The parameters added to this class can similarly be accessed from within the vehicle's main code.
```

Hwurzburg (changes requested):
some wording changes

dev/source/docs/code-overview-adding-a-new-parameter.rst
shiv-tyagi: Thanks! I've pushed the changes. Please check.

PR #3837, opened 2021-10-15
Add last build date

TunaLobster:
This adds a date code to the wiki saying when it was last build. I put it on the front index page, but it could go anywhere. Tested with sphinx 4.2.0 and 1.8.5 with python 3 and 2 all in WSL 1 Ubuntu 20.04. This would be a quick way to tell if the wiki is building correctly without checking for recent commits to show up or looking at logs on the build server.

It would look like this:
image

### DIFF
--- a/ardupilot/source/index.rst
+++ b/ardupilot/source/index.rst
@@ -207,6 +207,7 @@ Features
 
 
 
+Last build: |today|
 
 --------------
 


brunoolivieri:
I do not have a strong opinion regarding the built date.

There was a discussion with a similar topic: https://github.com/ArduPilot/ardupilot_wiki/pull/2278

At the time, We argued that inserting the last updated date for each file was not attractive. 

To know when the wiki was built last time is possible to look for the hidden folders, for example, the "docs" for each site (https://ardupilot.org/ardupilot/docs/)

PR #3838, opened 2021-10-15
clarify value for inverting button logic

mattdholmes00:
clarifying how to invert button logic, so user doesn't need to understand that bit 1 = second bit = decimal 2

### DIFF
--- a/common/source/docs/common-buttons.rst
+++ b/common/source/docs/common-buttons.rst
@@ -19,7 +19,7 @@ Alternatively, a PWM signal can be used on the pin for switching, just like the 
 
 There is no MIDDLE position for the function, if it has one, using a button.
 
-The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>`` parameter bit 1 (e.g. BTN_OPTIONSx = 2). This applies to both PWM and switch control of the pin.
+The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>``  bitmask bit 1 (ie adding 2 to the decimal value of the parameter). This applies to both PWM and switch control of the pin.
 
 The function that will be controlled by the button is determined by its ``BTN_FUNCx`` parameter. If set to -1, no function is performed. Otherwise, the number corresponds to the :ref:`Auxiliary Function <common-auxiliary-functions>` number.
 


common/source/docs/common-buttons.rst
```diff
@@ -19,8 +19,8 @@ Alternatively, a PWM signal can be used on the pin for switching, just like the
 
 There is no MIDDLE position for the function, if it has one, using a button.
 
-The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>`` parameter bit 1. This applies to both PWM and switch control of the pin.
+The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>`` parameter bit 1 (e.g. BTN_OPTIONSx = 2). This applies to both PWM and switch control of the pin.
```
Hwurzburg, line 22: this is incorrect....since its a bit mask, the VALUE should have +2 added to it, not set to 2 which possibly destroys the setting of bit 0..the best clarification would be :
```suggestion
The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>``  bitmask bit 1 (ie adding 2 to the decimal value of the parameter). This applies to both PWM and switch control of the pin.
```

common/source/docs/common-buttons.rst
```diff
@@ -19,8 +19,8 @@ Alternatively, a PWM signal can be used on the pin for switching, just like the
...
 
 The function that will be controlled by the button is determined by its ``BTN_FUNCx`` parameter. If set to -1, no function is performed. Otherwise, the number corresponds to the :ref:`Auxiliary Function <common-auxiliary-functions>` number.
 
-The :ref:`BTN_REPORT_SEND<BTN_REPORT_SEND>` sets the length of time that a button state message is sent to the ground control station for display purposes, each time the button changes state. This is a MAVLink2 message.
\ No newline at end of file
+The :ref:`BTN_REPORT_SEND<BTN_REPORT_SEND>` sets the length of time that a button state message is sent to the ground control station for display purposes, each time the button changes state. This is a MAVLink2 message.
```
Hwurzburg, line 26: no obvious change here, but doesn't hurt anything

Hwurzburg (changes requested):

PR #3847, opened 2021-10-19
Fix build error caused by Hondex page

Hwurzburg:
@rmackay9 really need to get your build environment up again


rmackay9:
ah, shoot, sorry!

PR #3848, opened 2021-10-20
Mateksys H743-Wing: Add RC input section

ybeyer:
I connected my Matek H743-Slim to an SBUS receiver and had difficulty configuring it because I couldn't find the appropriate documentation.
For other FCs there is also a "RC inputs" section. I have now added this section for the H743 as well and hope it helps.
By the way, there is an open thread that could possibly be closed by this:
https://discuss.ardupilot.org/t/cannot-calibrate-radio-no-green-bars/67928

### DIFF
--- a/common/source/docs/common-matekh743-wing.rst
+++ b/common/source/docs/common-matekh743-wing.rst
@@ -85,9 +85,22 @@ Default UART order
 - SERIAL6 = USER = UART4
 - SERIAL7 = USER = UART6 (TX only unless :ref:`BRD_ALT_CONFIG<BRD_ALT_CONFIG>` = 1, then RX available also)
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
Sorry, but this is incorrect for many things...
1. RX6 is by default compatible with all RC protocols that are not full duplex (that indeed needs to be updated in the wiki ,in all the FCs and some other places . FPort, SRXL2 and CRSF (recent additions in 4.1) require a UART. 
2. The ALT config is only to allow those to be accessed using the existing RCIN UART (which by default is NOT a UART input, its a timer interrupt input) to avoid consuming a different UART.
3. Any UART RX can be used for RCinput for single wire  and full duplex  RC protocols, if using protocol 23, BUT it must be able to invert its inputs for SBUS and FPort (F4 cpus cannot) or use an external inverter but it cannot decode PPM.
4. The normal RCin configuration can decode SBUS, either inverted or non-inverted from the RX, nothing special need be done.

Hwurzburg:
closing this...I have updated the autopilot page with info on this when using the bdshot firmware as well as detailed RC setup info

PR #3865, opened 2021-10-26
add auto generated Lua docs

IamPete1:
This adds a new submodule: https://github.com/boolangery/sphinx-lua I have then done a patch on top of that to get what we need. 

This then adds two new extensions and uses them to auto generate a nearly ( just missing global methods ) complete set of documentation based on that added to AP recently. 

Takes a little messing, will need to install https://github.com/boolangery/py-lua-doc and install the new submodule from its `setup.py`

Formatting is not fantastic, but we can improve that, this is more just a proof that we can grab the docs from the AP repo and generate the wiki. 

I have left in the existing API documentation, since this is not fantastically useful until we flesh it out a little more with descriptions and such. 

This is what the page ends up looking like:
[Lua Scripts — Copter documentation.pdf](https://github.com/ArduPilot/ardupilot_wiki/files/7422004/Lua.Scripts.Copter.documentation.pdf)

For example:
image

One of the classes that has good documentation:
image


IamPete1:
Updated the templates to better fit our wiki style. Split out into new page for API docs. Currently this includes the existing manual docs because there better than the auto generated ones. The plan in the future would be to migrate those across and for the new page to be completely auto generated.

The auto generation is still missing things like class and method descriptions that are present in `doc.lua` so there are some future enhancements to be done. I'm not sure if that issue is in the lua parser or the docs generator.

Hwurzburg:
does not work on my local build of this PR....ran sphinxsetup.sh   then try to build wiki...get this error:
Extension error:
Could not import extension sphinxcontrib.luadomain (exception: No module named 'sphinxcontrib.luadomain')
Makefile:56: recipe for target 'html' failed

TunaLobster:
Need to rerun Sphinxsetup.sh to install the new python packages.

@IamPete1 Might be helpful to add messages like this one in update.py.
https://github.com/ArduPilot/ardupilot_wiki/pull/3746/commits/7be6cd3051394c9fc64e8fa3b98be45297450555

common/source/docs/common-lua-API.rst
```diff
@@ -0,0 +1,400 @@
+.. _common-lua-API:
+
+[copywiki destination="plane,copter,rover,dev"]
+=====================================
+API Documentation
+=====================================
+
+The API documentation described here is not a complete list, but rather some examples.  For a **full list of the methods** currently available, the LUA auto-generated document file is a complete list, It can be found `here <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/docs/docs.lua>`__ and lists all the available bindings and their parameters.
+
+See :ref:`API Docs Syntax<common-lua-API>`.
```
Hwurzburg, line 10: this points to itself???

common/source/docs/common-lua-scripts.rst
```diff
@@ -18,9 +18,14 @@ While scripting normally requires at least some LUA programming and editing, the
...
 
+.. toctree::
+    :maxdepth: 1
+
+    API Documentation <common-lua-API>
+    LUA Bindings <common-lua-binding-syntax>
 
 Getting Started
-===============
+---------------
```
Hwurzburg, line 28: why are you changing these?

Hwurzburg:
@IamPete1 lets chat on this when you have a chance...been a while...shall we close...dont think we can get the format pretty, right?

IamPete1:
We could get the format better, just is a weird template method I was having a hard time with.

PR #3885, opened 2021-11-06
Wiki:Doc error: param load cmd, changed '\' to '/'

khanasif786:
Problem: param load command was having backslashes ' \ ' instead of forward slashes ' / ' . causing error in parameters loading.
Changes: Changed Backslashes to forward slashes


Hwurzburg:
thanks

PR #3888, opened 2021-11-07
Frontend performance improvement

khancyr:
This brings some small html and css fixes for frontend.

It also add webp images version for much better performance. All images have been converted to webp and scalled correctly to the max size allowed on carrousel: 920x480


brunoolivieri:
LGTM

PR #3894, opened 2021-11-12
Docs: added more info

AlksSAV:
this is a PR to add information about different modifications of the JSN-SR04T


Hwurzburg:
file names cannot have spaces in Linux....rename the image files....this will not build properly

AlksSAV:
> file names cannot have spaces in Linux....rename the image files....this will not build properly

Accepted, I will rename

Hwurzburg:
thanks...merging

PR #3903, opened 2021-11-15
Update common-lua-scripts.rst, remove duplicates

Dan-Moss:
remove duplicates


Hwurzburg:
thanks...I have squashed and changed commit message to one that has info....merging...

PR #3909, opened 2021-11-17
Dev: add VSCode GDB Debugging Profiles

### LINKED ISSUE #3629: VS Code Debugging: add how to debug startup tasks
Here: https://ardupilot.org/dev/docs/debugging-with-gdb-on-linux.html#vs-code

We should add how to "Launch" debugging in VSCode so users can debug object construction, SITL parameters. 

We may also want to create a new header in the Debugging tree that lists "Debugging with an IDE" new users seem not be seeing those directions.

hendjoshsr71:
Closes #3629

Adds VSCode GDB Debugging Profile Information for autotests, attaching, constructor debugging, and unit tests

@Hwurzburg  Here you go. It probably needs a little help in terms of formatting and wording but it should be most of the way there. 

I still need to figure out how to add the JSON file I placed on this page into the actual ArduPilot repo.


dev/source/docs/debugging-with-gdb-using-vscode.rst
```diff
@@ -0,0 +1,223 @@
...
+.. image:: ../images/VSCode-config1.png
+    :target: ../_images/VSCode-config1.png
+
+
+
+
+Example launch.json File
+------------------------
+
+.. code-block:: json
```
Hwurzburg, line 47: just use ::     sphinx does not lex json as a code block and gives build error...otherwise its fine...

Hwurzburg (changes requested):

dev/source/docs/debugging-with-gdb-using-vscode.rst
hendjoshsr71: I've changed the code block to javascript  as that will at least parse nearly the same. It seems newer versions of sphinx can parse json. But not important enough to go down that hole...

PR #3913, opened 2021-11-18
addition of horizon31 turnkey products

climr:
Add details for horizon31.com turnkey companion computer solutions, PixC4-Jetson and PixC4-Pi


Hwurzburg:
LGTM...merging

PR #3914, opened 2021-11-18
Horizon31 PixC4-Jetson hardware added to wiki

climr:
Addition of the closed hardware PixC4-Jetson turnkey FMU/companion computer hardware.


Hwurzburg:
squashed and merging

PR #3929, opened 2021-11-24
frontend: use lazyloading for images

khancyr:
This allow the browser to load images only on need.
This doesn't work on Safari for now


brunoolivieri:
What happens in Safari?  

The page loads ok in Safari?

khancyr:
According to https://caniuse.com/loading-lazy-attr lazy loading isn't supported

Hwurzburg:
@brunoolivieri can you approve so it can be merged?

brunoolivieri:
LGTM

PR #3932, opened 2021-11-25
4.1:Add EKF sources page

Hwurzburg:


rmackay9:
Thanks for this.

There's some overlaps with this page, https://ardupilot.org/copter/docs/common-non-gps-to-gps.html.  We should link each to the other I think.

PR #3936, opened 2021-11-28
Update radio-failsafe.rst

geofrancis:
This scenario happened on a commercial hex using this failsafe method.

### DIFF
--- a/copter/source/docs/radio-failsafe.rst
+++ b/copter/source/docs/radio-failsafe.rst
@@ -174,3 +174,5 @@ Although this mostly works it is not recommended because it will not
 trigger if the receiver loses power or if the wires between the receiver
 and autopilot are broken.
 
+This will Cause rapid flight mode changes and a crash if you attempt to land beyond radio range. as the receiver looses signal it will change to RTL mode but as soon as it climbs and regains connection it will change back to your landing mode, it will rapidly cycle between the mode your attempting to land in and RTL until it runs out of battery or crashes.
+


Hwurzburg:
you should be able to recover from this situation by simply changing the flight mode on TX to RTL......so that when RC is recovered it continues back in RTL mode....

copter/source/docs/radio-failsafe.rst
```diff
@@ -174,3 +174,5 @@ Although this mostly works it is not recommended because it will not
 trigger if the receiver loses power or if the wires between the receiver
 and autopilot are broken.
 
+This will Cause rapid flight mode changes and a crash if you attempt to land beyond radio range. as the receiver looses signal it will change to RTL mode but as soon as it climbs and regains connection it will change back to your landing mode, it will rapidly cycle between the mode your attempting to land in and RTL until it runs out of battery or crashes.
```
Hwurzburg, line 177:
```suggestion
.. note :: This can cause rapid flight mode changes if you attempt to land beyond radio range. As the receiver looses signal it will change to RTL mode but as soon as it climbs and regains connection it will change back to your landing mode and re-attempt the land. To avoid this simply change the mode on the TX to RTL so that it will continue to return back into reliable radio range when RC link is re-established.
```

Hwurzburg (changes requested):

geofrancis:
> you should be able to recover from this situation by simply changing the flight mode on TX to RTL......so that when RC is recovered it continues back in RTL mode....

that doesnt help if your battery is running out and your trying land it before it falls out the sky.

Hwurzburg:
this is a case of improper setup then...you do not have the critical battery failsafe set to land...which will override the RC failsafe

geofrancis:
> I understand ...battery failsafe is being overriden by active mode changes...will discuss with others...may need to make a stronger warning on this setup than what was suggested....like do not use this method! @rmackay9 comments?

yes there is no reason it should ever be used. the only other time I have seen the mode switch used for failsafe was on the old walkera x350 quads,

Hwurzburg:
I have merged changes to plane and copter failsafe sections which address this after discussing with vehicle leads..for copter we are changing to a hard do not do this...for plane adding a warning since for non-quaplanes its not as big an issue...closing...thanks for pointing this out

PR #3938, opened 2021-11-29
Update common-glossary.rst

MaxHearnden:
Change altitude (height) to attitude (orientation) bringing it inline with the [common abbreviations](https://github.com/ArduPilot/ardupilot_wiki/blob/master/common/source/docs/common-abbreviations.rst) and the text describing the AHRS


Hwurzburg:
nice catch...thanks...merging

PR #3946, opened 2021-11-30
added omni 

geofrancis:
added omni and re arranged so it's less confusing.


Hwurzburg:
thought about it a while and my view is unchanged.....closing...thanks though

PR #3947, opened 2021-11-30
Re-org copter tuning into bite-size sections,remove obsolete

Hwurzburg:
basically just took Leonards page and sectioned it up, no content is lost...added some rationale for setup requirements...added notch filter before final tuning...archived obsolete tuning pages

@lthall you say in manual tune instructions :  that :ref:`ATC_THR_MIX_MAN <ATC_THR_MIX_MAN>` must be set properly but no where do we tell them what or how it should be set....need to add this


lthall:
We just need to make sure they set it to 0.1 as I say here:
image

lthall (approved):
Looks good to me.

PR #3953, opened 2021-12-02
Add warning about using receiver mode failsafes

Hwurzburg:
Added some warning about using the


rmackay9:
This seems ok to me.  The point, "will not trigger if the receiver loses power" is still valid though so maybe put that back if you want.

Hwurzburg:
> This seems ok to me. The point, "will not trigger if the receiver loses power" is still valid though so maybe put that back if you want.
Actually this was a bit miss-leading IMHO since it will do a failsafe action in that case if FS_THR_ENABLE is set...independent of whether this is used or not

PR #3954, opened 2021-12-03
Plane: add plane records

peterbarker:
The numbers in here are completely made up - I'm not even sure these are the true records.


hendjoshsr71:
I really like this idea. But maybe it should be for more than just plane and a link prominent on the Wiki to it? Basically to show off it can handle challenging problems both in breadth and depth. 

Oh and maybe a smallest/lightest category too.

peterbarker:
> Oh and maybe a smallest/lightest category too.

How do we measure "smallest"?  Wingspan?  Displacement?

hendjoshsr71:
> > Oh and maybe a smallest/lightest category too.
> 
> How do we measure "smallest"? Wingspan? Displacement?

Can i punt and say both? ;) If I had to chose one do wingspan.

MichelleRos:
Great idea 👍

Hwurzburg:
Be sure to get permission from the record holder first.... Some might not be legal and they do not wish them published further

peterbarker:
> Be sure to get permission from the record holder first.... Some might not be legal and they do not wish them published further

I'll do that when populating this for real.

Seems to be general agreement that this would be nice.

Any idea where we would link to this from elsewhere?

peterbarker:
I've added a note into here with instructions on how to update, a note to check legality of flight, and a note that we won't accept any flights which would violate the Dev CoC...

plane/source/docs/records.rst
```diff
@@ -0,0 +1,35 @@
...
++-------------------------+-----------------------+--------------+----------------+--------------------------------------------------+
++ Lightest                | ?                     +              +                +                                                  +
++-------------------------+-----------------------+--------------+----------------+--------------------------------------------------+
++ Longest                 | Krauss Aerospace      + 26 hours     +                +                                                  +
++-------------------------+-----------------------+--------------+----------------+--------------------------------------------------+
++ Furthest                | ?                     + km           +                +                                                  +
++-------------------------+-----------------------+--------------+----------------+--------------------------------------------------+
++ Higest $ Value Delivery | ?                     + US$          +                +                                                  +
```
hendjoshsr71, line 34:
```suggestion
+ Highest $ Value Delivery | ?                     + US$          +                +                                                  +
```

peterbarker: Done

Issue #3273, opened 2021-01-13
combine or group FFT and Harmonic Notch pages


rmackay9:
I think we should combine or group these two pages under a landing page:

- [Managing Gyro Noise with the Static Notch and Dynamic Harmonic Notch Filters](https://ardupilot.org/copter/docs/common-imu-notch-filtering.html)
- [Managing Gyro Noise with In-Flight FFT](https://ardupilot.org/copter/docs/common-imu-fft.html)

I'm definitely not an expert in this harmonic notch and FFT feature but I think these two pages have duplicate information and should be rationalised somethow.

FYI @andyp1per

hendjoshsr71:
Just a poke. We have had a user mention issues about these pages and difficulty in parsing them as a whole idea as stated in this issue.

Hwurzburg:
done

PR #3960, opened 2021-12-04
Re-org notch filtering into easier to follow pages

Hwurzburg:
same approach as taken in #3947...makes is easier to follow for a given setup choice...trimmed it down a bit to essentials so as not to lose a reader with too many words... addresses #3273.....next will trim down the In-Flight FFT to remove duplications

### DIFF
--- a/common/source/docs/common-rpm-based-notch.rst
+++ b/common/source/docs/common-rpm-based-notch.rst
@@ -1,0 +1,16 @@
+.. _common-rpm-based-notch:
+
+====================================
+RPM Sensor Based Dynamic Notch Setup
+====================================
+
+- Set :ref:`INS_HNTCH_ENABLE <INS_HNTCH_ENABLE>` = 1 to enable the harmonic notch
+- Set :ref:`INS_HNTCH_REF <INS_HNTCH_REF>` = 1 to set the harmonic notch reference value to unscaled, unless the RPM sensor is reporting a fraction of the actual rotor speed. This might occur in a geared drive train setup. If the reported RPM is 1/3 the actual, for example, set this to 0.33.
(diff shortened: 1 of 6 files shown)


common/source/docs/common-rpm-based-notch.rst
```diff
@@ -0,0 +1,14 @@
...
+RPM Sensor Based Dynamic Notch Setup
+====================================
+
+- Set :ref:`INS_HNTCH_ENABLE <INS_HNTCH_ENABLE>` = 1 to enable the harmonic notch
+- Set :ref:`INS_HNTCH_REF <INS_HNTCH_REF>` = 1 to set the harmonic notch reference value to unscaled, unless the RPM sensor is reporting a fraction of the actual rotor speed. This might occur in a geared drive train setup. If the reported RPM is 1/3 the actual, for example, set this to 0.33.
+- Set :ref:`INS_HNTCH_FREQ <INS_HNTCH_FREQ>` = below the hover frequency - you can easily determine this by performing a gentle hover and looking at the RPM sensor log data (RPM.rpm1)
+- Set :ref:`INS_HNTCH_BW <INS_HNTCH_BW>` = half of INS_HNTCH_FREQ
+
+Checking Harmonic Notch Effectiveness
+=====================================
```
andyp1per, line 13: This is a general section that belongs elsewhere

Hwurzburg: both esc and rpm (and eventually in-flight FFT) should  (esc doesnt...will add) have this section saying that post flight harmonic notch effectiveness checking is not required, but...if you want, use the method in the throttle based section to do so....this needs a link added to : ref:`common-imu-notch-filtering-post-configuration-flight-and-post-flight-analysis` just like the heli section and esc section have....so that if someone follows those paths they get the link to the method of checking effectiveness and will add the section to esc based also...

Hwurzburg: updated

PR #3964, opened 2021-12-05
Correct sentence that can confuse beginner.

Pushpendra766:
Corrected the spelling of 'step' and transformed that point like the second one.


Hwurzburg:
@Pushpendra766 thanks for the formatting change, but the warning is still confusing....what "older" version is this referring to? 
and there is no git submodule init step shown on this page, perhaps it should be explicitly referred to?

Pushpendra766:
After reading the document, I found that git 1.8.x or earlier are old versions here. And init step is referring to initial steps.

Hwurzburg:
thanks...merging

PR #3969, opened 2021-12-07
Blheli32 current monitoring capabilities

geofrancis:
Clearing up the confusion about the capabilities of blheli32 current monitoring. Should be added to plane and rover too.

### DIFF
--- a/common/source/docs/common-dshot-blheli32-telemetry.rst
+++ b/common/source/docs/common-dshot-blheli32-telemetry.rst
@@ -4,11 +4,9 @@
 BLHeli_32 and BLHeli_S ESC Telemetry
 ====================================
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
adding to the common file automatically will appear in all vehicles wikis...
I think I would rather reword the intro paragraph, which does state that monitoring depends on the escs capabilites already....
I will pull this PR down locally and rework and push back up to your repo for merging as soon as I can get to it...thanks

geofrancis:
ok, this is to resolve an old running post on ardupilot.org where a number of users were posting complaining about mission planner, ardupilot etc not displaying current values on their ESC telemetry when the simple answer was the ESC was not capable of it. After checking, almost no 4 in 1 escs do current over blheli telemetry so I think the problem is users just see in the specs that it supports telemetry and current sensing not realizing it was only using telemetry for rpm.
Re-word it however you think will work to make this much clearer to users

https://discuss.ardupilot.org/t/copter-3-6-and-blheli-32-no-curr-info/34657/13

hendjoshsr71:
@geofrancis  Thanks for this!

Hwurzburg:
pushed changes...thanks..merging

PR #3972, opened 2021-12-09
fix RCx_OPTION value for walking height in document (203-> 211)

hfujikawa77:
I fixed the document based on this commit.

ArduPilot/ardupilot@a97786c#diff-4394277d4bfb7b48f27084825c024642a3522592a67620c50fcea4877dbe1ea1


Hwurzburg:
thanks....I have also submitted a PR for the code's metadata for the RCx_OPTIONS to reflect this, that was omitted when the change was made to the function.

PR #3995, opened 2021-12-20
Add section on notch slewing

Hwurzburg:


common/source/docs/common-esc-telem-based-notch.rst
```diff
@@ -18,6 +17,19 @@ Set the :ref:`INS_HNTCH_REF <INS_HNTCH_REF>` parameter to 1, which will disable
...
+
+By default the update rate is 200Hz and where the source of frequency information is slower than that - for instance when using ESC telemetry where the maximum rate that can be sustained is about 100Hz - ArduPilot will slew the frequency changes at 200Hz to avoid large steps. The slewed rate is the rate that is reported by ESC telemetry, although the raw rate can be seen in the logs as well.
+
+On systems with faster CPUs (H7 based autopilots) it is possible to update the harmonic notch at the main loop rates used for VTOL aircraft (typically 300-400Hz set by :ref:`SCHED_LOOP_RATE<SCHED_LOOP_RATE>`) by setting bit 3 of :ref:`INS_HNTCH_OPTS<INS_HNTCH_OPTS>`, i.e. 
+
+- :ref:`INS_HNTCH_OPTS<INS_HNTCH_OPTS>` = 4
+
+Slewing ensures that the step changes at each update tick are smooth, but for optimum system performance you can use bi-directional Dshot ESCs which can deliver frequency updates at 400Hz, using the above option, if possible .
```
andyp1per, line 31: If you want you can mention that the bi-dir update rate is dependent on PWM groups and dshot rate. If you have a single PWM group then then update rate is the dshot rate (1Khz bydefault) / number of channels (4) - so 250Hz. If you have two PWM groups then it would be 1Khz / 2 (channels per group). So on my KakuteF7 I have a 800Hz loop rate and a dshot rate of 2 == 1.6Khz and two timer groups, so the bi-dir update rate is 1.6Khz / 2 = 800Hz.

andyp1per (approved):

common/source/docs/common-esc-telem-based-notch.rst
Hwurzburg: I think that it would be too complex to cover all the variations

PR #3997, opened 2021-12-21
Dev: small update to mavlink interface section

rmackay9:
This PR attempts to make it easier for developers to find the mavlink interface messages they are looking for by splitting upC Copter's commands page into two pages.  One for movement related commands (which is what most devs want) and non-movement related commands (which actually apply to all vehicles I think)

I have tested these changes on my local WSL machine and they seem OK.


khancyr (approved):
yep make sens

PR #4002, opened 2021-12-22
Dev: mavlink interface for Plane

rmackay9:
This PR improves the "MAVLink Interface" page for Plane by:

- Reducing the scope of the page so it only covers movement commands.  Non-movement commands are mostly consistent across vehicles so we have new pages for these.  All commands removed from the Plane page have been added to other pages (if they weren't there already)
- clarifies that the MAV_CMD_NAV_WAYPOINT must be within a MISSION_ITEM_INT and provides details on what fields to set and even gives 3 examples that can be copy-pasted into SITL/MAVProxy
- adds DO_REPOSITION message to the list of MAV_CMDs supported

This PR also includes a couple of drive-by improvement:

- Improved formatting of the mavlink-mission-command-messages page by removing the blank lines between each command.  This makes it easier for readers to scan the page
- MAVLink Interface page for Copter loses the arm/disarm command which is already covered on another page.  Also adds DO_REPOSITION message to the list of MAV_CMDs supported

This has been tested by building and manually reviewing the pages on my local PC


Hwurzburg:
LGTM

PR #4006, opened 2021-12-23
Update SITL with RealFlight with some tips

Hwurzburg:


TunaLobster:
That's helpful enough to be put in other places too. Not just real flight. Is there a page for all of the sim_vehicle.py options?

Hwurzburg:
I will put it on my list to add here: https://ardupilot.org/dev/docs/using-sitl-for-ardupilot-testing.html

Hwurzburg:
done

PR #4008, opened 2021-12-24
Update Pix32V5 Wiki

vincentpoont2:
Add UART mapping


common/source/docs/common-holybro-pix32v5.rst
```diff
@@ -69,6 +69,17 @@ Specifications
      - Dimensions: 45mm x 45mm x 13.5mm
      - Weight: 33.0g
 
+UART Mapping
+=============================
```
Hwurzburg, line 73: Title underline must be the same length as title text.....I will fix

PR #4010, opened 2021-12-27
update Holybro RTK wiki

vincentpoont2:
add "If unable to preform normal compass calibration "compass dance" for any reason, set parameter COMPASS_ORIENT=6 (Yaw270) for proper compass orientation."


Hwurzburg:
minor cleanups....merging

PR #4014, opened 2022-01-01
Common: Fixing URL link errors in ThoneFlow-3901U

muramura:
URL link error 404 occurred in ThoneFlow-3901U.
I will fix the URL.


khancyr (approved):
LGTM

PR #4019, opened 2022-01-07
Dev: mavlink interface copter command examples of moving right without yawing

rmackay9:
At leaste one user struggled to stop the vehicle from yawing when providing body-frame velocities so this provides two specific examples (one for velocity, one for acceleration).

https://discuss.ardupilot.org/t/fly-in-circles-with-set-position-target-local-ned-command/80289/3

I've tested this locally on my machine and it looks ok.


hendjoshsr71:
Nice! You can also just specify yaw = 0 and probably (i've not tested this one) do the same thing with both yaw and yaw rate zero as well.

Hwurzburg:
lgtm...thanks

PR #4029, opened 2022-01-12
update.py: use unlink/rename in place of os.replace

peterbarker:
os.replace not in Python2.


peterbarker:
@stephendade could you check this hasn't broken Windows, please?

tridge (approved):

stephendade:
> @stephendade could you check this hasn't broken Windows, please?

A bit late, but all good on Windows. We're actually running the wiki CI over Windows and Ubuntu now, so should be easier to find these issues.

PR #4030, opened 2022-01-12
switched to py3 to fix blog post generator

tridge:


Hwurzburg:
seems to be building now without this...but cant see the harm

brunoolivieri:
I check on the server and it is ok.

The harm would be on frontend discourse links:

image

PR #4031, opened 2022-01-12
Common: add swag store to stores page

rmackay9:
This isn't as visible as it's old place on the wiki menu but this is the limit of my abilities :-).  I've tested this locally on my machine.


Hwurzburg:
LGTM

PR #4032, opened 2022-01-13
Reword APM to autopilot/ArduPilot as appropriate

nkruzan:
This is a follow up to previous PR's I have submitted, rewording APM to autopilot or ArduPilot as appropriate for the context. I don't think I have modified any of the pages that are supposed to be specific to APM hardware. Will do more updates as I find the time/motivation.


Hwurzburg:
LGTM...merging...thanks

PR #4034, opened 2022-01-13
Fix case error with log messages

stephendade:
The log message page had the wrong case (logmessages.rst vs LogMessages.rst).

This PR restores it back to all lower-case

### DIFF
--- a/common/source/docs/common-logs.rst
+++ b/common/source/docs/common-logs.rst
@@ -15,16 +15,16 @@ Topics related to logging and analysis
     Dataflash Logs <common-downloading-and-analyzing-data-logs-in-mission-planner>
 [/site]
 [site wiki="plane"]
-    Plane Log Messages <LogMessages>
+    Plane Log Messages <logmessages>
 [/site]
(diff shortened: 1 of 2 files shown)


peterbarker:
Merged, thanks!

Hwurzburg:
and this just re-breaks everything I had just fixed over the last week

Hwurzburg:
I reverted this so I can continue working without massive errors until I understand this.....there are other impacts

Hwurzburg:
okay, I have had time to track the issue:
- originally the file was lowercase
- in Dec Stephen changed to CamelCase which created massive build errors
- I fixed those build errors in the wiki and in gitignore last week....which also required removing the old lowercase files from local source since they were no longer being autoremoved
- this change recreates all the build errors for CamelCase files which needed to be purged locally after this fix and does not fix the gitignore
- I will re-re-revert and fix the gitignore
- perhaps next time I will get a chance to review wiki changes before merging, even if its not my area of expertise (update.py) so I can test build locally

PR #4052, opened 2022-01-18
Fix Typos #4 - Common and MavProxy

nkruzan:
Finishes these 2 sections of wiki


mavproxy/source/docs/modules/ublox.rst
```diff
@@ -1,5 +1,5 @@
 ==========================
-uBlox GPS Managment Module
+uBlox GPS Management Module
```
Hwurzburg, line 2: need to fix overline and underline length

Hwurzburg (changes requested):

mavproxy/source/docs/modules/ublox.rst
nkruzan: Fixed, Thanks.

PR #4064, opened 2022-01-23
Add details in video transmitter setup.

markfickett:
* Clarify CRSF setup when Vtx and RC receiver both use the protocol. I saw a thread w/ someone who was trying to use CRSF w/ both rx and Vtx and wanted to record the info here, I made the same mistake of trying to use CRSF directly from the FC for both. I ended up just using SmartAudio instead, which is working great.
* Add details to RCx_OPTION usage around PWM range and output power settings. The most important part here was PWM range for me (my tx sends 1000-2000 so a 3-position switch wasn't overlapping with the 6-position switch interpretation well). Short of copying the code for assigning vtx power levels into here as a table (which could be useful!) I figured one example would help folks figure out what's going on.

### DIFF
--- a/common/source/docs/common-vtx.rst
+++ b/common/source/docs/common-vtx.rst
@@ -28,7 +28,7 @@ CRSF (Crossfire)
 
 CRSF is a bi-directional protocol that requires both TX and RX of a serial port to be connected. Follow the instructions for CRSF telemetry to setup CRSF support (See :ref:`common-tbs-rc`). If you only wish to use CRSF for VTX support, then (again using SERIAL5 as an example) :ref:`SERIAL5_PROTOCOL <SERIAL5_PROTOCOL>` should be set to 29 rather than 23. You will also need to make sure your VTX is configured for CRSF control.
 
-When using CRSF for both RC and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions.
+When using CRSF for both RC/Telem and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions. In this scenario, you must connect your flight controller's UART to your CRSF receiver, and then connect your CRSF receiver to the VTX to forward CRSF commands, you cannot connect to the FC using CRSF directly for both the Receiver and VTX using two UARTs.
 
 Reboot your flight controller and power your video transmitter. At startup you should see a video message indicating the current video settings. All the video settings are stored in the ``VTX_x`` parameters and by default these will be changed to reflect the currently configured VTX settings. Once booted you can modify these settings and they will be reflected in the VTX configuration.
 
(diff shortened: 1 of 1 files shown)


common/source/docs/common-vtx.rst
```diff
@@ -28,7 +28,7 @@ CRSF (Crossfire)
 
 CRSF is a bi-directional protocol that requires both TX and RX of a serial port to be connected. Follow the instructions for CRSF telemetry to setup CRSF support (See :ref:`common-tbs-rc`). If you only wish to use CRSF for VTX support, then (again using SERIAL5 as an example) :ref:`SERIAL5_PROTOCOL <SERIAL5_PROTOCOL>` should be set to 29 rather than 23. You will also need to make sure your VTX is configured for CRSF control.
 
-When using CRSF for both RC and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions.
+When using CRSF for both RC and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions. In this scenario, you must connect your flight controller's UART to your rx, and then connect your rx to forward CRSF to your VTX, you cannot directly connect the FC using CRSF direclty to both the rx and VTX.
```
andyp1per, line 31: "directly"

andyp1per (changes requested):
Small typo

common/source/docs/common-vtx.rst
markfickett: Thanks for the quick review & catch, removed the extra "directly."

Hwurzburg:
this is still confusing, but now that I understand will create a couple of pictures to add to make it clear to users...on my list...

common/source/docs/common-vtx.rst
```diff
@@ -28,7 +28,7 @@ CRSF (Crossfire)
 
 CRSF is a bi-directional protocol that requires both TX and RX of a serial port to be connected. Follow the instructions for CRSF telemetry to setup CRSF support (See :ref:`common-tbs-rc`). If you only wish to use CRSF for VTX support, then (again using SERIAL5 as an example) :ref:`SERIAL5_PROTOCOL <SERIAL5_PROTOCOL>` should be set to 29 rather than 23. You will also need to make sure your VTX is configured for CRSF control.
 
-When using CRSF for both RC and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions.
+When using CRSF for both RC and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions. In this scenario, you must connect your flight controller's UART to your rx, and then connect your rx to forward CRSF to your VTX, you cannot connect the FC using CRSF direclty to both the rx and VTX.
```
Hwurzburg, line 31:
```suggestion
When using CRSF for both RC/Telem and VTX control it is important to disable "my VTX" support on Open TX transmitters. Follow your transmitters' instructions. In this scenario, you must connect your flight controller's UART to your CRSF receiver, and then connect your CRSF receiver to the VTX to forward CRSF commands, you cannot connect to the FC using CRSF directly for both the Receiver and VTX using two UARTs.
```

Hwurzburg:
added some figures in a previous PR...tweaked this a bit...rebased and squashed...merging...thanks

markfickett:
Thanks very much! The new diagrams look great too, definitely helpful.

PR #4075, opened 2022-01-27
Dev: Update ROS instructions

stephendade:
Some of the ROS instructions are quite out of date. This updates them.

### DIFF
--- a/dev/source/docs/ros-install.rst
+++ b/dev/source/docs/ros-install.rst
@@ -3,9 +3,7 @@
 Installing ROS
 --------------
 
-General instructions for `installing ROS Kinetic can be found here <http://wiki.ros.org/kinetic/Installation>`__.  Ubuntu users should follow `these instructions <http://wiki.ros.org/kinetic/Installation/Ubuntu>`__. (only for Ubuntu 16.04)
-
-Ubuntu 18.04 users will need to install ROS Melodic. General instructions are available `here <http://wiki.ros.org/melodic/Installation>`__. Replace ``kinetic`` with ``melodic`` in the instructions below.
+General instructions for `installing ROS can be found here <http://wiki.ros.org/ROS/Installation>`__.
 
  - if installing on a desktop PC, install ROS Desktop-Full, on a companion computer ROS-Base is enough
  - add ROS tool to your shell as stated in ROS wiki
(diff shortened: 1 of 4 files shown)


dev/source/docs/ros-install.rst
```diff
@@ -14,11 +12,13 @@ Ubuntu 18.04 users will need to install ROS Melodic. General instructions are av
...
 
-Instructions for `installing MAVROS can be found here <https://github.com/mavlink/mavros/tree/master/mavros#installation>`__ but in short involve running the following command
+MAVROS is compatible with all recent versions of ROS including Kinetic, Melodic and Noetic.
+
+Instructions for `installing MAVROS can be found here <https://github.com/mavlink/mavros/tree/master/mavros#installation>`__ but in short involve running the following command.
 
 .. code-block:: bash
 
-    sudo apt-get install ros-kinetic-mavros ros-kinetic-mavros-extras
+    sudo apt-get install ros-melodic-mavros ros-melodic-mavros-extras
```
khancyr, line 21: I think that is better to default to noetic  as it support py3 !

dev/source/docs/ros-timesync.rst
```diff
@@ -28,17 +28,7 @@ ArduPilot Configuration
...
-For rover, open `Rover.cpp <https://github.com/ArduPilot/ardupilot/blob/master/Rover/Rover.cpp>`__ and within the scheduler_tasks[] array ensure the gcs entries have their update rate set to 200 like below:
-
-::
-
-    SCHED_TASK(gcs_retry_deferred,    200,    500),
-    SCHED_TASK(gcs_update,            200,    500),
-    SCHED_TASK(gcs_data_stream_send,  200,   1000),
-
-After making this change you will need to compile rover and upload the new binary to the flight controller.  Hopefully this will not be necessary in future versions of Plane or Rover.
+For ArduPilot, the SCHED_LOOP_RATE parameter should be increased because mavros ignores TIMESYNC messages if the round trip time is more than 10ms.
```
khancyr, line 31: I don't think it should be raise on copter, no  ?

dev/source/docs/ros-install.rst
stephendade: Fixed!

dev/source/docs/ros-timesync.rst
stephendade: I've only ever seen it on Rover, so will put in a note that it will _generally_ only occur on rover

dev/source/docs/ros-timesync.rst
```diff
@@ -26,19 +26,9 @@ Mavros's `apm_config.yaml <https://github.com/mavlink/mavros/blob/master/mavros/
...
-For rover, open `Rover.cpp <https://github.com/ArduPilot/ardupilot/blob/master/Rover/Rover.cpp>`__ and within the scheduler_tasks[] array ensure the gcs entries have their update rate set to 200 like below:
-
-::
-
-    SCHED_TASK(gcs_retry_deferred,    200,    500),
-    SCHED_TASK(gcs_update,            200,    500),
-    SCHED_TASK(gcs_data_stream_send,  200,   1000),
-
-After making this change you will need to compile rover and upload the new binary to the flight controller.  Hopefully this will not be necessary in future versions of Plane or Rover.
+If the ``RTT too high for timesync`` message is shown in the Mavros console (most commonly when using ArduRover), the :ref:`SCHED_LOOP_RATE <SCHED_LOOP_RATE>` parameter should be increased. This is because Mavros ignores TIMESYNC messages if the round trip time is more than 10ms.
```
khancyr, line 31: thinking  about this we should propose a value for the loop_rate and what it implies. 
It seems that the 200Hz value was a safe choice before. But I suspect that raising the main loop implies some retuning.

khancyr (approved):
Beside my comments on the loop rate that is good now (not that blocking, it wasn't explained more before) ! Thanks for this

PR #4077, opened 2022-01-27
update.py: copy old-release param files if not already in output dir

peterbarker:


Hwurzburg:
@peterbarker any reason not to merge this? I cant test of course

PR #4080, opened 2022-01-28
Common: optflow setup page gets inflight cal video

rmackay9:
This completes the wiki instructions for the new optical flow inflight calibration feature

This has been tested on my local machine.


rmackay9:
I've added a 2nd commit as well to add some links to the GPS/Non-GPS transitions page because a question about this appeared int he comments of the video I've added.

Hwurzburg:
since this is 4.2 we'll wait the few days before its beta before merging

rmackay9:
ok, the feature is in master now. txs

PR #4094, opened 2022-02-03
common: add document on running ArduPilot without RC inputs

peterbarker:
@Hwurzburg what do you think of this as a concept?

### DIFF
--- a/common/source/docs/common-running-without-rc.rst
+++ b/common/source/docs/common-running-without-rc.rst
@@ -1,0 +1,40 @@
+.. _common-running-without-rc:
+
+========================================
+Running ArduPilot without an RC Receiver
+========================================
+
+This article provides an overview of how ArduPilot can be configured
+to safely run with no RC receiver attached.
+
+.. note::
+
+   Fitting your vehicle with an RC receiver allows for additional recovery measures should your vehicle misbehave.  Particularly while creating and developing a vehicle we strongly recommend flying with an RC receiver/transmitter pair providing this additional recovery mechanism.
+
+  Before making the decision to configure your vehicle without an RC receiver, consider options such as ignoring RC failure for failsafe purposes; see "Configuring for Emergency RC Receiver Use", below.
+
+Configuring for no RC Receiver
+==============================
+
+Configuring Failsafes
+---------------------
+ - this is somewhat vehicle-specific, but for ArduCopter FS_THR_ENABLE should be zero THR_FS_ENABLE should be zero
+ - it is strongly suggested that a GCS failsafe be configured in the case that an RC failsafe is not configured.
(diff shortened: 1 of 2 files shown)


proficnc:
Excellent. Quite a common configuration for swarming drones and delivery systems

Hwurzburg:
I dont have any issue with this as long as there are enough caveats 
I would go into a LOT more detail of safe alternate control of the vehicle...ie a GCS is necessary and setup for and using a joystick for anything other than AUTO missions...GCS failsafes setup, etc... the exception being rover with a KICK or GO switch, but LOTs of caveats about that one!
-should link in the Advanced configurations TOC

common/source/docs/common-running-without-rc.rst
```diff
@@ -0,0 +1,40 @@
...
+========================================
+
+This article provides an overview of how ArduPilot can be configured
+to safely run with no RC receiver attached.
+
+.. note::
+
+   Fitting your vehicle with an RC receiver allows for additional recovery measures should your vehicle misbehave.  Particularly while creating and developing a vehicle we strongly recommend flying with an RC receiver/transmitter pair providing this additional recovery mechanism.
+
+  Before making the decision to configure your vehicle without an RC receiver, consider options such as ignoring RC failure for failsafe purposes; see "Configuring for Emergency RC Receiver Use", below.
```
nkruzan, line 14: I don't see a section marked "Configuring for Emergency RC Receiver Use" below

Hwurzburg:
that would work, only Rishab, Randy or I usually do that page.......as it stands, I cant get an output file from the above...maybe if its made into a standalone script generating an rst file in the directory from which its executed?

nkruzan:
> that would work, only Rishab, Randy or I usually do that page.......as it stands, I cant get an output file from the above...maybe if its made into a standalone script generating an rst file in the directory from which its executed?

I think this is meant for #4087

Hwurzburg:
yep too many open windows...moved

common/source/docs/common-running-without-rc.rst
peterbarker: > I don't see a section marked "Configuring for Emergency RC Receiver Use" below

Yeah, this is a WIP... patches against it welcome.  It requires a chunk of research and *must* be right....

amilcarlucas:
This seams to duplicate this existing page: https://ardupilot.org/copter/docs/common-gcs-only-operation.html

Hwurzburg:
@peterbarker close this?

Hwurzburg:
@peterbarker since there is no response above, I am closing this based on the wiki already addresses this...feel free to reopen if you disagree...

Issue #4072, opened 2022-01-27
Rover: add "boat mode" page to rover wiki


rmackay9:
We should move the copter "boat mode" wiki page https://ardupilot.org/copter/docs/boat-mode.html to be common and then:

- make it appear on the rover wiki
- link it from the special page for boats: https://ardupilot.org/rover/docs/boat-configuration.html
- maybe rename it because "boat mode" for people using boats will be confusing... at the same time though the community calls this feature "boat mode" in the context of copters so I'm unsure if renaming it is helpful

Another alternative is to just add a new page for boats.

PR #4095, opened 2022-02-03
Boatmode

Hwurzburg:
addresses #4072


nkruzan (approved):
LGTM

common/source/docs/common-advanced-configuration.rst
```diff
@@ -83,6 +83,7 @@ tuning options for the vehicle.
 [site wiki="copter"]
     Motor Thrust Scaling <motor-thrust-scaling>
 [/site]
+    Moving Vehicle Initialization <common-moving-vehicle-initialization>
```
rmackay9, line 86: can we add "(aka Boat Mode)" here and on the page itself?

I wonder if "Moving Vehicle Initialisation" is clear enough.  How about "Starting from a moving platform"?... or "Initialising from a Moving Platform"?

rmackay9 (approved):
I don't feel strongly enough about my comments to block merging.. :-).

common/source/docs/common-advanced-configuration.rst
Hwurzburg: if someone searched they would find boat mode and be re-directed....but will change to "Initializing from a Moving Platform"

PR #4100, opened 2022-02-04
Dev: add note about VcXsrv  Windows firewall settings

hendjoshsr71:
Adds a note about windows firewall settings that can cause issues when setting up WSL


Hwurzburg:
thanks...

PR #4106, opened 2022-02-05
Update wiki for new collective pitch setup in 4.2

bnsgeyer:
This PR updates the heli collective blade pitch setup in 4.2

### DIFF
--- a/copter/source/docs/traditional-helicopter-swashplate-setup.rst
+++ b/copter/source/docs/traditional-helicopter-swashplate-setup.rst
@@ -159,6 +159,8 @@ angle measures the desired maximum collective blade pitch.
 Set :ref:`H_SV_MAN <H_SV_MAN>` to 4 to have swashplate move to minimum position.  Adjust :ref:`H_COL_MIN <H_COL_MIN>` until blade pitch
 angle measures the desired minimum collective blade pitch.
 
+.. note:: Overall collective measurements must be made using the :ref:`H_SV_MAN <H_SV_MAN>` parameter since these would be impacted and inaccurate if the autopilot stabilization was active.
+
 Version 4.2
 -----------
 Minimum and Maximum Collective Blade Pitch Angle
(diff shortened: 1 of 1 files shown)


copter/source/docs/traditional-helicopter-swashplate-setup.rst
```diff
@@ -159,8 +159,27 @@ angle measures the desired maximum collective blade pitch.
...
+
+Minimum Landed Collective
++++++++++++++++++++++++++
+
+The :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` parameter is used as the lower collective pitch limit in degrees for modes that use altitude hold in the vertical axis.  This keeps the autopilot from driving collective to low resulting in ground resonance. The default angle for :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` parameter is set to -2 degrees collective blade pitch angle.  If you have non-symmetrical blades, then set it to a blade pitch that is lower by the amount used as the zero thrust pitch angle.  For example, if the :ref:`H_COL_ZERO_THRST <H_COL_ZERO_THRST>` is determined to be -3 deg then :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` should be set to -5 degrees.
```
Hwurzburg, line 177:
```suggestion
The :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` parameter is used as the lower collective pitch limit in degrees for modes that use altitude hold in the vertical axis.  This keeps the autopilot from driving collective too low resulting in ground resonance. The default angle for :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` parameter is set to -2 degrees collective blade pitch angle, and would allow symmetrical blade equipped vehicles to descend at a reasonable rate, yet avoid being too negative on the ground .  If you have non-symmetrical blades, then set it to a blade pitch that is lower by the amount used as the zero thrust pitch angle.  For example, if the :ref:`H_COL_ZERO_THRST <H_COL_ZERO_THRST>` is determined to be -3 deg then :ref:`H_COL_LAND_MIN <H_COL_LAND_MIN>` should be set to ~ -5 degrees.
```

Hwurzburg (changes requested):
perhaps we should move the note on measuring with H_SV_MAN rather than just sitting in a stab mode up into the 4.2 section to be sure its not missed...

also, I dont know if you have made the landing detection bullet proof yet such that getting a big updraft can force detect and that land_min plays a role it that detection? if it still does, perhaps a note on that?

copter/source/docs/traditional-helicopter-swashplate-setup.rst
bnsgeyer: > perhaps we should move the note on measuring with H_SV_MAN rather than just sitting in a stab mode up into the 4.2 section to be sure its not missed...

I will move it up above 4.2 for the setting of H_COL_MIN and H_COL_MAX

> also, I dont know if you have made the landing detection bullet proof yet such that getting a big updraft can force detect and that land_min plays a role it that detection? if it still does, perhaps a note on that?

I think being able to set the default at -2 deg without user interaction makes it much more robust.  I will draft a note

bnsgeyer:
@Hwurzburg incorporated suggested changes please review.  If you are happy with them then feel free to merge.

Hwurzburg:
great...thanks

PR #4110, opened 2022-02-07
Improved wiki for system identification

fbredeme:
The new content is part of a Master thesis at IAV GmbH supervised by Amilcar


Hwurzburg:
great...we needed this content...can you squash to one commit...thanks

amilcarlucas:
Hey Henry, this does not add any pages, it just extends an existing page. We can squash this once it gets reviewed.

I would like to discuss this in the beginning of the DevMeeting today, but I can not add the DevCall label to it. Can you do it?

Hwurzburg:
sorry, just looked a add flag....in too much a hurry...DevCall labels only get looked at for the ardupilot repo, not wiki...have to tell Craig to add topic

CraigElder:
> sorry, just looked a add flag....in too much a hurry...DevCall labels only get looked at for the ardupilot repo, not wiki...have to tell Craig to add topic

@Hwurzburg I look at all of the repos that have a DevCallTopic Label including the wiki
image

Hwurzburg:
cool, did not realize that....super

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+Examining the first 45 seconds of the test flight, it can clearly be observed that there is not much of a response on the pitch axis, as its rate stays almost constant around 0 rad/s.
+Note that the initial frequency of the sweep (0.05 Hz) is hold for 40 seconds at the beginning of the test flight in order to run two full cycles with a low frequency.
+With the increase of the frequency afterwards, the magnitude of the resulting pitch rate also rises.
+
+.. image:: ../images/PitchRateAxis8IdentificationIOSignals.png
+:name: fig-io-signals-axis8
+
+By estimating the :ref:`frequency-response<fig-freq-resp>` of the system and plotting the Bode diagram, the described behaviour above becomes clear.
+For lower frequencies, the system response reacts with low amplitudes (magnitude decreases).
+Normally, one would expect a (nearly) constant gain around 1 (0 dB) of the closed-loop transfer function for low frequency excitations and a decreasing magnitude for high frequences (which does not show in the diagram because the highest frequency of the frequency-sweep was 5 Hz).
```
bnsgeyer, line 168: @amilcarlucas @fbredeme  The bode plot makes perfect sense now that I read your approach to obtaining the data.  In stabilize mode the user is requesting a desired attitude and the flight controller will try to maintain that desired attitude.  When you inject pitch, roll or yaw rates, the flight controller sees theses as disturbances and quickly tries to attenuate them through the attitude controller feedback.  Thus your bode plot shows reduced gain at low frequencies until you get above the response frequency of the attitude controller which is based on the Angle P gain.

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+The Parameter :ref:`SID_AXIS<SID_AXIS>` of the system identification mode feeds in the frequency-sweep at different spots inside Ardupilot's controlers.
+The locations of these injection points are shown in the following :ref:`figure<fig-ctrl-sys-ardupilot>` as small chirp symbols.
+At each injection point, the sweep can be added to either the roll, pitch or yaw axis. 
+
+.. image:: ../images/ControlSystem.png
+:name: fig-ctrl-sys-ardupilot
+
+For the purpose of obtaining a dynamic model of only the copter itself (aka plant model), it is first necessary to calculate a transfer function(s) between the control inputs of the copter and its measurable dynamic response.
+In this case, the angular rate controller outputs (:math:`p_{Thrust}`, :math:`q_{Thrust}` and :math:`r_{Thrust}`) represent the control inputs of the copter whereas the gyroscope's measured angular rates represent the system response.
+Since a common application of the created system model consists in optimizing the control parameters, the filtered gyroscope signals are used instead of the raw gyroscope data, as these are the direct inputs to the angular rate controller.
```
bnsgeyer, line 112: I don't necessarily agree with this approach.  Using the filtered gyroscope signals can cause issues if this model was used in SITL or in any other simulation environment when the controller expects raw gyroscopic data.  You are using simulink to wrap a controller around this flight dynamic model, why not just add the filters to the controller and use the unfiltered gyroscopic data to develop the raw flight dynamic model.

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+Thereby, the separate modelling of the filters becomes redundant, which makes it possible to directly connect a model of the angular rate controller to the plant model without any further effort.
+
+.. image:: ../images/RateControl.png
+:name: fig-rate-ctrl
+
+The generation of flight data for the system identification is conducted during hovering flight.
+Therefore, the three axes (namely roll, pitch and yaw) are considered as decoupled.
```
bnsgeyer, line 119: Is the intent of telling the reader to conduct the testing in hovering flight to ensure an uncoupled response? You may want to add a sentence that says the flight data will be generated at the trim conditions of interest.  So that could be at some forward flight speed or in a hover. But you chose hover because it was easier to model due to an uncoupled response.  however I think multicopters are uncoupled even at some flight speed.  Helicopters  are completely different story ;-)

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+According to the :ref:`figure<fig-rate-ctrl>` shown above, which illustrates the general structure of the rate control loop independent of any specific control axis, it is possible to construct a transfer function for each axis:
+
+.. math::
+
+G_{Roll}(s) &= \frac{Y(s)_{Roll}}{U(s)_{Roll}} = p_{Meas}/p_{Thrust} \\
+G_{Pitch}(s) &= \frac{Y(s)_{Pitch}}{U(s)_{Pitch}} = q_{Meas}/q_{Thrust} \\
+G_{Yaw}(s) &= \frac{Y(s)_{Yaw}}{U(s)_{Yaw}} = r_{Meas}/r_{Thrust} \\
+
+As the figure shows, the transfer function of each axis models the system consisting of the motor mixer, the motors, the airframe of the copter and the gyroscopes together with the filters.
+Initially, the frequency-sweeps of :ref:`SID_AXIS<SID_AXIS>` 7-9 are used to inject the test signal at the input of the angular rate controller.
```
bnsgeyer, line 129: To get around the problem you experienced in your analysis, I would recommend using inputs 1-3 or I suppose you could use 3-6 as well.  These are making inputs into the attitude controller and so it will recognize the input and not try to attenuate it.  Then you can still perform the open loop analysis using the mixer inputs.

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
++--------------------------------------+------------------+
+|:ref:`SID_T_FADE_IN<SID_T_FADE_IN>`   | 5 s              |
++--------------------------------------+------------------+
+| :ref:`SID_T_REC<SID_T_REC>`          | 70 s             |
++--------------------------------------+------------------+
+
+The next :ref:`figure<fig-io-signals-axis8>` shows the recorded input (pitch rate target) and output (measured pitch rate) signals over time.
+Below that, the power spectral densities of both signals are depicted, indicating how the power of the signal is distributed over the tested frequency range.
+Examining the first 45 seconds of the test flight, it can clearly be observed that there is not much of a response on the pitch axis, as its rate stays almost constant around 0 rad/s.
+Note that the initial frequency of the sweep (0.05 Hz) is hold for 40 seconds at the beginning of the test flight in order to run two full cycles with a low frequency.
```
bnsgeyer, line 160:
```suggestion
Note that the initial frequency of the sweep (0.05 Hz) is held for 40 seconds at the beginning of the test flight in order to run two full cycles with a low frequency.
```

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+Normally, one would expect a (nearly) constant gain around 1 (0 dB) of the closed-loop transfer function for low frequency excitations and a decreasing magnitude for high frequences (which does not show in the diagram because the highest frequency of the frequency-sweep was 5 Hz).
+It is probable that there is a systemic cause for this (undesirable) response of the closed-loop system, which could be a nonlinearity in the system or a high-pass filter that is applied to the gyro measurements and therefore attentuates pitch rates of low frequencies?
+
+.. image:: ../images/BodeplotPitchClosedLoop.png
+:name: fig-freq-resp-axis8
+
+The next :ref:`figure<fig-io-signals-axis11>` shows the signals that were used for the system identification, consisting of the actual pitch rate as system output and the motor mixer input of the pitch axis (aka pitch rate pid output) as system input.
+It is again observable that the system is merely reacting to an excitation at low frequencies.
+Therefore, the supressing of amplitudes at low frequencies is probably not located in the PID controller, but in the system to be identified.
```
bnsgeyer, line 176: Remove this sentence.  It was how you conducted the analysis

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,101 @@ main loop frequency divided by a sub-sample factor.
...
+The differences between t=55s and t=80s are mainly caused by shifted phases, the amplitude is nearly the same.
+At lower frequencies, the model output actually follows the target pitch rate shown :ref:`above´fig-io-signals-axis8>`, while the system response measured during the test flight does not alter in large amplitudes and mainly consists of noise.
+It is important to note that the input and output signals had to be truncated in order to generate a model that leads to acceptable results.
+To be precise, only the data from 55s till the end was used for the identification.
+Therefore, the system response at frequencies that appeared ealier is the extrapolation of the frequency-response that was identified with the used data.
+This indicates that there is some component or effect in the system, that has to be considered in the identification process in order to generate a linear model that is capable to reproduce the complete behaviour of the real system.
```
bnsgeyer, line 195: So this can be rewritten once you gather the data and redo your analysis.   To show how the model compares with flight test data, typically the bode plots of the frequency sweeps are compared.  Then the validation is shown through making a pulse input in the model axis and then showing a comparison of the flight test data to the simulation response with the pilot input fed into the simulation.  The comparison of the time domain data with a different excitation is a better way to show correlation.

bnsgeyer (changes requested):
I would recommend conducting your flight test again using inputs 1-3 or 4-6 for the Axis parameter.  use the mixer inputs to extract your model.  Then use pulse inputs to generate validation data and compare the response of the flight test data to the simulator data.

copter/source/docs/systemid-mode-operation.rst
fbredeme: What we are looking at is the frequency response of the closed-loop of the rate controller though. The injected sweep is added to the desired rate and thus to the desired value of the controller. So, the injected sweep does not pose a disturbance for the rate controller but only for the attitude controller. From my understanding, we would see the attentuation if we looked at the frequency response of the attitude controller, but not for the rate controller, since the rate controller is not exposed to an injected disturbance and just tries to follow the desired rate

bnsgeyer: Even though you are making inputs into desired rate for the rate controller, the attitude controller is still maintaining the desired attitude based on stick input. The attitude controller will see the changes in the response due to the change in desired rate and will then make a counter input to oppose that rate in order to make the attitude remain at what was commanded by the pilot. 

Rate and attitude are kinematically linked so if you see an attenuation in rate then you would see the same in attitude

Use the SITL and perform a frequency sweep as I said above and then perform the spectral analysis on the rate controller. You will see the low frequency will not be attenuated. Or you can do it with rare inputs and just make the Angle P gain zero.

bnsgeyer:
@fbredeme @amilcarlucas 
I was conducting frequency sweeps in SITL today.  This was a sweep in pitch using SID_AXIS of 2.  You can see in the plot that both bode plot of desired pitch attitude to pitch attitude (solid line) and desired pitch rate to pitch rate (dashed line) do not lose gain at low frequency.  
image

Here is a plot of motor output to pitch rate which would be used for model development taken from the same flight.
image

So it is best to gather the data using the attitude input into the frequency sweep.  At least for aircraft that don't have high coupling.  For helicopters, data would have to be gathered in a rate mode so it is difficult to use the Sys ID mode to get model development data because the off axis response is being attenuated by the attitude controller.

fbredeme:
@bnsgeyer 
Thanks for all the input. I tried the frequency sweeps in SITL as well. Below, you can see the result when exciting the pitch angle axis (sid_axis = 2) instead of the pitch rate axis (sid_axis = 8). The rate controller's closed-loop gain is approximately 0 dB, which corresponds to the result you predicted. The uncertainties at higher frequencies can be explained by the constant window size that I am using to create the Bode plot (I'm using a chirp-z-transform to create it and as for now, I have not yet implemented a merged frequency-response computing for several window width).
BodePlotPitchAngleAxisSweep
I then tried the sweep for the pitch rate axis (sid_axis 8) with the same settings as in the real test flight that we conducted. The result is shown in the next figure. The gain is not decreasing with lower magnitudes, which is contrary to what we would assume.
BodePlotPitchRateSweepSameMag
I then continued and lowered the magnitude of the pitch rate sweep. As illustrated below, the magnitude now decreases with lower frequencies. Also, the coherence is too low to actually trust the computed result at low frequencies. I guess the reason for this is that the influence of the attitude controller on the pitch rate target grows since the sweep's magnitude has been lowered. What is your take on this?
BodePlotPitchRateSweepLowerMag

fbredeme:
@bnsgeyer 
I conducted a test flight with an injected frequency-sweep on axis 2 (pitch angle) just as you recommended. The sweep has an amplitude of 20 deg and starts with 0.05 Hz. You can see the desired pitch rate and the actual pitch rate over time below.

The respective bode plot below again shows a gain lower than 0 dB for low frequencies. In between, the coherence is quite low and I'm not quite sure why that is, but I think since the coherence is high for the starting frequencies, we can trust the data at that point. What do you think?

fbredeme:
@bnsgeyer yes, the drastic frequency increase is definately a problem and I'll fix that for the next test flight. I thought about starting at different frequencies, conduct two periods of the lowest frequency as I already did and then fuse those data sets together to obtain the composite frequency-response. I'll also try to leave the low frequency dwell out or to increase the record length to obtain a flatter frequency increase.

The initial idea for the high amplitude was to achieve a good signal-to-noise ratio in the system response, but I'll try lower amplitudes as you recommended. 

Are you always using the attitude controller input to inject your frequency sweep or are there situations where you use the other injection points within the control loop to excite the system?

I'd be happy to share the log files. How do you want me to provide them to you?

bnsgeyer:
You can place a link here to cloud drive.

Hwurzburg:
@bnsgeyer so does this get your approval now Bill?

bnsgeyer:
@Hwurzburg yes there are still open comments above.  I will leave it up to @fbredeme  and @amilcarlucas  as to how they wish to proceed with this wiki.

fbredeme:
@Hwurzburg @bnsgeyer I still have to go over the wiki again and incorporate the changes that Bill proposed. There are also some new results and findings from my ongoing work that I would like to add.

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,142 @@ main loop frequency divided by a sub-sample factor.
...
+The frequency sweep inputs 10-12 are used for the identification, since they provide a direct input to the regarded system. Because of this, the input signal is not 
+directly influenced by the controllers compared to the other chirp inputs, thereby making it independent from the parameterization of the controller.
+
+.. image:: ../images/ControlSystemDiagram.png
+:name: fig-ctrl-sys-ardupilot
+
+Model Structure
+------------------------------
+
+.. image:: ../images/bodyDiagramQuadcopter.PNG
```
bnsgeyer, line 114: @fbredeme The positive sign of the rotation about the bx and by axes is not as I would expect and is counter to the aircraft reference systems that I have seen in texts.  Using the right hand rule, positive rotation about the bx axis would be roll to the right and positive rotation about the by axis would be pitch up.  This is also opposite to the positive rotations for the IMU data.  Any reason that you chose the positive directions like this?

copter/source/docs/systemid-mode-operation.rst
```diff
@@ -95,3 +95,142 @@ main loop frequency divided by a sub-sample factor.
...
 +--------------+--------------+---------------+
 
+Identification of a Multicopter
+===============================
+Model Scope
+-------------------------------
+
+The system and its components, that are included in the mathmatical model to be identified is shown in the following block diagram. It covers the motor mixer, the 
+motors and ESCs as well as the airframe of the copter, the gyroscopes and their lowpass filters. 
+The frequency sweep inputs 10-12 are used for the identification, since they provide a direct input to the regarded system. Because of this, the input signal is not
```
bnsgeyer, line 105: I still don't agree with using these inputs as there will be reduced aircraft response at low frequencies and potentially low coherence.  Users should use inputs 1-3 but then use the motor inputs for the input signal of the frequency response analysis.

fbredeme: I actually agree with you on this point and I also conducted sysid flight using axes 1-3. Though, the coherence was always lower compared to the results when using inputs 10-12, that is why I continued using these inputs. Here, the coherence is pretty good if the integrator gains are set to 0. Also, I have not tried inputs 1-3 with the default parameters, but my guess is that the coherence will be even lower since the default are pretty low compared to what autotune or the parameter tuning with the identified models produced

copter/source/docs/systemid-mode-operation.rst
fbredeme: @bnsgeyer No, I just adopted it incorrectly from the literatur. Thanks for the correction.

copter/source/docs/systemid-mode-operation.rst
bnsgeyer: > Though, the coherence was always lower compared to the results when using inputs 10-12, that is why I continued using these inputs.

That is interesting.  Ok, make the suggestion based on your experience.  I have not done much identification of multicopters.  Helicopter identification is very different as there is much coupling between axes.  So you either have to use special identification process that accounts for the cross coupling between axes and decouples the frequency responses or use a mode that does not feed back attitude or rate in order to uncouple the inputs.  It is very difficult (and scary) to fly.

bnsgeyer:
@fbredeme Did you intentionally close this PR?  Aren't you going to finish it?

fbredeme:
@bnsgeyer There is still some work to do until I can finish the wiki. I'll reopen the PR once I'm done, approximately next week

fbredeme:
Reopened after internal reviews.

PR #4114, opened 2022-02-09
Add c rtk2

cuhome:


cuhome:
add cuav c-rtk2

common/source/docs/common-cuav-c-rtk2.rst
```diff
@@ -0,0 +1,201 @@
...
+	:target: ../images/c-rtk2/rtk1.jpg
+
+- 1 PPK base station/Ntrip account 
+- 1 C-RTK 2 
+- 1 Mapping camera 
+- 1 Drone 
+- 1 Data link 
+- 1 Computer 
+
+.. note::
```
Hwurzburg, line 175: I do not understand this note...in the PPK figure you are using this as a PPK base station, why is it not recommended? what exactly are you trying to say?

Hwurzburg:
@cuhome I spent several hours this morning, correcting build issues, improving formatting, and improving grammar....still have one questions, see above...otherwise, its close to being ready for merge....please review my changes carefully...thanks

common/source/docs/common-cuav-c-rtk2.rst
cuhome: C-RTK2 BASE is another product under development; C-RTK2 is used as a PPK base station; it first needs to know the absolute precise position of the current point (latitude and longitude and the height of the pole); and it is not conducive to users Measure control points.

cuhome: Here is the configuration tutorial for the base station: https://doc.cuav.net/gps/c-rtk2/en/base-station.html; can you help me add some needed information?

common/source/docs/common-cuav-c-rtk2.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Preparation before use
+======================
+Only use PPK
+------------
+
+.. image:: ../../../images/c-rtk2/ppk1.jpg
+	:target: ../images/c-rtk2/ppk1.jpg
+
+- 1 PPK base station/Ntrip account
```
Hwurzburg, line 157:
```suggestion
- 1 PPK base station data or Ntrip account for post-processing
```

common/source/docs/common-cuav-c-rtk2.rst
```diff
@@ -0,0 +1,201 @@
...
+- 1 Mapping camera 
+- 1 Drone 
+
+PPK+RTK
+-------
+
+.. image:: ../../../images/c-rtk2/rtk1.jpg
+	:target: ../images/c-rtk2/rtk1.jpg
+
+- 1 PPK base station/Ntrip account
```
Hwurzburg, line 168:
```suggestion
- 1 PPK base station and Ntrip account 
```

common/source/docs/common-cuav-c-rtk2.rst
Hwurzburg: I want to make the docs clear to a naive user like myself...I dont understand a lot about PPK and RTK setups....but here is what I think I know, correct me if I am wrong:

 For PPK, you need the raw GPS data from the vehicle (always C-RTK2 in this case), and either an historical data base of RTK data or the RTK data from a calibrated PPK base station. The "calibrated" can be exact GPS coordinates of the base entered into the base so it can generate RTK data, or an RTK base that has "surveyed in" itself over time to establish its exact position. Apparently the C-RTK2 does not store the PPK  data in a base station friendly format or does not have the ability to survey itself, or cannot accept entered GPS location, and that is why its not recommended for PPK Base use.

Also, the labels on the two figures are confusing...we need to use exact naming to avoid confusion, so I propose that a third diagram be added, the the current ones modified as below. I would avoid labels as AIR unless you actually call the C-RTK2 AIR, especially since it can be on any vehicle type:
PPK
ppk1
RTK
rtk2
PPK + RTK
rtk1

let me know if my understanding is incorrect

cuhome: I recreated the picture and sent it to your skype;

Hwurzburg:
updated figures

PR #4125, opened 2022-02-13
Stream rates are slower while sending parameters and waypoints

eblanshey:
This PR clarifies that stream rates are 4x slower or more if parameters and waypoints are currently being sent. If you want me to link the docs to the [source ](https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L954) where this happens, let me know.


Hwurzburg:
LGTM.....thanks

PR #4128, opened 2022-02-14
Update common-accelerometer-calibration.rst

timtuxworth:
I've seen videos online showing new 'unboxed' flight controllers being calibrated before installing in a vehicle. I guess this might work ok in a quad, but I'm thinking that it might lead to sub-optimal performance in a plane. I'm suggesting making it clearer that the ideal is to calibrate in the vehicle, especially for a plane.

### DIFF
--- a/common/source/docs/common-accelerometer-calibration.rst
+++ b/common/source/docs/common-accelerometer-calibration.rst
@@ -15,7 +15,7 @@ This article shows how to perform basic accelerometer calibration (using
 
 .. important::
 
-   Accelerometer calibration cannot be performed while vehicle is armed.
+   Accelerometer calibration cannot be performed while vehicle is armed. 
 
 Calibration steps
 =================
@@ -44,9 +44,11 @@ Calibration steps
       Accelerometer Calibration Positions (Copter)
 
    It is important that the vehicle is kept still immediately after pressing the key for each step. This is more important than getting the angle exactly right, ie. left being 90deg to horizontal, etc. Except for the first "LEVEL", the positions can be within 20 degs of being exact. Being still in each position as you press the key is much more important.
-
+   
+   -  You should calibrate the board mounted in the vehicle if possible. 
    -  You may need to calibrate the board before it is mounted if the size/shape of the vehicle makes this difficult.
    -  The level position is the most important to get right as this will be the attitude that your controller considers level while flying.
+   -  You can recalibrate LEVEL using Mission Planner, after you have installed the flight controller and are ready to fly.
 
 [site wiki="plane"]
    .. tip:: For planes, the "level" angle for pitch is important for steady altitude flight. See :ref:`tuning-cruise` for more details.


Hwurzburg:
not accurate...accuracy of an acc cal depends primarily on how still each step in the cal is....+/- 10degs off during the cal steps (except level) is much less important than the stillness in each position to get accurate gyro biases...while it is also important that the resulting vehicle mounting be within 10 degs in each axis(to an orientation option), trying to cal in even a 1m vehicle unless propped absolutely still, will be less accurate than on a bench (potentially)...

I meant to add a note on this page a long time ago, but forgot...will do that now

common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -15,7 +15,7 @@ This article shows how to perform basic accelerometer calibration (using
 
 .. important::
 
-   Accelerometer calibration cannot be performed while vehicle is armed.
+   Accelerometer calibration should be performed with the the fligth controller installed in the vehicle. Accelerometer calibration cannot be performed while vehicle is armed.
```
Hwurzburg, line 18: this is too strong....please revert

common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -44,7 +44,8 @@ Calibration steps
       Accelerometer Calibration Positions (Copter)
 
    It is important that the vehicle is kept still immediately after pressing the key for each step. This is more important than getting the angle exactly right, ie. left being 90deg to horizontal, etc. Except for the first "LEVEL", the positions can be within 20 degs of being exact. Being still in each position as you press the key is much more important.
-
+   
+   -  You should calibrate the board mounted in the vehicle if possible.
```
Hwurzburg, line 48: this is acceptable

Hwurzburg (changes requested):

common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -15,7 +15,7 @@ This article shows how to perform basic accelerometer calibration (using
 
 .. important::
 
-   Accelerometer calibration cannot be performed while vehicle is armed.
+   Accelerometer calibration cannot be performed while vehicle is armed.
```
TunaLobster, line 18: Not sure why this whitespace change is needed.

common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -44,9 +44,11 @@ Calibration steps
       Accelerometer Calibration Positions (Copter)
 
    It is important that the vehicle is kept still immediately after pressing the key for each step. This is more important than getting the angle exactly right, ie. left being 90deg to horizontal, etc. Except for the first "LEVEL", the positions can be within 20 degs of being exact. Being still in each position as you press the key is much more important.
-
+   
+   -  You should calibrate the board mounted in the vehicle if possible. 
    -  You may need to calibrate the board before it is mounted if the size/shape of the vehicle makes this difficult.
    -  The level position is the most important to get right as this will be the attitude that your controller considers level while flying.
+   -  You can recalibrate LEVEL using Mission Planner, after you have installed the flight controller and are ready to fly.
```
TunaLobster, line 51: Drop Mission Planner and use "AHRS trim" instead.

common/source/docs/common-accelerometer-calibration.rst
timtuxworth: It was an accident. I wanted to revert my previous change as requested by @Hwurzburg. This is just me putting it back to what it was.

common/source/docs/common-accelerometer-calibration.rst
timtuxworth: Is this documentation intended for someone new to this? If I am new to this, this documentation is already pretty hard to follow. Dropping back to "AHRS Trim" might not help making this more accessible to folks trying to get started.

Hwurzburg: AHRS_TRIM is not a param that we recommend for a user to access....we used to recommend using it to adjust AOA after a bench or fuselage line accel cal, but it interacts with VTOL setup and TRIM_PITCH_CD is the only param we suggest a user touch....unfortunately, copter and heli dont have this param and must use it
I really need to scrub it from the two plane docs that still have it...

Hwurzburg: I think I will leave the refs in tuning-cruise since it shows using AHRS trims as being NOT the preferred method already...and just change the starting up and calibrating doc...

Hwurzburg: #4131

PR #4138, opened 2022-02-16
Replace string

RuffaloLavoisier:


rmackay9:
I think these days we prefer to just call it "Rover" instead of "ArduRover"...

RuffaloLavoisier:
@rmackay9 
Should I change this part to "Rover"?

RuffaloLavoisier:
@rmackay9 
What about the other parts? copter, boat, sub..Etc. Is it different for each part?

rmackay9:
@RuffaloLavoisier,  If you could change it to "Rover" that'd be great.  I think the same goes for all our vehicles these days.  So "ArduCopter" is now referred to as just "Copter", "ArduPlane" is just "Plane", etc.  We never used the word "ArduBoat" I think.. but if we did we shouldn't have and instead we should have just called it "Rover".

RuffaloLavoisier:
@rmackay9 
Then we will change all the words on the wiki? Replace Copter, Plane..etc

Hwurzburg:
I disagree....that would be a massive change requiring 1000's of individual checks for appropriateness since Ardu...is still used...
I have a PR to correct Arducopter to ArduCopter, etc. in the wiki...but going to Plane, Copter, and Rover is something we need to discuss in detail....

RuffaloLavoisier:
@Hwurzburg 
> ..requiring 1000's of individual checks..

You're right.
If discussion is needed, then we will switch this PR to draft.

Hwurzburg:
need to remove the WIP status and I will merge...thanks

RuffaloLavoisier:
@Hwurzburg 
I see, thank you

PR #4147, opened 2022-02-21
Dev: update gsoc ideas list for 2022

rmackay9:
Initial update of GSoC project ideas.  We need to make a bunch more updates to the details but at least this fixes the title and a number of the top ideas.


Hwurzburg:
sure

PR #4151, opened 2022-02-23
BMP280 barometer wiring fix

sq3ope:
Wiring diagram of Adafruit BMP280 external barometer is wrong. The yellow wire should go to SDI, not SDO pin.
Why do I think so?
1. I checked in original Adafruit manual: https://learn.adafruit.com/adafruit-bmp280-barometric-pressure-plus-temperature-sensor-breakout/pinouts
2. Confirmed by other on forums: https://discuss.ardupilot.org/t/how-to-apply-the-bmp280-sensor-to-pixhawk2/40300
3. I've checked it myself. The sensor works on my pixhawk setup.


Hwurzburg:
LGTM...thanks

PR #4164, opened 2022-02-26
Update and add Fport Settings

Niels-bt:
Setup guide

### DIFF
--- a/common/source/docs/common-matekf765-wse.rst
+++ b/common/source/docs/common-matekf765-wse.rst
@@ -106,6 +106,21 @@ With this option, :ref:`SERIAL7_PROTOCOL<SERIAL7_PROTOCOL>` must be set to "23",
 
 Any UART can be used for RC system connections in ArduPilot also, and is compatible with all protocols except PPM. See :ref:`common-rc-systems` for details.
 
+If using the Fport with a Frsky receiver you should consider having it on te version 2.1.0 FCC, the newest 2.1.1 seems to have telemtry issues
+
+Parameters:
+
+Use RX6 pin to connect your FPort pin. The configuration should be:
+
+BRD_ALT_CONFIG =1
+SERIAL7_PROTOCOL =23
+SERIAL7_OPTIONS =15
+RC_OPTIONS =8 (depending on receiver and its firmware rev, this may not be necessary)
+RSSI_TYPE =3
+
+Useful link:
+https://ardupilot.org/copter/docs/common-FPort-receivers.html
+
 Dshot capability
 ================
 


common/source/docs/common-matekf765-wse.rst
```diff
@@ -106,6 +106,21 @@ With this option, :ref:`SERIAL7_PROTOCOL<SERIAL7_PROTOCOL>` must be set to "23",
 
 Any UART can be used for RC system connections in ArduPilot also, and is compatible with all protocols except PPM. See :ref:`common-rc-systems` for details.
 
+If using the Fport with a Frsky receiver you should consider having it on te version 2.1.0 FCC, the newest 2.1.1 seems to have telemtry issues
```
Hwurzburg, line 109: If this is a bug in their firmware, we dont want to make it a permanent comment in our wiki....

common/source/docs/common-matekf765-wse.rst
```diff
@@ -106,6 +106,21 @@ With this option, :ref:`SERIAL7_PROTOCOL<SERIAL7_PROTOCOL>` must be set to "23",
...
+
+Parameters:
+
+Use RX6 pin to connect your FPort pin. The configuration should be:
```
Hwurzburg, line 113: this is not required...its an alternate setup....the setup using the TX pin is the normal connection and not doing a SWAP since F4 autopilots cannot even do that...

common/source/docs/common-matekf765-wse.rst
```diff
@@ -106,6 +106,21 @@ With this option, :ref:`SERIAL7_PROTOCOL<SERIAL7_PROTOCOL>` must be set to "23",
...
+
+BRD_ALT_CONFIG =1
+SERIAL7_PROTOCOL =23
+SERIAL7_OPTIONS =15
+RC_OPTIONS =8 (depending on receiver and its firmware rev, this may not be necessary)
```
Hwurzburg, line 118: this is highly unusual to need this pad...and its really only for autopilots that have special level shifters which delay the signals....have you found issues on this one? we would need to duplicate before adding to the wiki

Hwurzburg (commented):
overall, I cannot approve this....it might be worthwhile to add links to the common-FPort-receivers.rst page, but in all autopilot docs, not just  one...but on the other hand, every other RC protocol has a detailed document also, so this just gets into massive duplication....which we try to avoid for maintenance purposes...

common/source/docs/common-matekf765-wse.rst
Niels-bt: Okay I'm sorry, I'm new to GitHub so...
The thing is I wanted to make something to help others to get faster on the first setup 

I am using this bord and it took me a lot of time to find this and make it work. Same for the rest the overall documentation for the bord on matek is poor and this bord needs a special setup to make fport work. 
Since it doesnt work for the 2.1.1 version and updates are not quick at frsky idk
On the fport documentation it says for the F 765-Wing. Should I ad there that it's the same for the matek F765-WSE ?

Thanks for your time and hope this can help others Niels.

common/source/docs/common-matekf765-wse.rst
Niels-bt: yes but it's a f7 flight controller

Niels-bt:
> overall, I cannot approve this....it might be worthwhile to add links to the common-FPort-receivers.rst page, but in all autopilot docs, not just one...but on the other hand, every other RC protocol has a detailed document also, so this just gets into massive duplication....which we try to avoid for maintenance purposes...

common/source/docs/common-matekf765-wse.rst
Niels-bt: PS: the thing is there are so many topics on discuss ardupilot and something like this could help

common/source/docs/common-matekf765-wse.rst
Niels-bt: i just tried the other pin without the swap and it's not working..

common/source/docs/common-matekf765-wse.rst
Hwurzburg: Niels, aside from the FRSky firmware bug in the lastest ACCESS firmware and having to use the earlier version firmware....did you find that using the existing instructions ,ie use TX pin and the parameters mentioned in this page, did NOT work? if so, we would need to confirm and modify....but I personally have setup this controller with FPort on an FRSky R-XSR ACCst using the instructions in the article several times....

Niels-bt: Yes the tx doesn't work I first tested the settings described below. I think it's the same as the matek F765-Wing this is why on this site https://ardupilot.org/copter/docs/common-FPort-receivers.html. It says to use the RX pin

Niels-bt: No I already tested it and it's why I wanted to change the wiki....

common/source/docs/common-matekf765-wse.rst
Hwurzburg: I had Matek try it and it works for them using the current instruction for the RX6 pin....I have asked the dev team for another confirmation one way or the other....it works for me on a 765-wing also

PR #4165, opened 2022-02-27
Lua: Add a debug level

muramura:
The GCS message level has a debug level.


Hwurzburg:
LGTM..thanks

muramura:
@Hwurzburg san
Thank you for this.

PR #4103, opened 2022-02-04
Add page on scripting parameters

Hwurzburg:
for 4.2

PR #4166, opened 2022-02-27
LUA: Add a script configuration parameter

muramura:
Can create configuration parameters for scripts.
I learned about this method from a sample script.
I think I should post it in my WIKI.


Hwurzburg:
see #4103.....dont think this is needed since there will be a dedicated page once 4.2 wiki updates are merged...

muramura:
I understood. I'll close.

PR #4167, opened 2022-02-27
Dev: typo fix in precision-landing-with-irlock.rst

shiv-tyagi:


copter/source/docs/precision-landing-with-irlock.rst
```diff
@@ -103,7 +103,7 @@ demo is shown below (using an older firmware).
 If the vehicle does behave appropriately, download the dataflash logs
 and examine the PL messages.
 
--  If the "Heal" (meaining health) field is not "1" then there may be a communication issue between the Pixhawk and IR-LOCK sensor.
+-  If the "Heal" (meaning health) field is not "1" then there may be a communication issue between the Pixhawk and IR-LOCK sensor.
```
Hwurzburg, line 106: lets try when we make updates to correct things like this also, if in the same sentence:
```suggestion
-  If the "Heal" (meaning health) field is not "1" then there may be a communication issue between the autopilot and IR-LOCK sensor.
```

Hwurzburg (changes requested):

copter/source/docs/precision-landing-with-irlock.rst
shiv-tyagi: Sure! I've committed this change.

Hwurzburg (approved):

Hwurzburg:
had to resolve a merge conflict that I caused yesterday with some work I did...rebased and merging...thanks

PR #4186, opened 2022-03-02
Common: Add note about pulldown for FPort on Cube Orange

stephendade:
As per https://discuss.ardupilot.org/t/solved-cube-orange-and-x8r-frsky-s-port-telemetry-not-working/73556/2 and https://discuss.cubepilot.org/t/frsky-pass-through-protocol-on-serial-5/120/11

Tested and confirmed that this is required on Cube Orange.

### DIFF
--- a/common/source/docs/common-connecting-sport-fport.rst
+++ b/common/source/docs/common-connecting-sport-fport.rst
@@ -56,6 +56,8 @@ ______________________________________________________________________
 
 .. note:: F7/H7 boards can have the FPort connected to the UARTs RX pin, instead of the TX pin as shown above, and use the UART's SWAP option. F4 boards do not have this SWAP capability See :ref:`common-Fport-receivers` for more information.
 
+.. note::  some autopilots, like the Cube Orange flight controller, have level shifters on their UART pins used on their Telemetry ports. This requires a 10K pulldown resistor externally on the receiver's SPort signal to work properly. 
+
 .. _frsky_cables:
 
 Bi-Directional Inverter cables


stephendade:
I think the pad option is for F.Port (combined RC and telemetry). I was using S.Port (separate RC and telemetry pins)

Hwurzburg:
yep...brain glitch...but its probably the same issue of level shifters getting in the way....perhaps we should generalize the statement using Cube Orange as the example like the FPort pad bit statement...

common/source/docs/common-connecting-sport-fport.rst
```diff
@@ -56,6 +56,8 @@ ______________________________________________________________________
 
 .. note:: F7/H7 boards can have the FPort connected to the UARTs RX pin, instead of the TX pin as shown above, and use the UART's SWAP option. F4 boards do not have this SWAP capability See :ref:`common-Fport-receivers` for more information.
 
+.. note:: If using the Telemetry port on a Cube Orange flight controller, a 10K pulldown resistor is required on the SPort signal.
```
Hwurzburg, line 59:
```suggestion
.. note::  some autopilots, like the  Cube Orange flight controller, have level shifters on their UART pins used on their Telemetry ports that require a 10K pulldown resistor externally on the receiver's the SPort signal to work properly. 
```

Hwurzburg (changes requested):

stephendade:
Fixed up as per your suggestion

Hwurzburg:
Thanks

PR #4191, opened 2022-03-02
common: add note on servo/relay functions and -1

peterbarker:
We need to advertise this change much more widely or we're going to have a lot of user support questions.

### DIFF
--- a/common/source/docs/common-relay.rst
+++ b/common/source/docs/common-relay.rst
@@ -1,5 +1,9 @@
 .. _common-relay:
 
+.. note:
+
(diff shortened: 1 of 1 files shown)


nkruzan:
Does this need to be below the ===== lines? It doesn't show in preview, but i know github preview of rst is a little borked.

Hwurzburg:
Should be down around the brd param wording showing 4.1 way and new way 4.2 and later

Hwurzburg:
Although I added this to the GPIOs page already for the 4.2 update...I still have to touch each page that refs BRD_PWM_COUNT and update...also SERVO_GPIO_MASK comes into this and needs an explanation in the wiki....
now that I have 4.2 out (more or less) I will work on this...

PR #4194, opened 2022-03-03
comments for connecting to bt device in linux

nkruzan:
I am not sure correct way to add pics to wiki....If someone advise and want pics I will be happy to add them.

### DIFF
--- a/common/source/docs/common-mission-planner-bluetooth-connectivity.rst
+++ b/common/source/docs/common-mission-planner-bluetooth-connectivity.rst
@@ -148,6 +148,23 @@ Your Bluetooth device should now be ready and you can connect to the
 proper port at the proper baud rate of 57600 (red LED steady on and not
 flashing)
 
+Connecting With Linux
+=====================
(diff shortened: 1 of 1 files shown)


Hwurzburg:
I have still not succeeded using these instructions...I have tried with the HC-06 setup at both 57600 and 115200 baud...it works fine on QGC Android, QGC Windows, MP Windows at either baudrate....rfcomm0 is present....but I always get this:
Device /dev/rfcomm0 reopened OK
link 1 down
Device /dev/rfcomm0 is dead
Device /dev/rfcomm0 reopened OK
Device /dev/rfcomm0 is dead
Device /dev/rfcomm0 reopened OK
Device /dev/rfcomm0 is dead

BT module shows solid led indicating its connected....
Same with QGC in Linux....set up a Serial port as rfcomm0 at correct baudrate, connect....solid LED on dongle...no heatbeats or conenction on QGC...

am afraid others will have issues unless we figure this out...

PR #4195, opened 2022-03-03
Remove sphinxcontrib.vimeo plugin

TunaLobster:
This is in preparation for building LaTeX PDFs of the wiki. This will break all current build environments including the build server. Any environment that is currently out there will receive this error message when running `update.py`.

```
Traceback (most recent call last):
  File "./update.py", line 57, in <module>
    raise Exception("Module sphinxcontrib.vimeo was found. Please run python3 -m pip uninstall -y sphinxcontrib.vimeo")
Exception: Module sphinxcontrib.vimeo was found. Please run python3 -m pip uninstall -y sphinxcontrib.vimeo
```

A check for the module has also been added to the `Sphinxsetup.sh` script. In the Windows batch script it is in the list of removed modules at the top.

The functionality and tags that this plugin provides are now included in the youtube plugin. This module does not support LaTeX output, and the youtube plugin does.

Example vimeo video in the documentation now:
image


TunaLobster:
@peterbarker Can you give this one a review?

Talked with Henry about this one. Approach is good from his perspective.

I did think of one corner case that might cause some issue. If using python2 to build and the environment doesn't have the updated sphinx-contrib.youtube plugin that fixes support for python2, then the build will fail. I have not tested if pip will grab the correct module and update it on python2. I know they were looking at removing support, but don't know where that stands now. The get-pip.py script won't work on python2 anyways though...hmm. Might be nothing to worry about.

peterbarker:
> I did think of one corner case that might cause some issue. If using python2 to build and the environment doesn't have the updated sphinx-contrib.youtube plugin that fixes support for python2, then the build will fail. I have not tested if pip will grab the 

What version of `sphinx-contrib.youtube` is required - and can I just check `__version__`?

TunaLobster:
`python3 -m pip show sphinxcontrib-youtube` Should show a version greater than 1.0.1 https://github.com/sphinx-contrib/youtube/releases/tag/v1.0.1

peterbarker:
Chatted about this; I think we can merge this after the error message tells you to make sure your sphinxcontrib-youtube plugin is up-to-date.

We will need to coordinate removing the package from the build server with the merge of this PR.

peterbarker:
... further thought, we'll patch common_conf.py to include vimeo only if the youtube plugin is old.

TunaLobster:
This should work with any system that has setuptools installed as well as python >=2.7. Tested with python 3.8 with various configurations of package versions installed.

Fixed up all of the conf.py files to pass the flake8 check and now CI will check all of those files too. Also thought about adding the oldest supported python version to the matrix in CI.

TunaLobster:
The set up script should not remove the vimeo extension. Just ignore that it exists unless the youtube extension is old.

PR #4204, opened 2022-03-07
Dev: fixup SITL preclanding setup instructions

rmackay9:
I found that the SITL instructions were missing some critical parameter settings so here they are.  I've tested these changes on my local machine.

FYI @peterbarker


khancyr (approved):
Lgtm

PR #4218, opened 2022-03-11
Sphinxsetup.sh: cope with bionic pip specialisation

peterbarker:


Hwurzburg:
does someone need to review this? I cant

nkruzan (approved):
LGTM.......didn't test on bionic, but I verified url is good and curl command work correctly...assume will work as expected

PR #4226, opened 2022-03-14
Add COMPASS_AUTO_ROT explanation

Hwurzburg:


AndKe:
@Hwurzburg  thank you  - what is needed for success ful orintation?
- does it work by matching to an internal compass?
- does it require GNSS fix while calibrating?

I am asking because I tried to auto-set orientation on an external compass - the only one that was enabled - and compass calibration never finished (progerss stopped at 90-ish percent.)
Therefore, it would be nice to have a few words on what conditions must be met / how it works.

Hwurzburg:
>     * does it work by matching to an internal compass?
> 
no....by comparing mag field changes as attitude changes...need a calibrated acc and no mag field corruptions from ferrous objects to get a good offset and orientation
>     * does it require GNSS fix while calibrating?
> 
no, but its recommended for better accuracy (allows ellipticals and diagonals to be calculated better by knowing what region of the earth its located)....a high quality fix  is needed for inflight GSF based learning however...

AndKe:
Thank you. So - in other words, there is no special requirement for the orientation to work, one can just do the compass calibration with all the random orientations as usual, and no need to start/end in a specific heading or anything like that.

Hwurzburg:
correct

AndKe:
@Hwurzburg  Thank you. - and thanks for maintaining documentation - good documentation is a very important part of ArduPilot's success.

PR #4227, opened 2022-03-14
Add flake8 format checking

TunaLobster:
These should be non-functional changes in the code. Added specific noqa where needed are removed unneeded noqa.

Added flake8 to CI with the same .flake8 config as the main repo. Tested that this does return a CI failure if flake8 finds an issue. Right now it is clean.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,10 @@ jobs:
       - name: Check image formats
         run: |
           find ./images/ -exec file {} + | grep RIFF -vzq
+      - name: Check python flake8 formating
+        run : |
+          set -eux -o pipefail
+          flake8 .
       - name: Test build
         run: |
           ./update.py --parallel 4 --verbose --destdir=/tmp/web
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,20 @@
+[flake8]
+ignore =
+    # H301: one import per line
+    H301,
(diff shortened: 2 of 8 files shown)


peterbarker (approved):
LGTM

A follow-up commit to remove some of the noqa's would be nice.

A follow-up commit to add a CI task to enforce this would be nicer :-)

Thanks!

TunaLobster:
CI addition is included with this PR. Here's an expected failure run when I was testing CI. https://github.com/TunaLobster/ardupilot_wiki/runs/5545322486?check_suite_focus=true

I at least wanted to mark the noqa lines for what exactly is wrong with them so future changes aren't swept under the rug. Some of the ones that are currently there seem like logical overrides.

hendjoshsr71:
Love this! Will make my VSCode much happier..... 

This wont affect the ardupilotwaf directory will it? It gives mad problems with flake8 I just ignore because that is a job to fix up

.github/workflows/test.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       - name: Check image formats
         run: |
           find ./images/ -exec file {} + | grep RIFF -vzq
+      - name: Check python flake8 formating
```
peterbarker, line 32:
```suggestion
      - name: Check python flake8 formatting
```

peterbarker (approved):

peterbarker:
Merged, thanks!

PR #4237, opened 2022-03-20
Add warnings to read the "must reads" in vehicle indexes and minor tweaks to highlight tuning

Hwurzburg:


Hwurzburg:
not having gotten any negative responses, merging

PR #4239, opened 2022-03-21
Add red text to Copter, Rover, and Plane toc 1st time items

TunaLobster:
Looks a little something like this. Looks very similar on Rover and Copter too.
image

### DIFF
--- a/common/source/_static/common_theme_override.css
+++ b/common/source/_static/common_theme_override.css
@@ -45,3 +45,16 @@ table.partners-table td {
     padding-top: 40px;
     margin-top: -40px;
 }
+
+/* make certain toc sidebar items red */
+.wy-menu .toctree-l1 a[href*="arduplane-setup.html"],
+.wy-menu .toctree-l1 a[href*="first-flight-landing-page.html"],
+.wy-menu .toctree-l1 a[href*="initial-setup.html"],
+.wy-menu .toctree-l1 a[href*="flying-arducopter.html"],
+.wy-menu .toctree-l1 a[href*="apmrover-setup.html"],
+.wy-menu .toctree-l1 a[href*="rover-first-drive.html"],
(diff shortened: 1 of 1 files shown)


common/source/_static/common_theme_override.css
```diff
@@ -45,3 +45,14 @@ table.partners-table td {
...
+
+/* make certain toc sidebar items red */
+.wy-menu .toctree-l1 a[href="docs/arduplane-setup.html"],
+.wy-menu .toctree-l1 a[href="docs/first-flight-landing-page.html"],
+.wy-menu .toctree-l1 a[href="docs/initial-setup.html"],
+.wy-menu .toctree-l1 a[href="docs/flying-arducopter.html"],
+.wy-menu .toctree-l1 a[href="docs/apmrover-setup.html"],
+.wy-menu .toctree-l1 a[href="docs/rover-first-drive.html"]
+{
+    color: red;
```
Hwurzburg, line 57:
```suggestion
    color: red; font-weight: bold;
```
too bad it isnt persistent....also, what about the Quadplane TOC entries? can they be made to do the same thing?

Hwurzburg (changes requested):
thanks

common/source/_static/common_theme_override.css
TunaLobster: Yep! Just need to add the correct CSS class and attribute things. What do you mean by persistent?

Hwurzburg: if you click on another menu item, and then return (ie close that items expansion) the style reverts....its only red until you click something....not a biggie

TunaLobster: Fixed both

Hwurzburg:
playing with it some more, it seems to eventually refresh...thanks merging

PR #4245, opened 2022-03-23
AUTOTUNE available in stabilized modes from 4.1.3 not 4.2

CallanDaniel:
AUTOTUNE can be done in any stabilized flight mode from 4.1.3, not 4.2. I had a discussion with @tridge to confirm, and these changes were already merged to 4.1.3 in November last year.


Hwurzburg:
thanks

PR #4254, opened 2022-03-25
Fettec: Promote up one level in tree, shrink title image, fix two links

rmackay9:
This makes 3 small changes to the FETtecOne wiki page:

1. moves it up out of the "Serial ESC" page (which is then deleted).  I think we should avoid landing pages with just one item on them especially considering that the regular ESC-and-motors page still has lots of room on it.
2. fixes a couple of links on the page
3. removes white space from the title image

This has been tested on my local machine


amilcarlucas:
I have very high hopes that there will be other serial based ESCs in the future, so I do not agree with the first change.
But the other two are good to go.

amilcarlucas (commented):
I have very high hopes that there will be other serial based ESCs in the future, so I do not agree with the first change.
But the other two are good to go.

Thanks Randy

rmackay9:
@amilcarlucas,

We can always add the layer back if required in the future.  .. none of this is written in stone.. and even if we did have more serial ESCs why would we want to hide them one more layer down?

common/source/docs/common-fettec-onewire.rst
```diff
@@ -8,7 +8,7 @@ FETtec OneWire ESCs
 
 .. image:: ../../../images/FETtec_4in1_ESC.jpg
```
Hwurzburg, line 9:
```suggestion
.. image:: ../../../images/FETtec_4in1_ESC.jpg
    :target: ../../_images/FETtec_4in1_ESC.jpg
    :width: 450px
```

Hwurzburg (changes requested):
dont delete image and replace with lower resolution

rmackay9:
@Hwurzburg,

I don't think I made it lower resolution.. I just chopped off the white space around the important part of the image.

common/source/docs/common-fettec-onewire.rst
Hwurzburg: you're right...sorry

PR #4268, opened 2022-03-31
Copter: windspeed estimation clarification

rmackay9:
This PR improves the wind speed estimation documentation mostly by writing down a simplified version of what is discussed in the video.

I have actually run through this in SITL and on my vehicle to confirm that the steps basically work although I expect I will add some more detail to the "Calculate the Propeller Drag Coefficient" which is the most difficult part of the setup I think.

I have tested this locally on my machine.

@priseborough are you happy with this documentation?


copter/source/docs/airspeed-estimation.rst
```diff
@@ -1,9 +1,102 @@
...
+
+Calculate the position error coefficient for each direction
+
+- Calculate the air density (in kg/m^3) using an online calculator (`like this one <https://www.density.co.uk/calculators/density-of-air>`__)
+- Calculate the static pressure error (in Pa) = -height * gravity * air density
+- Calculate the dynamic pressure = 0.5 * air density * airspeed^2
+- Position error coefficient = static pressure error / dynamic pressure.  This value should be between -1 and +1
+
+Enable the compensation and copy the above calculated position error coefficients into the appropriate parameters
+
```
Hwurzburg, line 97: its a matter of style, but I have been trying to not link stuff like BAROx_ ...but rather say:
"for example for the first barometer: 
BARO1_...etc. "
so that the param label matches where you wind up...

Hwurzburg (changes requested):

copter/source/docs/airspeed-estimation.rst
rmackay9: Txs for the review, changed.

Hwurzburg:
thanks..LGTM...merging

rmackay9:
Thanks very much!

PR #4270, opened 2022-04-01
Blimp: Add fin construction photos

MichelleRos:
Adds a couple of photos to better explain how the fins are attached to the servos.


Hwurzburg:
thanks....merging

PR #4271, opened 2022-04-01
Add airspeed docs for rover and copter

Hwurzburg:


bnsgeyer:
@Hwurzburg  did you want to add how to set unit up in SITL?  Include that you have to set the sensor to analog and the pin to 1.

common/source/docs/common-airspeed-sensor.rst
```diff
@@ -0,0 +1,31 @@
+.. _common-airspeed-sensor:
+
+[copywiki destination="copter,rover"]
+
+Airspeed Sensor
+===============
+
+Copter, Rover, and Blimp can attach and enable an Airspeed Sensor. However, unlike Plane which uses the readings for navigation and control, these vehicles only log and report the readings for use in displays in Ground Control Stations, OSDs, and telemetry. The Plane airspeed sensor :ref:`documentation<plane:airspeed>` can followed for selecting and mounting the type of airspeed sensor, but be aware that it will not provide the functionality beyond measuring and reporting discussed in Plane.
```
hendjoshsr71, line 8:
```suggestion
Copter, Rover, Sub, and Blimp can attach and enable an Airspeed Sensor. However, unlike Plane which uses the readings for navigation and control, these vehicles only log and report the readings for use in displays in Ground Control Stations, OSDs, and telemetry. The Plane airspeed sensor :ref:`documentation<plane:airspeed>` can followed for selecting and mounting the type of airspeed sensor, but be aware that it will not provide the functionality beyond measuring and reporting discussed in Plane.
```

common/source/docs/common-airspeed-sensor.rst
```diff
@@ -0,0 +1,31 @@
...
+
+The one exception to the above is that Sailboats can use an airspeed sensor as part of a :ref:`Wind Vane<rover:wind-vane>` (:ref:`WNDVN_TYPE<WNDVN_TYPE>` = 4) and use its readings to calculate wind speed as a part of navigation.
+
+The following airspeed parameters have no effect and should always be set to "0" value:
+
+- :ref:`ARSPD_USE<ARSPD_USE>`
+- :ref:`ARSPD2_USE<ARSPD2_USE>`
+- :ref:`ARSPD_OPTIONS<ARSPD_OPTIONS>`
```
hendjoshsr71, line 16: I think for the wiki I would add that you can still use Disable_Voltage_Correction.

Like I stated in the other PR though the others here could be made functional by setting USE and the option bit I don't think we should advertise them for now.

hendjoshsr71 (changes requested):
This seems fine

hendjoshsr71:
The other way to test in SITL is ARSPD_TYPE 7 and ARSPD_BUS 2

Hwurzburg:
> @Hwurzburg did you want to add how to set unit up in SITL? Include that you have to set the sensor to analog and the pin to 1.

going to PR a defaults change to fix that for SITL...

common/source/docs/common-airspeed-sensor.rst
Hwurzburg: so I dont intend to mention it....no real value at this point....if airspeed is actually used for something, we can change it then for that vehicle

common/source/docs/common-airspeed-sensor.rst
Hwurzburg: Actually Sub is an independently maintained Wiki...you wont find SUB stuff in the other vehicle's wiki's

Hwurzburg:
> The other way to test in SITL is ARSPD_TYPE 7 and ARSPD_BUS 2

but  will still keep NONE as the default, will set the pin to 1 for analog so it works if chosen, and bus to 2 for dlvr, if someone choses it....

common/source/docs/common-airspeed-sensor.rst
hendjoshsr71, line 8:
```suggestion
Copter, Rover, and Blimp can attach and enable an Airspeed Sensor. However, unlike Plane which uses the readings for navigation and control, these vehicles only log and report the readings for use by the displays in Ground Control Stations, OSDs, and telemetry. The Plane airspeed sensor page :ref:`documentation<plane:airspeed>` can be followed for selecting and mounting an airspeed sensor. However, it will not provide the functionality discussed in Plane except for measurement and reporting.
```

Hwurzburg: modified similar to suggestion

PR #4281, opened 2022-04-05
Update common-airvolute-DroneCore-Suite.rst

lukaspalkovic:
Where To Buy section moved to the end of document where it is more appropriate.


Hwurzburg:
actually Randy prefers it at the beginning...I prefer it at the end....but I bow to Randy's wishes since its not a big deal to me...

PR #4286, opened 2022-04-07
common: fix links to piksi and septentrino for gps-for-yaw

peterbarker:


peterbarker:
Built and tested locally

PR #4289, opened 2022-04-12
Updated Cartographer launch "Remap odom and imu data"

snktshrma:
SLAM had an issue estimating position without remapping. Added remap tag with appropriate topics.


rmackay9:
I don't know this is correct but I think it might be :-)

snktshrma:
Thank you @rmackay9
This remapping is done to provide imu and odom input to cartographer so that it can estimate and map efficiently, especially in case of multi-copter where roll and pitch can cause distortion in mapping.

PR #4300, opened 2022-04-17
Add FW control surface gain scaling to tailsitter tuning

Hwurzburg:


plane/source/docs/tailsitter-tuning-guide.rst
```diff
@@ -47,3 +47,13 @@ Roll Tuning
 
 This should allow you to get a reasonable tune for Pitch and Yaw. Roll is tuned like STEP 10 for Roll in :ref:`QuadPlane<quadplane-vtol-tuning>`. Roll axis can even be AutoTuned using QAUTOTUNE, if restricted to only the roll axis using :ref:`Q_AUTOTUNE_AXES<Q_AUTOTUNE_AXES>`.
 
+VTOL Gain Scaling for Redundant Actuators
+=========================================
+
+Some tailsitters have redundant actuators for a given axis. Vectored thrust tailsitters may have both tilting motors for yaw as well as elevons. Copter tailsitters may have fixed wing control surfaces that produce pitch or yaw in addition to the copter style motors. In some cases, it would be desired to have those fixed wing control surfaces provide more attitude control in VTOL than the motors, reducing the thrust/throttle levels needed for that control.
+
+The following parameters allow the adjustment of now much control is produced either by the motors (and vectoring, if applicable) or the fixed wing control surfaces for each axis. Values above 1 weight more to the control surfaces, below 1, more to the motors. Default value is 1, which results in no behavior change from the past.
```
IamPete1, line 55:
```suggestion
The following parameters allow the adjustment of now much control is produced either by the the fixed wing control surfaces for each axis. Larger values apply more gain to the control surfaces, this will give the overall system more gain so the rate gains may have to be reduced. To reduce the response of the motors, one would for example, half the rate gains and use a control surface scale factor of two. This would result in the control surfaces responding as before with the motor outputs halved. Default value is 1, which results in no behavior change from the past.
```
I have reworded slightly, this is only the control surface gain. To get the motors to do less you have to lower the main PIDs.

IamPete1 (commented):
Thanks for adding this!

plane/source/docs/tailsitter-tuning-guide.rst
Hwurzburg: thanks...have changes....also Tridge said not to change the SMAX but rather leave as default, so removed that section

Issue #4301, opened 2022-04-18
Get and Set FlightMode - MAVLink doco - small edits


pauljeffress:
[Get and Set FlightMode](https://ardupilot.org/dev/docs/mavlink-get-set-flightmode.html) 
In the table for param1 it says "MAV_MODE_FLAG_CUSTOM_MODE_ENABLED=1(", should there be some additional text there after the open bracket?
Also, for param2, the "flightmode number", it would be cool if there were links to what those numbers are/mean, for the different ArduPilot platforms (Rover, Plane etc).
I think this is the one for Plane? https://ardupilot.org/plane/docs/flight-modes.html#
I think this is it for Rover? https://ardupilot.org/rover/docs/rover-control-modes.html
Thanks,
Paul

pauljeffress:
I just noticed that while the links I mentioned talk about each platforms modes, they don't actually give you the mode name to mode number mapping that you need when using the MAVLink MAV_CMD_DO_SET_MODE command as referenced on 
[Get and Set FlightMode](https://ardupilot.org/dev/docs/mavlink-get-set-flightmode.html)
What we really need is the info that I can only find in the code itself like for Rover
https://github.com/ardupilot/ardupilot/blob/master/Rover/mode.h#L21
where it shows the relevant enum.
`    // Auto Pilot modes
    // ----------------
    enum Number : uint8_t {
        MANUAL       = 0,
        ACRO         = 1,
        STEERING     = 3,
        HOLD         = 4,
        LOITER       = 5,
        FOLLOW       = 6,
        SIMPLE       = 7,
        AUTO         = 10,
        RTL          = 11,
        SMART_RTL    = 12,
        GUIDED       = 15,
        INITIALISING = 16
    };`

rmackay9:
@pauljeffress,

Txs for the report.  I'll remove the stray "(".

Near the top of the page it has links, "see [FLTMODE1](https://ardupilot.org/copter/docs/parameters.html#fltmode1) for flightmode numbers" but maybe we can repeat the link down in the param2 field.

pauljeffress:
@rmackay9 ,
Can't believe I didn't see the links at the top of the page..... :( opps.
[...]

PR #4302, opened 2022-04-18
Dev: add flight mode number ref to mavlink-get-set-flightmode

rmackay9:
Here's an attempt to resolve issue https://github.com/ArduPilot/ardupilot_wiki/issues/4301


rmackay9:
@Hwurzburg I think we should merge this.. what do you think?

Hwurzburg:
Should the blimp flight modes be added to the enum list?

Hwurzburg:
was not talking about this....about the xml file....

PR #4306, opened 2022-04-20
Common: dshot command parameters put in list

rmackay9:
This makes the DShot Commands section slightly more clear (I think) by putting the relevant parameters in a list.

I've tested this on my local machine and after this PR it looks like this
image

It's still not great because this section starts with the "how" (e.g. dshot commands) instead of the "what" or "why" (e.g. "Using DShot to reverse the direction of the motors, use the ESCs LEDs for autopilot status, use the motors as buzzers".


andyp1per (approved):

rmackay9:
Thanks!

PR #4308, opened 2022-04-21
Common: button clarification

rmackay9:
This makes the setup of buttons easier to follow at the cost of some duplication.

I've tested this on my local PC and the output looks like below.  I've also actually tested the feature on my CubeOrange to confirm that the instructions are correct.

image

### DIFF
--- a/common/source/docs/common-buttons.rst
+++ b/common/source/docs/common-buttons.rst
@@ -4,23 +4,53 @@
 Buttons
 =======
 
-Up to four button inputs to the autopilot may be defined that will execute an :ref:`Auxiliary Function <common-auxiliary-functions>` as if an RC channel switch had been used to execute it. The button input can be either a switch to ground, or a PWM input signal.
+Up to four external buttons or switches connected to the autopilot can be configured to trigger :ref:`Auxiliary Functions <common-auxiliary-functions>` similar to how RC channel switches can trigger them. These button inputs can be configured to use either digital logic level voltage inputs (pull-to-ground or pull-to-high) or PWM signal inputs.
 
-:ref:`BTN_ENABLE<BTN_ENABLE>` must be set to 1 to enable this functionality and see the ``BTN`` parameters.
+.. note::
 
-Setup
-=====
+    Some :ref:`auxiliary function <common-auxiliary-functions>` have a middle position (e.g "Parachute 3 Pos Switch") but the middle position cannot be triggered from a button
 
-A GPIO pin (see :ref:`common-gpios`) can be designated as a button input by setting one of the four ``BTN_PINx`` parameters to that GPIO pins' number, eg. :ref:`BTN_PIN1<BTN_PIN1>` = 51. Up to four pins can be assigned to buttons.
+Basic Setup
+-----------
 
-These pins will have an internal pullup assigned to them automatically, requiring only a switch to ground to change state from high to low. Open or high on the pin will correspond to the ``RCx_OPTION`` function it controls as being HIGH. Ground or logic low on the pin corresponds to LOW. There is no middle option.
+In order to use a button, an autopilot pin must be configured as a GPIO. Many autopilots have dedicated pins for this, but some do not. In those cases, a normal SERVO/MOTOR output must be configured as a GPIO pin instead of an :ref:`common-rcoutput-mapping`.
 
-Alternatively, a PWM signal can be used on the pin for switching, just like the PWM value for an RC channel option switch, by setting bit 0 of the button's ``BTNx_OPTIONSx<BTN_OPTIONSx>`` parameter. PWM values >1800us represent HIGH, and <1200us represent LOW.
+For autopilots with an IOMCU (e.g. Pixhawk, CubeOrange) the external button or switch should be connected to the autopilot's AUX outputs (not the MAIN outputs).  To ease configuration and avoid timer conflicts it may be best to use AUX outputs at the high end (e.g. AUX5 or AUX6) far from outputs used for motors and servos.  For autopilots without an IOMCU (e.g. pixracer, etc) the normal SERVO/MOTOR outputs may be used.
 
-There is no MIDDLE position for the function, if it has one, using a button.
+- Set :ref:`BTN_ENABLE<BTN_ENABLE>` = 1 (Enable) and refresh parameters
+- If using 4.2 (or later) set the ``SERVOx_FUNCTION`` = -1 (GPIO) where "x" is the servo output channel connected to the button . For example, if the AUX6 output on an autopilot with an IOMCU is to be used as a GPIO, set :ref:`SERVO14_FUNCTION<SERVO14_FUNCTION>` = -1.
+- If using 4.1 (or earlier) on an autopilot with an IOMCU (e.g. Pixhawk, CubeOrange) set ``BRD_PWM_COUNT`` low enough so that the servo desired and all other higher outputs may be used for a GPIO input.  E.g. ``BRD_PWM_COUNT`` = 4 means only AUX outputs 5 and 6 can be used as GPIOs, BRD_PWM_COUNT = 0 means AUX outputs 1 to 6 can be used. For boards not using an IOMCU, ``BRD_PWM_COUNT`` less than the total output number, will configure those higher outputs as GPIOs.
 
-The effective state of the button can be inverted by setting its ``BTN_OPTIONSx<BTN_OPTIONSx>``  bitmask bit 1 (ie adding 2 to the decimal value of the parameter). This applies to both PWM and switch control of the pin.
+Digital Logic/Analog Voltage Setup
+----------------------------------
 
-The function that will be controlled by the button is determined by its ``BTN_FUNCx`` parameter. If set to -1, no function is performed. Otherwise, the number corresponds to the :ref:`Auxiliary Function <common-auxiliary-functions>` number.
+Digital logic level analog voltages can be used for the button input.  The parameter changes listed below are for the 1st button but the 2nd, 3rd and 4th buttons can be setup similarly, simply replace the "1" in the parameter names with "2", "3" or "4". In order to determine which GPIO pin number corresponds to which autopilot SERVO/MOTOR output pin or dedicated GPIO pin, the autopilot's hwdef.dat file must be consulted. See :ref:`common-gpios` for how to determine this.
 
-The :ref:`BTN_REPORT_SEND<BTN_REPORT_SEND>` sets the length of time that a button state message is sent to the ground control station for display purposes, each time the button changes state. This is a MAVLink2 message.
+- Set :ref:`BTN_PIN1<BTN_PIN1>` to the number of the GPIO connected to the button/switch (e.g. "55" = "AUXOUT6")
+- Set :ref:`BTN_OPTIONS1<BTN_OPTIONS1>` = to "0" to trigger on high logic level voltage (>2.4V) and  be interpreted as a "HIGH" position by the Auxiliary function, "2" to trigger on low logic level voltage (<0.8V) and be interpreted as a "LOW" position by the Auxiliary Function.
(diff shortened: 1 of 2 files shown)


rmackay9:
OK, I'm done with this change now I think.  Sorry for all those who received spammy messages as I kept pushing  updates.

common/source/docs/common-buttons.rst
```diff
@@ -4,23 +4,51 @@
...
+For autopilots with an IOMCU (e.g. Pixhawk, CubeOrange) the external button or switch should be connected to the autopilot's AUX outputs (not the MAIN outputs).  To ease configuration and avoid timer conflicts it may be best to use AUX outputs at the high end (e.g. AUX5 or AUX6) far from outputs used for motors and servos.  For autopilots without an IOMCU (e.g. pixracer, etc) the main outputs may be used.
 
-Alternatively, a PWM signal can be used on the pin for switching, just like the PWM value for an RC channel option switch, by setting bit 0 of the button's ``BTNx_OPTIONSx<BTN_OPTIONSx>`` parameter. PWM values >1800us represent HIGH, and <1200us represent LOW.
+- Set :ref:`BTN_ENABLE<BTN_ENABLE>` = 1 (Enable) and refresh parameters
+- If using 4.2 (or later) set :ref:`SERVOx_FUNCTION<SERVO14_FUNCTION>` = -1 (GPIO) where "x" is the servo output channel connected to the button (e.g. SERVO14_FUNCTION is for AUX output 6)
```
Hwurzburg, line 19:
```suggestion
- In order to use an normal SERVOx output as a button input for firmware version 4.2 (or later), set :ref:`SERVOx_FUNCTION<SERVO14_FUNCTION>` = -1 (GPIO) where "x" is the servo output channel connected to the button (e.g. SERVO14_FUNCTION is for AUX output 6)
```

common/source/docs/common-buttons.rst
```diff
@@ -4,23 +4,51 @@
...
+- If using 4.1 (or earlier) on an autopilot with an IOMCU (e.g. Pixhawk, CubeOrange) set BRD_PWM_COUNT low enough so that the servo output may be used for reading an analog voltage (aka GPIO) or PWM input.  E.g. BRD_PWM_COUNT = 4 means only AUX outputs 5 and 6 can be used, BRD_PWM_COUNT = 0 means AUX outputs 1 to 6 can be used.
```
Hwurzburg, line 20:
```suggestion
- In order to use an normal SERVOx output as a button input for firmware version 4.1 (or earlier) on an autopilot with an IOMCU (e.g. Pixhawk, CubeOrange), set BRD_PWM_COUNT low enough so that the servo output may be used for reading an analog voltage (aka GPIO) or PWM input.  E.g. BRD_PWM_COUNT = 4 means only AUX outputs 5 and 6 can be used, BRD_PWM_COUNT = 0 means AUX outputs 1 to 6 can be used.
```

common/source/docs/common-buttons.rst
```diff
@@ -4,23 +4,51 @@
...
+- Set :ref:`BTN_FUNC1<BTN_FUNC1>` to the :ref:`Auxiliary Function <common-auxiliary-functions>` to be triggered.  If set to 0, no function is triggered
+
+.. note::
+
+    The valid PWM input range is 800us to 2200us.  If the PWM value falls outside this range it is equivalent to the RC auxiliary function switches position going to the low position.  Normally this means the auxiliary function will not be triggered.
+
+Additional Setup
+----------------
+
+:ref:`BTN_REPORT_SEND<BTN_REPORT_SEND>` sets the length of time that a `BUTTON_CHANGE <https://mavlink.io/en/messages/common.html#BUTTON_CHANGE>`__ mavlink2 message is sent to the ground control station each time the button state changes.  At the time of this writing we do not know of any GCS that make use of this message.
```
Hwurzburg, line 54:
```suggestion
:ref:`BTN_REPORT_SEND<BTN_REPORT_SEND>` sets the length of time that a `BUTTON_CHANGE <https://mavlink.io/en/messages/common.html#BUTTON_CHANGE>`__ mavlink2 message is sent to the ground control station each time the button state changes.  At the time of this writing we do not know of any GCS that makes use of this message.
```

Hwurzburg (changes requested):

common/source/docs/common-buttons.rst
```diff
@@ -4,23 +4,51 @@
 Buttons
 =======
 
-Up to four button inputs to the autopilot may be defined that will execute an :ref:`Auxiliary Function <common-auxiliary-functions>` as if an RC channel switch had been used to execute it. The button input can be either a switch to ground, or a PWM input signal.
+Up to four external buttons or switches connected to the autopilot can be configured to trigger :ref:`Auxiliary Functions <common-auxiliary-functions>` similar to how RC channel switches can trigger them. These button inputs can be either analog voltage inputs (pull-to-ground or pull-to-high) or PWM inputs.
```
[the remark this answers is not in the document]
rmackay9: Txs, I've updated the top paragraph.  I've left the bottom paragraph out for now though because the parameter descriptions don't include pin numbers for anything except the AUX OUT pins.

common/source/docs/common-buttons.rst
```diff
@@ -4,23 +4,51 @@
...
 
-:ref:`BTN_ENABLE<BTN_ENABLE>` must be set to 1 to enable this functionality and see the ``BTN`` parameters.
+.. note::
 
-Setup
-=====
+    Some :ref:`auxiliary function <common-auxiliary-functions>` have a middle position (e.g "Parachute 3 Pos Switch") but the middle position cannot be triggered from a button
```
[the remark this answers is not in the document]
rmackay9: .. is there any value is re-stating that the middle position cannot be triggered "either from analog voltage or PWM input signals"?  It sort of implies that maybe there is some 3rd method that does allow the middle position to be triggered.

common/source/docs/common-buttons.rst
rmackay9: Txs this is probably true but again, the param descriptions don't specify any pins except for the AUX OUTs so I'd like to keep it simple for now.

common/source/docs/common-buttons.rst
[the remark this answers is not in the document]
Hwurzburg: those are given as an aid  for IOMCU based autopilots...any appropriate pin can be input....if that is confusing we should change the value and descriptions to make it clear....those numbers for a MATEK flight controller are SERVO1-SERVO6, etc.

common/source/docs/common-buttons.rst
rmackay9: Same here, let's keep it simple and just focus on the aux outs.  Hopefully at some point someone will come through and re-test with other random GPIO pins and update the param descriptions and this wiki page at the same time (maybe I will..).

common/source/docs/common-buttons.rst
rmackay9: fixed, txs!

Hwurzburg:
@rmackay9 rather than forcing it, I just added my suggestions as a commit....if you agree, we can squash before merge..

common/source/docs/common-buttons.rst
```diff
@@ -4,7 +4,7 @@
 Buttons
 =======
 
-Up to four external buttons or switches connected to the autopilot can be configured to trigger :ref:`Auxiliary Functions <common-auxiliary-functions>` similar to how RC channel switches can trigger them. These button inputs can be configured to use either analog voltage inputs (pull-to-ground or pull-to-high) or PWM signal inputs.
+Up to four external buttons or switches connected to the autopilot can be configured to trigger :ref:`Auxiliary Functions <common-auxiliary-functions>` similar to how RC channel switches can trigger them. These button inputs can be configured to use either digital logic signal analog voltage inputs (pull-to-ground or pull-to-high) or PWM signal inputs.
```
rmackay9, line 7: Ah you're right here.. it's digital inputs not analog as I've said.  maybe we should remove the word "analog"?  Anyway, I'm happy either way.

common/source/docs/common-buttons.rst
```diff
@@ -13,20 +13,22 @@ Up to four external buttons or switches connected to the autopilot can be config
...
 
 - Set :ref:`BTN_ENABLE<BTN_ENABLE>` = 1 (Enable) and refresh parameters
-- If using 4.2 (or later) set :ref:`SERVOx_FUNCTION<SERVO14_FUNCTION>` = -1 (GPIO) where "x" is the servo output channel connected to the button (e.g. SERVO14_FUNCTION is for AUX output 6)
-- If using 4.1 (or earlier) on an autopilot with an IOMCU (e.g. Pixhawk, CubeOrange) set BRD_PWM_COUNT low enough so that the servo output may be used for reading an analog voltage (aka GPIO) or PWM input.  E.g. BRD_PWM_COUNT = 4 means only AUX outputs 5 and 6 can be used, BRD_PWM_COUNT = 0 means AUX outputs 1 to 6 can be used.
+- If using 4.2 (or later) set the ``SERVOx_FUNCTION`` = -1 (GPIO) where "x" is the servo output channel connected to the button (e.g. SERVO9_FUNCTION is for AUX output 1 for boards using an IOMCU)
```
rmackay9, line 21: we have a little difference of opinion on how to handle these "x" parameters.  I like to link to one of the parameters anyway while you like to put them in quotes.  Not a big deal of course either way but I (of course) like my way better.

common/source/docs/common-buttons.rst
```diff
@@ -37,16 +39,16 @@ The pin is automatically configured with an internal pullup meaning the button o
...
 
-- Set :ref:`BTN_PIN1<BTN_PIN1>` to the number of the pin connected to the button/switch (e.g. "55" = "AUXOUT6")
-- Set :ref:`BTN_OPTIONS1<BTN_OPTIONS1>` = to "1" to trigger the auxiliary function with a PWM > 1800us
-- Set :ref:`BTN_OPTIONS1<BTN_OPTIONS1>` = to "3" to trigger the auxiliary function with a PWM < 1200us
-- Set :ref:`BTN_FUNC1<BTN_FUNC1>` to the :ref:`Auxiliary Function <common-auxiliary-functions>` to be triggered.  If set to 0, no function is triggered
+- Set :ref:`BTN_PIN1<BTN_PIN1>` to the number of the GPIO connected to the button/switch (e.g. "55" = "AUXOUT6")
+- Set :ref:`BTN_OPTIONS1<BTN_OPTIONS1>` = to "1" to be interpreted as a "HIGH" position by the auxiliary function with a PWM > 1800us
+- Set :ref:`BTN_OPTIONS1<BTN_OPTIONS1>` = to "3" to be interpreted as a "LOW" position by the auxiliary function with a PWM < 1200us
```
rmackay9, line 46: Ah, I think this change is incorrect.  With BTN_OPTIONS1 = 3 the button's HIGH position is triggered when the PWM < 1200us.

rmackay9:
Thanks for this.. I've put a couple of suggestions in there but after that feel free to merge.

As a side note I uncovered [this bug](https://github.com/ArduPilot/ardupilot/issues/20609) while testing the out-of-range PWM input.

common/source/docs/common-buttons.rst
Hwurzburg: right...

common/source/docs/common-buttons.rst
Hwurzburg: the problem I have is the link arrives at a parameter that does not match the highlited text...alternatively we can explicitly show it as a stated example,and make them match...either way is fine with me...

Hwurzburg:
> Thanks for this.. I've put a couple of suggestions in there but after that feel free to merge.
> 
> As a side note I uncovered [this bug](https://github.com/ArduPilot/ardupilot/issues/20609) while testing the out-of-range PWM input.

interesting, when Tridge skimmed the code, he said Interrupts were not involved...

PR #4311, opened 2022-04-21
Dev: Update X-Plane docs

### LINKED ISSUE #4295: Add X-Plane data in and out settings to sitl-with-xplane.rst
There was a step left out of the instructions for sitl-with-xplane.rst . When I used the additional instructions below (from x-plane-hardware-in-the-loop-simulation.rst) sitl with xplane worked. With X-Plane's initial default settings, no data will go in or out of X-Plane.

These are the screen shots that show the necessary settings you need to set in the Settings Menu – Data Input and Output:
image
image

TunaLobster:
A few of the screenshots are older, but you can still make out what is going on between the newer Mission Planner and X-Plane 11 moving things around a little bit. Removed unused screenshot file.

Closes #4295


Hwurzburg:
thanks!

PR #4312, opened 2022-04-21
Rover: add note about clearing waypoints

yuri-rage:
SaveWP behavior while disarmed was not documented.


rmackay9 (approved):

Hwurzburg:
I put adding this function for Copter on my to-do-list

PR #4320, opened 2022-04-25
Add Chimera7 as a reference frame

andyp1per:
Add Chimera7 as a reference frame. @rmackay9 / @Hwurzburg not sure how much detail you want here as a lot is better covered in the video series / blog


rmackay9:
great to see this!

Hwurzburg:
thanks...

PR #4321, opened 2022-04-26
Update information on GPIO configuration for IOMCU

Hwurzburg:


rmackay9 (approved):
LGTM.  There's certainly some important new info added with this PR.  For example the pin numbers for the MAIN outputs and also the limitation that MAIN outputs can only be used for outputs.. not inputs.

IamPete1 (approved):
Thanks.

PR #4328, opened 2022-04-28
Update tailsitter-tuning-guide.rst

kd0aij:
fix typo
it's SLT in the source, and I assume "left" is supposed to be "lift" in the comment:

// Transition for separate left thrust quadplanes
class SLT_Transition : public Transition


Hwurzburg:
thanks

PR #4342, opened 2022-05-16
Wiki:ROS Removed Note for protobuf removal

khanasif786:
```make uninstall``` is no longer working i have tested many times. catkin build gives error everytime until the package is removed. I solved (or ignored) this problem by just removing this package from the list everytime when i do a ```catkin build```. (Note that compiled package is still in ROS build folder i am using that from my first compilation). until anyone get a solution we should remove this Instruction.


rmackay9 (approved):
ok, I don't know myself but I'm convinced @khanasif786 knows what he's talking about.

PR #4343, opened 2022-05-17
Deprecate Eclipse and Cygwin as Windows build environments

Hwurzburg:


rmackay9:
Let's have a go at adding a VSCode setup guide.. FYI @davidbuzz @hendjoshsr71 @lthall

hendjoshsr71:
Certainly can deprecate the Eclipse page. 

However, Cygwin still has it uses for Mission Planner SITL binaries.  perhaps move that page or reference it both in Mission Planner SITL  as well at the DEV "Using SITL" page?

rmackay9:
@hendjoshsr71, the only problem with deprecating the Eclipse page is that that leaves us with no explanation on how to setup an IDE.  Eclipse on Windows with cygwin gives the developer an environment where they can both edit and compile which leads to fast iterative development.  If we remove that we're not giving any instructions on how to setup an IDE environment so users will tend towards compiling on the command line and editing in another stand-alone tool which is much less efficient..

.. so we need a page to describe how to setup VSCode as an IDE enabling editing, compiling and ideally even debugging with GDB

dev/source/docs/building-setup-windows.rst
```diff
@@ -4,20 +4,7 @@
 Setup the Build Environment on Windows
 ======================================
 
-There are two options for building on Windows: Windows Subsystem for Linux (WSL) or Cygwin. New developers should use WSL as it offers faster compilation times, greater compatibility, and a larger support base from other developers.
+Cygwin and Eclipse are no longer recommended for building code on a Windows PC. WSL2 is now the preferred environment.
```
TunaLobster, line 7: Cygwin must be supported or SITL in Mission Planner would no longer work. The sentence that is already there covers what is in practice.

TunaLobster (commented):
I don't agree with these changes calling for removing Cygwin without replacement for equivalent features. I witness Windows native binaries being used multiple times a day and removing support for that workflow would not be beneficial. Eclipse can go though. However, VSCode does not have integration with Cygwin.

The knowledge on how to set up VSCode for ArduPilot development is scattered across several dev team members right now. I know Peter Hall and Josh have done the GDB attachment. I've done it once, but I also don't really understand what the tool tells me. Build commands are included in one of the suggested extensions. We can add more or modify the suggested extension list if we want too. I also thought about adding code snippets that match the ArduPilot code style.

dev/source/docs/building-setup-windows.rst
Hwurzburg: the general consensus so far has been to have the wiki direct new devs to use WSL in Windows for builds....and VSCode for an IDE in Windows...and to sim with SITL in WSL in Windows in conjunction multiple graphical/physics interfaces like RealFlight...non-devs are the ones usually asking to use MP sims, so they dont really need to build an ELF.... but info on that using github actions is in the wiki, although needs to be made clearer

I think we should structure the wiki to point down the above  path for Windows boxes...but not deprecate the CYGWIN/ECLIPSE setup and CYGWIN build but put those pages for advanced users...

We need some new info, which I think @rmackay9 is working on:
- setup VSCODE with links to WSL in windows for an IDE
- how to use the VSCODE deug hooks in that environment
also how to setup WSL2 in Win11 (which I have already done, but did note the steps....will have to create a VM and write down the steps for a wiki page)

We get very few support questions about setting up a Linux box, but do get them from Windows users...at least my impression over the last few years.....especially having issues with Cygwin builds...haven't kept track since when those arise someone points them to WSL

hendjoshsr71: This pages gives decently detailed debug info.  on the use cases I've ran across. https://ardupilot.org/dev/docs/debugging-with-gdb-using-vscode.html

I would still be tempted to deprecate Eclipse, if 0 devs are actually using it? (If we get Randy to convert ;) )

hendjoshsr71:
On this VSCode page I'd like to add info. about the LiveSession option. It has been very handy with help partners debugging code.

PR #4346, opened 2022-05-18
Fix wrong parametrer description for find_serial

mbuzdalov:
The `find_serial` binding has two arguments in C++, the protocol and the instance number. In Lua bindings, the first argument is hardwired to 28, while the second one becomes the first Lua argument. The documentation referred to the first C++ argument instead, which caused a lot of confusion.

Now it seems to reflect the current state better.

### DIFF
--- a/common/source/docs/common-lua-scripts.rst
+++ b/common/source/docs/common-lua-scripts.rst
@@ -446,7 +446,7 @@ RC Channels (rc:)
 Serial/UART (serial:)
 ~~~~~~~~~~~~~~~~~~~~~
 
-- :code:`find_serial(protocol)` - Returns the first UART instance that allows the given protocol, or nil if not found.
+- :code:`find_serial(port_number)` - Returns the UART instance that allows connections from scripts (those with :code:`SERIALx_PROTOCOL = 28`). For :code:`port_number = 0`, returns first such UART, second for :code:`port_number = 1`, and so on. If the instance is not found, returns :code:`nil`.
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
are the docs impacted?

mbuzdalov:
I believe only documentation is impacted. I don't know, however, how much of that is generated, and what the primary sources are.

mbuzdalov:
Could you please maybe point me to it, so I can fix that as well?

I have searched for both 'find_serial' and 'the first UART instance that allows', and found only the calls to that function (in both Lua and C++), the occurrence edited in this commit, and the doc in `AP_SerialManager.cpp` which seems to be unlikely to be converted to Lua documentation, because the wording is very different.

Hwurzburg:
https://github.com/ArduPilot/ardupilot/blob/53a7866ba3741634e0bb430ff89d263d705ce5cb/libraries/AP_Scripting/docs/docs.lua#L1119

IamPete1 (approved):
LGTM, Thanks. We could change the docs.lua to use variable name `instance` rather than `protocol`

https://github.com/ArduPilot/ardupilot/blob/53a7866ba3741634e0bb430ff89d263d705ce5cb/libraries/AP_Scripting/docs/docs.lua#L1120

mbuzdalov:
Made a related PR in ardu/ardu: https://github.com/ArduPilot/ardupilot/pull/20798.

Hwurzburg:
thanks

PR #4347, opened 2022-05-19
Copter: command list correction for Loiter Turns direction

rmackay9:
This corrects the Copter mission command list to remove the Dir field which isn't actually supported.

This is in response to [this discussion](https://discuss.ardupilot.org/t/loiter-turns-in-missions-clockwise-only/85690/3)

I have tested this locally and it looks OK.


Hwurzburg:
thanks

Hwurzburg:
I actually thought I had already merged this....was this an error?

PR #4355, opened 2022-05-22
Copter Motors Library: Post a site to determine each motor factor from the motor layout

muramura:
I will illustrate the complex motor layout with a web tool
The WEB tool determines each motor factor based on this diagram.
As shown in the video, the tool can support complex frames.

I will post this WEB tool.

### DIFF
--- a/dev/source/docs/code-overview-copter-motors-library.rst
+++ b/dev/source/docs/code-overview-copter-motors-library.rst
@@ -94,5 +94,7 @@ To add support for a new Multicopter Matrix frame type (i.e. a frame with downwa
 
 Detailed instructions are not yet available for adding more complex frame types (sorry!)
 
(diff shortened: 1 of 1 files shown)


IamPete1:
The page is a little out of date now, we have moved to a table format in `AP_MotorsMatrix`. Would be great to update it to output the lua scripting format. 

Maybe we could host on ardupilot.org, @iforce2d?

Hwurzburg:
@IamPete1 @muramura shall we close this now?

IamPete1:
Would be nice to host a updated version that would generate a lua script, but as it is the tool is a little too out of date I think. So closing is OK.

rmackay9 (approved):
looks good, sorry for the delay!

PR #4357, opened 2022-05-23
d-shot doc update

proficnc:
corrected docs to clearly indicate that only bi-directional dshot is linited to aux 2-6, all other channels are usable for single directional


rmackay9 (approved):

Hwurzburg:
thanks

PR #4368, opened 2022-05-27
Fixed code block styling Functions and variables.

HarleyRossetto:
Both 'Right' examples under 'Functions and variables' were missing a blank link below :: and not being styled as code blocks.


Hwurzburg:
thanks

PR #4370, opened 2022-05-28
LOITER_TURNS: Add a radius item

muramura:
Changed implementation for radius of LOITER_TURNS at "ArduCopter: support *10 multipler when storing/retrieving radius in NAV_LOITER_TURNS" .
I would like to note in the WIKI that the radius does not change smoothly.


Hwurzburg:
@rmackay9 was the fix back ported to 4.2?

rmackay9 (approved):

PR #4388, opened 2022-06-14
Update ESC calibration page for DShot and active braking

MichelleRos:
Updates the ESC calibration page to clarify that digital protocols like DShot don't need to be calibrated.
Also the difference between brake on stop and active braking in ESC settings as current settings are unclear as to which is referred to.
Also removed some references to the APMs.

@tridge I've added CAN ESCs in here as well as you suggested, though it seems like this would only apply to integrated CAN ESCs - ESCs that use PWM but via a CAN adapter would still need to be calibrated somehow, right? Not sure if we need to mention this on the page or if it's clear enough?

### DIFF
--- a/copter/source/docs/esc-calibration.rst
+++ b/copter/source/docs/esc-calibration.rst
@@ -26,6 +26,7 @@ fails try the "Manual ESC-by-ESC" method.
 
 -  Refer to :ref:`common-brushless-escs` for setup of the ESC protocol (:ref:`MOT_PWM_TYPE <MOT_PWM_TYPE>`) according to your type of ESC.
 -  Some ESCs like the DJI Opto ESCs do not require and do not support calibration, so skip this page completely.
+-  If you are using a digital esc protocol like DShot or are using CAN ESCs then you also don't need to calibrate them, so skip this page. The section at the end on ESC settings is still applicable, however.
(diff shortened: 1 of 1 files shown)


Hwurzburg:
LGTM....merging....also will bring that note to top of page rather than at bottom for Plane...and Rover should have a page....on my list

MichelleRos:
Thanks.

rmackay9:
txs for the fix.  It's all fine but "APM" normally refers to hardware.. the old APM board so I think it should normally be replaced with "autopilot" rather than "ArduPilot".

Hwurzburg:
thinks it works either way here... could replace ArduPilot board with autopilot, versus "autopilot board"....but I think its okay as is...

MichelleRos:
@rmackay9 I thought that "autopilot" might confuse users more - in common use an "autopilot" is more the software than the hardware - or both combined (e.g. the Teslas' "autopilot system"). Calling it an "ArduPilot board" makes it clear that we're referring to the hardware board.

rmackay9:
@MichelleRos,

Yeah, not sure about using "ArduPilot board".  Maybe it's all history now but we've had lots of confusion between hardware and software.  APM vs ArduPilot.. Pixhawk vs ArduPilot.  I haven't seen the term "ArduPilot board" much so that seems like a new pair of words and I'm not a fan of using new words when it's not necessary.

For me "autopilot" means hardware.  "board" certainly means hardware so maybe that's an even better choice.

MichelleRos:
@rmackay9 Makes sense to try not to use new terms. Probably "board" is a good one then. I'll remember for future reference. Thanks. 🙂

PR #4399, opened 2022-06-22
Fix build warnings

TunaLobster:
These are several warnings that have come up during CI. None of them are errors, just warnings. There are still warnings about the sphinxcontrib.vimeo plugin that are resolved by #4195.


TunaLobster:
I noticed it in the other PR is submitted today. grep for "WARNING:" https://pipelines.actions.githubusercontent.com/serviceHosts/2637c4db-a9cd-4358-9d3e-4cf63ba8c921/_apis/pipelines/1/runs/3836/signedlogcontent/2?urlExpires=2022-06-22T20%3A53%3A14.7456680Z&urlSigningMethod=HMACV1&urlSignature=NiKW%2BPpa1cR3IGWa2jMfWDlp%2FRxysjI24L4zmKTH1E8%3D

Hwurzburg:
okay, that warning is not output as an error output like the content warnings/errors are (like those params), so I never see it in the daily chron job...thanks

PR #4400, opened 2022-06-23
Add CAN port description to Zealot H743

timtuxworth:
The CAN ports are disabled by default. It's important to note that each of the CAN ports needs to be enabled if it is going to be used.

### DIFF
--- a/common/source/docs/common-qiotek-zealoth7.rst
+++ b/common/source/docs/common-qiotek-zealoth7.rst
@@ -535,7 +535,7 @@ Then reboot.
 
 DroneCAN capability
 ===================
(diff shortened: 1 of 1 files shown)


Hwurzburg:
with the exception of a few autopilots that have CAN which enable the ports with Driver 1 in their defaults, most CAN enabled autopilots default to having the CAN ports disabled....that's why the wiki on CAN setup instructs to set a driver for the port....
having ports enabled to a default driver ( and you still have to setup the driver appropriately) consumes system resources that might be needed elsewhere....I am not sure that adding this for just this one autopilot, especially when covered in material that would need to be read to use DroneCAN or CAN anyway meaningfully is the way to go....if we do it here, it should be done on every CAN capable autopilot page....but  seems unneeded duplication since it appears on a page someone needs to read to use it anyway...would be like having to say you need to set the protocol on a serial port to rangefinder serial to use it, when a user must read a page on rangefinders that says that already...

timtuxworth:
I see your point @Hwurzburg , but I didn't find the CAN setup page/parameters when I was trying to set this up. I still think the page should say something about the CAN ports, but maybe if I changed it to refer to / reference the CAN setup page instead?

Hwurzburg:
LGTM...thanks

fyi "but I didn't find the CAN setup page/parameters when I was trying to set this up"....searching for 'CAN' on a vehicle's wiki main page would have provided links to the parms and the wiki pages...
Screenshot from 2022-06-24 06-57-15

PR #4401, opened 2022-06-23
Add ARK CANnode

AlexKlimaj:


Hwurzburg:
has this been tested by anyone in AP? if so, do you intend to add AP compatibility to your product website description of the product?

TunaLobster:
There is already a target for ARK-GPS. How different is the CANnode from the GPS? Should be pretty easy to make a new target if they are similar.

AlexKlimaj:
The ARK CANnode and ARK GPS are very similar. Same processor and share some of the same IO.
https://github.com/ARK-Electronics/ARK_GPS/blob/main/Hardware/Rev%202/ARK%20GPS%20Rev%202%20Schematic.PDF
https://github.com/ARK-Electronics/ARK_CANNODE/blob/main/Hardware/Rev%201/ARK%20CANNODE%20Rev%201%20Schematic.PDF

Hwurzburg:
rather than having you rebase, I just added it with #4680....closing this one

PR #4407, opened 2022-06-26
Add info about memory hit from each CAN driver

timtuxworth:
AndrewTridgell — Today at 3:07 PM
@timtuxworth with defaults each CAN bus costs 8k of memory pool
AndrewTridgell — Today at 3:15 PM
make that 12k, each driver has a 4k stack


common/source/docs/common-canbus-setup-advanced.rst
```diff
@@ -50,6 +50,8 @@ parameter ``CAN_Px_DRIVER``, where x is the number of the CAN port.
 The value of this parameter is the id of driver that will be associated with this
 port (interface).
 
+Each enabled bus/driver will use 12k of memory (RAM not flash) with defaults.
```
Hwurzburg, line 53: actually the total pool is 16K by default....and serves all drivers...so two enabled CAN ports will be allocated 8K

Hwurzburg (changes requested):

common/source/docs/common-canbus-setup-advanced.rst
Hwurzburg, line 53:
```suggestion
Each enabled bus/driver will use RAM memory (not Flash) depending on the type of driver and if CANFD is enabled. For example, DroneCAN will allocate 12KB for its driver (24K if CANFD) by default, but can vary from board to board depending on its defaults, if set by its hardware definition file. The ``CAN_Dx_UC_POOL`` parameter can be used to change the pool size. Required pool size depends on bus traffic required by the attached DroneCAN peripheral and can sometimes be reduced for peripherals such as GPS or Compass, whereas peripherals such as ESCs require more bus traffic and therefore a larger pool size.
```

Hwurzburg (changes requested):

Hwurzburg:
@timtuxworth I made my suggested change and merged...thanks

timtuxworth:
That looks great - it adds a lot of very useful information that I didn't know. Thanks @Hwurzburg

PR #4412, opened 2022-06-28
correcting relay pin settings for omnibusf4pro.

geofrancis:
current wiki is incorrect , in HWDAT it says that the pins 54 and 55 for output 5 and 6.
PA1 TIM2_CH2  TIM2 PWM(5) GPIO(54)
PA8 TIM1_CH1  TIM1 PWM(6) GPIO(55)


Hwurzburg:
good catch....thanks

PR #4428, opened 2022-07-08
common: Update OEM customization options

### LINKED ISSUE #3528: OEM Customisation needs link to APJ tools page
The OEM Customisation page (https://ardupilot.org/copter/docs/common-oem-customizations.html) should mention the third option which is APJ Tools (https://ardupilot.org/dev/docs/apjtools-intro.html).  Perhaps the APJ Tools page should also get a note with a link back to the OEM Customisation page.

We should also add hyperlinks whenver it says "Lua" back to the Lua scripts page. https://ardupilot.org/dev/docs/common-lua-scripts.html

TunaLobster:
Updates OEM customization page to reference APJ Tools. Also changed the example to use an autopilot that has more RAM to use the Lua scripting features. Added Lua scripting references. Also a quick pass on grammar.

Closes #3528

### DIFF
--- a/common/source/docs/common-oem-customizations.rst
+++ b/common/source/docs/common-oem-customizations.rst
@@ -6,16 +6,16 @@ OEM Customization
 
 ArduPilot provides several ways for an OEM to provide firmware on their products that has been customized to their specific system configuration:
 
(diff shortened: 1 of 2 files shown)


Hwurzburg:
LGTM

PR #4437, opened 2022-07-12
common: Accelerometer calibration grammar

TunaLobster:


common/source/docs/common-accelerometer-calibration.rst
```diff
@@ -31,24 +31,24 @@ Calibration steps
 
 - Click **Calibrate Accel** to start the full 3-axis calibration.
 
-   *Mission Planner* will prompt you to place the vehicle each
+   *Mission Planner* will prompt you to place the vehicle on each axis during the
    calibration position. Press any key to indicate that the autopilot is
```
Hwurzburg, line 35: this needs fixing with the above change
```suggestion
   calibration. Press any key to indicate that the autopilot is
```

Hwurzburg (changes requested):

common/source/docs/common-accelerometer-calibration.rst
TunaLobster: Went ahead and fixed the old text width breaks too.

PR #4438, opened 2022-07-12
common: Use ifconfig directive in place of [site wiki=...]

TunaLobster:
This one might be interesting from a build perspective. In the conf.py files there is an extension for the ifconfig directive that has been there for 6 years. The extension allows for controlling the contents of the page based on a Python style expression and properties from conf.py. In each of the conf.py files there is the project property which is the name of the site. That makes it pretty useful to use in the ifconfig directive. There are 232 uses of [site wiki=...] in the common directory right now and I expect all of them could be replaced by this directive. It should speed up building some, but I haven't checked how much yet. Should also make incremental builds easier to attain.

https://www.sphinx-doc.org/en/master/usage/extensions/ifconfig.html

### DIFF
--- a/common/source/docs/common-accelerometer-calibration.rst
+++ b/common/source/docs/common-accelerometer-calibration.rst
@@ -52,14 +52,13 @@ Calibration steps
 
 .. note:: this **Calibrate Level** operation can only correct up to a 10 degree difference between the initial calibration and the final position in the vehicle, and only corrects pitch and roll differences, not yaw.
 
-[site wiki="plane"]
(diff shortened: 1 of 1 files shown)


TunaLobster:
Old on the left and new on right. Plane on top and Copter on the bottom. I don't know why bullet points are not showing up in my build, but it is not related to this change.

image

Hwurzburg:
my issue with using this is the lack of clarity of where the section ends covered by the ifconfig.....like the problem of using blocks....the current way is clear and easy to use....this is kinda if it aint broke dont fix it....and mixing the two would be confusing to new contributors...unless there is a clear advantage to this not mentioned above....build time is not long at the moment and I doubt this halves it

PR #4441, opened 2022-07-15
Add Q_LAND_ALTCHG parameter

Hwurzburg:


plane/source/docs/qrtl-mode.rst
```diff
@@ -21,6 +21,15 @@ The QuadPlane will immediately navigate towards the home location at :ref:`Q_WP_
 .. image:: ../images/QRTL.jpg
     :target: ../_images/QRTL.jpg
 
+.. note::
+
+    QuadPlane will recognize that it has landed if the motors are at
+    minimum but its altitude does not change more than 0.2m for one
+    second.  It does not use the altitude itself to decide whether to shut off the
+    motors except that the QuadPlane must also be below 10m above the home
```
tridge, line 29: i'm not sure where the 10m comes from. There is Q_LAND_FINAL_ALT which defaults to 6m, and that is where land final starts, where is where we start to look for landing. Also, it is not necessarily above home. NAV_VTOL_LAND can use non-zero alt, or can use AGL or AMSL frames

tridge (changes requested):

plane/source/docs/qrtl-mode.rst
Hwurzburg: me neither....should be Q_LAND_FINAL_ALT above landing point...will fix...thanks...

Hwurzburg: was an existing thing...must not have been updated when Q_LAND_ALT_FINAL...fixed

PR #4444, opened 2022-07-16
Add Matek F405 TE family of autopilots

Hwurzburg:


MATEKSYS:
only WTE has SD card slot
F405-WMN, HDTE, miniTE, use 16M flash for logging

Hwurzburg:
thanks....fixed...merging

PR #4460, opened 2022-07-21
Update Common-autopilots.rst

vincentpoont2:
Add Pixhawk 6X & 6C


Hwurzburg:
I will move when ported into Stable firmware...its already in the wiki

PR #4470, opened 2022-07-23
Add FlywooF745 nano and update AIO page

Hwurzburg:


common/source/docs/common-flywoo-f745.rst
```diff
@@ -42,17 +52,28 @@ Specifications
...
 
-   - 33.5mm x 333.5mm (25.6 x 25.6mm mount pattern)
+   - 33.5mm x 33.5mm (25.6 x 25.6mm mount pattern)
    - 8.5g
 
-Where to Buy
-============
+   **Size and Dimensions Nano**
+   
+   - 22mm x 235mm (26mm x16mm mount pattern)
```
andyp1per, line 62: This is 16mmx16mm mount pattern I believe

common/source/docs/common-flywoo-f745.rst
```diff
@@ -70,15 +91,30 @@ UART3 supports RX and TX DMA. UART1, UART2, UART4, and UART6 supports TX DMA. UA
 RC Input
 ========
 
-RC input is configured by default via the USART3 RX input. It supports all RC protocols except PPM. FPort and full duplex protocols will use both RX6 and TX6. The 4.5V output pin is powered when the USB is connected to allow setup of receiver without connecting to the main battery.
+RC input is configured on the RX3 (UART3_RX) pin. It supports all RC protocols except PPM. See :ref:`common-rc-systems` for details for a specific RC system. :ref:`SERIAL3_PROTOCOL<SERIAL3_PROTOCOL>` is set to "23", by default, to enable this.
+
+- SBUS/DSM/SRXL connects to the RX3 pin but SBUS requires that the :ref:`SERIAL3_OPTIONS<SERIAL3_OPTIONS>` be set to "3".
```
andyp1per, line 96: This means invert both TX and RX, presumably you just mean invert RX here?

common/source/docs/common-flywoo-f745.rst
```diff
@@ -70,15 +91,30 @@ UART3 supports RX and TX DMA. UART1, UART2, UART4, and UART6 supports TX DMA. UA
...
 
-.. note:: If the receiver is SBUS, then :ref:`SERIAL3_OPTIONS<SERIAL3_OPTIONS>` must be set to "1" to provide inversion for detection. If the receiver is FPort, then the receiver must be tied to the USART3 TX pin and :ref:`SERIAL3_OPTIONS<SERIAL3_OPTIONS>` = 7 (invert TX/RX, half duplex) and :ref:`RSSI_TYPE<RSSI_TYPE>` =3.
+- FPort requires connection to TX3 and :ref:`SERIAL3_OPTIONS<SERIAL3_OPTIONS>` be set to "7".
```
andyp1per, line 98: Ditto here 7 means invert rx and tx and use half-duplex

common/source/docs/common-flywoo-f745.rst
```diff
@@ -70,15 +91,30 @@ UART3 supports RX and TX DMA. UART1, UART2, UART4, and UART6 supports TX DMA. UA
...
+
+The GOKU GN 745 AIO supports up to 8 PWM outputs. The pads for motor output ESC1 to ESC4 on the above diagram are the first 4 outputs, there are four additional pads for PWM 5-8. All 8 outputs support DShot as well as all PWM types.
+
+The Nano version has 4 esc signals, current and voltage sense inputs on an external connector. The remaining outputs are on solder pads.
+
+The PWM are in 5 groups: 1/2/7/8, 3/4, 5, 6
+
+Channels within the same group need to use the same output rate. If
+any channel in a group uses DShot then all channels in the group need
+to use DShot. PWM 1-4 support bidirectional dshot in the AIO version, only PWM2-4 support bidirectional dshot in the Nano version.
```
andyp1per, line 117: I don't think this is right - should be 1-4 on both boards

andyp1per (changes requested):

common/source/docs/common-flywoo-f745.rst
Hwurzburg: typo...thanks

common/source/docs/common-flywoo-f745.rst
Hwurzburg: doesnt matter....only RX is used......TX cannot be with those protocols.....it could be either 2 or 3 and work....

common/source/docs/common-flywoo-f745.rst
Hwurzburg: just the values Tridge has stated to use for FPort...its like that everywhere...maybe the the rx does not inversion....depends on how things connect internally when half duplex is used...it is the inversion before or after the internal path mux to the fifos

common/source/docs/common-flywoo-f745.rst
Hwurzburg: why is the hwdef different then for the AIO and Nano....thought the second I2C needed a DMA and it was taken from PWM1....but I am not sure I understand how the hwdef for BIDIR tags work

andyp1per: Just different resource allocations. I would have had to move something around to make it work, but I just checked - bi-dir on PWM 1-4 on the nano

Hwurzburg: Will change

PR #4473, opened 2022-07-24
Fix FFT documentation: FREQ=>REF

mbuzdalov:
An obvious bug has been introduced when updating the docs to include HNTC2 parameters.

I know at least one person who followed the new doc and got a mystically unflyable copter. Effectively, every input was filtered out with FREQ=1 and BW=120.

This fixes it.


Hwurzburg:
thanks for the catch...

PR #4482, opened 2022-08-03
Updates to Mission Planner Flight Data docs

TunaLobster:
Also includes some drive-by formatting and grammar checks.


TunaLobster:
Added the updated screenshots.

TunaLobster:
There is still a lot from the last few dev conferences and other features missing. I added the right-click the connect button options.

Hwurzburg (commented):
this is going to take a while to review and fix....the heirarchy is changed (perhaps valid, perhaps not) and a doc unlinked...the content updates are great, but need a few hours to walk this...wont be until this weekend

TunaLobster:
Fixed the missing table of contents item. toctree directive is a little weird. I think I might read up on that and see if I can use the glob command to make the ToC for the PDF build.

TunaLobster:
@magicrub I think you know more of these tricks that are missing of the docs. I know I am missing a few that were mentioned during conferences past. I can write up the docs and get the screenshots, but any suggestions on what is missing would be helpful.

Hwurzburg:
@TunaLobster sorry it took so long....made some fixes and reorg'd for better side menu presentation...thanks...merging

Hwurzburg:
LGTM...tests okay

PR #4496, opened 2022-08-11
Common: Change from NKF1 to NKF6

muramura:
I think NKF6.


Hwurzburg:
there is no NKF6

peterbarker:
Well, they're all wrong, unless you're still on EKF2 :-)

Hwurzburg:
I have updated that line for EKF3, but its correct for EKF2

PR #4497, opened 2022-08-12
Common: correct Gremsy upload instructions for 4.3

rmackay9:
This is a minor fixup to the Gremsy setup instructions for AP 4.3.

In short there is no need to unplug/plug-in the USB port as part of the firmware upgrade.

This has been tested on my local machine.


Hwurzburg:
LGTM

PR #4498, opened 2022-08-12
Copter: add script_time and attitude_time mission command descriptions

rmackay9:
This also makes some minor formatting changes to the top of the page.

BTW, I actually wanted to add short-cuts at the top of the page to the description of each command below but couldn't figure out how to do that.


Hwurzburg:
Wish these were added to the generic commands document...but I have about given up trying to get that document straightened out except for Plane

PR #4499, opened 2022-08-12
Update ADSB with new device protocol and ADSB_OPTIONS

Hwurzburg:


magicrub (approved):
I made a tweak, looks good. Thanks!

PR #4515, opened 2022-08-18
Common: Gremsy setup fixes

rmackay9:
This PR makes two changes:

- Removes the Pixy SM and adds T3 and T7 to the list of supported gimbals
- Adds RC9_OPTION parameter setting for "lock" vs "follow"

This has been tested on my local machine


Hwurzburg:
LGTM..thanks

PR #4519, opened 2022-08-19
update.py: Add opt-in --fast option to use environment caching for in…

TunaLobster:
…cremental builds

Downloading all of the youtube thumbnails can take a while. The sphinxcontrib-youtube extension does support using a cache of thumbnails, but that gets wiped out when doing `make clean`. This changes the build process to skip that command and sphinx picks up where it left off last time `make html` was called. I went ahead and added skip fetching parameters and log messages to the fast method as well.

Also fixed the file moving stacktrace that comes up for the backup moving. Should not impact the build server because `update.sh` uses the default value of `--destdir` and uses `--enablebackups`.

I'm thinking I should add this option to CI just for 1 site to test the logic.

### DIFF
--- a/common/source/docs/common-wiki-editing-setup.rst
+++ b/common/source/docs/common-wiki-editing-setup.rst
@@ -15,17 +15,17 @@ There are several methods to setup a local build environment:
 #. Or simply install the required libraries on your local Linux or Windows machine, using the instructions below. 
 
 Fork, Clone, and Update Your Copy of the Wiki
----------------------------------------------
+=============================================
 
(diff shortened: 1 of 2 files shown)


Hwurzburg:
@peterbarker do you think we can merge this safely? seems to work locally

TunaLobster:
Added the `update.py` build options to the docs and gave the page a small update on formatting and a grammar pass.

tridge (approved):

peterbarker:
We believe this killed the Wiki pretty hard - it was failing to create the required directory structure when moving things down into /var/sites/wiki/web

We played around with the concept of reverting this patch - but for now the cronjob on the wiki server is disabled, and we restored the Wiki from a backup.

PR #4520, opened 2022-08-20
Common: SToRM32 gimbal setup improvements

rmackay9:
This makes a few changes to the SToRM32 gimbal setup instructions:

1. clarifies which users should use the MAVLink vs Serial interface ([see discussion here](https://discuss.ardupilot.org/t/does-anyone-use-storm32-serial-gimbal-driver/89697))
2. adds hyperlinks to all parameters
3. replaces "pixhawk" with "autpilot"
4. other minor wording changes

this has been tested on my local machine

FYI @olliw42

### DIFF
--- a/common/source/docs/common-storm32-gimbal.rst
+++ b/common/source/docs/common-storm32-gimbal.rst
@@ -4,49 +4,35 @@
 STorM32 Gimbal Controller
 =========================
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
LGTM....thanks

PR #4523, opened 2022-08-20
Update Partners Logos

rishabsingh3003:
This fixes the minor issue with my last PR. I have a script to auto-generate the file but it does not include the mandatory "- .." at the end of an odd number logo. I have updated it now


Hwurzburg:
looks good now

Issue #3688, opened 2021-08-07
4.2 Wiki tasks (should be in upcoming features)


Hwurzburg:
- [ ] 17539 New GUIDED mode position controller(lthall)
- [ ] 17613 Ride Along JSON (iampete)
- [ ] 17651 ONIF camera in SITL/Linux
- [ ] 17668 Script backend for interpolated matrix (iampete)
- [ ] 176978 Land reposition in Q_OPTIONS (tridge)
- [ ] 17790 Airspeed disabled AUX switch (pbarker)
- [ ] 17894 FRSKY_OPTIONS parameter (yaapu)
- [ ] 17945 Matek F765-SE board(tridge)
- [ ] 17975 --udp option for sim_vehicle (tridge)
- [x] 17979 Torqeedo motor (rmakay9)
- [ ] 17986 CHUTE_OPTIONS (tridge)
- [ ] 17990 CRSF v3 (andypiper)
- [ ] 17992 USB passthru
- [ ] 17998 MAMBA Basic MK3.5 board (hw)
- [ ] 18018 PWM_BRD_COUNT removal (tridge)
- [ ] 18043 EXPO param for Plane manual/acrp/trainer modes
- [ ] 18133 switch to gcc 10.2.1 (tridge)
- [ ] 18137 LOG_FILE_RATEMAX with note on DSF.Dp (tridge)
- [ ] 18184 ROMFS_custom integration into ROMFS (tridge)

rmackay9:
"New GUIDED mode position controller" is included in 4.1 actually.  we should update this page https://ardupilot.org/dev/docs/mavlink-commands.html

HefnySco:
"18337 Webots2012b compatibility (@HefnySco)"
yes. Webots2021b is replacement of old one and the old files are not compatible with it.

drtrigon:
18929 Cam control Lua applet: pull request was accepted. Is there any other (administrative) thing to do? @IamPete1 can you instruct me what else needs to be done?

bnsgeyer:
@Hwurzburg Tradheli autotune (#14619) is expected to be merged for 4.2.  I am writing the wiki and it will be ready to merge directly.  I will work on the other two after I get autotune done

PR #4524, opened 2022-08-22
Add AP_Periph DShot and PWM example

TunaLobster:
Adds a page to hold examples of using AP_Periph for different use cases. First one up is the DShot and PWM example.

Covers one item in #3688.

### DIFF
--- a/common/source/docs/common-ap-periph-usage-examples.rst
+++ b/common/source/docs/common-ap-periph-usage-examples.rst
@@ -1,0 +1,87 @@
+.. _common-ap-periph-usage-examples:
+
+======================
+DroneCAN Adapter Setup
+======================
+
+These are several ways that a DroneCAN adapter node running AP_Periph can be used and this page is a small selection of the possibilities. These sections should walk you through setting up the autopilot and the node so that they work together.
+
+.. note:: DroneCAN Adapter nodes usually come with firmware that supports a subset of ArduPilot's large peripheral capability. Be sure to check that the firmware supplied supports the peripherals you are attemptime to setup. If not, check the `Firmware <https://firmware.ardupilot.org/AP_Periph/>`__  page for possible alternative firmware for that adapter that may have the capability.
+
+GPS Setup
+=========
+Attach the GPS to a DroneCAN Adapter's serial port
+
+- set :ref:`GPS_TYPE<GPS_TYPE>` = 9 on the autopilot
+- set the DroneCAN adapter's ``SERIALx_PROTOCOL`` on which the GPS is connected to 5 (GPS)
+
+Compass Setup
+=============
+Attach the Compass to the DroneCAN Adapter's external I2C port. It will appear as a normal Compass in ArduPilot and normal :ref:`compass setup procedures<common-compass-calibration-in-mission-planner>` would apply.
+
+.. note:: The MatekL431 CAN Adapter comes setup for a SPI connected Compass (RM3100). If an I2C Compass is attached, different firmware must be loaded for it to function.
+
+I2C Airspeed Sensor
+===================
+Attach to the I2C bus pinned out from the DroneCAN Adapter.
+
+- set :ref:`ARSPD_TYPE<ARSPD_TYPE>` =8 (DroneCAN)
+- set the DroneCAN's ``ARSPD_TYPE`` parameter to match the type of airspeed sensor connected.
+
+Rangefinder Setup
+=================
+To use rangefinders, follow the instructions at  :ref:`DroneCAN Setup Advanced<common-uavcan-setup-advanced>` to set up the ArduPilot parameters. Using MissionPlanner or DroneCAN Gui, set the parameters on the adaptor node following the instructions for the relevant rangefinder.
+
+.. note:: The orientation of the rangefinder (RNGFND1_ORIENT) must be set to 0 on the adaptor node.
+
+.. note:: The RNGFNDx_ADDR ArduPilot parameter must be set above 0 and be equal to the number set on the DroneCAN adapter node.
+
+PWM or DShot Output Node
+========================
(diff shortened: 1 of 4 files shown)


Hwurzburg (changes requested):
I think its good to have pages for specific nodes and their setup, like the autopilot pages....but this appears as if generic, when its not....it does not show how to create a specific node example like the MatekL431-Schot...its specific to the MatekL431-DShot, specifically..which is not in stable, by the way, but master....and should be an upcoming feature for 4.3.......

We could have a landing page for the existing AP_Periph implementations with setup instructions like this for specific firmware/boards linked from it....L431-DShot could be the first link from that page...but it needs to be like the autopilot hardware pages...

Alternatively or in addition to, we could have a tutorial on how to create a specific example, like the MatekL431_Dshot or MatekL431-Airspeed, etc.

so, this needs some work to fit into the hardware board page category....

just another cursory comment,  I am not a big fan of the contents link at the top having sub-bullets....perhaps this can be eliminated whenever its used...maybe just use plain links to tags below?

assuming this evolves into a landing page, with specific node pages linked like the MatekL431-Airspeed or Dshot, then a normal TOC would be the links to those

Hwurzburg:
Thinking more about this...we already have a page that points to specific implementations....https://ardupilot.org/plane/docs/common-uavcan-peripherals.html

we need to add a MatekL431-DShot page (as well as more or the new ones)....so if we take the content of the blog post and turn it into a page similar to https://ardupilot.org/plane/docs/common-matek-uavcan-dlvr.html base on what you already started...I think that would be the best approach...

we still could use a dev page on exactly how to create a new DroneCAN AP_Periph (since they all are DroneCAN periphs)

TunaLobster:
The generic presentation was purposeful. Having to create a new page for each of the roles for each generic node (ARK CAN node, mRo node, and Matekl431 just to name a few) can play is a lot of pages and would require a new page for each board when a new feature that works on all of them is added. The setup of the parameters for the roles is mostly agnostic to board just as ArduPlane parameters work between a Cube Orange and an Omnibus F4. The restriction would come when AP_Periph features grow to be too large to fit on the smaller nodes.

Each feature can have its own section on this page and this page linked to from each node board. I just picked the dev docs for the page because the parameters for nodes are also hosted there (even though I did not include :ref: directives to each parameter). I could add the AP_Periph parameters as a common page and have this page become a common page too. Then it could be linked to from the DroneCAN Peripherals landing page. I also placed it here because it is a advanced user use case to be setting up PWM/DShot nodes on an aircraft. It is also only a use case for larger aircraft where PWM signals can get wonky due to distance and radio noise.

The contents directive was used to have a page contents table that is automatically generated when new headings are added to the page. To me that would make reviewing and editing a little easier. Also having all of the capabilities described in one place is nice. I think one other use case for this directive would be the RC Options page. Right now a lot of the options are not documented beyond just a name. It generates many questions and is a common reply that an RC option does what a user wants. Having a contents at the top of the page for that page would also be useful IMO.

Hwurzburg:
"I think one other use case for this directive would be the RC Options page. Right now a lot of the options are not documented beyond just a name."
if any are not documented, please point them out...there is a detailed explanation of each one following the table...

TunaLobster:
> "I think one other use case for this directive would be the RC Options page. Right now a lot of the options are not documented beyond just a name." if any are not documented, please point them out...there is a detailed explanation of each one following the table...

Sorry. That was the servo functions table. I was trying to work out where I was going to add the scaled RC pass through details. I'll be looking at https://ardupilot.org/plane/docs/servo-functions.html and https://ardupilot.org/copter/docs/common-rcoutput-mapping.html and https://ardupilot.org/copter/docs/common-servo.html for that update. Which does have things missing and scattered and not using [xsite wiki] tags and not referenced to each other. Future PR though.

The contents tag along with a [sectnum](https://docutils.sourceforge.io/docs/ref/rst/directives.html#automatic-section-numbering) directive might make the long table pages more navigable. Right now using crtl+f is the best option to find the needed information and that's just not a great UX design to rely on a a lot of scrolling or a browser feature to use the documentation.

Hwurzburg:
the common-rcoutput-mapping should be the central spot for ALL servo/motor functions......for all vehicles.....like your scaled pass through that will be in Future wiki changes until we beta 4.3...its probably missing a few like scripting outputs, and the RCMAP replacements.....any help in updating those would be appreciated...

...the individual vehicle pages in setup, should have the primary vehicle functions like planes servo/motor functions...of which their are many....copter is mostly motors and is covered in its motors setup page...rover has a motor and servo page for its vehicle setup...but the rcoutput-mapping page should have every output type listed...

....common-servo is a peripherals page and very old

Hwurzburg:
@TunaLobster just a follow up....I went thru the rcoutput mapping page, and every output except for the 2 WingSail and  2 Alarm outputs are documented...and those are waiting for wiki inputs from the authors....so 4 pins out of 139...

Hwurzburg:
@TunaLobster fyi, In May I had added the Matek L431 Dshot reference and link to Tridge's discuss post on his page https://ardupilot.org/plane/docs/common-uavcan-adapter-node.html..... but I think a tutorial on how someone actually goes about creating a specific AP_Perph device and hwdef would be really a good addition to the dev wiki...

TunaLobster:
I'll move this info to a common page then. It is still generic to each of the nodes that have outputs set up on headers like the Matek L431. Might be worth while to split the list of supported DroneCAN devices into generic and purpose built sensor categories.

Hwurzburg:
There are already split

Hwurzburg:
expanded, made nomenclature more consistent....will merge if your okay with it....

TunaLobster:
LGTM! Fixed a few punctuation things.

PR #4525, opened 2022-08-22
Add CUAV-Nora UART3 to serial list

TunaLobster:
Also a few drive by small changes for unicode to ascii characters.

Covers one item in #3688.


Hwurzburg:
LGTM..thanks

PR #4531, opened 2022-08-23
Handling mode change in failsafe period

Target0815:
So far it is not made clear that a mode change during the failsafe period (RC aborts, failsafe is active, user changes mode at the transmitter) results in this mode being activated directly after the connection is restored.


Hwurzburg (changes requested):
will fix it and merge

PR #4534, opened 2022-08-24
Airmode Plane: Add node about no mavlink disarm in airmode

TunaLobster:
Checks off one item in #3688.

### DIFF
--- a/plane/source/docs/airmode.rst
+++ b/plane/source/docs/airmode.rst
@@ -6,18 +6,20 @@ AirMode
 
 .. note:: this feature will be available in Plane 4.1 firmware and later.
 
-This is not an actual flight mode, but rather, an important feature of QACRO and QSTABILIZE modes. Other flight control software refer to it as a flight mode, so for consistency, its included in the QuadPlane flight modes documentation.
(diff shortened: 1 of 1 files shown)


Hwurzburg:
@TunaLobster sorry this is incorrect, but thanks for the attempt to help clear the 4.2 list!....the actual change was to disable MAVLink commanded disarms while flying in any mode...its added to the code which disallows a rudder disarm from being commanded while is_flying is active....this is really a bug fix....probably dont need anything in the wiki...it already states that the MP disarm button can only be used on the ground...

TunaLobster:
Just trying to check things off and get caught up before things get to be 2 release versions old.

Hwurzburg:
thanks for that! the big things missing are the board pages....I already have done the ones that are popular (CUAV 6x is still being verified and changed...they are sending a final sample to me soon).....a few of AndyPipers boards are still not done yet....they take a lot of time...you have to check that the readme is REALLY correct, checking against the hwdef as you construct the pretty page for it.... and fill in the missing holes...I try to do one a week...it takes about 2-3 hours of careful back and forth checking...

PR #4535, opened 2022-08-25
Dev: mavlink interface for gimbal control

rmackay9:
This adds a new gimbal page to the MAVLink interface section of our development wiki.

I've tested the contents and rendering fairly carefully on my machine so hopefully it is ok.


Hwurzburg:
LGTM...thanks

PR #4536, opened 2022-08-25
dev: SITL for multivehicles should use mcast

rmackay9:
This slightly modifies the Using SITL page so that it adds the "--mcast" to the command line when starting multiple vehicles.

I've tested this on my local machine.

FYI @tridge

### DIFF
--- a/dev/source/docs/using-sitl-for-ardupilot-testing.rst
+++ b/dev/source/docs/using-sitl-for-ardupilot-testing.rst
@@ -797,11 +797,11 @@ Multi-Aircraft SITL with MAVProxy
 
 Multiple aircraft can be simulated in SITL with MAVProxy. This feature owes much to prior work here: https://ardupilot.org/mavproxy/docs/getting_started/multi.html 
 
-To access this feature use both the number (-n) or count (--count) as well as the auto-sysid (--auto-syid) features of sim_vehicle.py like so:
+To access this feature use both the number (-n) or count (--count) as well as the auto-sysid (--auto-sysid) and multi-cast (--mcast) features of sim_vehicle.py like so:
 
 ::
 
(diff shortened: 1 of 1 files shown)


dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -797,11 +797,11 @@ Multi-Aircraft SITL with MAVProxy
 
 Multiple aircraft can be simulated in SITL with MAVProxy. This feature owes much to prior work here: https://ardupilot.org/mavproxy/docs/getting_started/multi.html 
 
-To access this feature use both the number (-n) or count (--count) as well as the auto-sysid (--auto-syid) features of sim_vehicle.py like so:
+To access this feature use both the number (-n) or count (--count) as well as the auto-sysid (--auto-syid) and multi-cast features of sim_vehicle.py like so:
```
khancyr, line 800: auto-sysid

khancyr (approved):
yes, I haven't tested but surely need the --mcast to work with default config
should probably need to be add to the other lines too

dev/source/docs/using-sitl-for-ardupilot-testing.rst
rmackay9: @khancyr txs for reviewing.  I think "auto-sysid" is already mentioned so I think this is OK.

khancyr: ha , it is just that `--auto-syid` is written not `--auto-sysid` : miss a s !

rmackay9: oh!  existing problem but fixed, txs!

rmackay9:
@khancyr, txs for the review and feedback.  I've added the "--mcast" to the other line as well. txs!

PR #4538, opened 2022-08-25
dev: update WSL2 on Windows11 instructions (WIP)

rmackay9:
I setup my WSL2 environment today for the first time in a long time and bumped into various problems.  One of them seemed to be that with Ubuntu 22.04 at least, it fails with a "something went wrong" message and I was forced to google around a bit to discover that "wsl --update" followed by "wsl --shutdown" was required to complete the installation.

The installation method seems to vary depending upon the Ubuntu version.  I tried 20.04 and it did not suffer from this problem so I slightly worry that we will be updating this page again when 22.x or 23.x is release but I guess this is better for now anyway.


rmackay9:
Let's delay the merge of this while I test it more.

Hwurzburg:
Many will desire to use another distribution on Ubuntu....I dont think we should force them into the latest....18.06 is still supported for example...I think we need to keep it generic re Ubuntut distro....add notes if certain ones need special commands to work

dev/source/docs/building-setup-windows10.rst
```diff
@@ -24,12 +24,12 @@ The official instructions are `here <https://docs.microsoft.com/en-us/windows/ws
     .. image:: ../images/build-on-windows10-wsl-install.png
         :target: ../_images/build-on-windows10-wsl-install.png
 
-#. From a web browser open `https://aka.ms/wslstore <https://aka.ms/wslstore>`_ which should open the Microsoft store to allow installing Ubuntu. Then Launch Ubuntu and fill in a username and login
+#. From a web browser open `https://aka.ms/wslstore <https://aka.ms/wslstore>`_ which should open the Microsoft store to allow installing and launching Ubuntu 22.04.  If the launch fails it may be necessary to open a command prompt (as administrator) and run "wsl --update" and "wsl --shutdown" before launching Ubuntu 22.04 again and then following the setup steps.
```
hendjoshsr71, line 27: I'm a bit surprised the order isn't `wsl --shutdown` then `wsl --update`. 

Also updates to wsl are now served through windows update itself.  (So I think that is how I had my client do an update at some point.)

Either way the `wsl --update` should achieve the same thing.

hendjoshsr71:
> Many will desire to use another distribution on Ubuntu....I dont think we should force them into the latest....18.06 is still supported for example...I think we need to keep it generic re Ubuntut distro....add notes if certain ones need special commands to work

They still can select a different version even if our directions say 22.04. But I do think our directions should in general point to the newest version. I got asked the exact question of which version to select on discord the other day.

Hwurzburg:
Sorry I disagree....those of us that resist upgrading OS would be led to believe it only works with Ubuntu22....which is totally wrong...the instructions can use 22 as an example, but must clearly point out that any supported distro (18.06LTS and beyond at this point) can be used

khancyr:
Support for 18.04 won't be eternal, maintening python 2 become harder as time pass. If you try today, the support for 18.04 is broken...
So recommending at least ubuntu 20.04 for newcomers should be the way IMHO.

dev/source/docs/building-setup-windows10.rst
```diff
@@ -24,12 +24,12 @@ The official instructions are `here <https://docs.microsoft.com/en-us/windows/ws
...
 
     .. image:: ../images/build-on-windows10-wsl-install-part2.png
         :target: ../_images/build-on-windows10-wsl-install-part2.png
 
-#. From the Start menu, start the "Ubuntu" application and then follow the :ref:`Ubuntu instructions to install ArduPilot development environment <building-setup-linux>`
+#. From the Start menu, start the "Ubuntu 22.04" application and then follow the :ref:`Ubuntu instructions to install ArduPilot development environment <building-setup-linux>`
 
     .. note::
```
hamishwillee, line 34: Below it says "Compilation speeds will be much faster . True. Just as important are potential permission issues.

hamishwillee: I can't link to it, but below it says "#. Install an XWindows application to run graphical programs such as SITL by installing `VcXs". You don't need to do this on Windows 11 FYI.

TunaLobster: #4398 added a note about the xserver difference between 10 and 11. Should be able to remove this step or just reference the later section.

rmackay9: Yes, I've also found that I don't need to run VcXs anymore.

rmackay9:
@Hwurzburg,

Even if we recommend a particular version I don't think that means we're dropping support necessarily for older versions... we're just giving advice for new installs.

In any case, the "default" (e.g. "Ubuntu" with no version specified) seems to install the latest version 22.04 so we all get what we want I think.

rmackay9:
I'm seeing various display issues with this latest WSL2 / Ubuntu 22.04.  For example MAVProxy's WP editor is now unusable because I can't scroll down in the list of available mission commands.  I only get to see the top 30 or so which isn't nearly enough to even get to NAV_TAKEOFF.

hendjoshsr71:
> I also haven't figured out how to allow MP (running natively on the same Windows box) to connect to SITL running under WSL2/Ubuntu.

The normal TCP at 127.0.0.1:5673 (or 5761, 5762) didn't work? I usally do that rather than bother adding an output for sitl. Though when connected to aircraft remotely I do add an output. 

Also I've added to my `.bashrc` the following to allow for easier forwarding out

```bash 
# Export WSL2 IP Address & Display for XWindows 
# For WSL2
export WSL2_IP_ADDR=$(grep -m 1 nameserver /etc/resolv.conf | awk '{print $2}')
export DISPLAY=$WSL2_IP_ADDR:0
```

`sim_vehicle.py -f flightaxis:$WSL2_IP_ADDR --map --console --debug`

Though my WIn10 WSL2 has a large number of dropouts with RealFlight currently.... :( (some sort of networking thing I think as my graphics card can do modern games at high quality and frame rates)

dev/source/docs/building-setup-windows10.rst
hamishwillee: In theory you this is might come to Windows 10. if you can run simulations headless then you don't need an Xwindow app.

Hwurzburg:
@rmackay9 what is needed to move on this?

adalta87:
Where does this differ from using windows 11 and downloading the ubuntu from the Microsoft store. I have just installed WSL twice on a new windows 11 laptop and had no issues at all getting things up and running other than step 4 of current instructions are not needed of on windows 11. It was a pretty straight forward install of ubuntu 20.04.5 and clone of the code into its file system. I was able to build code with no issues.

rmackay9:
@adalta87,

The two issues that I've found are:

1. there are multiple versions of Ubuntu available in the store.  Which one should users choose?
2. the method of connecting MP (or other GCS) to the SITL instance running under WSL has changed.  MP use to connect automatically but now the "output add xxxx" is required.

PR #4543, opened 2022-08-27
Re-org motor/esc page, add EFI page

Hwurzburg:


tridge:
worth adding:
 - how to display key info in GCS, eg. cylinder head temperature in HUD
 - setup RPM_TYPE for EFI RPM
 - setup notch filter based on RPM

Hwurzburg:
Added content as requested....merging

PR #4552, opened 2022-08-29
Fix characters before ref tags causing issues and add step to Linux CI job

TunaLobster:
Also one or two grammar things I caught on a quick glance. I also realized there are there are trailing spaces after refs sometimes. Doesn't break the creation of the hyperlink though. Just looks odd.

### DIFF
--- a/.github/workflows/test-Windows.yml
+++ b/.github/workflows/test-Windows.yml
@@ -8,7 +8,7 @@ env:
   TEST: "1"
 jobs:
 
-  build:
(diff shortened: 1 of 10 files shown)


Hwurzburg:
@khancyr can you review the .yml changes please...thanks

khancyr (approved):
tested and working
can merge @Hwurzburg !
The build name change just change the CI job name, and that is good.
Another grep regex is used
rest is just corrections

PR #4565, opened 2022-09-02
Move Mission Planner specific items from common to MP wiki

TunaLobster:
Moves the SLCAN page to the MP Initial Setup page in the planner wiki and changes the references to use intersphinx. Also does the same thing for the telemetry logs page. Both pages are specific to Mission Planner. Build warning are coming from the intersphinx inventories on the server being different than what is used in this change.


TunaLobster:
Caught a few build errors. This will require a double build on the server for the updated inventory tags.

Hwurzburg:
going to merge this and recheck for inventory errors tomorrow morning...thanks

PR #4573, opened 2022-09-07
Fix serial port number in telemetry section of Holybro KakuteF4 Mini

silentjet:
Subj. SP soldering pad is connected to UART1/Serial1 but not to Serial2

### DIFF
--- a/common/source/docs/common-holybro-kakutef4-mini.rst
+++ b/common/source/docs/common-holybro-kakutef4-mini.rst
@@ -94,8 +94,8 @@ to set the following parameters to enable support for FrSky S.PORT. It
 has built-in inverters and a diode to allow for operation from a single
 pin with no special adapters.
 
-  - :ref:`SERIAL2_PROTOCOL<SERIAL2_PROTOCOL>` 4 or 10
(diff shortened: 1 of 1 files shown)


common/source/docs/common-holybro-kakutef4-mini.rst
```diff
@@ -94,8 +94,8 @@ to set the following parameters to enable support for FrSky S.PORT. It
 has built-in inverters and a diode to allow for operation from a single
 pin with no special adapters.
 
-  - :ref:`SERIAL2_PROTOCOL<SERIAL2_PROTOCOL>` 4 or 10
-  - :ref:`SERIAL2_OPTIONS<SERIAL2_OPTIONS>` 0
+  - :ref:`SERIAL1_PROTOCOL<SERIAL2_PROTOCOL>` 4 or 10
```
Hwurzburg, line 97: good catch, but you have to change the number within the <   > also...since you are editing on the web, I will do this after I merge it...

silentjet:
Thank you very much, and you're right I did it from a smartphone ^.^

PR #4575, opened 2022-09-07
fix slight errors in ais docs

Autonomost:
as of 9/6/22 the docs were slightly outdated. corrected small discrepancies.


Hwurzburg:
thanks....

PR #4583, opened 2022-09-09
make backups and wiki runs safer

tridge:
- check that each wiki did build ok
- check for key python version and modules

### DIFF
--- a/update.py
+++ b/update.py
@@ -232,6 +232,9 @@ def check_build(site):
     """
     check that build was successful
     """
+    if platform.system() == "Windows":
+        print("Skipping check_build on windows")
+        return
     for wiki in ALL_WIKIS:
         if wiki in ['common', 'frontend']:
             continue
--- a/update.py
+++ b/update.py
@@ -104,6 +104,12 @@ def error(str_to_print):
     print("[update.py][error]: " + str(str_to_print))
 
 
+def fatal(str_to_print):
+    """Show and count the errors."""
+    error(str_to_print)
+    sys.exit(1)
+
+
 def remove_if_exists(filepath):
     try:
         os.remove(filepath)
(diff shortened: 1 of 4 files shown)


update.py
```diff
@@ -782,7 +807,7 @@ def copy_static_html_sites(site, destdir):
     )
     parser.add_argument(
         '--destdir',
-        default="/var/sites/wiki/web" if platform.system() != "Windows" else os.path.join(os.path.dirname(__file__), "..", "wiki"),  # noqa: E501
+        default=None,
```
peterbarker, line 810: Are we supplying this on the server?

tridge: yes, see change to update.sh

update.py
```diff
@@ -750,8 +751,29 @@ def copy_static_html_sites(site, destdir):
             pass
 
 
-#######################################################################
+def check_imports():
+    '''check key imports work'''
+    import pkg_resources
+    requires = ["sphinx_rtd_theme>=0.1.8", "sphinxcontrib.vimeo>=0.0.1"]
```
TunaLobster, line 757: sphinxcontrib.vimeo does not exist. sphinxcontrib-vimeo does. However, neither are installed by the setup script any more. See #4195.

TunaLobster:
Checking for the python modules shouldn't be necessary. Sphinx fails to attempt a build if a module isn't present. The real trick is modules with the same name but different behavior. There have been multiple sphinxcontrib-youtube extensions through the years. Checking the version and other details of the module installed in the current environment is more important. But in order to do that we need to lock down the versions used in the python environment. Future PR though.

tridge:
> Checking for the python modules shouldn't be necessary. Sphinx fails to attempt a build if a module isn't present.

its a lot less confusing to give a clear error at startup of update.py than fail deep in the build after thousands of other lines of output

tridge:
> The real trick is modules with the same name but different behavior

I've changed the install to use ArduPilot forks of the modules

PR #4586, opened 2022-09-09
Pin down python package versions and fix build errors

TunaLobster:
There are a few thing going on here. This is to get local environments functional again after the changes yesterday. This also makes it safer to run Sphinxsetup.sh on the build server because more of the package versions match what is used locally and in CI. There might still be issues, but at least the big name ones are solved by this PR.

- Pin down sphinx to a known working version and so that the build server will match local.
- Pin down docutils so that bullet points will be rendered when using the older sphinx rtd theme. See https://stackoverflow.com/a/68685753/2578171 for details.
- Fix a build error due to destdir being None by default now.
- Add the correct package name for sphinxcontrib-youtube
- Add version checking for docuilts, sphinxcontrib-youtube, and sphinx
- Add a message at the end of the build to convey that there were no errors during the build.

### DIFF
--- a/Sphinxsetup.bat
+++ b/Sphinxsetup.bat
@@ -2,7 +2,7 @@ rem remove any existing packages that may cause conflicts
 pip uninstall -y sphinx lxml sphinx-rtd-theme sphinxcontrib-youtube beautifulsoup4
 
 rem Install sphinx
-pip install --upgrade sphinx
+pip install --upgrade sphinx==5.1.1 docutils==0.16
 
 rem lxml for parameter parsing:
 pip install --upgrade lxml
@@ -12,7 +12,6 @@ pip install --upgrade git+https://github.com/ArduPilot/sphinx_rtd_theme.git
 
 rem and a youtube plugin:
 pip install --upgrade git+https://github.com/ArduPilot/sphinxcontrib-youtube.git
-pip install --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
 
 rem and a parser to use getting posts from Discourse (forum) and insert in FrontEnd
 pip install --upgrade beautifulsoup4
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -48,8 +48,11 @@ curl "$GET_PIP_URL" -o get-pip.py
 python3 get-pip.py
 rm -f get-pip.py
 
-# Install sphinx
-python3 -m pip install --user --upgrade sphinx
+# Install python packages using known working versions
+# Install sphinx with a specific docutils version
+# Docutils version is for correct bullet point rendering. Can be rolled forward after theme is updated to >=0.5.1
+# See https://stackoverflow.com/a/68685753/2578171
+python3 -m pip install --user --upgrade sphinx==5.1.1 docutils==0.16
 
 # lxml for parameter parsing:
 python3 -m pip install --user --upgrade lxml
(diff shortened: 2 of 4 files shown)


Sphinxsetup.bat
```diff
@@ -12,7 +12,6 @@ pip install --upgrade git+https://github.com/ArduPilot/sphinx_rtd_theme.git
 
 rem and a youtube plugin:
 pip install --upgrade git+https://github.com/ArduPilot/sphinxcontrib-youtube.git
-pip install --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
```
tridge, line 15: why remove the install of the vimeo plugin?

Sphinxsetup.sh
```diff
@@ -58,8 +61,9 @@ python3 -m pip install --user --upgrade lxml
 python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinx_rtd_theme.git
 
 # and youtube and video plugins:
+# This command might require a --force option if you have and older extension installed
+# Rerun Sphinxsetup.sh after doing that
 python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib-youtube.git
-python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib.vimeo.git
```
tridge, line 62: why remove vimeo?

update.py
```diff
@@ -756,22 +755,15 @@ def copy_static_html_sites(site, destdir):
 def check_imports():
     '''check key imports work'''
     import pkg_resources
-    requires = ["sphinx_rtd_theme>=0.1.8", "sphinxcontrib.vimeo>=0.0.1"]
+    # package names to check the versions of. Note that these can be different than the string used to import the package
+    requires = ["sphinx_rtd_theme>=0.1.8", "sphinxcontrib.youtube>=1.2.0", "sphinx==5.1.1", "docutils==0.16"]
```
tridge, line 759: why no vimeo?

tridge (changes requested):

Sphinxsetup.bat
TunaLobster: It doesn't support the latexpdf builder. The sphinxcontrib.youtube extension does support Vimeo and the latexpdf builder. It's one of the steps towards getting a PDF version of the docs.

See also #4195 where it was removed. The version checking added in that PR can be removed now.

tridge (approved):

update.py
rishabsingh3003, line 759: @TunaLobster the youtube check is making it impossible for me to run the script: 
image

I think it might be an Ubuntu22 thing. I have made sure I have all the required packages. Any ideas what this might be? Currently just removing that check makes the script work like normal

TunaLobster: You will have to force install sphinxcontrib.youtube from the AP repo. `python3 -m pip install --user --force --upgrade git+https://github.com/ArduPilot/sphinxcontrib-youtube.git` Then rerun Sphinxsetup.sh to get everything else set to the correct versions.

rishabsingh3003: @TunaLobster worked like magic. Thank you!!

TunaLobster: I am trying to figure out a way to avoid the `--force` option. I think one way is to bump the version in the AP repo by a lot. I think that might be the least intrusive way to do it. I can't seem to find examples of telling pip to install forked packages from git over pypi packages that are already installed.

PR #4588, opened 2022-09-09
Fix build warning for common-dock-mode

TunaLobster:
Dock mode is only included in the rover wiki toctree right now. Changed the copy destination to match that. Was that page meant to be in the rover wiki and not in the common files?


Hwurzburg:
thanks...got rushed when checking and test building (retested only Rover)and did not notice the file was in the wrong directory....rather than asking you to change this, I just fixed it...thanks for catching

PR #4590, opened 2022-09-10
fix build errors from MP re-org

Hwurzburg:


TunaLobster:
Oops. A few of those got mixed up when I did a rebase. Thanks!

PR #4597, opened 2022-09-12
mavproxy: fix appending path to bashrc

rsaxvc:
When adding ~/.local/bin to the path in .bashrc,
bash will first expand $PATH in the following:
    echo "export PATH=$PATH:$HOME/.local/bin" >> ~/.bashrc
to something like:
    export PATH="/usr/bin:/bin:/home/user/.path/to/remove/later:/home/user/.local/bin"
which nullifies future modifications to PATH made earlier in .bashrc, which can be confusing.

### DIFF
--- a/mavproxy/source/docs/getting_started/download_and_installation.rst
+++ b/mavproxy/source/docs/getting_started/download_and_installation.rst
@@ -43,7 +43,7 @@ For Python 3 on Debian based systems (including Ubuntu, WSL, Raspian):
 
     sudo apt-get install python3-dev python3-opencv python3-wxgtk4.0 python3-pip python3-matplotlib python3-lxml python3-pygame
     pip3 install PyYAML mavproxy --user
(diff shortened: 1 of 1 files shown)


khancyr:
Looks right, I wonder why those lines were added in first place as we shouldn't need them on a standard ubuntu install

rsaxvc:
I'm running Debian 11, and did need to append it to my path.

It's functional as is, but creates a slight maintenance annoyance over time.

peterbarker (approved):
Change is correct, but this does need to pass CI before merge.

peterbarker:
@rsaxvc I've pinged somebody who might know instantly how to fix the CI tests failing here.  Stand by :-)

Hwurzburg:
Will need to run Sphinxsetup.sh loally,rebate and push back up...then Linux ci will pass...ignore Windows...it's broken right now

rsaxvc:
I have done so and the Linux build is now green.

PR #4602, opened 2022-09-13
Move ref directive regex check from CI to update.py

TunaLobster:
Move the regex for the ref directive formatting from a Linux only command to Python in update.py. This is working on Windows and Linux in CI.


Hwurzburg:
so now on local and server builder, it will flag but not stop building, correct?

TunaLobster:
Correct. It is not fatal, but you will see a none zero exit code at the end.

peterbarker (approved):

PR #4614, opened 2022-09-14
add FFT averaging info

Hwurzburg:


andyp1per (approved):
Probably could sat that values from 2-5 would be appropriate

PR #4618, opened 2022-09-15
removed windows CI test

tridge:
this is failing, and we don't think anyone actually does do doc builds on windows, they use WSL2


Hwurzburg (approved):

khancyr:
There are also vagrant and docker solution that are working on windows, so not supporting it explicitly should be a big lost

TunaLobster:
I fixed it in #4609.

tridge:
@TunaLobster thanks, I've merged yours and will close this

PR #4621, opened 2022-09-16
copter: add note on ATC_ANG_YAW_P and ACRO_YAW_P for older releases

peterbarker:


Hwurzburg:
really should delete the ones before 4.1 also

PR #4627, opened 2022-09-17
Update copter for new RATE_TC params

bnsgeyer:
This PR updates the wiki for the new parameters introduced in Copter 4.3.  These parameters adjust the time constant for achieving steady state rate for axes that are rate command for a given mode.  
PILOT_Y_RATE_TC - yaw rate time constant for all  modes except acro
ACRO_RP_RATE_TC - acro roll/pitch rate time constant
ACRO_Y_RATE_TC  - acro yaw rate time constant

I also updated the Input shaping wiki for the max rate parameters which were updated in 4.1 ( i think).

### DIFF
--- a/copter/source/docs/input-shaping.rst
+++ b/copter/source/docs/input-shaping.rst
@@ -8,12 +8,15 @@ Copter has a set of parameters that define the way the aircraft feels to fly. Th
 
 The most important of these parameters is:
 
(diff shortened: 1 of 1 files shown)


bnsgeyer:
@Hwurzburg this also probably needs to be changed for quadplane.  I think @IamPete1 added these into quadplane

IamPete1:
Yes, we have it in quadplane also now. Not quite so complicated as we only have the pilot yaw params and not the acro ones.

Hwurzburg:
thanks

PR #4633, opened 2022-09-20
update.py: cope with mismatch between build_options.py and features.json

peterbarker:


TunaLobster:
I wouldn't be opposed to adding E741 to the .flake8 file.

peterbarker:
> I wouldn't be opposed to adding E741 to the .flake8 file.

Bit of a silly check in this case.  I expect it would help if you were to be assigning the variable into other things, however.

I've made the thing a bit more descriptive.

Hwurzburg:
@peterbarker can I merge this?

peterbarker:
> @peterbarker can I merge this?

I was waiting for you to do so :-)

I've merged it now.

PR #4647, opened 2022-09-26
update heli param page with 4.3 params

bnsgeyer:
This updates the trad heli Params page with the new Params in Copter 4.3.0


Hwurzburg (changes requested):
we should add note beside no longer used params as to which firmware version they apply to

bnsgeyer:
Added applicable version for params no longer used.

Hwurzburg:
added version note on param missed,squashed

PR #4653, opened 2022-09-28
add tkoff_rpm_min param info

Hwurzburg:


rmackay9 (approved):
Looks ok to me although I'm not sure this is the best place for a description of this feature.  It's really more of a failsafe I think so maybe we should create a new page on the copter failsafe landing page.

rmackay9 (approved):
Great, thanks very much!

PR #4666, opened 2022-10-13
Common: add SIYI ZR10 gimbal page

rmackay9:
This adds a basic setup page for the SIYI ZR10 gimbal and adds it both to the existing cameras-and-gimbals landing page (with a warning that it is only supported in 4.4 and above) and the upcoming features.  The reason I've added it to the cameras-and-gimbals landing page is that I worry that users will not be able to find this important enhancement if it is only hidden away on the upcoming features page.


Hwurzburg:
LGTM...thanks

PR #4675, opened 2022-10-19
Add warning re RC rx impacts when upgrading to bdshot

Hwurzburg:
Andy, any updates to the list of inherently bdshot enabled boards?

### DIFF
--- a/common/source/docs/common-blheli32-passthru.rst
+++ b/common/source/docs/common-blheli32-passthru.rst
@@ -143,6 +143,8 @@ Newer versions of BLHeli32 (32.7 and higher) and BLHeli_S (16.73 and higher) sup
 
 Only the lowest 4 DShot capable servo outputs can be used.  For autopilots with an IOMCU (e.g. Pixhawk, CubeOrange) this means AUX1 to AUX4 can be used.  For Pixracer and other autopilots without a separate IOMCU coprocessor this means outputs 1 to 4 can be used.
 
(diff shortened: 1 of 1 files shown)


common/source/docs/common-blheli32-passthru.rst
```diff
@@ -143,6 +143,8 @@ Newer versions of BLHeli32 (32.7 and higher) and BLHeli_S (16.73 and higher) sup
 
 Only the lowest 4 DShot capable servo outputs can be used.  For autopilots with an IOMCU (e.g. Pixhawk, CubeOrange) this means AUX1 to AUX4 can be used.  For Pixracer and other autopilots without a separate IOMCU coprocessor this means outputs 1 to 4 can be used.
 
+.. warning:: on any autopilot using NON Bi-directional firmware and then the "-bdshot" version is loaded, if the an RC receiver using PPM, SBUS, IBus, DSM, or SRXL has been connected to an RC in using a timer interrupt, rather than a configuration using a UART with ``SERIALx_PROTOCOL`` = 23, the RC receiver will no longer function since that timer based RC input function has been eliminated to free up more DMA for Bi-directional DShot. The receiver will need to be moved to a UART or, if that RC input has an alternate configuration as a true UART input, the :ref:`BRD_ALT_CONFIG<BRD_ALT_CONFIG>` will need to be changed and that UART's ``SERIALx_PROTOCOL`` changed to "23".
```
andyp1per, line 146: I don't think this is correct. On the majority of bdshot boards the UART protocol is set correctly and uses the same pins as rcout so the switchover should be seamless. The only reason this would not be the case would be if you were using PPM

andyp1per (changes requested):

common/source/docs/common-blheli32-passthru.rst
Hwurzburg: sorry but you are mistaken...I have not checked them all but the MAtekF765-Wing and MatekF405-Wing, if configured with DSM,SBUS,PPM,IBUS, or SRXL will fail immediately upon loading the bdshot firmware and I am sure there are others...but those first two I checked will definitely fail

Hwurzburg: checking all the variations, also the KakuteH7 will have the problem

Hwurzburg: it appears that only the MatekH743 variants have this issue, so closing this and adding a defaults.parm for it...thinking that all RCIN timers were being reused in bshot led me to the wrong analysis...doing a default for this one board that has the issue is probably the best approach

PR #4676, opened 2022-10-19
Add joytsick cal info and warning

### LINKED ISSUE #4667: Joystick page improvement suggestions
At the recent [Japan Innovation Challenge](https://japan-innovation-challenge.com/en/) my team used Joysticks extensively and found a few problems that are not covered in [the documentation](https://ardupilot.org/copter/docs/common-joystick.html).  Below are some suggested improvements

1. users should calibrate the joystick using Window's Game Controller's screen
image

2. remove the bottom part of the note, "This error is caused because Mission Planner maps the Joystick exactly to the 1100 - 1900 range, but the pre-arm checks assume that if the values are not at least 1 pwm off the century value, that calibration has not been done." because this is no longer true.  Mission Planner uses the RC1_MIN/MAX, etc ranges as an input when calculating what values to send in the RC_OVERRIDES commands to the vehicle

3. We should mention that users should not perform an RC calibration or at least they should not do it repeatedly.  The problem is related to step 2 above where MP uses the RCx_MIN/MAX values as **inputs** when calculating what RC_OVERRIDE values to send.  Normally after step 1 (Windows level joystick calibration) the joystick still does not provide outputs that are in the full range (0 to 100?  0 to 1?).  This causes MP to send RC_OVERRIDE values that in a narrow range than RCx_MIN/MAX.  If the user does an MP RC calibration then RCx_MIN/MAX are changed to this narrower range.  Somewhat circularly MP's RC_OVERRIDES will again become more narrow.  This means each time the RC calibration is done the output range narrows.

4. we should add a note that if users find they are getting a very narrow range of RC Inputs from the joystick (they could check by looking at MP's Setup >> Mandatory Hardware >> Radio Calibration page) they should perform step1 (calibrate in Windows).  If this does not resolve the problem, the user may wish to manually modify RCx_MIN/MAX values to widen the range.

FYI @meee1

Hwurzburg:
closes #4667


rmackay9 (approved):
Thanks!

PR #4678, opened 2022-10-20
Common: minor mount updates

rmackay9:
This makes two small changes in the Cameras and Gimbals section:

- Clarify that the new SIYI A8 is supported with the same driver as the Z10
- Fixup the SToRM32 image of user parameters to match 4.3

This has been tested on my local machine and seems fine.


Hwurzburg:
thanks

PR #4694, opened 2022-10-31
Add clarification to HNTCH_FREQ for RPM based

Gone4Dirt:
Added a few extra words to make it clearer that the INS_HNTCH_FREQ is a lower limit constrain.


Hwurzburg:
LGTM..thanks

PR #4703, opened 2022-11-05
update.py: remove warning about feature not in build_options.py

peterbarker:
the features extract script is moving useful past its use on the Wiki page.

With this patch we won't include the feature in the Wiki if we can't find it in the build_options.py (just like before the patch, but with less complaining).

### DIFF
--- a/update.py
+++ b/update.py
@@ -866,6 +866,7 @@ def create_features_page(features, build_options_by_define, vehicletype):
         sorted_platform_features_in = []
         sorted_platform_features_not_in = []
         features_in = {}
+        feature_not_in_build_options = set()
         for feature in platform_features:
             feature_in = not feature.startswith("!")
             if not feature_in:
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -875,7 +875,7 @@ def create_features_page(features, build_options_by_define, vehicletype):
                 build_options = build_options_by_define[feature]
             except KeyError:
                 # mismatch between build_options.py and features.json
-                print("feature %s not in build_options.py" % feature)
+                # print("feature %s not in build_options.py" % feature)
```
tridge, line 878: use a set() to not emit same warning lots of times

tridge (changes requested):

update.py
peterbarker: Done.

@Hwurzburg could you test, please?

peterbarker:
We've decided to do this by adding a blacklist to the features-extraction script instead.

PR #4707, opened 2022-11-08
Common: move Gremsy setup instructions for AP4.3 above AP4.2

rmackay9:
Now that AP4.3 has gone live we can move the Gremsy setup instructions for 4.3 above the 4.2 instructions.


Hwurzburg:
LGTM..thanks

PR #4708, opened 2022-11-08
Documentation: Update QuadPlane setup for Tricopter

timtuxworth:
Mainly being a bit clearer about the initial setup for a Tricopter, but also adds the important note that setting Q_FRAME_CLASS and TYPE requires a reboot.

Requires the image posted by @Hwurzburg in Discord - I don't know how to include an image.


plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -85,9 +88,21 @@ You can optionally move the quad motors to be on any other channel above
 
 Tricopter
 ---------
+Frame Type 7 is Tricopter and can be either non-Tiltrotor configuration, or :ref:`Tiltrotor<guide-tilt-rotor>` configured using either Vectored or Non-Vectored yaw control.
```
Hwurzburg, line 91: its fine to add the default motor setups (its incorrect as shown), but changing the existing line 89 is not required...add it below

plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -85,9 +88,21 @@ You can optionally move the quad motors to be on any other channel above
...
+
+With the default Tricopter frame the motors are on outputs
```
Hwurzburg, line 93: this may be dependent on number of outputs on the controller....

Hwurzburg: I would word this section (to added below existing):
The Tricopter frame will attempt to assign the following output functions, if not already assigned , to unassigned outputs 5 and above:
Motor1: Right Front, looking from above
Motor2: Left Front 
Motor4: Rear
Motor7: Rear motor tilt servo (not used in vectored yaw tilt-rotor configurations)

timtuxworth: > I would word this section (to added below existing):
> The Tricopter frame will attempt to assign the following output functions, if not already assigned , to unassigned outputs 5 and above:
I don't think that's correct. I tested disabling all outputs and it will definitely attempt to assign to output 1 if it is the first unassigned output. Another real example, the T1 Ranger doesn't have a rudder, so I disabled output 4 - as a result auto assign put Motor 4 on output 4 (i.e. < 5).

plane/source/docs/quadplane-frame-setup.rst
timtuxworth: I'm not sure how to say this Henry, but if by "the existing line 89" you mean this 

`Frame Type 7 is Tricopter and can be either non-Tiltrotor configuration, or :ref:`Tiltrotor<guide-tilt-rotor>` configured using either Vectored or Non-Vectored yaw control. If using non-Tiltrotor or Non-Vectored Yaw Tilt-rotor, the yaw control output is setup as Motor 7 (``SERVOn_FUNCTION`` = 39) using a tilt mechanism for the yaw motor, Motor 4. You should set up the yaw servo’s maximum lean angle in degrees with :ref:`Q_M_YAW_SV_ANGLE<Q_M_YAW_SV_ANGLE>`. This lean angle assumes that ``SERVOn_MIN`` and ``SERVOn_MAX``, represent +/- 90 degrees, with ``SERVOn_TRIM`` representing 0 degrees lean.`

It is unreadable. My changes are an attempt to make sense of it, because as it stands it is incredibly hard to decipher.

plane/source/docs/quadplane-frame-setup.rst
Hwurzburg: all bets are off if you have altered output functions...plane sets outputs 1-4 by default, then qplane enable sets 5-8 since 1-4 are already set, then changing frame sets the appropriate outputs...ie hex adds 9 and 10, tri, drops 8 and adds 11...but if you have messed with things before you enable a frame, all bets are off and it will try to add motors as it can...

Hwurzburg:
I noticed some other things on the page that need updating, and the motor assignment should be in another section...so rather than trying to adjust this one, I just started one to replace this with you as co-author to get commit credits....#4709....feel free to comment....closing this one

PR #4709, opened 2022-11-09
Update QPlane frame page with tricopter motor defaults

Hwurzburg:
Co-authored-by:TimTuxworth(tim@objectivedata.com)

replaces #4708 with more clarifications

### DIFF
--- a/plane/source/docs/quadplane-frame-setup.rst
+++ b/plane/source/docs/quadplane-frame-setup.rst
@@ -7,16 +7,15 @@ QuadPlane Frame setup
 The QuadPlane code supports several frame arrangements of quadcopter,
 hexacopter, octacopter and octaquad multicopter frames which use lifting motors in addition to the normal forward motor(s). Also configurations in which the VTOL motors tilt for transitions or control, as well as choice between horizontal VTOL stance or vertical (Tailsitters).
 
-
 Tailsitters
 ===========
 
 Frame setup for Tailsitters is in this section: :ref:`guide-tailsitter`. Once the frame is configured, proceed with the other :ref:`quadplane-setup` instructions.
 
-Plane VTOL Stance
-=================
+Plane VTOL Motor Configurations
+===============================
 
-These configurations add multicopter style lifting motors to a conventional fixed wing configuration. Some or all of these motors may also be configured as tilting motors to be used in fixed wing flight.
+These configurations add multicopter style lifting motors to a conventional fixed wing configuration. Some or all of these motors may also be configured as tilting motors to be used in fixed wing flight instead of the normal fixed forward motor(s).
 
 Frame Types and Classes
 -----------------------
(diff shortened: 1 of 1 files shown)


plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -7,16 +7,15 @@ QuadPlane Frame setup
...
 
-Plane VTOL Stance
-=================
+Plane VTOL Motor Configurations
+===============================
 
-These configurations add multicopter style lifting motors to a conventional fixed wing configuration. Some or all of these motors may also be configured as tilting motors to be used in fixed wing flight.
+These configurations add multicopter style lifting motors to a conventional fixed wing configuration. Some or all of these motors may also be configured as tilting motors to be used in fixed wing flight instead of the normal fixed forward motor(s).
 
 Frame Types and Classes
```
timtuxworth, line 20: I think we should include the note about rebooting required no?

.. note:: You should reboot the autopilot after setting :ref:`Q_FRAME_CLASS<Q_FRAME_CLASS>` and/or
 :ref:`Q_FRAME_TYPE<Q_FRAME_TYPE>` and before making any other changes as they will change the default assignment of output channel.

plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -41,53 +40,59 @@ layout. For Tri and Y6 this parameter is ignored.
...
+Frame Type 7 is Tricopter and can be either non-Tiltrotor, or a :ref:`Tiltrotor<guide-tilt-rotor>` configuration for the front motors, but using a separate servo to sideways tilt the rear motor for yaw control, or a :ref:`Tiltrotor<guide-tilt-rotor>` using Vectored Yaw control via the front tilting motors.
 
+If using a tilting yaw control, an output is setup as Motor 7 (``SERVOn_FUNCTION`` = 39) for a servo-controlled sideways tilt mechanism for the yaw motor, Motor 4. You should set up the yaw servo’s maximum lean angle in degrees with :ref:`Q_M_YAW_SV_ANGLE<Q_M_YAW_SV_ANGLE>` to prevent prop strikes to the ground or frame. This lean angle assumes that ``SERVOn_MIN`` and ``SERVOn_MAX``, represent +/- 90 degrees, with ``SERVOn_TRIM`` representing 0 degrees lean.
```
timtuxworth, line 95: I think we need to mention that the Tail Tilt servo is ignored if using Vectored Yaw. I think this is important because it's confusing that this output gets created, so important to note that even if it appears in the outputs it will not be used in this case.

timtuxworth (changes requested):
Thanks for this @Hwurzburg , I think it makes it a lot easier to understand. 

I'm not sure how to update as a co-author, so I just made comments for now, but I'll try to make a change (maybe fix the typo) so I can figure out how to do it.

plane/source/docs/quadplane-frame-setup.rst
Hwurzburg: I will have to check, but I do not think that the servo is disabled....in other words, you could conceivably use both yaw methods I think at the same time....

plane/source/docs/quadplane-frame-setup.rst
Hwurzburg: yep missed adding that...will do

plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -41,53 +40,59 @@ layout. For Tri and Y6 this parameter is ignored.
...
+-  **Output 6:** Motor 2 - Front left motor
+-  **Output 8:** Motor 4 - Rear motor
+-  **Output 11:** Motor 7 - Tail Tilt Servo (see below)
+
 The normal plane outputs are assumed to be on 1 to 4 as usual. Only
-vertical lift outputs (5 to 8 on a quad setup) run at high PWM rate
-(400Hz). In a quad setup you can also use channels 9 to 14 in any way
-you like, just as with the normal Plane code.
+vertical lift outputs (5 to 8 on a quad setup) normally should be run at high PWM rate
+(400Hz). When using these default configurations, you can assigne other outputs to whatever functions you desire.
```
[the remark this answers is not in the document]
Hwurzburg: usually I have more than one...I am a dyslexic typist

plane/source/docs/quadplane-frame-setup.rst
Hwurzburg: tilt mask has nothing to do with the yaw tilt servo....its really not a tilt motor....tilt motors become fixed wing propulsion motors...the tricopter yaw servo tilt is not involved

timtuxworth: It might be good to say exactly that. If I find it confusing I'm sure others will.

Hwurzburg: doen

PR #4710, opened 2022-11-10
Rover: balancebot tuning shows how to speed up PID_TUNING messages

rmackay9:
Add an image that shows how users can speed up the PID_TUNING messages which helps with real-time tuning.

I've tested this on my local machine and it seems fine


Hwurzburg:
LGTM, thanks

PR #4711, opened 2022-11-11
docs: small correction in terminology

shubham-shahh:
MAVROS is a collection of rosnodes, and a collection of rosnodes is often referred as a package


Hwurzburg:
thanks

PR #4723, opened 2022-11-15
Update: remote ID compatible devices

adalta87:
added in the mro remoteID that is based on the ESP32-C3


Hwurzburg:
LGTM.....merging

PR #4724, opened 2022-11-15
Update common-mro-nexus.rst: fix broken link

adalta87:
fixed broken link to the power module. this is the new link as they have updated their product numbering.  https://store.mrobotics.io/product-p/m10026a.htm


Hwurzburg:
thanks

PR #4727, opened 2022-11-17
Hexsoon: EDU450 and 650 updates

rmackay9:
These are updates requested by Hexsoon to the EDU450 and EDU650 reference pages and also an update to the partners page logo.

I've tested these on my local machine and they seem fine.


Hwurzburg:
LGTM

PR #4732, opened 2022-11-20
benewake: Add TF02 Pro

muramura:
I would add TF02 Pro.
http://en.benewake.com/product/detail/5c345c9de5b3a844c4723299.html
The TF02 Pro frame does not have SIG.
The TF02 Pro is at device temperature.

Screenshot from 2022-11-19 20-38-48

I have the TF02 EOL.
I removed the TF02 link.


muramura:
I made a mistake.

rmackay9 (approved):
LGTM

Hwurzburg:
thanks

PR #4735, opened 2022-11-22
Update mission-command-list.rst

tatsuy:
The Dir description of CONDITION_YAW has been updated according to the following commit.
https://github.com/ArduPilot/ardupilot/commit/feae762e645e517e86a7fd9d61bca2cfc6b72070#diff-0e64ce1a1b5db8c13f7213375fb6ec4f06046136b39e1cbb312bc006284a9970R113


copter/source/docs/mission-command-list.rst
```diff
@@ -292,12 +292,13 @@ OR the change in heading (in degrees) when "rel/abs" field is "1"
 **Sec** - not supported.  meant to limit the rotation speed (in deg/sec)
```
lthall, line 292: This is now supported. It does set the maximum rotation speed provided it is not larger than the limits set by the aircraft.

copter/source/docs/mission-command-list.rst
```diff
@@ -292,12 +292,13 @@ OR the change in heading (in degrees) when "rel/abs" field is "1"
...
 as the vehicle turns to the desired heading
 
-**Dir (1=CW, -1=CCW)** - Used only when relative heading is specified
+**Dir (1=CW, -1=CCW)** - Used when relative heading is specified
```
lthall, line 295: I found this very difficult to read and a little confusing because of the word "Used" at the beginning I would suggest a slight change:

**Dir (1=CW, -1=CCW)** - When relative heading is specified
(i.e. "rel/abs" = "1") denotes whether the autopilot should add
(CW) or subtract (CCW) the degrees (Deg) from the current heading to
calculate the target heading. When absolute heading is specified
(i.e. "rel/abs" = "0") denotes CW/CCW direction. If this is set to "0",
the vehicle will always turn in direction that most quickly gets it
to the new target heading.

Maybe a clearer discription might be:

**Dir (1=CW, -1=CCW)** - Used to denote the direction of 
rotation to acheive the target angle. If this is set to "0",
the vehicle will always turn in direction that most quickly gets it
to the new target heading.

lthall (commented):
Thanks!!

copter/source/docs/mission-command-list.rst
tatsuy: Thank you.
I prefer simple one.

tatsuy:
@lthall Thank you for the review.
I renamed the field name "Sec" to "Speed" in Mission Planner here.
https://github.com/ArduPilot/MissionPlanner/pull/3018

The field name has been changed to reflect the above changes.

Hwurzburg:
@tatsuy I pushed some grammatic changes....before I merge however, what happens if DIR=0 and ABS/REL=1???

lthall:
It should just take the shortest path. 

Or are you asking about going negative angle. I think you can also provide a negative angle. The code behaves like you set DIR = 1.

tatsuy:
@Hwurzburg 
Thank you for the changes.

If ABS/REL=1, DIR: ≧0 (zero or positive) is CW, <0 (negative) is CCW.
> direction >= 0 ? 1.0 : -1.0
https://github.com/ArduPilot/ardupilot/blob/Copter-4.3/ArduCopter/autoyaw.cpp#L113

Hwurzburg:
thanks...pushed an expansion based on this...merging

PR #4736, opened 2022-11-22
Update: Fixed AMA FPV link

adalta87:
the link on the AMA FPV statement was broken. fixed to now take you to correct page instead of page not found.


Hwurzburg:
thanks!

PR #4737, opened 2022-11-22
Update: **Skid-Steer Rover**  video

adalta87:
the video that was tagged had been made private @Hwurzburg not sure if this one will do but right off the bat in this video it has good example of skid-steer rover? https://www.youtube.com/watch?v=LWc7Gkazwn4


rmackay9:
For the skid steering vehicle we might want to use one of our active users's video.  I think this one from Yuri would be good
https://www.youtube.com/watch?v=2mk8fhbTN3Y&t=11s

It's scope creep so doesn't need to be handled by this PR but the snowblower used a few lines up is an awesome example but not representative of most AP rovers so we should look for an example ackermann steering rover video instead.  Also we should replace the description, "Pivotting steering wheels" with "Ackermann steering vehicle"

Hwurzburg:
fixed ackerman video as suggested...also added vehicles types to ardupilot entry page....merging this now

PR #4738, opened 2022-11-22
update: dronekit python link

adalta87:
the link for dronekit python was broken (http://python.dronekit.io/) changed link to https://github.com/dronekit/dronekit-python is this what it should be or should we just link it to https://dronekit.io/


Hwurzburg:
thanks...merging

PR #4739, opened 2022-11-22
Update: broken links dronekit and 3dr robotics

adalta87:
@Hwurzburg  
again the dronkit python link was broke so I updated it to the dronkit webpage. also 3dr links broken and I removed the link completely as they will never be back. not sure if you want the dronekit link to go to their webpage or the dronekit github?


Hwurzburg:
LGTM

PR #4745, opened 2022-11-23
Common: ModalAi setup

rmackay9:
This adds a new ModalAI page in the Non-GPS Navigation section.

Although we will probably need some more details added in the future this is the basic setup that should at least help users get started using the camera.  These instructions are based on my own experiences setting up the camera.

I've built this page locally on my machine and it seems OK.


Hwurzburg:
thanks

PR #4746, opened 2022-11-24
Update common-cuav-x7-family-overview.rst

onurerden:
Mispelling on servo output channels (which may lead misunderstanding.)


Hwurzburg:
thanks

PR #4747, opened 2022-11-24
Update the list of supported MCUs

MichelleRos:
Previous list used eg STM32F41x notation, which implied that certain MCUs with less than 1 MB of flash were supported.


peterbarker (approved):
If it saves some confusion for people trying to put e.g. Copter somewhere it won't fit it's good.

I would point out that we *do* run on boards with less than 1MB - but AP_Periph only.

Hwurzburg:
okay merging this...and will put a note after about AP_Periph

PR #4748, opened 2022-11-25
Common: add video to modalai wiki page

rmackay9:
This adds my latest modalai testing video to the modalai wiki page.

I've tested this locally and it seems OK.


Hwurzburg:
thanks

PR #4750, opened 2022-11-26
Copter: update hexsoon edu450 and 650 reference frames

rmackay9:
Some minor updates requested by hexsoon to the EDU450 and EDU650 reference frame pages.


Hwurzburg:
LGTM

PR #4751, opened 2022-11-26
Update common-connecting-sport-fport.rst

JoshWelsh:
Setting the value to 15 includes "SWAP" which is incorrect


Hwurzburg:
thanks

PR #4752, opened 2022-11-27
plane: updating ArmDisarm for 4.2 and higher

arikrupnik:
I understand that `RCx_OPTION` `ArmDisarm` changed from 41 to 153 in 4.2. If it's desirable to note that older versions use 41, I can amend the PR.


Hwurzburg:
no need....thats two revs back in history..thanks

arikrupnik:
You're welcome. BTW, what's the story with the change from 41 to 153?

Hwurzburg:
the old way could arm with airmode active and get high motor rpm

arikrupnik:
Ah, I remember now. Thank you.

PR #4767, opened 2022-12-02
Common: add brushless pwm gimbal setup

rmackay9:
This adds a Brushless PWM gimbal setup page which is supported in AP 4.3.

I've reused the Tarot gimbal images for now because this is all I have.  The Tarot gimbal is a decent example of this kind of setup though although sadly the one shown only provides pitch control.  Still, I think this is an improvement and will at least give me a page to point users at when they struggle to setup brushless pwm gimbals with AP-4.3.

I have tested this on my local machine and it seems ok.


Hwurzburg:
thanks

rmackay9:
Thanks!

PR #4768, opened 2022-12-02
Add how funds are spent on the donations page

mohit-marathe:
Added a link to the page explaining how funds are spent at the bottom of the donations page


Hwurzburg:
LGTM....thanks

PR #4770, opened 2022-12-02
Common: update servo gimbal for 4.3

rmackay9:
This updates the servo gimbal setup page for 4.3 which is important because Mission Planner's camera gimbal setup screen doesn't work with 4.3.

While I think these changes are an improvement there are still numerous issues with this page:

1. the 4.2 section has a mix of instructions that are only valid for 4.3.  In particular the parameters that must be set to allow the pilot to control the gimbal angles via RC are incorrect.
2. the 4.2 section should mention that userrs should never check the "stabilize yaw" checkbox.
3. We should remove instructions on camera setup because we have another page for that already
4. We should move the "Camera Mount Mode/Targeting" to a separate page because this applies to all gimbals


Hwurzburg:
@rmackay9 if these additions cover your notes let me know and I will merge

rmackay9 (commented):
definitely an improvement, thanks!

rmackay9:
normally I don't merge wiki PRs anymore but I took the liberty this time 'cuz it's a mix of your work and mine.

PR #4773, opened 2022-12-03
Add Foxeer ReaperF745 board

Hwurzburg:
once we get the V3 hwdef changes up I will move this back to active board with V3 included


common/source/docs/common-foxeerf745aio.rst
```diff
@@ -0,0 +1,146 @@
...
+   -  I2C port for external compass, airspeed sensor, etc.
+   -  Micro USB port
+   -  Camera input/ VTX output
+   -  DJI Goggle connector
+
+
+   **Size and Dimensions**
+
+   - 32.5mm x 32.5mm (25.5 x 25.5mm mount pattern)
+   - 8.6g
```
andyp1per, line 54: This is an AIO - we should mention the ESCs as well

Hwurzburg: it mentioned in the lead and the feature list

Hwurzburg: what setup...the default params?

andyp1per: image

andyp1per: How about something like this, right here:
```
 **ESCs**

   -  4 integrated BLHeli_S ESCs
   -  Firmware - BlueJay
   -  2S-4S input voltage, 45A burst current
   -  DShot 300/600
```

Hwurzburg: okay...but 2s-6s

PR #4782, opened 2022-12-06
Common: clarify Mount control page

rmackay9:
This is a follow up to PR https://github.com/ArduPilot/ardupilot_wiki/pull/4770 that attempts to clearly explain how users may control a camera gimbal using RC inputs, GCS commands or Auto mode mission commands.

This has been tested on my local machine and seems OK.


Hwurzburg:
thanks

PR #4783, opened 2022-12-06
Linux board's don't support all features

LaneaLucy:
As I learned from the discord, Linux board's don't support all features because there isn't enough demand for Linux board's. As I didn't know this at the beginning, I now have a flight controller I can't use for my purposes. 
I think other should know this too before deciding eventually for a Linux based controller and don't make the same error as me

### DIFF
--- a/common/source/docs/common-autopilots.rst
+++ b/common/source/docs/common-autopilots.rst
@@ -6,6 +6,8 @@ Autopilot Hardware Options
 
 This section provides information about ArduPilot Autopilot Hardware options.  The list below is sorted by manufacturer and product name.
 
+Linux boards don't support all features as of low demand.
+
 Open hardware
 -------------
 


rmackay9:
@LaneaLucy,

Txs for the contribution.  By the way, which missing features were especially important?

LaneaLucy:
For the moment the only missing feature I know of, is dshot and ESC passthrough, but from what I heard, there is more, so future will show...
Or not, because I must buy a not Linux based FC. But it would be good to know in the first place, then I would have buyed a not Linux based one at the first place

Hwurzburg:
we will discuss at next devcall, but the caveat needs to be reworded a bit I think....perhaps:

ArduPilot provides code libraries for Linux board suppliers to port ArduPilot to their boards. The exact feature set is dependent on each board's hardware capabilities and how the board supplier implements ArduPilot functionality on their board. Contact the board supplier for support questions and details of their implementation.

rmackay9:
Suggested wording from the dev call 

Note: Linux boards have more CPU power and memory than many other boards but do not support DShot, ESC passthrough and easy upload from the ground stations.

davidbuzz:
Every board has a different set of fettures, doesnt matter if it stm32/chibios, arm/linux, intel/linux, esp32/freertos, all boards have different featurres, some dont have a fast-enuf cpu, some dont have enuf ram, some dont have tightly integrated peripherals, or drivers with the right dma/channel/option. A stm32f401 has less features than a stm32f427 which has less than a stm32f7, let alone the other hardware and they all have different selections of sensors,pinouts, power-managment, software , integrion etc.

Hwurzburg:
#4809 addresses this and more....closing

PR #4784, opened 2022-12-06
Common: add warning to Robsense swarm link

rmackay9:
This adds a warning to the Robsense telemetry page to warn users away from this product.  The reason I'm adding this warning is that I've received a complaint from a user that their order was cancelled and the email that we had on file for the CEO of the company no longer works.  There is also no longer an easy way to purchase the telemetry radio.

I've tested this on my local machine.


Hwurzburg:
thanks

Issue #4787, opened 2022-12-06
If ACRO mode uses rudder now, update acro-mode.html


arikrupnik:
https://ardupilot.org/plane/docs/acro-mode.html says: 

> In the current implementation the controller won’t use rudder while the plane is on its side to hold pitch, which means horizontal rolls won’t be as smooth as a good manual pilot, but that should be fixed in a future release. This also means that it won’t hold knife-edge flight.

Is this still accurate as of 4.3?

Hwurzburg:
yes its true unless YAW_RATE_ENABLE is enabled and the axis properly tuned...unless acro locking is also enabled, only the rate controller is active and attitude drift in body-frame yaw can still occur...

Hwurzburg:
see #4791

PR #4788, opened 2022-12-06
Plane: ACRO uses yaw controller now #4787

arikrupnik:
I understand that as of 4.2, yaw controller is available in ACRO mode. This page describes its configuration lower down. This paragraph appears to be a holdover from previous versions of ArduPilot.


Hwurzburg:
#4791 replaces this

PR #4789, opened 2022-12-06
Plane: how to save AUTOTUNE results

arikrupnik:
This keeps tripping me up. I don't use AUTOTUNE often enough, and when I do, I forget to disarm in AUTOTUNE. I believe this behavior is the same for plane as for copter in current versions.

### DIFF
--- a/plane/source/docs/automatic-tuning-with-autotune.rst
+++ b/plane/source/docs/automatic-tuning-with-autotune.rst
@@ -135,7 +135,9 @@ coping well with wind, or not holding altitude well. Keep flying in
 AUTOTUNE mode well past the point where you think the plane is flying
 well.
 
-If you have an OSD or Ground Control Station, then messages for D and P tuning phases for the axis, as well as a "Tuning Completed" message will be displayed. Once tuning is completed, it is saved.
+If you have an OSD or Ground Control Station, then messages for D and P tuning phases for the axis, as well as a "Tuning Completed" message will be displayed.
+
+.. note:: To save the new tune, you must disarm while still in AUTOTUNE mode. If you switch out of AUTOTUNE, new settings will be lost and previous tune restored. You can use this feature if you're unhappy with AUTOTUNE's results and want to discard them.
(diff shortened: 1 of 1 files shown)


plane/source/docs/automatic-tuning-with-autotune.rst
```diff
@@ -135,7 +135,9 @@ coping well with wind, or not holding altitude well. Keep flying in
 AUTOTUNE mode well past the point where you think the plane is flying
 well.
 
-If you have an OSD or Ground Control Station, then messages for D and P tuning phases for the axis, as well as a "Tuning Completed" message will be displayed. Once tuning is completed, it is saved.
+If you have an OSD or Ground Control Station, then messages for D and P tuning phases for the axis, as well as a "Tuning Completed" message will be displayed.
```
Hwurzburg, line 138:
```suggestion
If you have an OSD or Ground Control Station, then messages for D and P tuning phases for the axis, as well as a "Tuning Completed" message will be displayed. The values are saved as the tuning progresses and P and D values are saved automatically. The FF (feed-forward) values are continuously adjusted as long as the plane remains in AUTOTUNE mode.
```

Hwurzburg (changes requested):

plane/source/docs/automatic-tuning-with-autotune.rst
arikrupnik: Are you sure? That means there is no way to undo an AUTOTUNE in plane?

arikrupnik:
> you only have to change mode to save the completed Autotune...not like copter

Then the wiki should probably say that. Is it enough to switch out of Autotune (if you have it on a separate switch) or do you need to change modes, like from FBWA to MANUAL?

plane/source/docs/automatic-tuning-with-autotune.rst
Hwurzburg: I is applied based on FF

PR #4791, opened 2022-12-07
update ACRO mode page

Hwurzburg:


arikrupnik:
Thank you for adding these clarifications and updates. Looks excellent.

PR #4792, opened 2022-12-07
cygwin64 needs a 64bit OS, and

davidbuzz:
the cygwin project has recently dropped support for 32bit, microsoft has stopped allowing new 32bit OS installs approx two years ago, windows 7 went end-of-life in 2020.


Hwurzburg:
LGTM

PR #4794, opened 2022-12-08
Add joystick mode2 setting

csc-takeda:
Added joystick mode 2 setting.
Attention is reverse of Ch3.


Hwurzburg:
not sure that this is a good addition....it will vary depending on what joystick or TX (in joystick output mode) is being used and be incorrect for them....my intention is to close this unless @rmackay9 has an alternate suggestions

rmackay9:
Thanks for the wiki update.

I think I agree with Henry.  Although this addition is useful when using a Logitech F310 in Mode2 it may not apply to all joysticks and also some users want Mode1.  I think if we are going to add one specific example we need to ad many and then it just becomes too complex.  I think users can just follow the instructions that are there now to "waggle the sticks" to assign the right channels.

PR #4795, opened 2022-12-09
Add supported model number to VectorNav

Gone4Dirt:
Not all VectorNav units that are on the linked products page are currently supported.  Adding model number of the unit that is supported to prevent confusion.

Other VectorNav units may already work.  But I know that at least one of them doesn't (VN-100) at this time.


Hwurzburg:
LGTM

PR #4799, opened 2022-12-10
Common: small updates to Gremsy and Siyi gimbal pages

rmackay9:
This PR has two changes:

1. Gremsy ZIO is added to the Gremsy page
2. Siyi ZR10 and A8 image get voltage input requirements

I've tested these changes on my local machine and they seem OK.


Hwurzburg:
thanks

PR #4810, opened 2022-12-13
Update terrain tile webpage that was broken

adalta87:
this is what it was "https://terrain.ardupilot.org/data/" this gives a 404 error. this is what I believe should be the new link "https://terrain.ardupilot.org/continentsdat3/"


Hwurzburg:
thanks....needs some more changes which I will do now

PR #4811, opened 2022-12-13
add note that Sport mode is not built by default

Hwurzburg:


rmackay9:
Thanks very much!

rmackay9:
Ah, but there's a typo in the link to the firmware server.. 
https://fimrware.ardupilot.org/

Hwurzburg:
fixed....was wrong on the source I copied and pasted from also

PR #4812, opened 2022-12-14
Common: minor formatting fixes to autpilots page

rmackay9:
Just a few formatting fixups to the autopilot page:

- fix the bullet points on the unsupported boards section
- make the use of * consistent across categories

I've tested this on my local machine and it seems ok.


Hwurzburg:
thanks

PR #4813, opened 2022-12-14
Fix Copter Flight Mode Channel Description

pwnept:
Currently, on Copter, the wiki states that the flight mode channel is always channel 5. But this parameter has been added since Jan 27, 2018 on commit 89674482a717bd280c594c408c6feec7d569fac9 of the main firmware repo.

### DIFF
--- a/common/source/docs/common-rc-transmitter-flight-mode-configuration.rst
+++ b/common/source/docs/common-rc-transmitter-flight-mode-configuration.rst
@@ -80,10 +80,11 @@ The flight mode channel is the input radio channel that ArduPilot
 monitors for mode changes.
 
 [site wiki="copter"]
-On Copter this is always channel 5.
+On Copter this is configurable using the :ref:`FLTMODE_CH <FLTMODE_CH>`
+parameter. 
 [/site]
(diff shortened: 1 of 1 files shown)


common/source/docs/common-rc-transmitter-flight-mode-configuration.rst
```diff
@@ -80,7 +80,8 @@ The flight mode channel is the input radio channel that ArduPilot
 monitors for mode changes.
 
 [site wiki="copter"]
-On Copter this is always channel 5.
+On Copter this is configurable using the :ref:`FLTMODE_CH <plane:FLTMODE_CH>`
```
Hwurzburg, line 83:
```suggestion
On Copter this is configurable using the :ref:`FLTMODE_CH <FLTMODE_CH>`
```
its in the copter wiki parameters page now...no need for the redirection to plane....

Hwurzburg (changes requested):

Hwurzburg:
Since this was a web entered PR, I just fixed it and also made another update and merged ...thanks

PR #4824, opened 2022-12-20
Add info on JSON ride-along simulation

Hwurzburg:


dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -590,6 +590,26 @@ Other SITL vicon settings are hel in SIM_VICON_xx parameters:
 
    param show SIM_VICON*
 
+Simulating a second flight controller
```
IamPete1, line 593:
```suggestion
Simulating multiple flight controllers on a single vehicle
```

dev/source/docs/using-sitl-for-ardupilot-testing.rst
```diff
@@ -590,6 +590,26 @@ Other SITL vicon settings are hel in SIM_VICON_xx parameters:
...
+-------------------------------------
+
+The STANDBY Mode function (``RCx_OPTION`` = 76) allows an autopilot to be put into a soft standby mode so that a second autopilot or companion computer can take control, but allow the first to resume later with smooth transition. In order to help in the code development of such systems, you can simulate two flight controllers operating in the simulation at the same time. The "master" instance, which has the active outputs into the sim is started declaring that there will be a "slave" instance which will be monitoring the sensors and running its code in parallel, but whose outputs are not connected into the simulation physics:
```
IamPete1, line 596:
```suggestion
The STANDBY Mode function (``RCx_OPTION`` = 76) allows an autopilot to be put into a soft standby mode so that a second autopilot or companion computer can take control, but allow the first to resume later. In order to help in the code development of such systems, you can simulate two flight controllers operating in the simulation at the same time. The "master" instance, which has the active outputs into the sim is started declaring that there will be a "slave" instance which will be monitoring the sensors and running its code in parallel, but whose outputs are not connected into the simulation physics:
```

Transitions will not be smooth, the key thing is that they don't crash ;)

PR #4833, opened 2022-12-25
Improved Webots simulation build instructions

IanBurwell:
Added a note about Webots version that is currently working and improved the instructions on how to build the controller and physics plugin. Using the updated instructions I have a basic working Webots environment.

### DIFF
--- a/dev/source/docs/sitl-with-webots.rst
+++ b/dev/source/docs/sitl-with-webots.rst
@@ -18,8 +18,8 @@ Running the Examples
 ====================
 
 #. Open Webots using the command line.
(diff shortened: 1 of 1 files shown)


dev/source/docs/sitl-with-webots.rst
```diff
@@ -33,8 +33,8 @@ Running the Examples
 .. warning::
 
    It is important to run Webots before ArduPilot's SITL , otherwise the socket connection may not establish.
-
-
+   
+   As of writing, these instructions work with Webots 2021b. Newer versions may or may not work correctly.
```
HefnySco, line 37: Newer versions will not work correctly.
I am in the process of preparing one for R2023a.

Hwurzburg: @HefnySco so should we merge this as is or wait?

HefnySco: I suggest we can merge this, and we will keep this next to the newer version to keep compatibility with older versions in case someone  is using this old version with his/her world files.

PR #4838, opened 2022-12-27
Fix typo

ShevaDavid:


dev/source/docs/copter-commands-in-guided-mode.rst
```diff
@@ -288,12 +288,12 @@ The message definition can be found `here <https://mavlink.io/en/messages/common
 
 Valid options are:
 
-- MAV_FRAME_GLOBAL (0): alt is meters above sea level
-- MAV_FRAME_GLOBAL_INT (5): alt is meters above sea level
-- MAV_FRAME_GLOBAL_RELATIVE_ALT (3): alt is meters above home
-- MAV_FRAME_GLOBAL_RELATIVE_ALT_INT (6): alt is meters above home
-- MAV_FRAME_GLOBAL_TERRAIN_ALT (10): alt is meters above terrain
-- MAV_FRAME_GLOBAL_TERRAIN_ALT_INT (11): alt is meters above terrain
+- MAV_FRAME_GLOBAL (0): alt in meters above sea level
```
Hwurzburg, line 291: all should say "altitude is in meters....", rather than existing or your proposal

Hwurzburg (changes requested):
if we are going to change, might as well make a complete sentence

dev/source/docs/copter-commands-in-guided-mode.rst
rmackay9: "alt" in this case is actually a field name so I don't think we should change it to "altitude".

Hwurzburg: that is fine also...

ShevaDavid: actually I didn't think that in this particular context `alt` is variable name, not the short form of `altitude`. So if my change does not make sense, feel free to discard it

Hwurzburg:
squashed...merging...thanks

PR #4843, opened 2022-12-29
add SURFTRAK_MODE info

Hwurzburg:
@rmackay9 we need to explain how terrain following and surface tracking co-exist...ie hand-offs....does terrain following take over once out of downward facing RF range?what happens if it goes out of upward facing range?

### DIFF
--- a/copter/source/docs/terrain-following-manual-modes.rst
+++ b/copter/source/docs/terrain-following-manual-modes.rst
@@ -4,9 +4,9 @@
 Surface Tracking
 ================
 
-Copter supports low altitude Surface Tracking in nearly all modes including pilot controlled modes like :ref:`AltHold <altholdmode>`, :ref:`Loiter <loiter-mode>` and :ref:`PosHold <poshold-mode>` and also fully autonomous modes including :ref:`AUTO <auto-mode>` and :ref:`Guided <ac2_guidedmode>`.
+Copter supports low altitude Surface Tracking, either a floor/ground or a ceiling, in nearly all modes including pilot controlled modes like :ref:`AltHold <altholdmode>`, :ref:`Loiter <loiter-mode>` and :ref:`PosHold <poshold-mode>`.
 
-Additionally, Terrain Following is supported in autonomous modes using a terrain height data base stored on the autopilot's SD card and/or obtained real-time from the Ground Control Station.
+Additionally, :ref:`Terrain Following<terrain-following>` is supported in autonomous modes using a terrain height data base stored on the autopilot's SD card and/or obtained real-time from the Ground Control Station.
 
 This page describes the Surface Following setup .  For autonomous modes using downloaded data for Terrain Following please see the :ref:`terrain following for autonomous modes <terrain-following>` wiki page.
 
@@ -16,8 +16,10 @@ This page describes the Surface Following setup .  For autonomous modes using do
 Setup and Configuration
 -----------------------
 
-- Connect a downward facing :ref:`lidar or sonar <common-rangefinder-landingpage>` to the vehicle
-- An :ref:`auxiliary switch <common-auxiliary-functions>` (Copter 4.0 and later), or  :ref:`channel-7-and-8-options`, can be configured to turn on/off use of the rangefinder.
+- Connect a downward facing rangefinder(for floor/ground tracking) and/or an upward facing rangefinder (for ceiling tracking) :ref:`lidar or sonar <common-rangefinder-landingpage>` to the vehicle. 
+
+- Which is being used is configured by the :ref:`SURFTRAK_MODE<SURFTRAK_MODE>` parameter or by: 
+- An :ref:`auxiliary switch <common-auxiliary-functions>` (function "75"), or  :ref:`channel-7-and-8-options`, can be configured to turn on/off use of the rangefinder.
 
 .. warning::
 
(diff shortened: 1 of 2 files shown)


copter/source/docs/terrain-following-manual-modes.rst
```diff
@@ -4,7 +4,7 @@
 Surface Tracking
 ================
 
-Copter supports low altitude Surface Tracking in nearly all modes including pilot controlled modes like :ref:`AltHold <altholdmode>`, :ref:`Loiter <loiter-mode>` and :ref:`PosHold <poshold-mode>` and also fully autonomous modes including :ref:`AUTO <auto-mode>` and :ref:`Guided <ac2_guidedmode>`.
+Copter supports low altitude Surface Tracking, either a floor/ground or a ceiling, in nearly all modes including pilot controlled modes like :ref:`AltHold <altholdmode>`, :ref:`Loiter <loiter-mode>` and :ref:`PosHold <poshold-mode>` and also fully autonomous modes including :ref:`AUTO <auto-mode>` and :ref:`Guided <ac2_guidedmode>`.
```
rmackay9, line 7: The new additions are fine but the original wording should be corrected to clarify that "Surface Tracking" only applies to pilot controlled modes.  So this part should be removed at least:

> and also fully autonomous modes including :ref:`AUTO <auto-mode>` and :ref:`Guided <ac2_guidedmode>`

copter/source/docs/terrain-following-manual-modes.rst
```diff
@@ -4,7 +4,7 @@
...
 
 Additionally, Terrain Following is supported in autonomous modes using a terrain height data base stored on the autopilot's SD card and/or obtained real-time from the Ground Control Station.
```
rmackay9, line 9: We should change "Terrain Following" to be a hyperlink to the terrain following page.  This description that Terrain Following only uses the terrain database is also incorrect because a lidar can also be used.

copter/source/docs/terrain-following-manual-modes.rst
```diff
@@ -16,7 +16,7 @@ This page describes the Surface Following setup .  For autonomous modes using do
 Setup and Configuration
 -----------------------
 
-- Connect a downward facing :ref:`lidar or sonar <common-rangefinder-landingpage>` to the vehicle
+- Connect a downward facing rangefinder(for floor/ground tracking) and/or an upward facing rangefinder (for ceiling tracking) :ref:`lidar or sonar <common-rangefinder-landingpage>` to the vehicle. Which is being used is configured by the :ref:`SURFTRAK_MODE<SURFTRAK_MODE>` parameter.
```
rmackay9, line 19: I think we should separate out, "Which is being used is configured by the :ref:`SURFTRAK_MODE<SURFTRAK_MODE>` parameter." to its own bullet point.  There are actually two ways to select which is being used.  The parameter can be used but there's also an auxiliary switch option 75 ("Surface Tracking Up/Down")

copter/source/docs/terrain-following.rst
```diff
@@ -4,7 +4,7 @@
...
+Copter supports "terrain following" in nearly all modes including autonomous modes like :ref:`AUTO <auto-mode>`, :ref:`Guided <ac2_guidedmode>`, :ref:`RTL <rtl-mode>` and :ref:`Land <land-mode>`.  This feature allows the vehicle to climb or descend to maintain a specified distance above the terrain using either a :ref:`downward facing Lidar or Sonar <common-rangefinder-landingpage>` or from `SRTM <https://en.wikipedia.org/wiki/Shuttle_Radar_Topography_Mission>`__ data (aka terrain altitude data) provided by the ground station using a mapping service such as Google maps or stored on the autopilot's SD card in the "Terrain" subdirectory.  Details of how the SRTM data is used and stored can be found on the :ref:`plane terrain following page <plane:common-terrain-following>`
```
rmackay9, line 7: "in nearly all modes" should be replaced with "in nearly all autonomous modes".

copter/source/docs/terrain-following.rst
```diff
@@ -13,7 +13,7 @@ See :ref:`common-understanding-altitude` for altitude definitions.
 
 .. note::
 
-   :ref:`Loiter <loiter-mode>`, :ref:`PosHold <poshold-mode>` and :ref:`AltHold <altholdmode>` modes also support low altitude terrain following, called Surface Tracking.  Please refer to the :ref:`Surface Tracking <terrain-following-manual-modes>` wiki page.
+   :ref:`Loiter <loiter-mode>`, :ref:`PosHold <poshold-mode>` and :ref:`AltHold <altholdmode>` modes also support low altitude terrain following, called Surface Tracking. Surface tracking uses a rangefinder(s) oriented either pointing down or up or both and which surface is tracked (floor or ceiling) is controlled by the :ref:`SURFTRAK_MODE<SURFTRAK_MODE>` parameter.  Please refer to the :ref:`Surface Tracking <terrain-following-manual-modes>` wiki page.
```
rmackay9, line 16: I think we should keep this line as it was.  No need to go into the details of up/down here.

rmackay9:
Hi, thanks very much for this.  The new additions are mostly fine, the existing wording is not great though.  I'm also happy for this to be merged mostly as-is (I think the last change to the terrain-following page should be removed) and then I will rework the existing wording if you like.

copter/source/docs/terrain-following.rst
Hwurzburg: so terrain following is not valid in other manually set altitude hold modes??

rmackay9:
Thanks!

PR #4851, opened 2023-01-02
common: added note on APD ESC bug

tridge:


Hwurzburg:
builds and LGTM...merged

PR #4854, opened 2023-01-04
Common: siyi gimbal param change to control camera

rmackay9:
This adds a missing parameter setting required to allow autonomous or manual control of the Siyi cameras.

I've tested this on my local machine and it seems fine.


rmackay9:
Thanks!

PR #4858, opened 2023-01-06
Update ac2_guidedmode.rst

PeterQFR:
The Guided mode timeout isn't documented here but would be worth mentioning as users may want to shorten the default timeout.

### DIFF
--- a/copter/source/docs/ac2_guidedmode.rst
+++ b/copter/source/docs/ac2_guidedmode.rst
@@ -83,6 +83,8 @@ Bit 	Meaning
 
 Bit 3 makes interpretation of SET_ATTITUDE_TARGET MAVLink command's ``thrust`` field as pure thrust from 0 to 1 , instead of a climb rate. See `Commands in Guided Mode <https://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html>`__
 
+The :ref:GUID_TIMEOUT<GUID_TIMEOUT> parameter holds the timeout (in seconds) when the vehicle is being controlled using attitude, velocity and/or acceleration commands. If no commands are received from the companion computer for this many seconds, the vehicle will slow to a stop (if velocity and/or acceleration commands were being provided) or hold a level hover (if attitude commands were provided). The default setting is 3 seconds.
+
 .. _guided_nogps:
 
 Guided_NoGPS


copter/source/docs/ac2_guidedmode.rst
```diff
@@ -83,6 +83,9 @@ Bit 	Meaning
 
 Bit 3 makes interpretation of SET_ATTITUDE_TARGET MAVLink command's ``thrust`` field as pure thrust from 0 to 1 , instead of a climb rate. See `Commands in Guided Mode <https://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html>`__
 
+The :ref:`GUID_TIMEOUT<GUID_TIMEMOUT>` parameter sets the timeout for receiving commands from the companion computer, after which the system will zero out all commands, and transition into z_vel mode at zero velocity if using thrust. The default setting is 3s.
```
rmackay9, line 86: Thanks for this.

It is definitely good to mention this configurable timeout.  Two things though, we should simplify the description of what happens to "vehicle will come to a stop" (or similar) and also mention that it only applies if the caller is providing velocity or acceleration targets.

Maybe the wording should be something like this

The :ref:`GUID_TIMEOUT<GUID_TIMEOUT>` parameter holds the timeout (in seconds) when the vehicle is being controlled using angle, velocity and/or acceleration commands.  If no commands are received from the companion computer for this many seconds, the vehicle will slow to a stop.  The default setting is 3 seconds.

rmackay9 (changes requested):

copter/source/docs/ac2_guidedmode.rst
PeterQFR: Sure, though I thought it also applies when in attitude thrust guided mode, such as guided_no_gps, and it switches the z control from thrust to z vel. 
https://github.com/ArduPilot/ardupilot/blob/54a4b7a98b0bd96b81bf1026d2b4fbafa62d3842/ArduCopter/mode_guided.cpp#L924

If thats the case then, saying the system will slow to the stop is incorrect, in attitude/thust the system will drift but hopefully not increase or decrease in height. 

I had been speaking with James patterson about this, It would be good if it would automatically go into another mode, like the GCS Timeout does. Landing mode would at least allow the pilot to have RC control and for the aircraft to descend.

rmackay9: @PeterQFR, yes, you're right that in angle control mode it won't stop I think.  It just levels out and maintains altitude so perhaps the wording in my suggestion could be improved a bit.  Mostly I want to keep the wording simple enough that non-native speakers can those without a super strong tehcnical background can understand what will happen.

PeterQFR: I was just reading your first comment, so this timeout only applies to velocity, and acceleration (where acceleration is attitude/thrust/zvel). It doesn't apply in position mode? Ill pop a draft back up.

rmackay9: @PeterQFR, that's right.  When only position is provided the vehicle will just fly to it without requiring constant updates.

copter/source/docs/ac2_guidedmode.rst
```diff
@@ -83,6 +83,13 @@ Bit 	Meaning
 
 Bit 3 makes interpretation of SET_ATTITUDE_TARGET MAVLink command's ``thrust`` field as pure thrust from 0 to 1 , instead of a climb rate. See `Commands in Guided Mode <https://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html>`__
 
+The :ref:GUID_TIMEOUT<GUID_TIMEOUT> parameter holds the timeout (in seconds) when the vehicle is being controlled using angle, velocity and/or acceleration commands. If no commands are received from the companion computer for this many seconds, the vehicle will slow to a stop. The default setting is 3 seconds.
```
rmackay9, line 86: I don't want to be too picky but I think we should leave the Note out and instead just try and get all the info into the single paragraph.  How about this?

The :ref:GUID_TIMEOUT<GUID_TIMEOUT> parameter holds the timeout (in seconds) when the vehicle is being controlled using attitude, velocity and/or acceleration commands. If no commands are received from the companion computer for this many seconds, the vehicle will slow to a stop (if velocity and/or acceleration commands were being provided) or hold a level hover (if attitude commands were provided). The default setting is 3 seconds.

PeterQFR: No problems, that reads better

rmackay9 (approved):

Hwurzburg:
squashed and rebased...merging...thanks

PR #4862, opened 2023-01-07
Update SITL information

Hwurzburg:
needs a careful read, extensively re-organized into evoking the sims by sim_vehicle or in Missionplanner, separating params that control the sim to a new page, and a page on how to add simulated devices

also update the Mission Planner SIM page with new command line options that are useful

### DIFF
--- a/dev/source/docs/SITL_simulation_parameters.rst
+++ b/dev/source/docs/SITL_simulation_parameters.rst
@@ -1,0 +1,200 @@
+.. _SITL_simulation_parameters:
+
+Using Simulation Parameters to Control the Simulation
+=====================================================
+
+SITL has a large set of parameters to control the simulation. Wind speed/direction, Sensor malfunctions and parameters, RC failures, etc. can be emulated and controlled. They are visible and present when a simulation is running and prefixed by ``SIM_``. A few are detailed below.
+
+Setting parameters
+------------------
+
+.. note:: If the attached GCS is Mission Planner, you would use the CONFIG/Full Parameter List or Tree to change the parameters. The following gives examples using MAVProxy commands to change the parameter values.
+
+Many of the following tasks involve setting parameter values over
+MAVLink, which you do using the ``param set`` command as shown:
+
+::
+
+    param set PARAMETERNAME VALUE
+
+All available parameters can be listed using ``param show``. The
+SITL-specific parameters start with ``SIM_``, and can be obtained using:
+
+::
+
+    param show SIM_*
+
+.. tip::
+
+   A table of user settable SITL-specific parameters can be found in `SITL.cpp <https://github.com/ArduPilot/ardupilot/blob/master/libraries/SITL/SITL.cpp>`__
+   When you change a parameter the value remains in the virtual EEPROM
+   after you restart SITL. Remember to change it back if you don't want it
+   any more (or :ref:`reload/reset the parameters <using-sitl-for-ardupilot-testing_loading_a_parameter_set>`). 
+
+
+Changing the speed of the simulation
+------------------------------------
+
+Most of the simulator backends (native, X-plane, etc.) support changing the speed while
+running (NOT Realflight). Just set the ``SIM_SPEEDUP`` parameter as needed. A value of 1
+means normal wall-clock time. A value of 5 means 5x realtime. A value
+of 0.1 means 1/10th of real time.
+
+Running with high-levels of speedup can result in enough network traffic that MAVProxy can’t keep up. An error such as “Set RC override timeout” or the vehicle entering GCS failsafe are typical of these failures. Reducing the ``SIM_SPEEDUP`` factor is typically sufficient to avoid this problem. Also, using the ``--udp`` options will use UDP instead of the slower TCP communication protocol and can help avoid this issue at higher speedups.
+
+Loading a parameter set
+-----------------------
+
+When starting SITL the first time, the device will be configured with
+default parameters for the vehicle/frame type. Typically, you will want to replace these with
+values that simulate the desired vehicle and environment conditions.
+
+::
+
+    param load <path to parameter file>
+
+the working directory is the directory in which MAVProxy starts, not the simulation
+
+You can re-load the parameters later if you choose, or revert to the
+default parameters by starting SITL (**sim_vehicle.py**) with the
+``-w`` flag.
+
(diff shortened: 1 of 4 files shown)


dev/source/docs/SITL_simulation_parameters.rst
```diff
@@ -0,0 +1,192 @@
...
+   When you change a parameter the value remains in the virtual EEPROM
+   after you restart SITL. Remember to change it back if you don't want it
+   any more (or :ref:`reload/reset the parameters <using-sitl-for-ardupilot-testing_loading_a_parameter_set>`). 
+
+
+Changing the speed of the simulation
+------------------------------------
+
+Most of the simulator backends (native, RealFlight,etc) support changing the speed while
+running (NOT Realflight). Just set the ``SIM_SPEEDUP`` parameter as needed. A value of 1
```
peterbarker, line 39: Better make up your mind on RealFlight :-)

dev/source/docs/SITL_simulation_parameters.rst
```diff
@@ -0,0 +1,192 @@
...
+Loading a parameter set
+-----------------------
+
+When starting SITL the first time, the device will be configured with
+default parameters for the vehicle/frame type. Typically, you will want to replace these with
+values that simulate the desired vehicle and environment conditions.
+
+::
+
+    param load <<path to parameter file>
```
peterbarker, line 54:
```suggestion
    param load <path to parameter file>
```

dev/source/docs/SITL_simulation_parameters.rst
```diff
@@ -0,0 +1,192 @@
...
+``-w`` flag.
+
+.. note:: You may have to load them again, after typing ``param fetch``, in order to load parameters that require enabling before presenting their parameter set.  And, in some cases, you may even need to restart SITL in order for some new parameters, such as output function changes, to take effect. This can be avoided if you add the parameter file during the start of SITL with the "--add-param-file=*pathtofile* "....ie: sim_vehicle.py -f flightaxis:192.168.x.x - -map - -console --add-param-file=*pathtofile* -w. This adds the param file as a default and then wipes any previous param changes that may exist in the simulation directory. 
+
+Parameters can also be saved. For example, to save the parameters into
+the present working directory you might do:
+
+::
+
+    param save ./myparams.parm
```
peterbarker, line 69: Perhaps mention `param help` for other options?

dev/source/docs/SITL_simulation_parameters.rst
```diff
@@ -0,0 +1,192 @@
...
+
+.. note:: the "default" parameter set emulates the parameter default values contained in the firmware. These used at startup for parameters, UNLESS the user has previously changed them for a parameter(s), in which case those changed values are used. In SITL, these changes are stored in a file in the simulation startup directory named ``eeprom.bin``. The above commands loads user set values for the parameters, not the default values created at simulation start.
```
peterbarker, line 71:
```suggestion
.. note:: the "default" parameter set emulates the parameter default values contained in the firmware. These are used at startup for parameters, UNLESS the user has previously changed them for a parameter(s), in which case those changed values are used. In SITL, these changes are stored in a file in the simulation startup directory named ``eeprom.bin``. The above commands loads user set values for the parameters, not the default values created at simulation start.
```

I think this should be rewritten - I'm having trouble following it...

dev/source/docs/SITL_simulation_parameters.rst
```diff
@@ -0,0 +1,192 @@
...
+
+    param set SIM_RC_FAIL 1
+
+This simulates the complete loss of RC input. If you just want to
+simulate low throttle (below throttle failsafe level) then you can do
+that with the RC command:
+
+::
+
+    rc 3 900
```
peterbarker, line 89: There's actually a `SIM_RC_FAIL` value of 2 which is used to simulate receiver-goes-to-bind-time-values.

dev/source/docs/adding_simulated_devices.rst
```diff
@@ -0,0 +1,319 @@
...
+
+Adding a Gimbal
+===============
+
+SITL can simulate a virtual gimbal.
+
+.. note::
+
+   Gimbal simulation causes SITL to start sending
+   `MOUNT_STATUS <https://mavlink.io/en/messages/ardupilotmega.html#MOUNT_STATUS>`__
```
peterbarker, line 29: This is no longer true; we send a `GIMBAL_DEVICE_ATTITUDE_STATUS` message.  And it only has a quaternion in it, so good luck! :-)

( @rmackay9 )

dev/source/docs/adding_simulated_devices.rst
```diff
@@ -0,0 +1,319 @@
...
+
+::
+
+   ls -lt logs
+
+Choose the youngest, then:
+
+::
+
+   mavlogdump --type PL logs/<youngest>
```
peterbarker, line 225:
```suggestion
   mavlogdump.py --type PL logs/<youngest>
```

peterbarker (changes requested):
I wonder if we should mention the comments I've been dropping into the simulated devices to aid in use them.  

e.g. https://github.com/ardupilot/ardupilot/blob/master/libraries/SITL/SIM_RF_Lanbao.h#L18

dev/source/docs/SITL_simulation_parameters.rst
Hwurzburg: fixed

dev/source/docs/SITL_simulation_parameters.rst
Hwurzburg: existing...fixed

dev/source/docs/SITL_simulation_parameters.rst
Hwurzburg: added

dev/source/docs/SITL_simulation_parameters.rst
Hwurzburg: added....actually I misunderstood this when updating the RC_FAIL param....will fix that now

dev/source/docs/adding_simulated_devices.rst
Hwurzburg: can fix later once Randy updates

dev/source/docs/adding_simulated_devices.rst
Hwurzburg: fixed

Hwurzburg:
> I wonder if we should mention the comments I've been dropping into the simulated devices to aid in use them.
> 
> e.g. https://github.com/ardupilot/ardupilot/blob/master/libraries/SITL/SIM_RF_Lanbao.h#L18

so are the rangefinder instructions incorrect? or just incomplete for some RF? as to mentions, not sure how to definitively instruct someone since we have a mix of places the added devices exist, right?

dev/source/docs/SITL_simulation_parameters.rst
Hwurzburg: done

peterbarker:
> 

The rangefinder instructions use the in-built analogue rangefinder, but we simulate pretty much every rangefinder we support - so we get testing of the protocol parsers.  I don't think we simulate their maximum ranges, however.

PR #4863, opened 2023-01-07
mavproxy.py performs poorly on slow PIs

timtuxworth:
Added a note about performance of mavproxy.py on low powered PIs. mavlink-routerd works much better.

### DIFF
--- a/dev/source/docs/raspberry-pi-via-mavlink.rst
+++ b/dev/source/docs/raspberry-pi-via-mavlink.rst
@@ -116,6 +116,11 @@ MAVProxy
 MAVProxy can be used to send commands to the flight controller from the Pi. 
 It can also be used to route telemetry to other network endpoints.
 
+.. note::
+
+    If the Raspberry PI is heavily loaded, mavproxy.py might not provide a reliable connecton. This is more likely on older/slower
+    devices like the Raspberry PI Zero. If this happens, consider using mavlink-routerd (see below).
+
 This assumes you have a SSH connection to the Pi. If not, see see the 
 the `RPi Documentation <https://www.raspberrypi.org/documentation/remote-access/ssh/>`__.
 
--- a/dev/source/docs/raspberry-pi-via-mavlink.rst
+++ b/dev/source/docs/raspberry-pi-via-mavlink.rst
@@ -116,11 +116,6 @@ MAVProxy
 MAVProxy can be used to send commands to the flight controller from the Pi. 
 It can also be used to route telemetry to other network endpoints.
 
-.. note::
-
-    If the Raspberry PI is heavily loaded, mavproxy.py might not provide a reliable connecton. This is more likely on older/slower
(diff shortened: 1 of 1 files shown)


timtuxworth:
I'm not sure this looks how I intended. I was trying to make a "note:' in set like the one a few lines above.

peterbarker:
There are other options apart from "use `mavlink-routerd` (I don't mind `mavlink-routerd` as a tool, mind you).

The most important one it to unload any modules you do not need.  We load a lot by default, and if you're only using mavproxy to shuffle packets around then we will be doing a *lot* of extra work.

timtuxworth:
> There are other options apart from "use `mavlink-routerd` (I don't mind `mavlink-routerd` as a tool, mind you).
> 
> The most important one it to unload any modules you do not need. We load a lot by default, and if you're only using mavproxy to shuffle packets around then we will be doing a _lot_ of extra work.

How would I know which modules I need?

khancyr:
@timtuxworth I wrote a blog post about this some time ago : https://discuss.ardupilot.org/t/mavlink-routing-with-a-router-software/82138#solution-1-3

Hwurzburg:
this PR should link that blog post instead of the current wording...nice post Pierre

timtuxworth:
If I use mavproxy.py --default-modules='' link will be loaded anyway (but nothing else). 
But it didn't help, CPU maxed out at 100%, parameter download on connection fails every time. 

Whereas using mavlink-routerd parameter download completes successfully in 3 seconds flat, using about 15% of the CPU on the pi.

timtuxworth:
> this PR should link that blog post instead of the current wording...nice post Pierre

Yes, agreed - it's very comprehensive.

timtuxworth:
> > If I use mavproxy.py --default-modules='' link will be loaded anyway (but nothing else). But it didn't help, CPU maxed out at 100%, parameter download on connection fails every time after about 45 seconds using QGroundControl.
> > Whereas using mavlink-routerd parameter download completes successfully in 3 seconds flat, using about 15% of the CPU on the pi.
> 
> that does not stop the interactive shell creation....try the complete command

I did - specifically 
`mavproxy.py --daemon --default-modules='' --continue --master=/dev/serial0 --baudrate 1500000 --out=udp:pro:14550 --aircraft Binary`

timtuxworth:
> this PR should link that blog post instead of the current wording...nice post Pierre

Added - it's a new commit.

timtuxworth:
I think my "note" was in the wrong place. It was mixing the discussion on routing in with the initial discussion on how to set up mavproxy.py and getting it to work. I've moved it down to the bottom of the MAVProxy section, added an example and also moved the "note" that I started with - which now has a home, as it hangs off the explanation of mavproxy telemetry routing. It's a 3rd commit.

Hwurzburg:
LGTM...merging

PR #4877, opened 2023-01-17
Update mission-planner-advanced-installation.rst

nkruzan:
Remove very obsolete note about requiring winzip. Zip decompression has been built into Windows since at least win7.


Hwurzburg:
LGTM...thanks

PR #4880, opened 2023-01-18
update AIRMODE page

Hwurzburg:
@bnsgeyer Bill, just want to be sure Trad Heli still does not use AIRMODE per note here....


bnsgeyer:
@Hwurzburg as far as I know, heli’s can’t select airmode.

andyp1per (approved):
funciton typo

PR #4887, opened 2023-01-19
Update common-servo.rst for digital servos

timtuxworth:
Digital servos support higher update rates using the SERVO_RATE parameter, but this is only a good idea if all the servos on a vehicle are digital

### DIFF
--- a/common/source/docs/common-servo.rst
+++ b/common/source/docs/common-servo.rst
@@ -15,15 +15,15 @@ Either
 or `digital servos <https://www.amazon.com/Hitec-RCD-35076S-HS-A5076HB-Digital/dp/B003RXHMNK>`__
 may be used.
 
-Connecting the servo to a Pixhawk
-=================================
+Connecting the servo to an AutoPilot
+====================================
 
 .. image:: ../../../images/Servo_Pixhawk.jpg
     :target: ../_images/Servo_Pixhawk.jpg
 
 -  If using Copter Connect the servo to AUX OUT 1, 2, 3 or 4.  MAIN OUT
    1 ~ 8 should be avoided because these update at 400hz.
--  If using Plane or Rover where all pins update at 50hz, any unused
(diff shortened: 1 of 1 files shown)


common/source/docs/common-servo.rst
```diff
@@ -15,15 +15,15 @@ Either
...
+Connecting the servo to an AutoPilot
+====================================
 
 .. image:: ../../../images/Servo_Pixhawk.jpg
     :target: ../_images/Servo_Pixhawk.jpg
 
 -  If using Copter Connect the servo to AUX OUT 1, 2, 3 or 4.  MAIN OUT
    1 ~ 8 should be avoided because these update at 400hz.
--  If using Plane or Rover where all pins update at 50hz, any unused
+-  If using Plane or Rover where all pins update at ``SERVO_RATE`` which defaults to 50hz, any unused
```
Hwurzburg, line 26:
```suggestion
-  If using Plane or Rover where all PWMN outputs update at :ref:`SERVO_RATE<SERVO_RATE>`, which defaults to 50hz, any unused
```

Hwurzburg (changes requested):
a few clarifying nit picks

timtuxworth:
@Hwurzburg - unfortunately my change from "PixHawk" to "AutoPilot" - makes some of the details in this section about setting up a standard autopilot incorrect I think. I thought it made sense to change it because many people these days are using a variety of other controllers, but the comments about limitations/defaults on the relay pinouts are not correct as they are specific to the PixHawk (and probably only the PixHawk 1). Should I change this too, or leave it for a separate PR?

Hwurzburg:
nah, we talk about MAIN and AUX a lot of places and that applies to most of the high end autopilots that use an IOMCU, so its fine...I will squash this and merge tomorrow when I have time....thanks

Hwurzburg:
squashed and merging...thanks

PR #4899, opened 2023-01-26
Plane: QRTL: add interactive return altitude graph

IamPete1:
Adds a interactive graph that changes with user inputs to show return altitude.

image

Changing Q_RTL_ALT:
image

Someone who knows HTML better than I do could probably make it neater. Its possible there are more integrated ways to do this, I have just embeded HTML into the page and the script into the HTML. 

I do think this could be quite a powerful way to show things. It also provides a graph that could be copied into a operations manual.


IamPete1:
I should probably add that the feature that this shows is only in master ATM.

Hwurzburg:
looks good! I would put this on the future-wiki-changes.rst under the Plane site....I can help if you wish...also, is the gitignore change part of this, I think not, right?might tweak the size of the legend lines....

could we do one of these for the hybrid RTL also??? pretty please....

IamPete1:
@Hwurzburg yeah, gitignore changes are unrelated, I just had a bunch of new stuff show up after a build.

Yeah I can help with more, would be good to see if there is a way to make this more integrated, just including a bunch of HTML is not ideal, having said that this does seem to work..... I just wanted to see if it was possible to add interactive elements.

IamPete1:
I did a little reading, I think we could make it so you could insert the canvas like a image and then insert each param with user input box. You would then just need to write the custom javascript stuff to populate it. However to do that we would need to write a custom sphinx extension with I have never done, so it might not be worth it.....,

IamPete1:
After a good deal of messing about I have improved the formatting of the labels and removed most of the boiler plate html that was embeded.

Now looks like this:
image

Formatting does get messed up with a narrow window doing wrapping.

image

Does come good again if you go very narrow....
image

I have no idea how to fix that..... I think its better than starting out with one on each line.

IamPete1:
Getting carried away now....

I have added a return path, you can click the graph and it show the path home. 

image

image

image

IamPete1:
Moved to future section. Got the legend looking better and fixed the axis lables.

image

If you click it adds the return path. Sometimes with FW sometimes not:

image

image

image

Hwurzburg:
@IamPete1 I think the second graph is wrong....long FW return shoots for Q_RTL_ALT at 1.5X maxrad, then airbrakes in POS1, moves to POS2, then decends, correct?

IamPete1:
@Hwurzburg I have removed the sloped FW return, so now it looks like this:
image

There is other stuff that will happen at larger radius's. It will climb up to RTL alt and then descend down again. But that all happens between 2 and 20 times the radius. This graph stops at 2 times the radius. I don't think we can show that on this graph without making the bit close to home very tiny.

We could show that on a second graph of course. Or make the max display radius display configurable but default to as it is now.

Hwurzburg:
@IamPete1 lets get rid of anyting FW because what is shown is not accurate....lets just talk to what happens  beyond 2X ...ie transition to FW and do a hybrid RTL

IamPete1:
@Hwurzburg FW is correct now more or less, once you start the transition it will hold altitude. So there might be a little wobble at the start but then it will maintain altitude.

Hwurzburg:
Sorry, but that is not correct....here is what your current graph in this PR shows if you switch to QRTL from VTOL at alt above Q_RTL_ALT and beyond 1.5X:
Screenshot from 2023-02-21 07-39-13
here is what it actually does in SITL (as I would expect):
qrtl_fs_return

lets just remove FW anything from the graph and make the right margin just under 1.5X from home...it will be less confusing

Hwurzburg (changes requested):
lets remove the FW line

Hwurzburg:
@IamPete1 bump....if we get rid of the FW return line we could add this

Hwurzburg:
@IamPete1 closing this since its been forever and never worked on...feel free to reopen

PR #4904, opened 2023-01-28
Common: fix torqeedo typo

rmackay9:
replaced 483 with 485

This has been tested on my local machine and seems ok


Hwurzburg:
thanks

PR #4911, opened 2023-01-30
update.py: Reprint update.py error messages at the end

TunaLobster:


Hwurzburg:
something is wrong with this....I tested locally by creating an error in the airspeed sensor page and building plane and do get the report:
Reprinting error messages:
[update.py][error]: Remove character before ref directive in "common/source/docs/common-airspeed-sensor.rst" on line number 9
[update.py][error]: 1 errors during Wiki build

but then fixing it and re-running a build for plane I get this:
Build finished. The HTML pages are in build/html.
Reprinting error messages:
[update.py][error]: Remove character before ref directive in "rover/source/docs/common-airspeed-sensor.rst" on line number 9
[update.py][error]: Remove character before ref directive in "copter/source/docs/common-airspeed-sensor.rst" on line number 9
[update.py][error]: 2 errors during Wiki build

so something is broken
also, if you can print the report on the std error output, it will appear in my daily cron job that I check the build daily for build errors (although I don't look EVERY day at it...maybe every other day)

update.py
```diff
@@ -1068,8 +1074,12 @@ def create_features_page(features, build_options_by_define, vehicletype):
     # --allow-file-access-from-files". Otherwise it will appear empty
     # locally and working once is on the server.
 
+    error_count = len(error_log)
     if error_count > 0:
-        fatal("%u errors during Wiki build" % (error_count,))
+        print("Reprinting error messages:", file=sys.stderr)
+        for msg in error_log:
+            print(f"[update.py][error]: {msg}", file=sys.stderr)
```
Hwurzburg, line 1081:
```suggestion
            print(f"\033[1;31m[update.py][error]: {msg}", file=sys.stderr)
```

Hwurzburg (changes requested):
looks good....one small change to make it perfect

TunaLobster:
IIRC, color might be hit or miss in some terminals. Added the color reset tag to the end of the string.

Hwurzburg:
thanks

PR #4912, opened 2023-01-30
Common: add djirs2 gimbal support to upcoming features

rmackay9:
This adds a wiki page to go with the new DJI RS2 gimbal support. https://github.com/ArduPilot/ardupilot/pull/22776

Note that the Siyi A8 support has been released with 4.3 so I've removed it from the upcoming features.


Hwurzburg:
@rmackay9 we need to understand why your local build is not showing the significant wiki build error that this creates when you tested this PR locally...it leaves the SIYI page "floating" (not in a TOCTREE)....it must be linked on the camera and gimbals page in the bottom toctree....

Hwurzburg (changes requested):
fix toctree build error

rmackay9:
@Hwurzburg,

Txs for the review.  I've fixed the missing toc now and I don't see build errors any more.

Hwurzburg:
thanks...

PR #4913, opened 2023-01-30
Rover: link maker boat advanced from reference frames page

rmackay9:
This reference frame is actively maintained as can be seen from [this discussion](https://discuss.ardupilot.org/t/maker-boat-advanced-build-info/73585).

I've tested this on my local machine and it seems fine.


Hwurzburg:
thanks..

PR #4917, opened 2023-02-01
Add note about Vicon being available on 1MB boards

MichelleRos:
Vicon can still be used on 1MB boards.


Hwurzburg:
LGTM...thanks

PR #4918, opened 2023-02-02
Common: T265 setup recommends setting serialx_options

rmackay9:
This comes from a report on my YouTube channel in which a user said their vehicle's telemetry stopped working after they turned on the RPI/T265.

I've also removed some old 4.0 instructions and clarified that the GPS can be left enabled if users wish to have it there.  Leaving it enabled means users may be able to skip setting the home location.

I have tested this on my local machine and it seems ok and no build errors.


Hwurzburg:
LGTM...thanks

PR #4919, opened 2023-02-03
Add ThePeach FCC-K1, FCC-R1

ThePeach-kr:
Sorry for the delay.

We add rst files and images for ThePeach FCC-K1/R1.

Thank you. @Hwurzburg


Hwurzburg:
Needs a few tweaks, but I will do it after I merge...thanks

PR #4925, opened 2023-02-04
Update guide-four-channel-plane.rst

aliyasirh7:
~ The Rudder movement controls the Yaw of a standard 4 channel plane configuration. It does not make the plane roll.

The correct control configuration of the rudder is as 👇:
 - Yaw Plane Right  |  Rudder moves Right
 - Yaw Plane Left    |  Rudder moves Left


Hwurzburg:
sorry but this is actually incorrect....the section is talking about how the surfaces move in FBWA with the default values for KF_RDDRMIX which moves the rudder, together with the ailerons, to counteract any bank away from level in FBWA...there is a note further down that if you do NOT want rudder involved in the roll axis stabilization, then set it to zero.

PR #4934, opened 2023-02-09
MAVPRoxy documentation arm throttle force

timtuxworth:
Document how to force arm in mavproxy


Hwurzburg:
might as well put disarm force in also ....

tridge (approved):

PR #4942, opened 2023-02-15
Fix for SITL build

phokur:
./waf configure --board=sitl --toolchain i686-pc-cygwin was causing a crash because it couldn't find i686-pc-cygwin. Adding these links allows the build script to complete as written.


magicrub:
> i686-pc-cygwin has been deprecated from cygwin

To add to that, at some cygwin will be deprecated from ArduPilot in favor of WSL.

magicrub:
@rmackay9 are you still using cygwin that you can test this?

magicrub:
Actually, @meee1 would be interested in this because I think Mission Planner relies cygwin a little bit for SITL

Hwurzburg:
different thing....MP takes a github action build from CI....this is a developer thing

khancyr:
Setting x64 as i686 is not a good things IMHO.
I have test that the x64 build is building following CI instructions so this PR doesn't look right

robertlong13:
This whole section "Build SITL binaries for Windows" can be deleted, I think. This section is for building 32-bit cygwin binaries, which nobody can do now (as the author notes, this is being deprecated). As I understand it, this documentation PR has the user create symlinks so that the 64-bit compiler runs when you ask waf to build with the 32-bit toolchain. That's not necessary.  Just run with the default build options in cygwin and it will create the 64-bit binaries that Mission Planner now uses (at least, MP beta).

phokur:
As much as I prefer WSL as well, I believe this is the only way to get the Ardupilot.exe build for SITL locally.

phokur:
> 

That is interesting! I think it'd still be useful to keep some mention of using cygwin for Mission Planner's exe SITL builds though.

PR #4945, opened 2023-02-17
add PixPilot-V3 and PixSurveyA2

Cedric0489:


Hwurzburg (changes requested):
nice...thanks...both should have a section on Loading Firmware like this:
https://ardupilot.org/plane/docs/common-holybro-pixhawk6C.html#loading-firmware

PR #4948, opened 2023-02-19
Note for dynamic harmonics and narrower bandwidth

MichelleRos:
Add a note about dynamic notches and latency. 

@tridge @andyp1per let me know if this makes sense. 
Does it nead a suggested notch width?

### DIFF
--- a/common/source/docs/common-esc-telem-based-notch.rst
+++ b/common/source/docs/common-esc-telem-based-notch.rst
@@ -16,6 +16,12 @@ If ESCs with telemetry are used (either via a separate telemety wire or bi-direc
 - Set :ref:`INS_HNTCH_FREQ <INS_HNTCH_FREQ>` and/or :ref:`INS_HNTC2_FREQ <INS_HNTC2_FREQ>` = below the hover frequency - you can easily determine this by performing a gentle hover and looking at the ESC telemetry data
 - Set :ref:`INS_HNTCH_BW <INS_HNTCH_BW>` and/or :ref:`INS_HNTC2_BW <INS_HNTC2_BW>` = half of INS_HNTCH_FREQ
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


andyp1per:
Yes I recommend starting with f/4 in this instance

andyp1per (approved):

common/source/docs/common-esc-telem-based-notch.rst
```diff
@@ -16,14 +16,20 @@ If ESCs with telemetry are used (either via a separate telemety wire or bi-direc
 - Set :ref:`INS_HNTCH_FREQ <INS_HNTCH_FREQ>` and/or :ref:`INS_HNTC2_FREQ <INS_HNTC2_FREQ>` = below the hover frequency - you can easily determine this by performing a gentle hover and looking at the ESC telemetry data
 - Set :ref:`INS_HNTCH_BW <INS_HNTCH_BW>` and/or :ref:`INS_HNTC2_BW <INS_HNTC2_BW>` = half of INS_HNTCH_FREQ
 
+.. note::
+
+   If you set the dynamic harmonics option (i.e. a notch per motor) mentioned below, the bandwidth should not be half the frequency. It should be greatly reduced as more notches cause more phase lag (i.e. latency), thus the bandwidth needs to be reduced to maintain a reasonable phase lag, else oscillation and a poorer tune will result despite the more accurate filtering compared to throttle-based filtering.
```
Hwurzburg, line 21:
```suggestion
   If you set the dynamic harmonics option (i.e. a notch per motor) mentioned below, the bandwidth should not be half the frequency. It should be greatly reduced as more notches cause more phase lag (i.e. latency), thus the bandwidth needs to be reduced to maintain a reasonable phase lag, else oscillation and a poorer tune will result despite the more accurate filtering compared to throttle-based filtering. For example, with four motors, :ref:`INS_HNTCH_FREQ<INS_HNTCH_FREQ>` / 4 is suggested as a starting point.
```

Hwurzburg (changes requested):
per Andy's comment

MichelleRos:
I've updated it.

Hwurzburg:
thanks...

PR #4949, opened 2023-02-20
Use autopilot-agnostic language

peterbarker:


rmackay9:
Looks good to me.  I wonder if there is a difference between "autopilot" and "flight controller" and whether we should tend to use one or the other.

Hwurzburg:
I try to use autopilot everywhere....Antenna Trackers and Rovers usually do not fly

nkruzan:
Nice to see tracker getting some love ;)

peterbarker:
I've changed this to just be "autopilot", not "flight controller".

Hwurzburg:
thanks!

PR #4954, opened 2023-02-21
tradheli-update for angle P autotune test

bnsgeyer:
This PR updates the autotune wiki to include changes that were made to the angle p autotune test (#20780).  This change included a tune check test to allow the user check the tune once they completed the autotune testing.

### DIFF
--- a/copter/source/docs/traditional-helicopter-autotune.rst
+++ b/copter/source/docs/traditional-helicopter-autotune.rst
@@ -38,8 +38,8 @@ User can specify the tuning desired using the :ref:`AUTOTUNE_SEQ<AUTOTUNE_SEQ>` 
 | +-------+------------------------------------------+| |         |  & ``ATC_RAT_xxx_P``                   ||
 | | 3     | Max Gain Determination                   || +---------+----------------------------------------+|
 | +-------+------------------------------------------+| | 4       | ``ATC_ANG_xxx_P``                      ||
-|                                                     | +---------+----------------------------------------+|
-|                                                     | | 5       | ``ATC_RAT_xxx_FF`` & ``ATC_ANG_xxx_P`` ||
+| | 4     | Tune Check                               || +---------+----------------------------------------+|
+| +-------+------------------------------------------+| | 5       | ``ATC_RAT_xxx_FF`` & ``ATC_ANG_xxx_P`` ||
 |                                                     | +---------+----------------------------------------+|
 |                                                     | | 6       | ``ATC_RAT_xxx_D``, ``ATC_RAT_xxx_P``,  ||
 |                                                     | |         | & ``ATC_ANG_xxx_P``                    ||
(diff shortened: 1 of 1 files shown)


copter/source/docs/traditional-helicopter-autotune.rst
```diff
@@ -49,6 +49,8 @@ User can specify the tuning desired using the :ref:`AUTOTUNE_SEQ<AUTOTUNE_SEQ>`
 |                                                     | +---------+----------------------------------------+|
 |                                                     | | 8       | Max Gain Determination                 ||
 |                                                     | +---------+----------------------------------------+|
+|                                                     | | 16      | Tune Check                             ||
+|                                                     | +---------+----------------------------------------+|
 +-----------------------------------------------------+-----------------------------------------------------+
 
 ** Max gain determination is always done before tuning Rate D and Rate P gains. This test determines the maximum allow values of ``ATC_RAT_xxx_D`` & ``ATC_RAT_xxx_P`` so that the helicopter never experiences servere oscillations due to raising these gains too high.  Since the maximum allowable values of ``ATC_RAT_xxx_D`` & ``ATC_RAT_xxx_P`` are found, tuning is done by incrementing these gains by 5% of their maximum allowable value.
```
Hwurzburg, line 56: Evidently combinations of bits 3 or 4 with the others are not allowed...are there code protections?in any case, it should be mentioned that bit 3 or 4 must be set without any other bits, or the precedence explained...

Hwurzburg (changes requested):

copter/source/docs/traditional-helicopter-autotune.rst
bnsgeyer: @Hwurzburg bits 3 and 4 just complementary tests. Bit 3 is done as part of the rate D and rate P test.  Bit 3 is just provided for people that just want to see what the max gains are without actually tuning. Bit 4 is provided for people that want to evaluate the final tune. It can be combined with others without harm.

Hwurzburg: then we should state that...the values listed imply otherwise

Hwurzburg: just to be clear, any or all of the bits can be set in any combination

bnsgeyer: Yes that is true but it doesn't make sense from a tuning perspective to do it that way.  The combinations that I have shown to this point are combinations that I would expect to be used.  The third bit is just for information.  So it could be used with lets say the FF test if the user just wanted to investigate the limit of the rate P and D gains after tuning the FF gain.  You would not use it with the Rate P and D gain because it is already part of this test.

I would really prefer you just clarify what you meant above for the 3rd and 4th bits and leave the other combinations as they are.

Hwurzburg: I think we need to be clearer about when or when not you should set bit 3 in the MAX GAIN section....and what to look at in the logs for TUNE CHECK....is it just the items in the LOG Analysis section? if so, we need to directly state that in the TUNE CHECK section

PR #4962, opened 2023-02-23
update cube black and orange;include orange+

Hwurzburg:


TunaLobster:
I might have done `(+)`, but the `/+` looks fine too. I do appreciate seeing the ADC pin numbers in the table. That would have been helpful to have as a quick reference many times in the past.

PR #4967, opened 2023-02-25
Add UM982 Moving Baseline

Hwurzburg:


tridge (dismissed):
the diagrams show "MAG_" where it should be "GPS_"
There is also a "!" where there should be a "1" in an offset

Hwurzburg:
fixed...thanks

PR #4968, opened 2023-02-26
Remove roverauto docking from GSOC ideas list

shiv-tyagi:
This project was successfully completed during last summer. Having this here is creating confusion among potential gsoc students. Many students have been repeatedly asking to take this project for GSoC 2023. Hence, we should remove it from here to avoid any further confusion.


Hwurzburg:
Thanks

peterbarker:
We do have a "completed projects" section which is becoming increasingly misleading :-)

Might be worth moving some of this data rather than just removing it in future.

PR #4972, opened 2023-02-26
clarify and expand safety state operation

Hwurzburg:


common/source/docs/common-safety-switch-pixhawk.rst
```diff
@@ -7,25 +7,27 @@ Safety Switch
 .. image:: ../../../images/pixhawk_safetyswitch.jpg
     :target: ../_images/pixhawk_safetyswitch.jpg
 
-A safety switch can be used to enable/disable the outputs to motors and servos.
+A safety switch can be used to enable/disable the outputs to motors and servos. The switch controls the "Safety" state of the vehicle. When in this state, motors are prevented from operating (except in Plane's MANUAL mode if force armed), a pre-arm error condition is generated to prevent accidental arming, and servo outputs have there PWM disabled.
```
tridge, line 10: safety state stops motors and servos even in manual. The exception you are thinking of is because the SAFETY_MASK was set to allow motors and servos to bypass the safety state. In that case the outputs where safety mask is set will allow for control surface movement and manual throttle if force armed.

tridge (changes requested):

common/source/docs/common-safety-switch-pixhawk.rst
Hwurzburg: thanks...fixed

PR #4985, opened 2023-03-03
Add DO_JUMP_TAG cmds

Hwurzburg:
when it appears in MP I will capture a screenshot


rmackay9:
My guess is that someone will need to do a PR to MP to get these DO_JUMP_TAG commands appearing in the command list.  It's not hard to do though.  FYI @magicrub

PR #4995, opened 2023-03-06
add mission option for batt FS distance calc

Hwurzburg:
I recently came across this undocumented option...investigated the code and I think this is the correct documentation...please review and approve or suggest changes...thanks


Hwurzburg:
yep...it got accidentaly added as part of another PR..note that it has conflicts now.....I started that one( #5012) based off this one by mistake...I will close this PR...if there is an issue with the note on MIS_OPTIONS bit 1

Gone4Dirt:
The comment as you have it looks correct to me anyhow, so all good anyway.

PR #5004, opened 2023-03-09
Update Installation Instructions for Linux

billbonney:


Hwurzburg:
I dont think this is correct....all APMPlanner releases should be via the firmware server, I think....  if they are not there, then there is a problem...@peterbarker please confirm....thanks

billbonney:
> I dont think this is correct....all APMPlanner releases should be via the firmware server, I think.... if they are not there, then there is a problem...@peterbarker please confirm....thanks

Why would that be a problem? You requested the links be updated. And this matches that request.

see https://github.com/ArduPilot/apm_planner/issues/1280

planner2/source/docs/installation-for-linux.rst
```diff
@@ -18,9 +18,11 @@ System Requirements
 Download
 --------
 
-Download the latest deb file for your machine from 
+Download the latest deb file for your machine from GitHub.
```
peterbarker, line 21: This doesn't look right - looks like it's all AppImage now.

peterbarker (changes requested):
@Hwurzburg releasing straight-from-github is a legitimate project thing now-adays.

I reckon it's a change, so I've marked it DevCallTopic to make sure everyone's aware - but I'm pretty sure everyone will be, "Someone's doing fresh releases?!  Great stuff!" and we'll want to merge this :-)

peterbarker:
@billbonney I've dismissed my stale review.  I wonder if the mention of OpenSuse repo could also go away - I guess it's probably very stale.  Not trying to stretch this PR, just observing.

billbonney:
> @billbonney I've dismissed my stale review. I wonder if the mention of OpenSuse repo could also go away - I guess it's probably very stale. Not trying to stretch this PR, just observing.

Yep, we can remove that. Since AppImages are the 'run anywhere' solution

PR #5005, opened 2023-03-09
Update Install Instructions macOS

billbonney:


Hwurzburg:
same comment here as other PR

PR #5019, opened 2023-03-14
Cube ID and Cube ID_CAN

chrisksl:
Add Cube ID and Cube ID_CAN to the list of "OpenDroneID Compatible devices"


Hwurzburg:
LGTM...thanks

PR #5028, opened 2023-03-17
Copter: deadreckoning link to lua script updated

rmackay9:
lua script was moved from examples directory to applets directory.

Ideally this should be merged soon after this flight code PR goes in https://github.com/ArduPilot/ardupilot/pull/23233


Hwurzburg:
Need to add to common-scripting-applets also

Hwurzburg:
thanks

PR #5043, opened 2023-04-01
Common: Add Siyi A8 warning

tajisoft:
It does not support zoom at 4K resolution.

Here is the [A8 mini User Manual v1.2](https://drive.google.com/drive/folders/19Hs92ClHiRWIjE9xS32Y6GToXNvGveLO)

```
Record Resolution: Switch camera record resolution between HD (720p), Full HD (1080p),
2K, and 4K (under 4K recording resolution, A8 mini camera does not zoom).
```


rmackay9 (approved):
Great, thanks for this!

PR #5044, opened 2023-04-02
Cartographer: Fix syntax problems with launch file. Avoid modifying code.

zhouzhiwen2000:
This PR fixes some syntax problems with the launch file and avoids modifying robot_pose_publisher.cpp and mavros's node.launch.
Avoiding direct modifications to codes can improve forward compatibility.


rmackay9:
Thanks for this.  I suspect this is correct.  Certainly the missing "=" seems like a bug fix.

@jmachuca77, @srmainwaring, what do you think?

srmainwaring:
> Certainly the missing "=" seems like a bug fix.

Agree.

I don't have a working ROS (1) environment to run this up on at present so can't confirm it's working (almost exclusively using ROS 2 now). I'll see if I can set one up in VM later.

rmackay9 (approved):
I think we should go ahead and merge this.  I'm unsure if it is completely correct but it at least fixes some issues.

PR #5045, opened 2023-04-04
Use recurse-submodules when cloning ardupilot

Ryanf55:
This flag is available on Ubuntu 18.04 and up. It reduces the number of commands to type


Ryanf55:
> did you leave out https://ardupilot.org/dev/docs/skyviper.html?highlight=skyviper intentionally?
> 
> also, do you think we should change #6 here:https://ardupilot.org/dev/docs/git-rebase.html to include the --init ? I just realized my alias for that command always has the --init included....probably not needed, but doesn't hurt....any case where someone might need it?

Yes, I can do both of these. Skyviper was an oversight; I will fix it. 

In step 6, since you have already cloned ardupilot earlier with submodules initialized, you don't need to pass in `--init` again. It seems there is no ill effect, I can add it in.

Ryanf55:
> did you leave out https://ardupilot.org/dev/docs/skyviper.html?highlight=skyviper intentionally?
> 
> also, do you think we should change #6 here:https://ardupilot.org/dev/docs/git-rebase.html to include the --init ? I just realized my alias for that command always has the --init included....probably not needed, but doesn't hurt....any case where someone might need it?

Skyviper fixed in 8ef3f2d37cfc745a1ba668591c57b21fea72063d
Rebase instructions fixed in 777be76017193c114262f4989069b7a552e2da72

common/source/docs/common-beagle-bone-blue.rst
```diff
@@ -287,7 +287,6 @@ use either the stable tag ``ArduPlane-stable`` or the current Plane branch ``Pla
 
     ./Tools/environment_install/install-prereqs-ubuntu.sh
     git checkout Plane-4.1
-    git submodule update --init --recursive
```
peterbarker, line 290: This doesn't look right; you've jsut checked out an older version of the software, a submodule update is often required,

peterbarker (changes requested):

common/source/docs/common-beagle-bone-blue.rst
Ryanf55: Proposal: init once, during clone. Then, just call update each time you change branches in this guide.

peterbarker: > Proposal: init once, during clone. Then, just call update each time you change branches in this guide.

I think that flow would break if we add a new submodule.

Pretty sure the `--init` is essentially free anyway.

I *do* like the premise of this PR - when cloning add the extra argument.   I *think* `--recurse` may be sufficient vs `--recurse-submodules`... less typing?

Ryanf55: > > Proposal: init once, during clone. Then, just call update each time you change branches in this guide.
> 
> I think that flow would break if we add a new submodule.
> 
> Pretty sure the `--init` is essentially free anyway.
> 
> I _do_ like the premise of this PR - when cloning add the extra argument. I _think_ `--recurse` may be sufficient vs `--recurse-submodules`... less typing?

Yes, I can add `--init` in other areas as needed in the guide. 

The `--recurse` option is not available for me under the `git clone --help` listing.

Ryanf55: > > Proposal: init once, during clone. Then, just call update each time you change branches in this guide.
> 
> I think that flow would break if we add a new submodule.
> 
> Pretty sure the `--init` is essentially free anyway.
> 
> I _do_ like the premise of this PR - when cloning add the extra argument. I _think_ `--recurse` may be sufficient vs `--recurse-submodules`... less typing?

Attempted to resolve this in efc85f176dd1638267812379e7ccb4ac5ce7e5ef.

peterbarker: > The `--recurse` option is not available for me under the `git clone --help` listing.

It works, 'though :-)  But not fussed.

dev/source/docs/git-rebase.rst
```diff
@@ -31,12 +31,14 @@ the "Git Shell" or "Git Bash" utility that should already be :ref:`installed <gi
    ::
 
        git checkout topic-branch
+       git submodule update --init --recusive
```
peterbarker, line 34:
```suggestion
       git submodule update --init --recursive
```

similarly elsewhere

dev/source/docs/skyviper.rst
```diff
@@ -40,15 +40,15 @@ Otherwise:
...
+   git clone --recurse-submodules https://github.com/SkyRocketToys/ardupilot
 
 Once you have the repository:
 
 ::
 
    cd ardupilot
    git checkout skyviper-stable # or skyviper-latest
-   git submodule update --init --recursive
+   git submodule update
```
peterbarker, line 51: switching branches here - need the init


... also, these steps are probably broken; better to run master?

peterbarker (changes requested):

dev/source/docs/skyviper.rst
Ryanf55: I'm not familiar with skyviper, but that branch is 30 commits in front of their master. I assume for good reason. 
https://github.com/SkyRocketToys/ardupilot/tree/skyviper-stable

Ryanf55: Added the init in 13d09cd0b61855ed5583280067b66d5fdd8e7b47

dev/source/docs/git-rebase.rst
Ryanf55: c700e88026d53e64d8089456516395321066f7ef

Hwurzburg (changes requested):
please squash...thanks

Ryanf55:
> please squash...thanks

done

Hwurzburg:
> > please squash...thanks
> 
> done

thanks, when @peterbarker approves, will merge

Hwurzburg (approved):

PR #5049, opened 2023-04-06
Copter: add another dead reckoning failsafe video

rmackay9:
This adds a new video to the Copter dead reckoning failsafe page.  I've tested this on my local machine and it looks OK.


Hwurzburg:
LGTM...thanks

PR #5054, opened 2023-04-07
dev: update for new support.ardupilot.org MissionPlanner UI

tridge:


Hwurzburg (commented):
are we sure that we wont want to explain the use of Mavlink Mirror in the future? you are removing the images for that. No need unless you are sure we don't need any wiki info on it...and if that's the case we should remove it as an option in Mission Planner, shouldn't we?

tridge:
we could separately document the other method, but not on the support.ardupilot.org doc page

Hwurzburg:
I would just not delete the images in case we wanted them in the future....

tridge:
> I would just not delete the images in case we wanted them in the future....

ok, done

Hwurzburg:
thanks! merging

PR #5061, opened 2023-04-14
Fix typo on doc landing page

jhairssteven:
Fix "... system can be **as a** short-range substitute..." to be "... system can be a short-range substitute..."


Hwurzburg:
lgtm...thanks

PR #5064, opened 2023-04-15
Use python3

Ryanf55:
This updates all the docs to use python3. There were a few tutorials which are super old and still use python2, so I didn't update them. 

I also removed the `python-is-python3`. That's a hack package IMO, let's just update the calls to `python3` and deprecate python2.


peterbarker (approved):
LGTM

PR #5072, opened 2023-04-20
Common: remove storm32 gimbal video

rmackay9:
This removes a very old video I made of the SToRM32 gimbal.  It's not a bad video but its very out of date now especially considering all the recent code and mission command changes.

This has been tested locally.


Hwurzburg:
thanks

PR #5074, opened 2023-04-22
sim: clarify a required setting in RealFlight

arikrupnik:
It is necessary to set "Pause Sim When in Background" to "No" for RealFlight to work with Mission Planner. Otherwise, the simulation stops as soon as you switch to MP (to arm, to set parameters or anything else) and MP cannot communicate to FlightAxis.

### DIFF
--- a/dev/source/docs/sitl-with-realflight.rst
+++ b/dev/source/docs/sitl-with-realflight.rst
@@ -25,7 +25,7 @@ See the last section on this page for setup instructions, if using an OpenTX tra
 Enabling RealFlight Link Feature
 ================================
 
-On RealFlight go to Settings->Physics and enable the FlightAxis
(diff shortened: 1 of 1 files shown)


Hwurzburg:
see line 48 further down....already covered in this article

arikrupnik:
I encourage you to either take the duplication out or duplicate it in full.

arikrupnik:
In master, "change the option for 'Pause Sim When in Background' to No" appears under "From within RealFlight, Reduce graphics options to improve performance." disable pause in background is not a performance improvement. It is a hard requirement. I recommend putting it under "Enabling RealFlight Link Feature." There is no harm in keeping it in both places.

Hwurzburg:
I will remove it lower down to avoid duplication after this merges...thanks

PR #5085, opened 2023-04-26
Update common-holybro-kakuteh7.rst

vincentpoont2:
Fix IMU model error.


Hwurzburg:
this does not match the hwdef for the board!

vincentpoont2:
I was told the name in the hwdef is irrelevant
The icm and mpu shares the same driver in ardupilot

Hwurzburg:
okay....will make this a V1 page...will also correct the hwdef and its readme....even if it uses the same driver the documentation should accurately reflect reality...

PR #5099, opened 2023-05-02
Video demo of send_text debugging

flatherz:


Hwurzburg:
@flatherz sorry for the slow response....as it happens we already had created some videos on debugging and this topic was actually the first one....been editing and they are now posted and linked....sorry again...please continue to contribute to the wiki!

PR #5113, opened 2023-05-08
Expand info on frame and class types for quadplane

Hwurzburg:
Co-Authored-by:Jet<taijunjet@hotmail.com>

### DIFF
--- a/plane/source/docs/quadplane-frame-setup.rst
+++ b/plane/source/docs/quadplane-frame-setup.rst
@@ -21,24 +21,34 @@ Frame Types and Classes
 -----------------------
 
 To use a different frame type you can set :ref:`Q_FRAME_CLASS<Q_FRAME_CLASS>` and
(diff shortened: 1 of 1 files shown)


plane/source/docs/quadplane-frame-setup.rst
```diff
@@ -21,24 +21,34 @@ Frame Types and Classes
...
 
 -  1 for quad
 -  2 for hexa
 -  3 for octa
 -  4 for octaquad
--  5 for Y6
--  7 for Tri
--  10 for Tailsitter
+-  5 for Y6 (ignores all following frame classes except 10 and 11, all others result in Y6A configuration)
+-  7 for tri (ignores all following frame classes except 6 for reversed tricopter)
```
kd0aij, line 36: shouldn't "ignores all following frame classes" be "ignores frame type"?
And class 10 completely ignores frame type, I think.

PR #5120, opened 2023-05-10
add example of  SERVO_AUTO_TRIM adjustiment range

Hwurzburg:


kd0aij:
Note that in your example of min/max = 1100/1900 that
if the initial trim value is not within the range [1420,1080] then auto trim will not function.
So there is an implicit assumption that optimal trim midway between min and max.

It would take an additional parameter to support a control surface which has "zero" trim outside of that range.
I assume that case is unusual or non-existent in the auto-trim universe.

Hwurzburg:
in my example you will note the initial trim is 1500

PR #5129, opened 2023-05-16
copter: add betaflight-x to connect-escs-and-motors

yuri-rage:
Added Betaflight/X frame type (image) to the ESC/motor connection page.
Also corrected a very minor typo.


Hwurzburg:
LGTM...probably should add the BF X REV too at some point

yuri-rage:
> LGTM...probably should add the BF X REV too at some point
Doh! I should've included that here. I can probably get to that later today.

PR #5135, opened 2023-05-21
Update dfu-util instructions

tormodvolden:
dfu-util is available for most platforms, not only Linux.

The "with some versions of dfu-util you may need this instead" is nonsense, also because -s and --dfuse-address is the same option.


Hwurzburg:
LGTM..thanks

PR #5142, opened 2023-05-23
Copter: dynamically generate motor diagrams (WIP)

yuri-rage:
WIP PR for discussion/review.
Creates dynamically displayed motor diagrams based on JSON output from https://github.com/ArduPilot/ardupilot/pull/23821

Known issues (that I am working to correct):
* [FIXED] V, H, and I frame types draw incorrect frame arms
* Coaxial frames are not displayed in pseudo-3D
* Frame types missing from JSON output (Bicopter and Tricopter) need to be integrated
* Javascript `fetch` is not compatible with local dev builds. JSON frame definitions need to be available to the script without the need to dynamically fetch/parse.

TODO:
* Automatically create/download updated JSON during wiki build

### DIFF
--- a/copter/source/_static/FrameNotes.json
+++ b/copter/source/_static/FrameNotes.json
@@ -1,0 +1,130 @@
+{
+    "1": {
+        "Name": "QUAD",
+        "6": {
+            "Name": "PLUS (REVERSED)"
+        },
+        "12": {
+            "Name": "X (BETAFLIGHT)"
+        },
+        "13": {
+            "Name": "X (DJI)"
+        },
+        "14": {
+            "Name": "X (CLOCKWISE)"
+        },
+        "16": {
+            "Name": "PLUS (NO YAW TORQUE)",
+            "Notes": "Motor rotation direction does not matter"
+        },
+        "17": {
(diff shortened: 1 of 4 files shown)


Hwurzburg (changes requested):
Sorry this displays nothing but two drop down selection boxes that do not do anything on my local test builds....and I am not 100% convinced that making the reader select his frame before showing it is the best idea...many would like to see all the options at once if they are just exploring what options are available....if this displayed all the configurations without the drop boxes required, then this would be a nice addition

yuri-rage:
@Hwurzburg, the issue may be the fact that I'm using `fetch` calls in Javascript, which are blocked on local page loads for obvious security reasons. I used VS Code's live server plugin to overcome that.

I understand your concern, but I'm not certain that I agree. I will probably cease work on this until we come to a consensus on proper display of the page. Please see the Discuss topic for the traction gained there, and I'd certainly appreciate your comments there, as well!

yuri-rage:
Closing in favor of https://github.com/ArduPilot/ardupilot_wiki/pull/6215

PR #5145, opened 2023-05-23
Update common-external-ahrs.rst

antholuo:
Added information for VN-300 Configuration


IamPete1:
These commands are sent and part of the auto config, 

https://github.com/ArduPilot/ardupilot/blob/59b24c304fe043de247d79991530add9690cbb65/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp#L422-L425

I'm not sure having them in the wiki really helps. Or at least you would need to give more info about how to connect to and send the commands, and also look up the save command so they stick. 

Basically if they don't work in the auto-config I would not expect them to work if sent manually either.

antholuo:
> These commands are sent and part of the auto config,
> 
> https://github.com/ArduPilot/ardupilot/blob/59b24c304fe043de247d79991530add9690cbb65/libraries/AP_ExternalAHRS/AP_ExternalAHRS_VectorNav.cpp#L422-L425
> 
> I'm not sure having them in the wiki really helps. Or at least you would need to give more info about how to connect to and send the commands, and also look up the save command so they stick.
> 
> Basically if they don't work in the auto-config I would not expect them to work if sent manually either.

That makes sense - do you think it would be relevant to include information about _how_ EAHRS expects the VN-300 to be configured (ie update rates of information) and to not include the actual commands to configure them?

Hwurzburg:
@IamPete1 awaiting your input

IamPete1:
I think the key thing in this PR is the addition of the notes on setting GPS type. I don't think users should ever have to get into the message config, that should auto config so long as the baud rate matches up. We could add a some info on how to set that correctly. What messages we actually use could be interesting, but would be too much for most users.

antholuo:
Page should now look like this (screencap from preview)
image

PR #5149, opened 2023-05-24
Add style for enums and prefer enum classes

Ryanf55:
Recommend enum classes since they are commonly used. Provide a style recommendation for them. 
https://github.com/ArduPilot/ardupilot/pull/23877
https://github.com/ArduPilot/ardupilot/pull/23830
https://github.com/ArduPilot/ardupilot/pull/23768/files#diff-ed061f26ab421c524f35bb9112b56bc1322892589f95b77e9497345eb48eafbe
https://github.com/ArduPilot/ardupilot/pull/23119/files#diff-a98f7c976f44f4a82c8c4882b638ca053d70a95f603861216630e20243689dd2

I did not cover the aspect of using `#DEFINE` to enable/disable enums, and to be sure to pin all the numbers so they don't change value when you change your build options. Seems too advanced for the style guide.

### DIFF
--- a/dev/source/docs/style-guide.rst
+++ b/dev/source/docs/style-guide.rst
@@ -356,6 +356,31 @@ underscores.
 
     class ap_compass { };
 
+Enums
+---------------
+
+Prefer enum classes over raw enums. Enums should be PascalCase, and be singular.
+All entries should have trailing commas, even the last one.
+
+**Right:**
+
+::
+
+    enum class CompassType {
+        FOO,
+        BAR,
+    };
+
+**Wrong:**
+
+::
+
+    enum compass_types {
+        FOO,
+        BAR
+    };
+
+
 Functions and variables
 -----------------------
 


rmackay9 (approved):
In the code we have a mix of camel case and underscore for the enum names but I think we are tending towards camel case so I think this looks right.  @peterbarker you agree?

dev/source/docs/style-guide.rst
```diff
@@ -356,6 +356,30 @@ underscores.
...
+
+Prefer enum classes over raw enums. Enums should be CapitalCase, and be singular.
+
+**Right:**
+
+::
+
+    enum class CompassType {
+        FOO,
+        BAR
```
peterbarker, line 370:
```suggestion
        BAR,
```

Do we need some words around this?  Adding hthe trailing comma simplifies the diffs when scanning down the page in github etc

peterbarker (changes requested):

peterbarker:
> In the code we have a mix of camel case and underscore for the enum names but I think we are tending towards camel case so I think this looks right. @peterbarker you agree?

This is actually PascalCase - not camelCase :-)  camelCase is the one that has the hump in the middle :-)

dev/source/docs/style-guide.rst
TunaLobster: Black would definitely add the commas. Not sure about flake8.

dev/source/docs/style-guide.rst
```diff
@@ -356,6 +356,30 @@ underscores.
 
     class ap_compass { };
 
+Enums
+---------------
+
+Prefer enum classes over raw enums. Enums should be CapitalCase, and be singular.
+
+**Right:**
```
TunaLobster, line 364:
```suggestion
**Correct:**
```

dev/source/docs/style-guide.rst
Ryanf55: > Black would definitely add the commas. Not sure about flake8.

Those are both python formatters, this is a c++ style guide

dev/source/docs/style-guide.rst
```diff
@@ -356,6 +356,30 @@ underscores.
...
+
+::
+
+    enum class CompassType {
+        FOO,
+        BAR
+    };
+
+**Wrong:**
```
TunaLobster, line 373:
```suggestion
**Incorrect:**
```

dev/source/docs/style-guide.rst
TunaLobster: Yep. I've spent too long writing slots and enums python and c++ recently. It's all blurring together.

dev/source/docs/style-guide.rst
Ryanf55: Can you explain why you want this change? The rest of the style guide uses right/wrong wording, so my changes are consistent.

TunaLobster: My assumption was that Correct would feed through a translator more accurately. Right is an idiom where correct is more definitive. Disregard if it matches the pattern.

Ryanf55: It matches the existing pattern, feel free to submit a PR though. Makes sense.

dev/source/docs/style-guide.rst
Ryanf55: I tested `astyle`, it does not care about trailing commas, so there is no way to enforce it. Sure, I'll add it to the wiki, because I agree it makes sense, but it's not something we can test for.

peterbarker (approved):

PR #5151, opened 2023-05-24

vincentpoont2:
Add Holybro RM3100


Hwurzburg:
Sry for delay on this...we should add it also on the  DroneCAN peripherals page...will update this pr and merge in a couple of days

PR #5168, opened 2023-06-07
Common: Siyi ZR10 yaw angle min/max fix

rmackay9:
Fix siyi setup instructions to specify yaw min/max to +-135 deg which matches the datasheet.

image

This has been tested local and seems ok.


Hwurzburg:
thanks

PR #5183, opened 2023-06-14
Update partners page

rishabsingh3003:


Hwurzburg:
LGTM....interesting...Scorpion....thought they were just motors and escs....are they doing something AP'sh?

PR #5188, opened 2023-06-16
Esp8266 detail esptool

Iragne:
Adding more detail for mac user to how to flash the ESP8266

### DIFF
--- a/common/source/docs/common-esp8266-telemetry.rst
+++ b/common/source/docs/common-esp8266-telemetry.rst
@@ -66,7 +66,29 @@ If you are using the Adafruit HUZZAH ESP8266 board and are flashing from a Windo
     .. image:: ../../../images/esp8266-telemetry-flash.jpg
         :target: ../_images/esp8266-telemetry-flash.jpg
 
-- Mac and Linux users should use the `esptool <https://github.com/espressif/esptool>`__
+- Mac users should use the `esptool <https://github.com/espressif/esptool>`__
+
+  - First, find the COM of your connected device by using the following command line in the terminal. The USB COM port that the ESP8266 is attached will appear something  "/dev/ttyusbserial-A600eto2" 
+
+    ..  code-block:: shell
+
+      $ ls /dev/tty\.*
+
+  - Next, you need to erase all content in your ESP8266. Put the ESP8266 in bootloader mode, and run this command in your terminal window, replacing the *COM* with the ESP8266 COM port you found previously above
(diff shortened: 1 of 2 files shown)


Iragne:
Hello any update here ?

Iragne:
Hi @Hwurzburg  thanks for the reply. 
I will redouble check but the note are regarding how to use esptool for Esp8266 and it the same for linux. 
I can re double check for linux but it's the same command line. 
Sorry I wrongly describe my PR and it should be for both (Linux and Mac).

Also i will update the command line with code block

Hwurzburg:
your command to find the USB COM port that the device is attached will not work in Linux....the /dev directory will have an ttyACMx file for each USB...of zero length and information....and the only simple way to tell which is which is to unplug the USB device, run "dmseg -w", plug back in and see the TTYACMx assigned for that device name....so this part is incorrect for Linux...hence my skepticism at the information

Iragne:
Your right I'm too much used of Mac that I mixed that part. I will push an update on it to correct it. Sorry but thanks for the previous info. I really would like to have a better Mac wiki due to the Tim I passed on it.

common/source/docs/common-esp8266-telemetry.rst
```diff
@@ -66,15 +66,37 @@ If you are using the Adafruit HUZZAH ESP8266 board and are flashing from a Windo
...
+
+      $ ls /dev/tty\.*
+
+  - After, you need to erase all content in your ESP8266, put your device in bootloader mode, and replace the *COM* with the one you found previously by running this command line in your terminal 
+
+    ..  code-block:: shell
+
+      $ esptool.py   --baud 921600  --port COM erase_flash
+
+  - Last, you flash the board with the correct expected firmware, put your device in bootloader mode, and replace the *COM* with the  port you used before
```
Hwurzburg, line 83:
```suggestion
  - Finally,  flash the board with the correct expected firmware: Put your device in bootloader mode again, and use the following command, again replacing the *COM* with the  port you used in the previous command.
```

Iragne:
Thanks just validate your commit suggestions

Iragne (commented):
commit approved

Hwurzburg:
LGTM...just squash and I will merge

Iragne:
Sorry just went back, any update here ?

Hwurzburg:
sorry dropped off the radar....merging...thanks

PR #5194, opened 2023-06-19
common: add warning against purchase of CrossFlight product

peterbarker:


rmackay9:
I wonder if we could combine the warnings?

common/source/docs/common-radiolink-minipix.rst
```diff
@@ -7,7 +7,7 @@ RadioLink MiniPix
 
 .. warning:: This autopilot is not recommended because some versions of the board are not compatible with the official ArduPilot software despite multiple efforts to work with the manufacturer to make them compatible. The manufacturer is also apparently not abiding by the GPLv3 license which requires releasing the modified source code to its customers.  "V1.0" and "V1.2" probably work, "V1.0 II" and "V1.1" definitely do not work.
```
Hwurzburg, line 8:
```suggestion
.. warning:: This autopilot and the RadioLink CrossFlight are not recommended for use, not supported by ArduPilot, because some versions of the board are not compatible with the official ArduPilot software despite multiple efforts to work with the manufacturer to make them compatible. The manufacturer is also not abiding by the GPLv3 license which requires releasing the modified source code to its customers.  "V1.0" and "V1.2" probably work, "V1.0 II" and "V1.1" definitely do not work.
```

Hwurzburg (changes requested):
Agree...proposed combined wording

common/source/docs/common-radiolink-minipix.rst
peterbarker: I don't agree; I think the two-paragraph approach is better.  There's just two much being crammed into one paragraph otherwise.

The more red at the top of the page the better.

Hwurzburg: we could just remove the page and any firmware builds....its not a supported autopilot by ArduPilot....given them years to comply....and they have not....

peterbarker: > we could just remove the page and any firmware builds....its not a supported autopilot by ArduPilot....given them years to comply....and they have not....

These pages (and binaries) are not for the vendors.  These pages are here for the users.

We'd just like to have fewer of those, thus these warnings.

PR #5197, opened 2023-06-21
cube overview - I2C2 port might be wrong

hamishwillee:
Looking at the schematics, it might be that the I2C port numbers for GPS1 and GPS2 are reversed https://github.com/proficnc/The-Cube/blob/master/CB_REV_C_Altium/PCB%203D%20Print%20from%20Above.PDF

It is possible the schematic is wrong, but this does align with the cube docs I am working off: https://docs.cubepilot.org/user-guides/autopilot/the-cube-module-overview


Hwurzburg:
verified on the bench...thanks...merging

PR #5201, opened 2023-06-22
Common: add xacti to upcoming features

rmackay9:
This adds a Xacti camera gimbal setup page to the "upcoming features" area.

The corresponding flight code PR is here https://github.com/ArduPilot/ardupilot/pull/24117


Hwurzburg:
@rmackay9 you are the man! wish everyone created docs as they create the code....looks good

Hwurzburg:
this does not cover all the new params in 24117

rmackay9:
@Hwurzburg,

PR 24117 adds a new parameter but that param should not be set by users so I think this is OK as-is... although I guess you're right that if the user swapped out their Xacti camera gimbal for another camera which used a different Node Id then they'd need to reset that parameter to zero.  That's a real edge case though.

PR #5203, opened 2023-06-22
Update common-ads-b-receiver.rst with correct fields and note for XP …

Danf96:
…transponders

SERIALX_PROTOCOL value for ADSB in the wiki was incorrectly set to 28 instead of 35 as well as an incorrect default baud rate of 115600 for the XP series transponders rather than 57600. A note was also added to mention that the XP transponders do not currently work with Mission Planner's operating mode messages, but will still work with Mission Planner if operating mode is changed in an external program.


Hwurzburg:
LGTM..thanks

PR #5209, opened 2023-06-23
 add throttle curve examples

Hwurzburg:
based on some question I been getting from users...needed to clarify what when how the throttle curve params are used....


images/aerobatic_throttle_curve.jpg
bnsgeyer, line 1: @Hwurzburg This figure needs to be fixed.  The lines are skewed and don't exactly cross at the correct point.  I can make this figure if you like.

images/normal_throttle_curve.jpg
bnsgeyer, line 1: Again these lines don't exactly cross at the correct points. I'm not sure I follow this comment.  Why do you mention zero thrust?  I can make this figure if you like.

bnsgeyer (commented):
I think what you have added is good.  I just think the figures could be better.   Let me know if you want me to do them.

Hwurzburg:
Would be happy to see your takes on the figures...yes, some lines were skewed, but if you want to make changes I wont update them...

images/aerobatic_throttle_curve.jpg
Hwurzburg: because the collective is asymmetric about zero thrust -2 to +12 with 0 usually 0 thrust (unless the airfoil is assymetrtic) as opposed the aero setup with asymmetric collective on either side of the 0 thrust point...

bnsgeyer:
@Hwurzburg I drafted the graphs in excel.  I will post them tomorrow after I have annotated them.

images/aerobatic_throttle_curve.jpg
bnsgeyer: @Hwurzburg attached are my suggested figures
[3d setup.pdf](https://github.com/ArduPilot/ardupilot_wiki/files/11860725/3d.setup.pdf)
[Normal setup.pdf](https://github.com/ArduPilot/ardupilot_wiki/files/11860726/Normal.setup.pdf)

Let me know if you want any changes or what format.  If you want the original excel spreadsheet, I can provide that too.

Hwurzburg: thanks...except we dont do the curved segments....the output is via straight line interpolation from the 5 points....right?

bnsgeyer: @Hwurzburg  we have had 5 point spline throttle curves since 2017.

Hwurzburg:
@bnsgeyer thanks for the nicer graphs! merging

PR #5215, opened 2023-06-26
Common: update pozyx EKF3 param settings

rmackay9:
This updates the Pozyx page to remove references to very old AP versions and fixup the EKF settings.


Hwurzburg:
LGTM...thanks...

PR #5250, opened 2023-07-03
Reorg ESC pages

Hwurzburg:
best way to review is build locally and navigate from the ESC and Motors peripherals page


common/source/docs/common-esc-guide.rst
```diff
@@ -0,0 +1,40 @@
...
+
+- Single signaling wire to autopilot servo/motor output which can use pulse or digital communication protocols
+- :ref:`DroneCAN <common-uavcan-escs>` or CAN, which provides a more robust connection with greater length limits at the expense of more wiring.
+
+Communication Protocols
+=======================
+
+The ESC may be compatible with one or more of the following protocols:
+
+- :ref:`PWM, OneShot, OneShot125<common-brushless-escs>` are pulse based protocols. Almost all ESCs are compatible with PWM signaling. And this is the default configuration of ArduPilot firmware for serov/motor outputs.
```
rmackay9, line 22: small typo, "serov" should be "servo".

common/source/docs/common-esc-telemetry.rst
```diff
@@ -0,0 +1,98 @@
+.. _common-esc-telemetry:
+
+=============
+ESC Telemetry
```
rmackay9, line 4: I wonder if we're trying to cover all the various protocols that support ESC Telemetry (in the wider sense) or do we just want to cover BLHeli's ESC telemetry feature.  If it's the former then we should expand this to include the CAN protocols.  If it's the latter then we should perhaps change the title to BLHeli ESC Telemetry.

common/source/docs/common-esc-telemetry.rst
```diff
@@ -0,0 +1,98 @@
...
+- :ref:`SERVO_BLH_POLES <SERVO_BLH_POLES>` defaults to 14 which applies to the majority of brushless motors and normally does not need to be changed.  Adjust as required if you're using motors with a pole count other than 14 to calculate true motor shaft RPM from ESC's e-field RPM.
+
+
+Bi-directional DShot
+====================
+
+[site wiki="rover"]
+.. warning::
+
+   Bi-directional DShot does not work with Rover (`see issue here <https://github.com/ArduPilot/ardupilot/issues/20372>`__)
```
rmackay9, line 39: I know this is existing text that has just been moved but in any case, this Bi-directional DShot issue for Rover has been fixed now.  https://github.com/ArduPilot/ardupilot/issues/20372

common/source/docs/common-escs-and-motors.rst
```diff
@@ -29,19 +29,45 @@ Motors
...
+
 .. toctree::
     :maxdepth: 1
 
-    BLHeli/BLHeli32 Capable ESCs <common-blheli32-passthru>
-    Currawong Velocity ESCs <common-velocity-can-escs>
-    DroneCAN ESCs <common-uavcan-escs>
+    PWM, OneShot and OneShot125 ESCs <common-brushless-escs>
     DShot ESCs <common-dshot-escs>
+    DroneCAN ESCs <common-uavcan-escs>
```
rmackay9, line 48: Previously it was in alphabetical order.. I guess it is in order of popularity now.. I don't mind but just pointing this out.

common/source/docs/common-escs-and-motors.rst
```diff
@@ -29,19 +29,45 @@ Motors
...
+    BLHeli/BLHeli32 Capable ESCs <common-blheli32-passthru>
+
+
+Telemetry
+---------
+
+.. toctree::
+    :maxdepth: 1
+
+    ESC Telemetry <common-esc-telemetry>
```
rmackay9, line 69: This should probably be BLHeli ESC Telemetry

rmackay9:
Thanks for having a go at improving the flow of the ESC setup pages.

Some of this PR is definitely an improvement but TBH, I'm not sure it's an improvement to split up the BLHeli information into multiple pages.  My guess is that this is done because there is confusion about terminology.. so "ESC Telemetry" and "Bi-drectional DShot" sound like they're more general purpose but actually they're only supported by BLHeli.  I think could try and resolve the terminology confusion (e.g. specify that these features are BLHeli specific despite their name) instead of reorganising the pages.

common/source/docs/common-esc-guide.rst
Hwurzburg: will fix

common/source/docs/common-esc-telemetry.rst
Hwurzburg: I disagree....there are many ESCs that provide telemetry that are _**not**_ BLHeli and that ArduPilot currently does not support (we probably should)...Castle, HobbyWing, etc. I can add CAN and notes on the BiDir that it is _**currently**_ only in BLHeli escs (I have been told recently that its coming on non-BLHeli ESC shortly...but I cant find that note right now)

common/source/docs/common-esc-telemetry.rst
Hwurzburg: will do

common/source/docs/common-escs-and-motors.rst
Hwurzburg: Historical, but I should move KDE can up above Currawong

common/source/docs/common-escs-and-motors.rst
Hwurzburg: As stated for the reasons above, I disagree...but will add a note about current ARduPilot implementations being only BLHeli ESCs

rmackay9 (approved):
Well, I think you've provided a lot more support on the forums in this area than I have so ..

PR #5254, opened 2023-07-04
Update testing-with-replay.rst

kd0aij:
I think there were some stray back-tics in this line, but not sure about the syntax and I'm not set up to build it.


Hwurzburg:
that link does not exist....there is no such tag in the wiki as "setting-up-sitl-on-sitl"...the fact that it was broken format prevented it from crating build error...which this correction does...I suggest just removing the link or sentence entirely

kd0aij:
I think this should be good... don't see a way to squash using this interface though.

Hwurzburg:
I will squash it and merge.....thanks

PR #5263, opened 2023-07-07
Common: Add warning about QGC and MP with RockBLOCK

stephendade:
To prevent user confusion, I've added a note recommending users to use MAVProxy instead of MP or QGC with RockBLOCK telemetry.

@rmackay9


rmackay9:
Thanks very much for this!

PR #5269, opened 2023-07-10
Top level menu reorganization for Copter, Plane, and Rover

brianmusslewhite:
@Hwurzburg, I've updated things based on your feedback from the last pull request. I still feel like the different variants could be better organized but I'm not sure the best way to do it. Let me know your thoughts overall.

I'm also thinking about how to simplify the main page menu and group together items under advanced configuration and peripheral hardware. For now, I think this should be a good incremental improvement though.


Hwurzburg:
@brianmusslewhite please DM me on https://discord.com/invite/M346dY8 just search for henry.wurzburg....we can chat if you have audio on your computer and screen share

Hwurzburg:
@brianmusslewhite Copter merged, Plane is pending...have not reviewed the Copter intro changes...will do that today...thanks

PR #5287, opened 2023-07-16
Add warning about picoblade connectors to Pixhawk

Zipdox2:
Background: https://www.rcgroups.com/forums/showthread.php?1493712-JST-connector-confusion-the-real-story/page3#post29192314


Hwurzburg:
LGTM, thanks

PR #5290, opened 2023-07-18
Added links to Qiotek M10 GNSS and fix expired issues 

QioTek:
Added links to Qiotek M10 GNSS and fix expired issues


Hwurzburg:
Thanks Jason....LGTM...merging

PR #5300, opened 2023-07-22
updated autopilot menu order and page intro

brianmusslewhite:
I updated the menu level so that when "Choosing an Autopilot" is clicked, it does not require another click to see the categories. I also reorganized the intro of the page accordingly.

### DIFF
--- a/common/source/docs/common-autopilots.rst
+++ b/common/source/docs/common-autopilots.rst
@@ -4,26 +4,26 @@
 Choosing an Autopilot
 =====================
 
-ArduPilot runs on many different autopilot boards, which are linked from the topic below: AutoPilot Hardware Options.
-
-Selecting the right board depends on the physical restraints of the
+ArduPilot runs on many different autopilot boards. Selecting the right board depends on the physical restraints of the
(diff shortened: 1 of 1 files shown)


common/source/docs/common-autopilots.rst
```diff
@@ -4,26 +4,26 @@
 Choosing an Autopilot
 =====================
 
-ArduPilot runs on many different autopilot boards, which are linked from the topic below: AutoPilot Hardware Options.
-
-Selecting the right board depends on the physical restraints of the
+ArduPilot runs on many different autopilot boards. Selecting the right board depends on the physical restraints of the
```
Hwurzburg, line 7:
```suggestion
ArduPilot runs on many different autopilot boards. Selecting the right board depends on the physical constraints of the
```

Hwurzburg (changes requested):

brianmusslewhite:
@Hwurzburg This should be ready to merge. Good catch! I overlooked that

Hwurzburg (approved):

PR #5307, opened 2023-07-28
Dev/Common: camera and gimbal updates

rmackay9:
This PR includes camera and gimbal related updates for both the developer and user wikis.

The developer wiki gets details on new commands supported in AP-4.5.0

The user wiki gets warnings and fixes for the Siyi and Gremsy gimbals.

This has been tested locally and seems OK.

### DIFF
--- a/common/source/docs/common-gremsy-pixyu-gimbal.rst
+++ b/common/source/docs/common-gremsy-pixyu-gimbal.rst
@@ -12,7 +12,12 @@ Gremsy `Mio <https://gremsy.com/products/mio>`__, `Pixy F <https://gremsy.com/pr
 .. warning::
 
    MIO gimbals are unable to communication with ArduPilot.  We hope for a fix with Gremsy software version 7.7.3.
+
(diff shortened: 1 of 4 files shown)


rmackay9:
I'd accidentally committed some changes to the build options file which I've removed now.  Sorry about that.

rmackay9:
Hi @Hwurzburg,

Thanks for that.  I think we should consider merging this before 4.5 is ready because many developers work off of master.  Also the non-dev changes are valid for 4.4.

Hwurzburg:
the non dev changes state 4.5 or higher....not 4.4....a mistake? if not, we can split them out as 4.5 and merge the dev only stuff...I am okay with that

rmackay9:
Hi @Hwurzburg,

I've removed the Gremsy 4.5.0 related change and will put that in a separate PR.  I think the rest of the changes are development (which can go in even though the stable version is not 4.5 yet) or are user wiki changes not related to 4.5.

Hwurzburg:
Thanks...merging

PR #5317, opened 2023-08-01
4.5:Add RADIX2HD

Hwurzburg:
Draft....firmware section needs work


mluessi:
@Hwurzburg I improve the instructions in the readme for how to obtain/create the firmware file. See PR below. 

ArduPilot/ardupilot/pull/24724

Hwurzburg:
will review when I return from vacation

Hwurzburg:
@mluessi ready for wiki once we get a firmware on your website...as I PM's I can create a 4.4.1 copter with this hwdef for it to bridge until 4.5 stable is out (the firmware is already in the 'latest' folder)

mluessi:
@Hwurzburg support for it has been added to Copter/Plane 4.4.1-beta2.

The beta and development releases packed for the bootloader the RADIX2HD uses are now available on the [BrainFPV website](https://www.brainfpv.com/firmware) (it uses the the binaries from the ArdPilot build server). So users don't have to run the "packing script" unless they are using a custom build.

PR #5319, opened 2023-08-02
Common: gremsy gets ZIO cam1_type param setting

rmackay9:
This is a small update to tell users how the Gremsy ZIO camera should be configured.

This covers the wiki update required for this autopilot flight code PR https://github.com/ArduPilot/ardupilot/pull/23138


Hwurzburg:
added link to this in futures..thanks

PR #5331, opened 2023-08-07
mavproxy: fix startup option dashes

rsaxvc:
--option renders as -option
-\\-option renders as -\-option
-\-option renders as --option

Switch from -\\-option to -\-option


Hwurzburg:
LGTM...thanks

PR #5337, opened 2023-08-08
sitl: Details for running with X-Plane 11 and WSL

Maarrk:
Updated SITL instructions to explain usage with X-Plane 11 and connecting with SITL compiled in WSL2. It was possible to guess from the previous documentation, but it took me a bunch of tries to get it right and the settings screen looks different.

_(Contribution on behalf of @FlyfocusUAV)_


Hwurzburg:
Dont think we have any XPlane11 experts around, so will trust you on this...builds okay...merging...thanks

PR #5343, opened 2023-08-10
Common: viewpro gets image and video download instructions

rmackay9:
This adds a little section to the bottom of the ViewPro page with instructions on how images and videos can be downloaded from the gimbal over ethernet.  Eventually I'd like to add this info for all our cameras but so far this is the only one that I have confirmed can do it.

I have tested this on my local machine and it looks OK.


Hwurzburg:
thanks....LGTM...

rmackay9:
Thanks!

PR #5360, opened 2023-08-18
Common: Flir vue page gets warning

rmackay9:
This PR adds a warning to the Flir Vue Pro camera setup page.  I've also adding a missing setting and removed a very old setting (the "CH8_OPT" part).

I've recently re-tested the setup and found that the MAVLink interface has somehow been broken and I strongly suspect it is due to some change on the FLIR firmware side.  [There is a discussion here](https://discuss.ardupilot.org/t/flir-vue-pro-gets-no-mavlink-info/96072) where other users are also apparently struggling.


Hwurzburg:
LGTM...thanks

PR #5367, opened 2023-08-22
Remove/normalize.css

khancyr:
remove normalize.css request that is a leftover from the previous update pr.

Make the json file no cache, so we always have updated files, we could probably use something like 1h caching time.
make blog youtube previous a bit nicer (larger in most screen)
Add as small dev container to test frontend changes
Refactor a bit the python script that gather blog post. 
Fix youtube video in blog post handling


Hwurzburg:
@khancyr needs some CI work

Hwurzburg:
lgtm

PR #5369, opened 2023-08-24
Dev: mavlink camera page gets absoluite zoom info

rmackay9:
This adds details on the mavlink command to control absolute zoom.

I've added some more details about which versions support which commands but this shouldn't block this from being merged because this is all in the developer wiki and developers often develop against "latest".


rmackay9:
Aha, I wasn't expecting a merge for another week or so! thanks!

PR #5371, opened 2023-08-25
Dev: Add link to newer SITL serial info

TunaLobster:
There are still links floating out there to the old page. Redirect to the newer info and the consolidated page.


Hwurzburg:
LGTM..thanks

PR #5373, opened 2023-08-28
Rover: minor fix to turn rate tuning advice

rmackay9:
This is a minor fix to the advice re tuning a Rover/Boat's turn rate P and I terms.

While helping a user I noticed the shocking advice (that I wrote years ago) that leaving P at zero is often a good idea.

This has been tested locally and looks OK.


Hwurzburg:
Lgtm

PR #5374, opened 2023-08-28
Update common-rc-systems.rst to include SBUS -> Serial wiring

antholuo:


Hwurzburg:
I dont think this is required...there is a section below about serial port connections....and this statement is incomplete as is for this connection....adding this here would require adding serial port connection info duplicated for the other protocol sections also and remove the need for the rc in  on serial ports section...unnecessary duplication

antholuo:
I think that makes sense, it also seems like my patch isn't up to date with master (It doesn't match what is on the current website?) 

I need to take a bit longer look at this...

PR #5380, opened 2023-09-01
common: marvelmind parameter update

rmackay9:
This slightly updates the marvelmind page to bring the EKF parameters up-to-date.

I've also removed references to very old versions of AP.

I do not have a marvel mind to test with but I've had a quick look at the driver and I think this should be correct.

FYI @karthikdesai.  I know this was a long time ago but at least FYI.


karthikdesai:
> This slightly updates the marvelmind page to bring the EKF parameters up-to-date.
> 
> I've also removed references to very old versions of AP.
> 
> I do not have a marvel mind to test with but I've had a quick look at the driver and I think this should be correct.
> 
> FYI @karthikdesai. I know this was a long time ago but at least FYI.

thanks! LGTM, i dont have access to MM sensors myself :(

PR #5381, opened 2023-09-02
updated bullets to caps

brianmusslewhite:
I updated the lower case bullet list to capital bullet lists


Hwurzburg:
LGTM...thanks

PR #5387, opened 2023-09-04
Update quadplane-parameters.rst SCHED_LOOP_RATE

timtuxworth:
It seems that 400 should be the recommended SCHED_LOOP_RATE fora aquaplane. If anyone knows more about this please add details.

### DIFF
--- a/plane/source/docs/quadplane-parameters.rst
+++ b/plane/source/docs/quadplane-parameters.rst
@@ -27,6 +27,7 @@ Key Parameters
    as one of your available flight modes (as that will shut down the
    quad motors). Please be careful not to use hard left rudder and zero
    throttle while flying or you risk disarming your motors.
+-  The default :ref:`SCHED_LOOP_RATE<SCHED_LOOP_RATE>` for a QuadPlane is to 300 (Hz). Most QuadPlanes do not need this to be rasied. Some very small vehicles (< 1Kg) might benefit from setting it to 400. In heavier vehicles, their higher inertia results in lower effective control response rates, so they do not benefit from a higher loop rate. Raising above 300 only leads to larger log files in these vehicles.
 
 Q_OPTIONS
 =========


timtuxworth:
Perhaps the parameter documentation for SCHED_LOOP_RATE needs updating too?

Hwurzburg:
@tridge needs to weigh in on this

tridge:
no, most quadplanes do not need 400. Some very small ones might benefit. The larger the vehicle the less a high loop rate matters, plus the inertia of quadplanes makes their control bandwidth lower so need lower loop rate.
Raising to 400 just leads to larger log files

timtuxworth:
> no, most quadplanes do not need 400. Some very small ones might benefit. The larger the vehicle the less a high loop rate matters, plus the inertia of quadplanes makes their control bandwidth lower so need lower loop rate.
> Raising to 400 just leads to larger log files

Would it make sense to say this on the wiki page?

peterbarker:
> > no, most quadplanes do not need 400. Some very small ones might benefit. The larger the vehicle the less a high loop rate matters, plus the inertia of quadplanes makes their control bandwidth lower so need lower loop rate.
> > Raising to 400 just leads to larger log files
> 
> Would it make sense to say this on the wiki page?

I think it would - this isn't the first time it's come up.  Care to put together a PR?

timtuxworth:
> > > no, most quadplanes do not need 400. Some very small ones might benefit. The larger the vehicle the less a high loop rate matters, plus the inertia of quadplanes makes their control bandwidth lower so need lower loop rate.
> > > Raising to 400 just leads to larger log files
> > 
> > 
> > Would it make sense to say this on the wiki page?
> 
> I think it would - this isn't the first time it's come up. Care to put together a PR?

I've updated this PR based on @tridge 's comment above. Will that do, or are you thinking of something more?

Hwurzburg:
LGTM...will merge once squashed...thanks

PR #5395, opened 2023-09-06
Common: add Tattu DroneCAN battery setup page

rmackay9:
This adds a page with setup information for the Tattu DroneCAN battery.

I've also done a driveby fix to the underline on the marvelmind page.

I've tested this on my local machine and it seems OK.


rmackay9:
Thanks!

PR #5407, opened 2023-09-08
Add YJUAV A6SE Flight Controller Wiki

yunjiuav:


common/source/docs/common-yjuav-a6se.rst
```diff
@@ -0,0 +1,291 @@
...
+
+UART Mapping
+============
+
+- SERIAL0 -> USB (OTG1)
+- SERIAL1 -> USART1 (Telem1)
+- SERIAL2 -> USART6 (Telem2)
+- SERIAL3 -> USART3 (GPS1), NODMA
+- SERIAL4 -> USART2 (GPS2), NODMA
+- SERIAL5 -> UART8 (SBUS)
```
Hwurzburg, line 221: where is UART8 pinned out? its say SBUS....is the SBUS out? TX only?

Hwurzburg (commented):
Aside from some formatting issues, looks very good...one question on UART8
once answered I can reformat and clean up as needed....thanks!

common/source/docs/common-yjuav-a6se.rst
yunjiuav: yes, This is defined as SBUS.out and only TX.

Hwurzburg: where in the hwdef or default params do you actually set the UART8 protocol to SBUS servos out?have you tested this???

yunjiuav: Yes, I have tested this output and the data is normal.
100Kps, 8E2
image

yunjiuav: > 

I didn't explain this serial port well above. In fact, the UART8 RX is also connected to the RC IN, and the SBUS OUT is connected to the SAFE identification interface. You can see it there, so the explanation will be clearer. So UART8 is input&output.

Hwurzburg: I see SBUS_OUT on the SAFETY connector and understand that its UART8TX, but SERIAL5's default protocol is NONE, so it will not output SBUS on its TX, nor will it detect RC on its RX input, even if tied to the RCIN pin...so a user will hook up an SBUS device to the SBUS_OUT and it will do nothing, unless he sets SERIAL5 protocol....this will result in customer support calls...that is why I suggest that a defaults.param file be added to the code to set it by default to enable SBUS outputs on UART8TX....in the wiki page we can explain an alternate use as RCIN (with a protocol change and losing the SBUS out feature)....since the UART8 is tied to the RCIN pin, its use as a general purpose UART is prevented and that also needs explanation in the wiki page....

yunjiuav: How can you help modify the wiki content? The default parameters will not be modified. The wiki explains how this UART is used by users.

Hwurzburg: I will work on the page today....also need to know which I2C bus is on the I2C connector and which is on the GPS connector? (1/2/3 or 4)

Hwurzburg: just fyi, your readme and this page have errors regarding SERIAL port vs UART...which I am correctly in this page also

Hwurzburg:
@yunjiuav I have updated the document, corrected the format errors, and the UART->SERIAL port errors, and added information on how to use UART8 TX ....please review and I will merge

yunjiuav:
> @yunjiuav I have updated the document, corrected the format errors, and the UART->SERIAL port errors, and added information on how to use UART8 TX ....please review and I will merge

The supplementary content is very good, thanks. can merge.

PR #5410, opened 2023-09-08
4.5:add rtl mode

Hwurzburg:


blimp/source/docs/flight-modes.rst
```diff
@@ -25,6 +25,7 @@ The table below shows for each flight mode whether it provides altitude or posit
     <tr><td>Manual</td><td>m</td><td>m</td><td>No</td><td>Manual output via motor/servo mixer.</td></tr>
     <tr><td>Velocity</td><td>s</td><td>s</td><td>Yes</td><td>Stick positions set a desired velocity. Mostly intended for tuning.</td></tr>
     <tr><td>Loiter</td><td>s</td><td>s</td><td>Yes</td><td>Holds altitude and position.</td></tr>
+    <tr><td>RTL</td><td>s</td><td>s</td><td>Yes</td><td>Returns at current altitude to position set when first armed (Home)</td></tr>
```
MichelleRos, line 28: Doesn't return at current altitude.

It returns directly to the X Y Z that it was  at when it armed. Also returns to 0 yaw.

Hmm. I wonder if it would be better for it to just keep pointing the direction it was when it armed?

MichelleRos (changes requested):

blimp/source/docs/flight-modes.rst
Hwurzburg: fixed

MichelleRos (commented):
LGTM. Thank you.

PR #5414, opened 2023-09-11
modules: graph: remove whitespace to fix code blocks

magate:
There was whitespace causing the `tickresolution` line to be included in the code block above it.


Hwurzburg:
LGTM...thanks

PR #5417, opened 2023-09-12
Feature/faster build

khancyr:
correct some python code on update.py
convert wget call to pure python equivalent that are faster. Make the download files from https://autotest.ardupilot.org in parallele to speed up the build

### DIFF
--- a/common_conf.py
+++ b/common_conf.py
@@ -59,8 +59,6 @@ intersphinx_mapping = {'copter': (intersphinx_base_url % 'copter',
 #  http://stackoverflow.com/questions/12772927/specifying-an-online-image-in-sphinx-restructuredtext-format
 #  And https://github.com/sphinx-doc/sphinx/issues/2429
 
-# Set False to re-enable warnings for non-local images.
-disable_non_local_image_warnings = True
 
 if disable_non_local_image_warnings:
     import sphinx.environment
--- a/update.py
+++ b/update.py
@@ -55,6 +55,8 @@ from datetime import datetime
 # without the following import on old versions of Python:
 from distutils import dir_util  # noqa: F401
 
+from frontend.scripts import get_discourse_posts
+
 if sys.version_info < (3, 8):
     print("Minimum python version is 3.8")
     sys.exit(1)
@@ -740,16 +742,13 @@ def put_cached_parameters_files_in_sites(site=None):
                 pass
 
 
-def update_frotend_json():
+def update_frontend_json():
     """
     Frontend get posts from Forum server and insert it into JSON
     """
     debug('Running script to get last posts from forum server.')
     try:
-        if platform.system() == "Windows":
-            subprocess.check_call(["python", "./frontend/scripts/get_discourse_posts.py"])
-        else:
-            subprocess.check_call(["python3", "./frontend/scripts/get_discourse_posts.py"])
+        get_discourse_posts.main()
     except Exception as e:
         error(e)
         pass
@@ -761,7 +760,7 @@ def copy_static_html_sites(site, destdir):
     """
     if (site in ['frontend', None]) and (destdir is not None):
         debug('Copying static sites (only frontend so far).')
-        update_frotend_json()
+        update_frontend_json()
         folder = 'frontend'
         try:
             site_folder = os.getcwd() + "/" + folder
--- a/update.py
+++ b/update.py
@@ -25,7 +25,7 @@ Build notes:
     using the "site" shortcode:
     [site wiki="plane,rover"]conditional content[/site]
 
-Parameters files are fetched from autotest using a Wget
+Parameters files are fetched from autotest using requests
 
 """
 from __future__ import print_function, unicode_literals
@@ -46,6 +46,11 @@ import shutil
 import subprocess
 import sys
 import time
+import requests
+from urllib.parse import urlparse
+from concurrent.futures import ThreadPoolExecutor
+from typing import Optional, Dict, List
+
 
 import rst_table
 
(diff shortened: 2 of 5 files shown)


common_conf.py
```diff
@@ -59,8 +59,6 @@
 #  http://stackoverflow.com/questions/12772927/specifying-an-online-image-in-sphinx-restructuredtext-format
 #  And https://github.com/sphinx-doc/sphinx/issues/2429
 
-# Set False to re-enable warnings for non-local images.
-disable_non_local_image_warnings = True
```
Hwurzburg, line 63: is this now undefined?

Hwurzburg:
Seems a lot faster on my local builds....even without the --fast option

update.py
```diff
@@ -487,46 +524,11 @@ def fix_site_shortcode(matchobj):
...
-        print("%s: except 7" % prefix)
-    try:
-        print("%s m8: %s" % (prefix, matchobj.group(8)))
-    except Exception:  # FIXME: narrow exception type
-        print("%s: except m8" % prefix)
+    for i in range(9):
+        try:
+            print("%s m%d: %s" % (prefix, i, matchobj.group(i)))
+        except IndexError:  # The object has less groups than expected
+            print("%s: except m%d" % (prefix, i))
```
TunaLobster, line 531: Maybe update to f-strings while doing this?

common_conf.py
khancyr: it is still present L19

scripts/cap_params.py
```diff
@@ -14,18 +14,18 @@
...
-    txt = open(f, 'r').read()
+    txt = open(f).read()
     matches = re.findall(r'[,.\s][A-Z][A-Z0-9]+_[A-Z_]+[,.\s]', txt)
     matches = re.findall(r'[,.\s][A-Z]+_[A-Z_]+[,.\s]', txt)
     changed = False
     for m in matches:
         s = str(m)
         param = s.strip()
-        s2 = "%s:ref:`%s<%s>`%s" % (s[0], param, param, s[-1])
+        s2 = f"{s[0]}:ref:`{param}<{param}>`{s[-1]}"
```
peterbarker, line 24: Was this change tested?

update.py
```diff
@@ -132,69 +137,110 @@ def remove_if_exists(filepath):
             raise e
 
 
-def fetch_url(fetchurl):
-    '''fetches content at url and puts it in a file corresponding to the filename in the URL'''
-    if platform.system() == "Windows":
-        subprocess.check_call(["powershell.exe", "Start-BitsTransfer", "-Source", fetchurl])
-    else:
-        subprocess.check_call(["wget", fetchurl])
+def fetch_and_rename(fetchurl: str, target_file: str, new_name: str) -> None:
```
peterbarker, line 140: This yieldssyntax isn't in *all* of Python3, is it?  When did it come in?

update.py
```diff
@@ -132,69 +137,110 @@ def remove_if_exists(filepath):
...
+    fetch_url(fetchurl, fpath=new_name, verbose=False)
+    # move in new file
+    if os.path.exists(target_file):
+        os.unlink(target_file)
+    try:
+        os.rename(new_name, target_file)
+    except Exception as e:
+        print(f"Failed to rename file: {e}")
```
peterbarker, line 148: Why should we continue if this fails?!

update.py
```diff
@@ -132,69 +137,110 @@ def remove_if_exists(filepath):
...
+        total_size = get_request_file_size(fetchurl)
 
-    """
-    # remove any parameters files in root
-    remove_if_exists('Parameters.rst')
+    try:
+        response = requests.get(fetchurl, stream=True)
+        response.raise_for_status()
+    except Exception as e:
+        print(f"Failed to fetch URL: {e}")
```
peterbarker, line 162: Isn't this just going to die later?

update.py
```diff
@@ -132,69 +137,110 @@ def remove_if_exists(filepath):
...
+    return 0
+
+
+def fetchparameters(site: Optional[str] = None, cache: Optional[str] = None) -> None:
+    dataname = "Parameters"
+    fetch_ardupilot_generated_data(PARAMETER_SITE, f'https://autotest.ardupilot.org/{dataname}', f'{dataname}.rst',
+                                   f'{dataname.lower()}.rst', site, cache)
+
+
+def fetchlogmessages(site: Optional[str] = None, cache: Optional[str] = None) -> None:
```
peterbarker, line 204: These types make things rather less readable.

peterbarker (changes requested):
I think I might quite like the meat of these changes.

But they are mixed in with a whole heap of changes to update the Python code.  Most of which are good - unicode string stuff, utf-declarations, fstrings etc.  This just makes it far more difficult to review than it should be.

This is only a partial review as I got frustrated with working through changes unrelated to the changes in the parameter and log information download.

peterbarker:
> @peterbarker I didn't intend to make a huge python3 style update, but Charlie said : hey why not use f-string. So I ran pyupgrade to modernize all .py in one go.
> 
> Beside the file downloading, I didn't intend to modify the logic nor refactor everything.

You could probably do that style update and rebase on top of that, then.

I'm still a touch worried about (a) the "yields" syntax which I thought was relatively new, and the readability with the Types system.   OTOH, I was tired and a bit grumpy on first review, so we'll see what happens on a second review :-)

Love f-strings :-)

TunaLobster:
Sphinx 5.1.1 is the current minimum for the ArduPilot wiki. That requires python 3.6 or newer. Generators showed up before then so everything is good there.

Type hints are also great for quickly coming up to speed with what is happening. They don't do much until coupled with a linter or static analyzer.

peterbarker:
> Sphinx 5.1.1 is the current minimum for the ArduPilot wiki. That requires python 3.6 or newer. Generators showed up before then so everything is good there.

Ah, sorry, I was referring to what's apparently known as "return type annotations".  I was using the older meaning of "yields" as synonymous with "returns", as opposed to the Pythonesque "yields" for generators.

> Type hints are also great for quickly coming up to speed with what is happening. They don't do much until coupled with a linter or static analyzer.

Are they, 'though?  It's harder to just glance at a function definition and see what it's taking.  I do think we should find out - but introducing their syntax really should be done in a separate PR.  If we can show that flake8 can now pick up more bugs, bringing it in would be a no-brainer IMO.  I don't think one could argue that it does make the syntax more scary, 'though.

Please note that you don't need my approval to merge this PR.  If you want to merge it with Henry's blessing that's workable.  It's just that if it breaks its likely to stay broken until you fix it :-)

We're not running out of PR numbers.  I would have merged the unicode and f-string changes instantly.

khancyr:
@peterbarker flake8 doesn't do type checking. The best tool for now is mypy or IDE if it support typing. 
One issue thus is that without setting at least the return type, most analyzers will just consider that the function take anything as input and return anything so they only report big error. Using typing, you have a better check on the return condition and avoid some strange casting.
I am not totally fine with typing as is could be really annoying to type and read, but it provides good bug tracking ... so I try to add it by default now

Hwurzburg:
@peterbarker Peter, still needs changes? or can I merge?

Hwurzburg:
@khancyr  @peterbarker says his review stands....I wont merge unless you two come to an understanding...sorry

khancyr:
no problem, I will improve this

khancyr:
@peterbarker removed py3 upgrade change, just the speedup now, will make an other PR for py3

PR #5419, opened 2023-09-13
Common: Tattu battery gets multiple battery explanation

rmackay9:
This adds a new section to the Tattu setup page to explain how multiple batteries can be connected.


Hwurzburg:
LGTM..thanks

PR #5420, opened 2023-09-13
Update ac_throttlemid.rst to take into account lower hover throttles and the max hover throttle hard limit in the code.

MichelleRos:
Very high-powered copters have a hover throttle lower than 0.2, so I've updated it to say "usually" instead of stating it as a requirement, which may make users think something is wrong if their hover throttle is that low.

I've also changed the max of 0.8 to 0.6 since the code limits the maximum hover throttle to 0.6875


Hwurzburg:
LGTM...thanks

PR #5424, opened 2023-09-14
AP_AdvancedFailsafe: wiki on AFS_OPTIONS

eppravitra:


plane/source/docs/advanced-failsafe-configuration.rst
```diff
@@ -262,6 +262,13 @@ the aircraft is a danger to people or other aircraft. To force an
 immediate termination you should use the :ref:`AFS_TERMINATE<AFS_TERMINATE>` parameter. By
 setting that parameter to 1 the aircraft will immediately terminate.
 
+AFS_OPTIONS
+------------------
+In addition, the behavior of AFS can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set, by default):
```
peterbarker, line 267:
```suggestion
In addition, the behavior of AFS can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set by default):
```

plane/source/docs/advanced-failsafe-configuration.rst
```diff
@@ -262,6 +262,13 @@ the aircraft is a danger to people or other aircraft. To force an
...
+
+- bit 0 (+1), if set, the aircraft will continue the mission even after data link is recovered. If not set, the aircraft will return to the mission item right before GCS failsafe occurs 
+- bit 1 (+2), if set, will force entering AUTO when GCS failsafe occurs. This is to make sure AFS is entered while Plane is flying in other autonomous modes (such as GUIDED).
```
peterbarker, line 270:
```suggestion
- bit 1 (+2), if set, will force entering AUTO when GCS failsafe occurs. This is to make sure AFS is entered if the Plane is flying in other autonomous modes (such as GUIDED).
```

plane/source/docs/advanced-failsafe-configuration.rst
```diff
@@ -262,6 +262,13 @@ the aircraft is a danger to people or other aircraft. To force an
 immediate termination you should use the :ref:`AFS_TERMINATE<AFS_TERMINATE>` parameter. By
 setting that parameter to 1 the aircraft will immediately terminate.
 
+AFS_OPTIONS
+------------------
+In addition, the behavior of AFS can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set by default):
+
+- bit 0 (+1), if set, the aircraft will continue the mission even after data link is recovered. If not set, the aircraft will return to the mission item right before GCS failsafe occurs
```
Hwurzburg, line 269: "continue the mission" ?doesnt this mean "continue the mission sequence jumped to by the AFS"?
what is data link? RC?GCS? either?

plane/source/docs/advanced-failsafe-configuration.rst
```diff
@@ -262,6 +262,13 @@ the aircraft is a danger to people or other aircraft. To force an
 immediate termination you should use the :ref:`AFS_TERMINATE<AFS_TERMINATE>` parameter. By
 setting that parameter to 1 the aircraft will immediately terminate.
 
+AFS_OPTIONS
+------------------
+In addition, the behavior of AFS can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set by default):
+
+- bit 0 (+1), if set, the aircraft will continue the mission even after data link is recovered. If not set, the aircraft will return to the mission item right before GCS failsafe occurs 
+- bit 1 (+2), if set, will force entering AUTO when GCS failsafe occurs. This is to make sure AFS is entered if the Plane is flying in other autonomous modes (such as GUIDED).
```
Hwurzburg, line 270: so currently if in GUIDED nothing happens on an AFS event? what other "autonomous" modes are there?

Hwurzburg (commented):
Needs some more explanation....probably partly due to historical documentation not being explict enough also

plane/source/docs/advanced-failsafe-configuration.rst
eppravitra: Not quite, continuing the mission means it will keep flying with the current mission item. If set, there would be no jumping. I reword it slightly. I hope it's a bit better.
 
The phrase data link has been changed.

plane/source/docs/advanced-failsafe-configuration.rst
eppravitra: That's correct. By default, AFS will not trigger on GUIDED.

I'm not sure if I want to list out all the modes. I kind of want to keep the description as it is. This is because autonomous modes might refer to different modes for Plane, Copter, and Rover.  In the code, it is whatever mode is interpreted as AP_AdvancedFailsafe::AFS_AUTO.

Hwurzburg: unfotunately, there is no place in the wiki that designates a vehicles modes as autonomous or non-autonomous, so we will have to list them here since this is the only place in the wiki that this designation is mentioned associated with a mode...even I am uncertain which modes would fall under this category....for example in Plane is LOITER??? you cant say that autonomous is a mode where pilot input is denied, because you can stick mix control even in AUTO....so how do you designate "autonomous"?

plane/source/docs/advanced-failsafe-configuration.rst
```diff
@@ -262,6 +262,13 @@ the aircraft is a danger to people or other aircraft. To force an
 immediate termination you should use the :ref:`AFS_TERMINATE<AFS_TERMINATE>` parameter. By
 setting that parameter to 1 the aircraft will immediately terminate.
 
+AFS_OPTIONS
+------------------
+In addition, the behavior of AFS can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set by default):
```
Hwurzburg, line 267:
```suggestion
In addition, the behavior of AFS during GCS communication loass can be modified by the setting of the :ref:`AFS_OPTIONS<AFS_OPTIONS>` bitmask parameter (no bits are set by default):
```
based on your change this param only affects GCS loss....not GPS(which also changes mission sequence)

Hwurzburg (changes requested):
Also under the "Ground station communications loss" section there is this incorrect sentence:
**If GPS lock is lost at the same time as GCS communications is lost then that is considered a “dual loss”, and the aircraft will immediately terminate.**
lower it states this only happens if the AFS_DUAL_LOSS is enabled...correct? or is the order of GCS/GPS loss important?

plane/source/docs/advanced-failsafe-configuration.rst
eppravitra: I'm not sure if we should change this sentence though. In the future, someone could use other bit mask to change behavior of RC failsafe or GCS failsafe.

Hwurzburg: if and when the behavior is changed, we can modify the sentence...as it is it could be assumed to be used for GPS mission changes also, which It is not

plane/source/docs/advanced-failsafe-configuration.rst
Hwurzburg: discussed with Tridge....its not autonomous....its auto-throttle modes in Plane..using AFS in other vehicles is something that is possible but only for an advanced user compiling their own code...so the wiki article will remain Plane focussed for now....I will update this PR accordingly...thanks

Hwurzburg:
I have updated and squashed....I think I am going to go ahead and merge this since it improves the existing info as well as add the new option, rather than putting it as an upcoming 4.5 feature...and the same with the other AFS 4.5 addition....users of this are advanced and it wont confuse normal users...thanks for documenting your code addition!

eppravitra:
Thank you!

PR #5426, opened 2023-09-14
Common: Xacti can and devid param info

rmackay9:
This adds some extra info on the Xacti setup discovered during testing.

I've tested this locally on my machine and it looks OK


Hwurzburg:
LGTM...thanks

PR #5436, opened 2023-09-19
Common: link Xacti page from cameras-and-gimbals landing page

rmackay9:
This PR makes three changes:

1. Add a link to the Xacti page from the cameras-and-gimbals landing page.  I know this may be slightly controversial because we normally leave new features in the "Upcoming Features" area until we start beta testing the next version but I think that it would be good to have more visibility on this series of cameras.  I think these cameras could be quite popular because although slightly expensive they are high quality and very small.  I'm sure it will take time for the number of users to ramp up and I would like to get that process started as quickly as possible.
2. Add note on which models are supported
3. Add firmware update process instructions


rmackay9:
Thanks so much!

PR #5439, opened 2023-09-22
Dev: MAVLink interface page for camera gets start-capture, stop-capture and fov-status info

rmackay9:
We've recently enhanced the flight code to allow start-capture and stop-capture commands to specify which camera is affected and also to specify a time interval between pics.  This PR adds the corresponding details to our wiki's mavlink interface page for cameras.

We've also added support for the CAMERA_FOV_STATUS message so that's included as well.

Finally I've removed mention of two deprecated messages that we will never support.

I've tested these changes locally and they look good to me.


Hwurzburg:
LGTM...thanks

PR #5448, opened 2023-09-26
Peripherals: Daiwa winch gets line length reset explanation

rmackay9:
This eases uses of the Daiwa winch in a few ways:

1. The Line Length zero reset feature is explained.  This includes a change to the wiring diagram to connect the zero-reset cable to the autopilot's SERVO11 output and an explanation of how to use MP's Servo/Relay tab to perform the reset
2. Do-Winch mission command is added
3. A link to the winch-test.lua script is added which allows easier manual deployment of the winch when no RC is available


Hwurzburg:
fixed a format error....added winch command to master list...thanks, merging

rmackay9:
Thanks!!

PR #5455, opened 2023-09-29
common: add instructions on how to flash-from-sdcard

peterbarker:
Please note that https://github.com/ArduPilot/ardupilot/pull/25143 is required for the "bootloader works in latest versions" thing to actually be true!  It should be merged before this PR is!

This work sponsored by Kraus Hamdani

### DIFF
--- a/common/source/docs/common-install-sdcard.rst
+++ b/common/source/docs/common-install-sdcard.rst
@@ -1,0 +1,111 @@
+.. _common-install-sdcard:
+
+[copywiki destination="copter,plane,rover,planner,blimp"]
+
+===========================
+Loading Firmware via sdcard
+===========================
+
+It is possible to update the ArduPilot firmware on certain autopilots by placing a specifically-named file onto an SD card and running the autopilot's bootloader (e.g. by power-cycling the board.
+
+.. note:: at this time only a few autopilots have this capability. See instructions below to determine if an autopilot has this feature. Only H7 based autopilots using MMC connected SD cards are capable of this and not all that do have had the feature added as of yet.
+
+Why would I want to do this?
+============================
+
+There are several reasons this technique may be of use:
+
+  - There is no convenient, reliable access to the vehicle's USB port
+  - It is desired to update the vehicle firmware over a telemetry radio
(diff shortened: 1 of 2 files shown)


Hwurzburg:
will need a few tweaks and build error fixes, which is easier for me to do than write comments..cool feature!

peterbarker:
> will need a few tweaks and build error fixes, which is easier for me to do than write comments..cool feature!

Thanks Henry.

I've also just tested the Pxihawk6X bootloader when it definitely contains the relevant code, and it all works too.  Hopefully we can be less autopilot-specific now in our descriptions. 

This is now also a related PR: https://github.com/ArduPilot/ardupilot/pull/25144

peterbarker:
We've merged the precursor PRs into ArduPilot master.

Hwurzburg:
are ALL the autopilots with this feature enabled updated on their bootloaders, including bdshot derivations?

peterbarker:
> are ALL the autopilots with this feature enabled updated on their bootloaders, including bdshot derivations?like CUAV Nano which lists it as an included feature?etc.

Sorry, which feature are you refering to?  Where does CUAVNano list it as an included feature?

The easiest way to tell if a bootloader has the support built in is:
```
pbarker@fx:~/rc/ardupilot/Tools/bootloaders(master)$ for i in *bin; do if grep -q 'ardupilot.abin' $i; then echo "$i: yes"; fi; done
CubeOrange_bl.bin: yes
CubeOrangePlus-bdshot_bl.bin: yes
CubeOrangePlus_bl.bin: yes
Pixhawk6X-bdshot_bl.bin: yes
Pixhawk6X_bl.bin: yes
pbarker@fx:~/rc/ardupilot/Tools/bootloaders(master)$ 
```

The bdshot variants of those boards were added sufficiently recently the bootloaders were built with the support in them on initial commit.  Notable absent is the CUAVv5Nano....

I haven't tested the bdshot boards in the above list.

Hwurzburg:
I think I confused this with another feature....would like to chat when you have a chance

Hwurzburg:
okay...made a few edits and build error fixes....even though the bds that have it are only in latest, going to merge this into the wiki now since someone could build their own now with stable

PR #5457, opened 2023-09-30
common: upgrade orientation note to warning

peterbarker:
two instances recent with users not setting orientation before magcal


Hwurzburg:
lgtm...thanks

PR #5471, opened 2023-10-08
Common: xacti fix to wiring diagram

rmackay9:
This is an important fix from Xacti for the camera <-> autopilot wiring.  I'm sure the new diagram is correct.

I have not been able to locally test this change but from inspection I think it should be OK.


Hwurzburg:
the connector pinout table and photo on the page is wrong also, then..probably should get everything coordinated...changing this appears insufficient and will still confuse

rmackay9:
Hi @Hwurzburg,

So there are actually two connectors on the gimbal.  The one that comes from the mount (the top photo) and the connector within the mount that attaches to the camera.  Most users will have a mount so they will refer to the top photo while some small minority may make their own mount (perhaps make the mount part of the vehicle).

This existing line attempts to make clarity this.
The top image shows how to connect the gimbal’s CANL, CANH and GND pins to one of the autopilot’s CAN ports when using the gimbal mounting plate option. The bottom image shows the camera’s pinout which may be used with a custom designed mount.

Hwurzburg:
okay wiki shows them stacked..its clear that way....'view file' here had them side by side and I skimmed over that sentence...merging

PR #5474, opened 2023-10-10
Update building-setup-windows-cygwin.rst

chenyuanzhen:
cygwin waf use /usr/bin/python3 not /usr/bin/python

### DIFF
--- a/dev/source/docs/building-setup-windows-cygwin.rst
+++ b/dev/source/docs/building-setup-windows-cygwin.rst
@@ -127,7 +127,7 @@ Set up directories/paths and extra packages in Cygwin
 
 ::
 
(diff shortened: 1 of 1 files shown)


TunaLobster:
Not bad idea. I haven't had issues with this though. These commands are different/non-existent in the environment setup script. Was there a reason to deviate from what was in the install script? https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-windows.ps1

khancyr:
we should probably nuke the whole instruction and ask to use the script, that would be one less diff to maintain. 

otherwise the change look good as we are now correctly supporting python3 .

PR #5479, opened 2023-10-13
Common: Update aux switch table to be an RST simple table

TunaLobster:
This is a format change that should still allow for easy editting and addition of new switch options. This should also keep the table to a reasonable width on the page.

I haven't had a chance to build yet. Haven't finished settting up on this new machine.


Hwurzburg:
does not change appearance or scroll bar issue....it would be simpler to maintain

TunaLobster:
Figured out that rst tables don't word wrap in the cell. I added a couple of line blocks to force wrapping. So the scroll bar is shorter now.

Hwurzburg:
might as well do that on every long line to prevent it running more than 2x the width and have a short bar

Hwurzburg:
going to merge this...I am going to reformat so the slider is eliminated...will take some time...thanks

PR #5480, opened 2023-10-13
Add information on the mLRS project for rc and telemetry

brad112358:
This may sound a little bit like a testimonial, but I guess I should explain why I think the Ardupilot documentation should include a reference to mLRS.

A little over a year ago, I discovered mLRS which is an open source project intended to  provide integrated RC and MAVLink telemetry at a reasonably long range with comfortable data rates on low cost hardware.  It targets several popular STM32 processors and LoRa chips and currently runs on some off the shelf hardware as well as several DIY boards.  I realized that it could be a better solution for my very small (200 g) Ardupilot based craft than the TBS Crossfire radios I was using at the time.  After fixing a couple of pain points in the code, I switched to mLRS firmware running on FrSky R9 hardware and I have been using it for almost a year.  It has proven to be very reliable and stable even during development and I find it ideal for my use case with reliable MAVLink over a much longer range than I was ever able to get from the small Crossfire receivers which have long advertised MAVLink telemetry support, but never delivered a reliable solution.

The project has seen very active and rapid development over this time and the community has grown with more users contributing code and documentation as well.  mLRS has recently reached the milestone of a stable 1.0 release.  I think many Ardupilot users will also find it has advantages over other telemetry systems.


Hwurzburg (changes requested):

common/source/docs/common-mlrs-rc.rst
```diff
@@ -0,0 +1,23 @@
+.. _common-mlrs-rc:
+[copywiki destination="plane,copter,rover,blimp"]
+=====================
+mLRS project
+=====================
+
```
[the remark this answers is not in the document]
brad112358: OK.  Does turnkey mean no firmware install needed?  Commercially available hardware is available for 900 MHz (FrSky R9) and 2.4 GHz  (Flysky FRM303) which only needs installation of this firmware.  I usually think of DIY as more than just firmware install I guess.  I'll try to make this more clear.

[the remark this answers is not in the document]
Hwurzburg: Give links to ready to flash hardware and point or explain firmware loading,etc...ie.anything different than buying like an ELRS set off the web...

common/source/docs/common-rc-systems.rst
```diff
@@ -64,6 +64,12 @@ FPort/FPort2
 FPort is a bi-directional protocol, using SBus RC in one direction, and serial telemetry in the other. The RC portion can be decoded when attached to an autopilot as if it were SBus, but the embedded telemetry would be lost. See the :ref:`FPort setup documentation<common-FPort-receivers>` for details on connection to one of the autopilots Serial Ports.
 
 
+mLRS
```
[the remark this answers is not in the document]
brad112358: done

common/source/docs/common-mlrs-rc.rst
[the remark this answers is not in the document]
brad112358: Better now?

common/source/docs/common-rc-systems.rst
```diff
@@ -64,6 +64,12 @@ FPort/FPort2
 FPort is a bi-directional protocol, using SBus RC in one direction, and serial telemetry in the other. The RC portion can be decoded when attached to an autopilot as if it were SBus, but the embedded telemetry would be lost. See the :ref:`FPort setup documentation<common-FPort-receivers>` for details on connection to one of the autopilots Serial Ports.
 
 
+:ref:`mLRS <common-mlrs-rc>`
+--------------------------
+
+You can connect the RC output pin of your mLRS receiver to the RX pin of a serial port and configure it for CRSF or SBus (The Tx pin is not used).  Or, you can omit the RC connection and configure the mLRS receiver to output RC channels over the telemetry serial connection by injecting RC_CHANNELS_OVERRIDE messages.
```
Hwurzburg, line 70: question:how do you get telemetry if only the rx pin is connected??

brad112358: mLRS Rx doesn't handle CRSF telemetry so there is no point in connecting the Tx pin on a port configured for CRSF.  Telemetry on the Rx side of the radio link is always via MAVLink.  mLRS is not intended for use with FC firmware like Betaflight which can't do MAVLink so there is no point in bothering to with CRSF telemetry on the Rx side.  It is expected that a full MAVLink connection is always connected between the Rx and FC.

As stated here, even the CRSF or SBus Rx pin is not really required since mLRS can be configured to deliver the Rc channels to the FC via MAVLink RC_CHANNELS_OVERRIDE.  This is how I always configure mLRS so I can avoid one extra wire and serial port on the FC.  The only reason you really need to connect CRSF or SBus pin is if your use case might require the use of RC_CHANNELS_OVERRIDE message for it's normal purpose where it originates from the GCS.

On the Tx module however, mLRS does use a full CRSF connection.  It translates MAVLink received from the FC via the receiver into CRSF telemetry (sensors) for  use by the radio handset to support things like Yaapu Telemetry.

I didn't discus any of these telemetry details at length on this page because this page is about RC.  Do you have a suggestion for what we could change to prevent readers from wondering about this as you did?  Would it help to explicitly state here that mLRS does not support CRSF telemetry on the Rx end of the radio link?  Or, should I expand on this in common-mlrs-rc?

Hwurzburg: okay without fully studying mlrs it was not apparent that there are two rx ports....the telem port and an sbus/crsf output....why even have a crsf rc only protocol option as rc out?? and cant you mix tx generated overrides with gcs overrides? this really needs a typical wiring diagram fir rx<->fc and tx module to wireless gcs link....I can work on the verbiage on these pages to make things clearer (and fix a few formatting errors)....will take some time for me to generate graphics since one hand is in cast

brad112358: I was asking for suggestions, not for you to solve the problem.  If you think a wiring diagram would help, I'll include one in common-mlrs-rc.

As far as why include the option of CRSF and not just SBUS for RC, I'm not sure what the original motivation was.  But it strikes me that having a choice could be handy since SBUS might require an external inverter depending on the specifics of the FC and Rx.

If the FC receives overrides from both the Rx injected messages based on the RC channels and from the GCS, how would it know which ones to use?  Generating and injecting override messages in the Rx and sending them to the FC is a very handy hack which allows eliminating a wire, but it's a hack which assumes no use of override at the GCS end.  Adding other hacks to squash the RC somewhere in the radio link when the GCS sends overrides is tempting, but then you'd open other policy questions that the radio is not prepared to deal with.  The MAVLink protocol really should define a separate RC message for this use case.

brad112358: OK.  I pushed an update with a small clarification and added a connection diagram for CRSF.  Sorry I missed the formatting error earlier.  Did I miss any others?

brad112358: Also, I checked with Olli and he has given his permission to use this diagram under the CC license which the Wiki uses.

Hwurzburg:
okay thanks...made a few edits, squashed....merging

brad112358:
Thanks for the improvements you made and thanks very much for your work here!

Hwurzburg:
@brad112358 is there a forum or discord for questions?....I have one...can the M5stamp s3 be used in place of the m5stamp c3u for use with the r9m module?

brad112358:
Yes, there are links in the readme on git to both discord and rcgroups discussions.  Questions seem to always be welcome.

To try and answer yours:  Looking at the pinout, I suspect it might be made to work, but the sketch doesn't support it yet.  Also, keep in mind that the esp32 s3 and c3 do not support clasic Bluetooth serial.  I prefer Bluetooth over WiFi when connecting to my phone because QGC on my phone can connect automatically without my needing to connect to the WiFi access point first.
I'd recommend the M5stamp pico if you think you might want to use Bluetooth.

PR #5481, opened 2023-10-13
Setup: Add venv for Ubuntu Lunar

TunaLobster:
Taking from @peterbarker's PR for the main repo for a similar environment.


peterbarker:
Looks OK - have thee focal and bionic envs been tested as still working?

We can probably let bionic die if it doesn't work; focal perhaps not.

TunaLobster:
Probably. Following along with the LTS timeline? 

I just realized I have never tried this on Jammy.

Sphinxsetup.sh
```diff
@@ -10,11 +10,23 @@ fi
...
 if [ ${DISTRIBUTION_ID} = 'Ubuntu' ]; then
   DISTRIBUTION_CODENAME=$(lsb_release -c -s)
-  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ]; then
-    sudo add-apt-repository universe
+  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'lunar' ]; then
+    sudo add-apt-repository universe -y
   fi
 fi
 
+PIP_USER_ARGUMENT="$PIP_USER_ARGUMENT"
```
khancyr, line 18: you never add --user ?

TunaLobster: Correct! Left over from a copy paste. Fixed.

Sphinxsetup.sh
```diff
@@ -10,11 +10,21 @@ fi
...
-  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ]; then
-    sudo add-apt-repository universe
+  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'lunar' ]; then
+    sudo add-apt-repository universe -y
   fi
 fi
 
+# create a Python venv on more recent releases:
+if [ ${DISTRIBUTION_CODENAME} == 'lunar' ]; then
+    sudo apt install python3.11-venv
```
Ryanf55, line 20: Why not `python3-venv`?

Sphinxsetup.sh
```diff
@@ -38,32 +48,24 @@ if [ "$(python --version)" == "Python 3.6.9" ]; then
     GET_PIP_URL="https://bootstrap.pypa.io/pip/3.6/get-pip.py"
 fi
 
-curl "$GET_PIP_URL" -o get-pip.py
-python3 get-pip.py
-rm -f get-pip.py
+if [ ${DISTRIBUTION_CODENAME} != 'lunar' ]; then
+  curl "$GET_PIP_URL" -o get-pip.py
```
Ryanf55, line 52: Why not install pip from apt?

khancyr: because you don't have the latest version so it can have some side effect

Ryanf55: But then you can just 
`python3 -m pip install --upgrade pip` as it recommends in the terminal.

TunaLobster:
This PR should change scope a bit since Lunar is EOL. 3 months until 24.04, but most of what's in here is fine. Just needs a new string match.

Sphinxsetup.sh
khancyr: You cannot update systems package with pip, so apt package stay at the repo version.

Sphinxsetup.sh
TunaLobster: Fixed. I didn't realize the default python packages were there. Looks like those are linked to whatever the distro is currently using.

TunaLobster:
Rebased on master. Updated the strings for mantic. 2 months until nobel.

PR #5486, opened 2023-10-19
Recommend WSL instead of Vagrant

MichelleRos:
Seems like the Vagrant instructions are also a bit out of date?
In which case, it's probably best to add this warning to this one as well, like the Cygwin etc pages have.


Hwurzburg:
LGTM

PR #5514, opened 2023-10-30
Improved navabar links

Sentious:
Added hover effects to links in navbar for better user experience.

Screenshot for preview:
image


Hwurzburg:
building this PR locally gives no change in behavior...and the attached screenshot shows me nothing that I can see...what ae we expecting from this pr?

Hwurzburg:
thanks.....merging

PR #5515, opened 2023-10-31
frontend: add another regext to catch blog image in first 5 lines

khancyr:
allow to catch image if they are in the right img field but not have .jpg/.png type in url


rmackay9:
great to see this!!

Hwurzburg:
@khancyr failing CI

khancyr:
yep, should be fixed now

PR #5516, opened 2023-10-31
Fix spacing in hyperlinks quadplane-vtol-tuning-process.rst

magate:


Hwurzburg:
so your fix actually creates build errors...I have fixed pushed back over your changes...thanks...merging

PR #5520, opened 2023-11-01
Update sphinx to newer version as theme was updated

khancyr:


Hwurzburg:
does this require ubuntu22 locally to build the wiki?

khancyr:
@Hwurzburg no, it is just for CI as it is a bit faster

PR #5527, opened 2023-11-04
ESP8266 telemetry: change link to ArduPilot's repo rather than Tridge's

MichelleRos:
The doc page should probably point to the repo in the ArduPilot organisation rather than @tridge's one.

IIUC, this is also the one that gets built for the binaries in the firmware server.


Hwurzburg:
lgtm..thanks

PR #5530, opened 2023-11-05
Rework Horizon31 products

climr:
Horizon31 was acquired from EchoMAV. Product description, links and product names changed to reflect current available hardware.

### DIFF
--- a/dev/source/docs/turnkey-companion-computer-solutions.rst
+++ b/dev/source/docs/turnkey-companion-computer-solutions.rst
@@ -10,36 +10,29 @@ This article lists turnkey companion computer solutions that are advertised as w
 
    Please let us know if you discover a new companion computer, so we can add it to this list. 
 
-Horizon31 - PixC4-Jetson
+EchoMAV - EchoPilot AI
 =========================================================
 
-.. figure:: https://horizon31.com/wp-content/uploads/2020/03/pixc4-jetson_transparent-600x452.png
-   :target:  https://horizon31.com/product/pixc4-jetson/
+.. figure:: https://echomav.com/wp-content/uploads/2020/03/echopilot_ai_1.png
+   :target:  https://echomav.com/product/echopilot-ai/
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
this is a lot more complex than this....the old name appears other places in the wiki, but more importantly, the firmware is named 
PixC4-Jetson and probably will not be renamed and must be referred to on this page...

climr:
Thanks for the feedback. FWIW, the turnkey companion computer page seems independent from other areas of the wiki to me. There are currently no links to/from the firmware to/from this page (unless I missed them). I don’t see the harm or complications associated with removing a product which is no longer being manufactured and replacing it with a turn-key system which is. Seems a net gain for users who’d otherwise be looking at a product which doesn’t exist.

For clarity - Horizon31 PixC4-Jetson has been replaced by the EchoMAV EchoPilot AI (as part of a business acquisition). Although the hardware has been updated, so it is not a direct substitution. As such, I believe existing references in the wiki to the PixC4-Jetson should remain, as those products exist in the wild, but happy to follow guidance from the dev team in how they normally handle deprecated hardware. If you’d like the PixC4-Jetson to remain, it would seem prudent to add a note to those wiki pages indicating it is deprecated hardware no longer being supported. I am happy to create a new PR which does this while adding a wiki page for the current version of hardware - EchoPilot AI. IMO that PR would fix the disconnect between the EchoPilot being listed as a turn-key companion system, but not listed in the supported autopilot list. Also please note that a PR is coming very soon for firmware build support for the EchoPilot AI hardware.

Please advise on next actions. Thank you for the time.

Hwurzburg:
if the new version will have its own hwdef, then it needs to be treated as a different autopilot...I would leave the Horizon31 stuff as is since existing users will want to upgrade firmware over time...when the b=new EchoMav autopilot hwdef is merged we can add it to the wiki...and PixC4-Jetson moved to discontinued boards...I would just close this

PR #5531, opened 2023-11-06
Recommend Sensor Connect for configuration

Ryanf55:
ArduPilot software does not currently configure the MicroStrain External AHRS, yet the Wiki never told users that. Adding a link to Sensor Connect is a good way to let users know they should go configure. 

image


Hwurzburg:
LGTM...thanks

PR #5538, opened 2023-11-07
Common: Xacti gets camera feedback pin setup

rmackay9:
This updates the Xacti page to specify how to connect up the picture feedback pin which improves timing of the EXIF data by about 0.7 seconds.  Connecting this pin also ensure the GCS doesn't receive a message saying a picture was taken when it actually failed (due to missing SD card,etc).

I've tested this wiki update locally and it looks fine to me.


Hwurzburg:
LGTM2...thanks

PR #5539, opened 2023-11-07
companion-computers: add entry for BlueOS

Williangalvani:


Hwurzburg:
this isnt a companion computer...its cloud mangement app...we have a wiki section for those https://ardupilot.org/copter/docs/common-commercial-support.html?highlight=fleet+management#fleet-management-systems

Williangalvani:
> this isnt a companion computer...its cloud mangement app...we have a wiki section for those https://ardupilot.org/copter/docs/common-commercial-support.html?highlight=fleet+management#fleet-management-systems you can also link it in sub since its sub specific

It actually is!

Although the .cloud domain can be confusing, BlueOS itself runs entirely on the pi and is what has been running on BlueRobotics vehicles. We don't actually have a cloud service.

Hwurzburg:
skimming the link and docs there are no links to any hardware...just says your companion computer....what is it?.how to purchase??very confusing....I can merge it but a user will be totally confused

Williangalvani:
> skimming the link and docs there are no links to any hardware...just says your companion computer....what is it?.how to purchase??very confusing....I can merge it but a user will be totally confused

That is good feedback, thanks!
Do you think this would be better?
https://blueos.cloud/docs/blueos/1.1/overview/

Hwurzburg:
you should lead with it being currently implemented for RPI3/4 platforms adding that it could be ported (if it can) to other companion computers.

and the Companion computer wiki entry would be RPI3/4 (and others) with BlueOS <link>   that would not be confusing

PR #5541, opened 2023-11-08
Copter: do-change-speed supports climb and descent rates

rmackay9:
This fixes up Copter's mission page's do-change-speed description to clarify that the Type field is supported.  The image is also updated to show both the Type and Speed m/s fields.

BTW the shared mission list page already shows both type and speed fields correctly.


Hwurzburg:
I need to fix the master list also

PR #5542, opened 2023-11-08
Common: Xacti power on/off setup

rmackay9:
This adds details on how an autopilot relay pin can be used to power on/off the Xacti camera.

Apologies for the drip feed of Xacti changes but it's a journey of discovery with new features coming to light each day.


Hwurzburg:
LGTM..thanks

PR #5551, opened 2023-11-12
Be explicit that humble is the ONLY supported version right now

Ryanf55:
Users are trying to use deprecated ROS versions with ArduPilot. These are not supported.


Hwurzburg:
lgtm...thanks

PR #5552, opened 2023-11-12
Update wind-vane.rst

AlksSAV:
typo fixed in link


AlksSAV:
typo fixed

Hwurzburg:
thanks...but that does not fix it...I have corrected

PR #5557, opened 2023-11-13
Feature/cache params

khancyr:
allow to test param versionning on docker

make faster build_parameters.py by using parallele http call : gain about 2 mins per use


build_parameters.py
```diff
@@ -73,41 +275,156 @@
...
+            file_out.write(line)
+
+    file_in.close()
+    file_out.close()
+
+
+def patch_cgi_escape_for_old_versions(version, param_metadata_dir):
+    """
+    Live patch all Python files in param_metadata for older firmware versions that use cgi.escape()
+    which was removed in Python 3.8. This affects htmlemit.py, rstemit.py, and potentially other files.
```
peterbarker, line 347: 3.8 is our minimum now

khancyr: yes but the parse_param.py on version like 4.0.X got the cgi import and that prevent parameter generation!

PR #5558, opened 2023-11-13
global: add git attribute for easier file management

khancyr:


Hwurzburg:
@peterbarker bump

.gitattributes
```diff
@@ -0,0 +1,15 @@
...
+text eol=lf
+
+# Declare files that will always have LF line endings on checkout.
+*.py text eol=lf
+*.sh text eol=lf
+*.bash text eol=lf
+
+*.bin -diff
+*.elf -diff
+*.hex -diff
```
peterbarker, line 11: Are you expecting these in the Wiki?!

peterbarker (commented):
I really don't like losing the history on that one file, but it's been painful in the past, and if it's required to pass these checks then OK.

.gitattributes
khancyr: right, who know... removed for now

PR #5564, opened 2023-11-15
Rover: boogie board reference frame gets GPS

rmackay9:
Also added a link to the reference frames landing page from the Additional-information landing page.

Tested locally and looks ok to me.


Hwurzburg:
lgtm...thanks

PR #5571, opened 2023-11-18
ubiquitiwarehouse no longer stocks the 16dBi antenna but ipwireless h…

gmorph:
…as it so I updated the link


Hwurzburg:
LGTM...thanks

PR #5577, opened 2023-11-20
Fix inline-code in `the-ardupilot-autotest-framework.rst`

robertlong13:
In rst, code blocks need double backtick or :code:`example`


Hwurzburg:
lgtm

PR #5578, opened 2023-11-21
Common: hobbywing DroneCAN ESC param fix

rmackay9:
This fixes a problem with our Hobbywing DroneCAN setup instructions, namely the CAN_D1_UC_ESC_BM parameter must be set in  order for the ESCs/motors to actually spin.


common/source/docs/common-hobbywing-dronecan-esc.rst
```diff
@@ -29,7 +29,12 @@ Connect to the autopilot with a ground station and set the following parameters
 
 - Set :ref:`CAN_P1_DRIVER<CAN_P1_DRIVER>` to 1 (First Driver)
 - Set :ref:`CAN_D1_PROTOCOL<CAN_D1_PROTOCOL>` to 1 (DroneCAN)
-- Modify :ref:`CAN_D1_UC_OPTION<CAN_D1_UC_OPTION>` so that "Hobbywing ESC" is checked (or directly set to 128)
+- Set :ref:`CAN_D1_UC_ESC_BM<CAN_D1_UC_ESC_BM>` to configure which servo outputs are transmitted over CAN.  For most quadcopters this should be set to 15.
```
Hwurzburg, line 32:
```suggestion
- Set :ref:`CAN_D1_UC_ESC_BM<CAN_D1_UC_ESC_BM>` to configure which servo outputs are transmitted over CAN.  For most quadcopters this should be set to 15 to send outputs 1-4 to DroneCAN.
```
I think we should and the explanation of what the param value actually does,as we do in the other params above

Hwurzburg (changes requested):

PR #5581, opened 2023-11-21
4.5:add Hirth engine/EFI

Hwurzburg:


Hwurzburg:
@Lokesh-Carbonix please review

common/source/docs/common-efi.rst
```diff
@@ -22,6 +22,7 @@ The following units have been tested and are supported:
 - Lutan Serial
 - DroneCAN (adapted versions of the above serial units)
 - PiccoloCAN (`Currawong's ECU <https://www.currawong.aero/ecu/>`__ and `IntelliJect EFI <https://power4flight.com/uav-engine-products/uav-engine-control-units/intelliject-efi/>`__)
+- Hirth Engines
```
Lokesh-Carbonix, line 25: can add the link of the provider https://hirthengines.com/2-stroke-engines/41-series/

common/source/docs/common-efi.rst
```diff
@@ -106,3 +109,22 @@ First, set either RPM1 or RPM2 sensor as being sourced from the EFI (examples wi
 
 - set :ref:`RPM1_TYPE<RPM1_TYPE>` = 3 (EFI)
 - then setup the :ref:`common-rpm-based-notch`
+
+Throttle Linearization
+======================
+
```
Lokesh-Carbonix, line 115: do we need to mention about the AP_EFI_ThrLin Enable parameter?

Lokesh-Carbonix:
@Hwurzburg do we want to add a description for Hirth functionality? Also how it works with IC Engine?

common/source/docs/common-efi.rst
Hwurzburg: yes,will add that param

common/source/docs/common-efi.rst
Hwurzburg: then I will need to add links for the others also...which I will do

Hwurzburg:
> @Hwurzburg do we want to add a description for Hirth functionality? Also how it works with IC Engine?

not necessarily...not opposed...but would need a separate page created just for these enginesa since we don't go into those details on this page...

PR #5616, opened 2023-12-02
flight-modes.rst: remove mention of terrain following for Cruise

robertlong13:
This suggests that Cruise does terrain following out of the box, and that FBWB cannot do terrain following. Both of which are wrong.

### DIFF
--- a/plane/source/docs/flight-modes.rst
+++ b/plane/source/docs/flight-modes.rst
@@ -34,7 +34,7 @@ of functionality available.
    <tr><td>MANUAL</td><td>-</td><td>-</td><td>-</td><td></td><td>Y</td><td>Manual control surface movement, passthrough</td></tr>
    <tr><td>FBWA</td><td>s</td><td>s</td><td>-</td><td></td><td>Y</td><td>Roll and pitch follow stick input, up to set limits</td></tr>
    <tr><td>FBWB</td><td>s</td><td>A</td><td>SPD</td><td>Y</td><td>Y</td><td>like FBWA, but with automatic height and speed control</td></tr>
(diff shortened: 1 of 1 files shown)


plane/source/docs/flight-modes.rst
```diff
@@ -34,7 +34,7 @@ of functionality available.
    <tr><td>MANUAL</td><td>-</td><td>-</td><td>-</td><td></td><td>Y</td><td>Manual control surface movement, passthrough</td></tr>
    <tr><td>FBWA</td><td>s</td><td>s</td><td>-</td><td></td><td>Y</td><td>Roll and pitch follow stick input, up to set limits</td></tr>
    <tr><td>FBWB</td><td>s</td><td>A</td><td>SPD</td><td>Y</td><td>Y</td><td>like FBWA, but with automatic height and speed control</td></tr>
```
Hwurzburg, line 36:
```suggestion
   <tr><td>FBWB</td><td>s</td><td>A</td><td>SPD</td><td>Y</td><td>Y</td><td>like FBWA, but with automatic speed and altitude control, including terrain following, if enabled</td></tr>
```

Hwurzburg (changes requested):
I think terrain following should be mentioned...but clearly....since Cruise is inclusive of FBWB, dont have to mention in both

robertlong13:
If we must mention it, then it should be mentioned for every flight mode that supports it: RTL, Loiter, Cruise, FBWB, Guided and Auto. We can cover Cruise by mentioning it in FBWB, but the others don't really reference that.

We could do it with a column, "Supports Terrain Following" or "Optional Terrain Following". I don't think we should muddle the summary of flight modes with an unrelated optional feature.

We could add the reference to each mode's respective page, like Cruise does. That seems more appropriate.

Hwurzburg:
well I am going to merge as is and then do some work on this page and the "mission sub modes"   which dont exist immediately after

PR #5618, opened 2023-12-03
Fix connected with git config --global url."https://"

vahmelk99:
Valid command is `git config --global url."https://".insteadOf git://` not `git config --global url."https://"`


Hwurzburg:
thanks..

PR #5619, opened 2023-12-03
update.py: correct fetching of Periph parameters

peterbarker:


Hwurzburg:
thanks...fixes it for me now...merging

PR #5621, opened 2023-12-04
Expand introduction to system identification process to also explain …

amilcarlucas:
…how to process the data

This PR contains no :math: elements, those got moved to another PR

PR #5622, opened 2023-12-04
SID add math support

amilcarlucas:
Builds on top of #5621 and adds math support to it

requires #5621


amilcarlucas:
I tested it locally on my machine, it does not work and I still get:
 ```
WARNING: LaTeX command 'latex' cannot be run (needed for math display), check the imgmath_latex setting
```

amilcarlucas:
Now I replaced 'sphinx.ext.imgmath' with 'sphinx.ext.mathjax' and got rid of the error and the page renders math symbols correctly.

PR #5625, opened 2023-12-05
MiniPix: Remove radiolink url and store url

muramura:
ArduPilot expects strict adherence to the GPLV3 license.
For RADIOLINK FCs that ignore the GPLV3 license, we will remove the RADIOLINK URL link and remove the store URL link.


khancyr:
that is fair. Would need to have a proof they don't provide source on demand, even if I doubt they will. They put strong emphasis on the website that they have works with us to fix some issue on their boards.

Most of their firmware link are failing thus. So maybe remove the whole page won't be a big deal

rmackay9 (approved):

Hwurzburg:
closing this as they cooperated

PR #5638, opened 2023-12-08
Update common-radio-control-calibration.rst

Colechr56:
Adding text to point to the later mode channel parameter, for later releases and for Rover, adding MODE_CH.


Colechr56:
Hi Henry
Yes you are right. I haven't done this before, but the version with the
linked reference looks better.
Many thanks CC

Hwurzburg (approved):

PR #5639, opened 2023-12-08
Update SET_ATTITUDE_TARGET to support body rates

BTPankow:
It looks like SET_ATTITUDE_TARGET should fully support body rates with this pull request:
https://github.com/ArduPilot/ardupilot/pull/18833

Updated the SET_ATTITUDE_TARGET table accordingly.

### DIFF
--- a/dev/source/docs/copter-commands-in-guided-mode.rst
+++ b/dev/source/docs/copter-commands-in-guided-mode.rst
@@ -442,7 +442,10 @@ Bitmask to indicate which fields should be **ignored** by the vehicle
 
 bit1:body roll rate, bit2:body pitch rate, bit3:body yaw rate, bit7:throttle, bit8:attitude
 
-Should always be 0b00000111 / 0x07 / 7 (decimal)
+Two values are accepted:
+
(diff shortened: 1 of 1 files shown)


Hwurzburg:
@rmackay9  or @lthall please review and approve, or not...thanks

rmackay9 (changes requested):
Thanks for this.  It's not quite sufficient though I'm afraid.

1. in the Bitmask field we should be providing the two accepted values instead of simply deleting the one that is listed.  This field is the one that callers are most likely going to get wrong (I've seen at least 3 community devs report they couldn't get the call working and it was because of this field).  We can't just delete important information because it's inconvenient to figure out how it should be expanded.
2. examples at the bottom should be expanded to include setting the body rates.

Hwurzburg:
@rmackay9 addressed your comments:
1. type_mask now documents both accepted values (0x07 to use attitude quaternion, 0x80 to use body rates) instead of just deleting the old note.
2. Examples table now includes two body-rate examples (roll rate and yaw rate).

Please re-review, thanks.

PR #5645, opened 2023-12-10
update gps landing page with CubePilot gps

Hwurzburg:
@bugobliterator Sid, maybe I should archive the here+ page and just link to the docs like the others?

### DIFF
--- a/common/source/docs/common-here-plus-gps.rst
+++ b/common/source/docs/common-here-plus-gps.rst
@@ -1,8 +1,8 @@
 .. _common-here-plus-gps:
 
-=============
-Here+ RTK GPS
-=============
+=======================
+CubePilot Here+ RTK GPS
+=======================
 
 This article provides a brief overview of how to setup a Here+ RTK GPS receiver which is based on the Ublox M8P.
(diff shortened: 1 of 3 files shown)


common/source/docs/common-here-plus-gps.rst
```diff
@@ -11,11 +11,6 @@ These instructions focus on the setup using Mission Planner on Windows, but HERE
 .. image:: ../../../images/here-plus-gps.png
 	:target: ../_images/here-plus-gps.png
 
-Where to buy
```
bugobliterator, line 14: Might be better to update this to say something along the lines of: Newer generation of this product can be purchased through CubePilot resellers at https://cubepilot.org/#/reseller/list

bugobliterator (changes requested):

bugobliterator:
> @bugobliterator Sid, maybe I should archive the here+ page and just link to the docs like the others?

I think its best to leave the links as is, and adding where to get newer generation of the product.

Hwurzburg:
done...merging

PR #5654, opened 2023-12-13
dev: mavlink-requesting-data: add note about SERIALn_OPTION bit 12

peterbarker:
... used for ignoring streamrate sets from a GCS

... bit of a run-on sentence, sorry!


Hwurzburg:
thanks

PR #5670, opened 2023-12-21
Update guidance on IMU logging

timtuxworth:
Peter Hall pointed out that raw IMU logging can be used on H7 processors and might be more useful.

### DIFF
--- a/common/source/docs/common-imu-batchsampling.rst
+++ b/common/source/docs/common-imu-batchsampling.rst
@@ -92,6 +92,10 @@ Advanced Configuration and Analysis
 - :ref:`INS_LOG_BAT_LGIN <INS_LOG_BAT_LGIN>` interval between pushing samples to the dataflash log, in ms.  Increase this to reduce the time taken to flush data to the dataflash log, reducing cycle time.  This will be at the expense of increased system load and possibly choking up the dataflash log for other messages
 - :ref:`INS_LOG_BAT_LGCT <INS_LOG_BAT_LGCT>` Number of samples to push to count every :ref:`INS_LOG_BAT_LGIN <INS_LOG_BAT_LGIN>` ms.  Increase this to push more samples each time they are sent to the dataflash log.  Increasing this may cause timing jitter, and possibly choke up the dataflash log for other messages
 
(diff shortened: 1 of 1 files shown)


IamPete1 (approved):
LGTM, thanks.

common/source/docs/common-imu-batchsampling.rst
```diff
@@ -92,6 +92,10 @@ Advanced Configuration and Analysis
 - :ref:`INS_LOG_BAT_LGIN <INS_LOG_BAT_LGIN>` interval between pushing samples to the dataflash log, in ms.  Increase this to reduce the time taken to flush data to the dataflash log, reducing cycle time.  This will be at the expense of increased system load and possibly choking up the dataflash log for other messages
 - :ref:`INS_LOG_BAT_LGCT <INS_LOG_BAT_LGCT>` Number of samples to push to count every :ref:`INS_LOG_BAT_LGIN <INS_LOG_BAT_LGIN>` ms.  Increase this to push more samples each time they are sent to the dataflash log.  Increasing this may cause timing jitter, and possibly choke up the dataflash log for other messages
 
+.. note:: On an H7 based autopilot such as the CubeOrange or Zealot H743, raw IMU logging can be used and will provide better data for analysis. Set INS_LOG_BITMASK = 0 and sent the raw IMU bit (bit 19) in LOG_BITMASK. Raw logging is very helpful when doing filter tuning, but will give you really large logs which generally means you don't want to leave it on. Batch logging on the other hand can be left on all the time which helps with ad-hoc analysis.
```
Hwurzburg, line 95: params should be linked

Hwurzburg (changes requested):

common/source/docs/common-imu-batchsampling.rst
Hwurzburg: not sure I agree with leaving batch logging on all the time

timtuxworth: updated

Hwurzburg:
if batch sampling is okay to be on all the time we would need to change defaults....if @tridge approves, we can do so and this recommendation can remain, otherwise I think we need to NOT recommend doing this

timtuxworth:
> if batch sampling is okay to be on all the time we would need to change defaults....if @tridge approves, we can do so and this recommendation can remain, otherwise I think we need to NOT recommend doing this

After recent changes, it's only being recommended on H7 APs. Will that work?

Hwurzburg:
Its a matter of unnecessary log size

Hwurzburg:
Checked with Tridge and Tom, all agree we should not recommend batch logging when not actually needing the log data....drop that and I can merge

timtuxworth:
> Checked with Tridge and Tom, all agree we should not recommend batch logging when not actually needing the log data....drop that and I can merge

Fixed and squashed. Thanks @Hwurzburg @andyp1per and @IamPete1

Hwurzburg:
has a build error, but will fix that after I merge...thanks

PR #5672, opened 2023-12-22
modules: chat update to improve setup and usage instructions

rmackay9:
This improves the chat module setup and usage instructions.  I've tested this on my local machine and it looks OK to me.


Hwurzburg:
LGTM...thanks...

PR #5681, opened 2023-12-26
Update loading firmware page with custom firmware instructions

rmaia3d:
Noticed there was a lack of clear documentation on this procedure, so added the steps based on the ChibiOS loading page (which is very easy to overlook) and some recent Discord chats.


Hwurzburg:
thanks, I have re-org'd a  bit and squashed commits...merging....I need to revisit all the loading pages and update since MP has changed a bit...will put that on my list

rmaia3d:
> thanks, I have re-org'd a bit and squashed commits...merging....I need to revisit all the loading pages and update since MP has changed a bit...will put that on my list

Very nice, thanks a lot!!!

PR #5682, opened 2023-12-26
OEM neuter moving basline setup page

Hwurzburg:
How about this Sid?

### DIFF
--- a/common/source/docs/common-gps-for-yaw.rst
+++ b/common/source/docs/common-gps-for-yaw.rst
@@ -40,8 +40,8 @@ Common Parameter Setup
 
 .. note:: Do not use :ref:`GPS_AUTO_SWITCH<GPS_AUTO_SWITCH>` = 2 (Blend) when using Moving Baseline configurations.
 
-Dual Serial GPS
----------------
+Dual Serial F9P GPS
+-------------------
 
 - :ref:`SERIAL3_PROTOCOL<SERIAL3_PROTOCOL>` = 5 ("GPS") assuming the 1st GPS is connected to SERIAL3.
 - :ref:`SERIAL4_PROTOCOL <SERIAL4_PROTOCOL>` = 5 ("GPS") assuming the 2nd GPS is connected to serial port 4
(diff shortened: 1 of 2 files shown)


common/source/docs/common-gps-for-yaw.rst
```diff
@@ -73,65 +73,44 @@ the two GPS modules to go via the autopilot board.
...
+NMEA
+~~~~
 
 - :ref:`SERIAL3_PROTOCOL<SERIAL3_PROTOCOL>` = 5 ("GPS") assuming the GPS is connected to SERIAL3 (be sure any lower numbered port does not use this protocol unless a GPS is attached).
 - :ref:`GPS_TYPE<GPS_TYPE>` = 5 (NMEA)
 
-This system requires that the "Master" antenna and "Slave" antenna (see manufacturer's documentation for which is antenna is designated the "Master") be mounted on the vehicle front to back in line with the 0 degree yaw of the vehicle and at the same vertical level and be at least 30cm apart.
-
-.. note:: this unit can be used with only its "Master" antenna connected, if desired, but no yaw information should be used.
+Some of these systems require that the "Master" antenna and "Slave" antenna (see manufacturer's documentation for which is antenna is designated the "Master") be mounted on the vehicle front to back in line with the 0 degree yaw of the vehicle and at the same vertical level and be at least 30cm apart. Otherwise , the antenna offset distances in the x/y/z directions must be entered detailed in the :ref:`Master-Slave Antenna Offsets<antenna-offsets>` section below.
```
rmackay9, line 86: do we call them "master" and "slave" or is it now "base" and "rover"?

Hwurzburg: base and rover is used in fixed baseline....we have always used master and slave for moving baseline

bugobliterator: I don't think that is correct. In our parameters we use Base and Rover. And that is the role they serve even in moving baseline. Base is the one generating corrections, and rover consumes those corrections to create a baseline. So the role is not so different compared to fixed baseline. So I think to ensure consistency with our code, parameter and even GPS module docs best to use Base and Rover nomenclature.

Hwurzburg: sorry to disagree, many companies actually mark their ant connections as Master and Slave...

PR #5693, opened 2024-01-01
Re-org camera and gimbal pages for easier navigation

Hwurzburg:


common/source/docs/common-cameras-and-gimbals.rst
```diff
@@ -17,45 +17,56 @@ Gimbals
 
 ArduPilot supports both brushless direct drive gimbals (Tarot, SimpleBGC, SToRM32)
 that have their own self-stabilization controllers and the simpler servo-driven
-gimbals in which ArduPilot controls the stabilisation.
+gimbals in which ArduPilot controls the stabilisation. Some gimbals also integrate still or video cameras.
 
+-  :ref:`Servo Gimbals <common-camera-gimbal>` — older-style servo-driven gimbal where ArduPilot provides stabilisation
 -  :ref:`Brushless PWM <common-brushless-pwm-gimbal>` - brushless gimbals that accept PWM or SBUS input for angle control
+
```
rmackay9, line 24: I'd prefer we don't have this separation.  It disturbs the flow and I don't think that users will really care about the distinction between the more DIY (above) and the OEM below.

common/source/docs/common-cameras-and-gimbals.rst
```diff
@@ -17,45 +17,56 @@ Gimbals
...
+camera (including its shutter or video recording). The following topics explain how to configure camera shutter and controls, and list a number of different approaches for converting
+the output signal into the form expected by your particular camera. If the camera is integrated with the gimbal, control of the camera is detailed in that gimbals setup.
+
+General Camera Control and Parameters
+-------------------------------------
+
+-  :ref:`Camera General Parameters and Control<common-camera-controls>`
+-  :ref:`RunCam Camera Control <common-camera-runcam>`
+
+Shutter Control
```
rmackay9, line 53: Here too I'm not a real fan of splitting up the groups.  I think we should just keep them in alphabetical order and list as many as we have.  Thanks for finding the missing items!

rmackay9:
Hi, thanks for having a go at cleaning up the cameras page.  I definitely agree it needs some work and most of this is an improvement, just the separation into groups that I'm not sure about.

common/source/docs/common-cameras-and-gimbals.rst
Hwurzburg: I think that recombining the Gimbals is okay...and will do

common/source/docs/common-cameras-and-gimbals.rst
Hwurzburg: I disagree on this....but I think I need to move the geotagger up into shutter control section....there is really only one camera here, the FLIR, otherwise this whole section is about camera shutter/triggers, not cameras....I would re-title Shutter Control  to Shutter/Trigger Controllers and move the geotagger into it since it includes trigger controller...the "Control of Specific Camera Models" would only have the FLIR and be retitled to "Cameras with Electronic Trigger Integrated"

rmackay9: The RunCam is also a camera I think..

Hwurzburg: yep, will move down...thanks

PR #5706, opened 2024-01-10
Add note to mavinit.rst regarding startup script scope

kramer65:


Hwurzburg:
I have modified this a bit and pushed to your changes directly since you are probably using the Web editor and any suggestions would be more difficult for you to implement. Thanks for the contribution. Merging

PR #5725, opened 2024-01-16
Documenting Fractional Turn Behavior

jrbronkar:
https://github.com/ArduPilot/ardupilot/pull/25873#issuecomment-1892908613

PR to add documentation for the fractional turn behavior once the above linked PR merges.


Hwurzburg:
thanks....but also here https://ardupilot.org/copter/docs/mission-command-list.html

PR #5726, opened 2024-01-17
common: update prearm check explanations

rmackay9:
This adds a table of pre-arm checks that may be displayed to users along with a "Cause" and "Solution".

This covers pretty much all non-vehicle specfic arming checks and Copter specific checks.  I.e. Plane, Sub and Blimp specific checks are not yet covered.

I've tested this locally now and it looks OK to me.

@Hwurzburg maybe we could merge this after you move a few of the most common items to the top?  I might move a few up in a future PR.

It's hard to say if this massive table will be useful for not but I hope that if it is complete and provides enough detail for each failure that it will slightly reduce the number of support calls or at least give us a page to point newbies at when they ask what a failure message means.

### DIFF
--- a/common/source/docs/common-prearm-safety-checks.rst
+++ b/common/source/docs/common-prearm-safety-checks.rst
@@ -33,17 +33,228 @@ determine exactly which check has failed:
 
 Pre-arm checks that are failing will also be sent as messages to the GCS while disarmed, about every 30 seconds. If you wish to disable this and have them sent only when an attempt arm fails, then set the :ref:`ARMING_OPTIONS<ARMING_OPTIONS>` bit 1 (value 1).
 
-Most are obvious as to the issue, but some can have multiple sources and the message cant identify them all in its limited length. The most common of those are:
+.. table:: Pre-Arm Failure Messages
+    :widths: auto
 
-**AHRS: waiting for home** : autopilot has not gotten a home location, usually because the GPS is detached or has not gotten a fix yet with enough precision. Common experience while indoors.
-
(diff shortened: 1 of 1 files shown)


Hwurzburg:
I think getting a more complete listing of the pre-arm messages and suggested resolutions is a good thing...by table or by line items as is the current format....
I dont think tabularizing them will help anyone find their pre-arm message....the wiki search box will take each component word of the phrase and match EVERY instance in the wiki of each word, unfortunately....
I do think that they should be grouped by at least the most commonly encountered messages and ones with most unobvious solutions first ..... eg. AHRS: not using configured AHRS type, ...EKF3 Roll/Pitch inconsistent by x degs...etc.

btw, there is one  message I see often but cant seem to find in the code: Terrain data missing or unavailable.....or something to that effect until the GPS gets a fix.....I dont use Terrain for navigation but do display the alt in OSD...maybe someone can find it

rmackay9:
Hi @Hwurzburg,

OK, I'll continue and complete the list.  This PR already includes nearly everything from AP_Arming.cpp and the libraries that it calls so it is just the vehicle specific messages that are missing.

Re grouping, I'll just do it alphabetically then if that's OK.  Most related messages will end up grouped together but there will be a few which are out-of-place but then maybe those messages should be fixed instead.

Hwurzburg:
are you going to add Rover and Plane's?

rmackay9:
Hi @Hwurzburg,

I'm going to add Rover at least but in a follow-up PR if that's OK.  I might leave Plane out 'cuz I'd just be guessing the solution half the time.

Hwurzburg:
so its ready for merge as far as you are concerned? I will add the Plane arming this weekend and move a couple items up after its merged, if you are ready

rmackay9:
@Hwurzburg,

Yes, I think this is ready to merge.

PR #5727, opened 2024-01-18
Fix BATT3_SUM_MASK documentation

rishabsingh3003:
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_BattMonitor/AP_BattMonitor_Sum.cpp#L63
The code skips lower instances and only sums higher instances.
This tripped up a Partner and they were getting "Battery unhealthy" pre-arm


Hwurzburg:
thanks

PR #5731, opened 2024-01-21
Document waf tools for submodule, disable recursion

Ryanf55:
Document the following PR: https://github.com/ArduPilot/ardupilot/pull/25196

And, document a preventative fix for the following issue: https://github.com/ArduPilot/ardupilot/issues/26039

This will be discussed in Dev Call Jan 22


Hwurzburg:
what was the devcall conclusion?

Ryanf55:
> what was the devcall conclusion?

I didn't get a clear answer. Seemed like people didn't want to document advanced options in the wiki that aren't default. Tridge concluded this problem is due to a bug in git, but no one could find an upstream issue for it in `git`. I'm just going to close this.

PR #5746, opened 2024-01-31
fixed and clarified RealFlight configuration instruction

jordancormack:
In addition to the 'Configure RealFlight' section, there was a separate section for enabling RealFlight Link, however the instruction for RealFlight 8/9 was incorrect, and the instruction for RealFlight Evolution was repeated in the following section.

This change keeps only the 'Configure RealFlight' section, adding in the instruction for Link in RealFlight 8/9, and moving the restart instruction to the end of the bullet point list.


Hwurzburg:
lgtm,,,thanks

PR #5757, opened 2024-02-02
Common: correct Xacti wiring diagram

rmackay9:
This is quite a serious mistake so glad we caught it before too many users have started using these camera gimbals.


Hwurzburg:
tnx

PR #5763, opened 2024-02-04
Anyleaf h7

Hwurzburg:


David-OConnor:
LGTM, with the following change: This is a suitable image: https://www.anyleaf.org/static/images/h7_fc_top_angle.jpg 

The one linked is for a different FC, that is not compatible with AP.

Hwurzburg:
done...thanks

PR #5768, opened 2024-02-06
Dev: update GSoC ideas list

rmackay9:
I've removed three old projects and added two new ones.

I've run this locally and it looks OK to me.


Hwurzburg:
tnx

PR #5769, opened 2024-02-06
Copter: hexsoon edu650 minor fixes

rmackay9:
This fixes two minor issues with the EDU650 reference frame wiki page:

1. 40cm battery straps are required (not 50cm)
2. The LED wiring was backwards (I had taken a guess and of course, given a 50/50 chance of being right, I was wrong).  If you look closely at the digram you'll see the red and black lines have been swapped.

I've built the page on my local machine and it looks OK.


Hwurzburg:
@rmackay9 are those signal or power leads....if signal (which I suspect) change them from red/black to any other color otherwise someone will tie them to the servo power and ground rails!

Hwurzburg (changes requested):
confirmed they are signal not power....please change

rmackay9:
Hi @Hwurzburg,

Yes, they're signal wires, I agree the colours are not great but those are the colours of the wires... red and black.. not choice! :-).

rmackay9:
@Hwurzburg,

I added some white lines to the red and black wires to make them look a little more like the real thing. I'm not sure I totally succeeded.. I probably should have made the white lines a little thicker..

Hwurzburg:
@rmackay9 and I was about to add a written note on the drawing: Connect AUX5/6 to PDB connector pins 4/5
but then I realized that wiring does not match the parameter setup AUX 5 is data which is pin 5 on the connector, not 4 per the Cube PDB docs...we can change the drawing or change the param setup in the document

Hwurzburg:
tnx...LGTM

PR #5770, opened 2024-02-06
common-ekf-sources: Add comments around source options

Gone4Dirt:
I think it is worth adding some information about the source options bit in the switching page.

### DIFF
--- a/common/source/docs/common-ekf-sources.rst
+++ b/common/source/docs/common-ekf-sources.rst
@@ -23,8 +23,14 @@ The options for these are generally self-explanatory. However, the :ref:`EK3_SRC
 Source Switching
 ================
 
-Three set of EKF position and velocity source parameters are provided which can be selected via an :ref:`RC switch <common-auxiliary-functions>` set to "90". Otherwise, the SRC1 set is used.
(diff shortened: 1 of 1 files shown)


Gone4Dirt:
@rmackay9 does this seem correct and appropriate to you?

rmackay9 (approved):

common/source/docs/common-ekf-sources.rst
```diff
@@ -27,4 +27,10 @@ Three set of EKF position and velocity source parameters are provided which can
 
 This feature is especially helpful when using :ref:`common-non-gps-to-gps`.
 
+Source Switching
```
rmackay9, line 30: I think this title should be updated to EK3_SRC_OPTIONS? .. or something like that?  otherwise we have two identical titles.

rmackay9 (changes requested):
Sorry, I approved but then looked a little closer and noticed the title issue.

common/source/docs/common-ekf-sources.rst
Hwurzburg: I fixed this and some other things

PR #5771, opened 2024-02-07
Common: add camera and mount aux functions

rmackay9:
I thought that these were automatically generated but perhaps not.


rmackay9:
FYI @peterbarker

PR #5772, opened 2024-02-07
Common: fixup links and wording on raw-imu-logging

rmackay9:
This includes some slight re-wording and more importantly fixes broken links on the [Raw IMU Logging for FFT analysis page](https://ardupilot.org/copter/docs/common-raw-imu-logging.html).

As a side note, I think our notch filter setup wiki pages are overly complex and are in the wrong part of the wiki.  They're under logging but I think they should be in the Tuning or Advanced Configuration sections of the wiki.

FYI @andyp1per

### DIFF
--- a/common/source/docs/common-raw-imu-logging.rst
+++ b/common/source/docs/common-raw-imu-logging.rst
@@ -1,22 +1,22 @@
 .. _common-raw-imu-logging:
 
-======================================
-Using Raw IMU Logging for FFT Analysis
-======================================
+================================
+Raw IMU Logging for FFT Analysis
+================================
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
@rmackay9 Randy, I am not sure I agree with your statement about being in the wrong place...batch and raw sampled logging can be used as general vibration diagnosis tools....they are logging techniques.....the tuning sections of the wiki contain the common-imu-notch-filtering filter setup page, which refers the user to the two logging methods for center freq determination and how to gather the data and interpret them depending on which is used...
setting up the filters is complex due to the many center freq adjustment methods, the many options for harmonics configurations...I am not sure how you can simplify without reducing the number of configuration options

rmackay9:
@Hwurzburg,

Well, this PR doesn't try to resolve the placement issue.. it just does minor corrections on wording and links.

Hwurzburg:
lgtm

PR #5773, opened 2024-02-07
Fix image url on https://ardupilot.org/plane/docs/common-installing-3dr-ublox-gps-compass-module.html

valeriyvan:


Hwurzburg:
thanks

PR #5776, opened 2024-02-08
Common: fix hexsoon-edu650 param link

rmackay9:
This flight code PR changed the hexsoon edu650 parameter file name so this PR fixes the link in the wiki.

This has been tested locally and looks ok.


Hwurzburg:
tnx

Hwurzburg:
@rmackay9 shouldnt this have the ProfiLED setup also????

rmackay9:
@Hwurzburg,

Yes, you're right.  We should add the ProfiLED parameters to the [edu650 parameter file](https://github.com/ArduPilot/ardupilot/blob/master/Tools/Frame_params/hexsoon-edu650.param).  I think I was planning to do that with [this PR](https://github.com/ArduPilot/ardupilot/pull/26158) but then forgot.

PR #5783, opened 2024-02-13
Fix copy/paste setting-up-for-tuning.rst

timtuxworth:
This looks like a simple copy and paste from the quadplane docs, but this is a Q parameter, should be MOT I think.


Hwurzburg:
tnx

PR #5784, opened 2024-02-14
Common: xacti camera settings via dronecan

rmackay9:
This adds a little section to the bottom of the Xacti camera gimbal page to show users how to set additional settings via DroneCAN.


rmackay9:
Thanks!

PR #5788, opened 2024-02-18
Add bicopter frame type

cloudpresser:
Bicopter tilt type was introduced in https://github.com/ArduPilot/ardupilot/pull/10793 but the frame type required is not listed here.
Should probably include bicopter frame type here since it's used in the bicopter tilt type


IamPete1 (approved):

Hwurzburg:
thanks!

PR #5790, opened 2024-02-19
Remove reference to centidegrees in QuadPlane tips

timtuxworth:
The parameter name had been changed but not the following text that was still referring to the old usage (centidegrees).


Hwurzburg:
lgtm,tnx

PR #5793, opened 2024-02-21
Clarify X vs. H frame type selection for quadplanes

LoftedAero:
The existing wording regarding X vs H motor mixing for quadplanes recommend H mixing for all geometries. While it correctly outlines the effects of induced airframe twist on yaw performance for quadplanes with X mixing and fuselage-mounted motors, it omits that this effect would also occur when using H mixing with wing-mounted motors. 

This edit clarifies which geometries of quadplanes should use X or H mixing to avoid the effects of harmful airframe twist.

### DIFF
--- a/plane/source/docs/quadplane-tips.rst
+++ b/plane/source/docs/quadplane-tips.rst
@@ -31,14 +31,10 @@ For vectored thrust frames, fixed-wing mode alignment is also critical. Very sma
 X vs H Quad Mixing
 ==================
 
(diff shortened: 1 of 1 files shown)


robertlong13:
Your explanation and wording in your first paragraph is spot-on. Well put.

I have two minor suggestions, feel free to take them if you want. For each bullet point for the two frames,
1. I'd reverse the order to lead with which frame you *should* use.
2. I'd use "counteracts" instead of "combats"

Hwurzburg:
Since he did this from the web editor making changes are hare, so I incoproated @robertlong13  suggested changes...merging...thanks

PR #5798, opened 2024-02-28
Common: JRE 30 gets CAN support

rmackay9:
JRE emailed me recently to tell me that the JRE-30 has been extended to support CAN now so I've added this to the wiki.

I've built these changes locally and they look OK to me.


Hwurzburg:
tnx

PR #5800, opened 2024-02-29
Common: changed JRE 30 image files

jfbblue0922:
In order to make the connection image easier to understand, 
we have changed the image that has already been uploaded from the shape of the prototype to the shape of the product.


jfbblue0922:
We have added a description of the data rate (1Mbps) to this PR.

Hwurzburg:
lgtm

PR #5810, opened 2024-03-06
Update warning for centimeters on fly-by-wire-low-altitude-limit.rst

timtuxworth:
Fix warning that is no longer correct since parameter rework in 4.5.0 CRUISE_ALT_FLOOR is now in meters.


Hwurzburg:
tnx

PR #5812, opened 2024-03-08
Add Hargrave DroneCAN ESC documentation

Vanja-APD:
Add documentation for Hargrave DroneCAN ESCs


common/source/docs/common-escs-and-motors.rst
```diff
@@ -49,6 +49,7 @@ Protocols
     DroneCAN ESCs <common-uavcan-escs>
     Currawong Velocity ESCs <common-velocity-can-escs>
     Hobbywing DroneCAN ESCs <common-hobbywing-dronecan-esc>
+    Hargrave Technologies DroneCAN ESCs <common-hargrave-dronecan-escs>
```
Hwurzburg, line 52: remove this is not using a unique comm prtocol...its dronecan and listing it under dronecan escs is the correct location you will need a hidden toclist for it at the bottom of the page

Hwurzburg (changes requested):
if you need help moving the page to a hidden toclist, I can help...also the commits should be squashed to a single commit

PR #5813, opened 2024-03-08
Dev: update VSCode debug instructions

rmackay9:
This updates the images on the VSCode debug wiki page and slightly clarifies some of the wording.  For example it previously said, "Click and select C/C++ : (gdb) Attach option." but it didn't clarify **where** to click.

I've deleted one image, "VSCode-attach1.png" because I've combined the information it held with the "VSCode-config2.png" image.

@magicrub knows a lot more than I do about this setup but I think this makes it slightly clearer.  Also VSCode's layout has changed slightly since the page was first written.


magicrub (approved):
I agree with all those changes. Thanks @rmackay9 !

hendjoshsr71 (approved):
Ahh yes this is much better!

Hwurzburg:
tnx

PR #5820, opened 2024-03-11
Simplify clone by using URL method

Ryanf55:
Simplify clone by not needing to create `src` directory, and rely on `vcs` to clone from URL automatically.


Hwurzburg:
lgtm...tnx

PR #5823, opened 2024-03-14
Common: ModalAI VOXL page gets links to optflow+rangefinder transition applet

rmackay9:
This updates the ModalAI wiki page to include changes in PR https://github.com/ArduPilot/ardupilot/pull/26277 which is included in Copter-4.5.0-beta3.  The changes include:

- explanation of the VISO_QUAL_MIN parameter
- link to a new Lua applet that automatically switches to optical flow + rangefinder if the ModalAI fails

I've also included a bit of a drive-by change to the developer wiki to remove some references to EKF2.  EKF3 has been the default for quite some time now so I don't think we need to mention EKF2 anymore.

I've tested these changes locally and they look OK to me.


Hwurzburg:
tnx

rmackay9:
@Hwurzburg,

Thanks!  I actually just found another unrelated issue with the ModalAI so I will create another PR for this page.  Sorry for the noise!

PR #5824, opened 2024-03-14
Common: modalai gets optional SR2_EXTRA3 setting

rmackay9:
This adds a workaround to an obscure issue with the ModalAI camera's processing of the [SYSTEM_TIME message](https://mavlink.io/en/messages/common.html#SYSTEM_TIME).

It seems that the camera loses its position estimate after about a minute if sent a stream of SYSTEM_TIME messages where the "time_unix_usec" field is zero but the "time_boot_ms" field is non-zero.  The easiest workaround for now is to ensure that the SRx_EXTRA3 parameter is set to zero so the message is not sent to the camera at all.

I've also reported this to ModalAI and hopefully they can figure out the cause.

### DIFF
--- a/common/source/docs/common-modalai-voxl.rst
+++ b/common/source/docs/common-modalai-voxl.rst
@@ -54,6 +54,7 @@ Connect to the autopilot with a ground station (i.e. Mission Planner) and check 
 - :ref:`SERIAL2_PROTOCOL <SERIAL2_PROTOCOL>` = 2 (MAVLink2).  Note this assumes the camera is connected to the autopilot's "Telem2" port.
 - :ref:`SERIAL2_BAUD <SERIAL2_BAUD>` = 921 (921600 baud)
 - Optionally set :ref:`SERIAL2_OPTIONS <SERIAL2_OPTIONS>` = 1024 (Don't forward mavlink to/from) to disable the camera's odometry messages from being sent to the GCS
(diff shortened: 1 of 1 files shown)


common/source/docs/common-modalai-voxl.rst
```diff
@@ -54,6 +54,7 @@ Connect to the autopilot with a ground station (i.e. Mission Planner) and check
 - :ref:`SERIAL2_PROTOCOL <SERIAL2_PROTOCOL>` = 2 (MAVLink2).  Note this assumes the camera is connected to the autopilot's "Telem2" port.
 - :ref:`SERIAL2_BAUD <SERIAL2_BAUD>` = 921 (921600 baud)
 - Optionally set :ref:`SERIAL2_OPTIONS <SERIAL2_OPTIONS>` = 1024 (Don't forward mavlink to/from) to disable the camera's odometry messages from being sent to the GCS
+- Optionally set :ref:`SR2_EXTRA3 <SR2_EXTRA3>` = 0 to disable sending the SYSTEM_TIME message to the camera which has been known to cause the camera to lose its position estimate (e.g. quality falls to -1).  Note this assumes the camera is connected to the autopilot's second mavlink port (e.g. Telem2)
```
Hwurzburg, line 57:
```suggestion
- Optionally set :ref:`SR2_EXTRA3 <SR2_EXTRA3>` = 0 to disable sending the SYSTEM_TIME message to the camera which has been known to cause the camera to lose its position estimate (e.g. quality falls to -1).  Note this assumes the camera is connected to the autopilot's second mavlink port (e.g. usually Telem2)
```

Hwurzburg (dismissed):

Hwurzburg:
tnx

PR #5829, opened 2024-03-18
dev: mavlink interface for set-camera-source

rmackay9:
This adds mavlink interface info for the new SET_CAMERA_SOURCE mavlink command

This shouldn't be merged until this PR is merged https://github.com/ArduPilot/ardupilot/pull/26172


rmackay9:
Support for the flight code PR has gone into master/latest now so I think we can merge this.  Note that this is a change to the developer wiki so I think it's OK to go in even though the feature hasn't gone out in a stable release yet.

PR #5837, opened 2024-03-24
Dev: Update Support Engineer DroneCAN GUI Notes

hendjoshsr71:
The signing key in DroneCAN GUI refers to the password to generate the key.


Hwurzburg:
tnx

PR #5844, opened 2024-03-26
ADS-B Extlinks: Fixed links and removed dead link uav-store.de

anttiryt:
ADS-B Extlinks: Fixed links and removed dead link uav-store.de. 

Small fixes to other stores and removed dead uav-store.de. Domain killed.


Hwurzburg:
tnx...lgtm

PR #5846, opened 2024-03-27
Added ModalAI VOXL 2 as a compatible companion computer

lululauren101:
Hello, I've added ModalAI's VOXL 2 as an Ardupilot compatible companion computer. This is the link to our documentation: https://docs.modalai.com/voxl2-external-flight-controller/#voxl-2-as-companion-computer-ardupilot-sw-setup


rmackay9:
Hi @lululauren101,

I wonder if instead we should update our existing page for the VOXL1 to make it clear that the VOXL2 works as well.  https://ardupilot.org/copter/docs/common-modalai-voxl.html

.. or we could make a new page.

Hwurzburg:
I think that the VOXL2 should have a link in both companion and nongps lists

PR #5863, opened 2024-04-07
APM references in documentation.

geofrancis:
I dont think this page has been updated in a long time, It talks about APM controllers .


Hwurzburg:
I think the entire page should be removed.....tuning is better done using Autotunee for a model to better account for its setup/weight/throws, etc....will discuss with team

PR #5873, opened 2024-04-11
Update common-mount-targeting.rst to better describe SysID targeting

timtuxworth:
Just make it clearer how to target another vehicle.


Hwurzburg:
tnx

PR #5881, opened 2024-04-16
Common: hereflow gets brd-boot-delay setting

rmackay9:
We discovered soon after the 4.5.0 release that some users found the HereFlow sensor was not working with some autopilots and the workaround was to slow the board's boot up time.  This PR adds this advice to the HereFlow wiki page even though it will soon be resolved for at least some boards.


Hwurzburg:
tnx

PR #5883, opened 2024-04-18
Copter: improve GUID_OPTIONS parameter explanation

rmackay9:
This PR improves the Copter Guided mode GUID_OPTIONS parameter explanation.

This is in response to [this Copter 4.5 support issue](https://discuss.ardupilot.org/t/drone-not-avoiding-obstacles/116293) in which the user was attempting to get BendyRuler working in Guided mode.  It turns out that one of the GUID_OPTIONS bits needs to be set for this to work.

Some other changes are also included:

- Guide mode links changed to the telemetry landing page instead of the SiK radio.
- Guided mode link fixed for GUID_TIMEOUT
- Dijkstra's loses mention of ver 4.0
- Dijkstra's mentions the GUID_OPTIONS parameter setting (for Copter)
- Weathervaning is added to the Advanced Configuration landing page.  It's only in the Flight Features at the moment but the Advanced Configuration page is where we should have a full list of advanced-but-not-commonly-used features

This has been tested locally and looks good to me.

### DIFF
--- a/common/source/docs/common-advanced-configuration.rst
+++ b/common/source/docs/common-advanced-configuration.rst
@@ -143,5 +143,6 @@ tuning options for the vehicle.
     Vibration Damping <common-vibration-damping>
 [/site]
 [site wiki="copter"]
+    Weathervaning <weathervaning>
     Windspeed Estimation and Baro Compensation <airspeed-estimation>
 [/site]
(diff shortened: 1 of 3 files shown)


Hwurzburg:
tnx

PR #5887, opened 2024-04-19
Add steps for verbose build logs

Ryanf55:
We've had a ton of users have problems building because they skip steps in the Wiki. 
This command will increase verbosity of logs, which makes it easier to troubleshoot.
Users knew to both ROS 2 and ArduPilot are likely not familiar with colcon.

Upstream: https://github.com/ArduPilot/ardupilot_gz/issues/50#issuecomment-2065892858


Hwurzburg:
lgtm...thanks

PR #5897, opened 2024-04-24
Common: gremsy gets new ZIO warning

rmackay9:
It seems like the ZIO cannot be used as a 2nd gimbal at least according to testing on [this 4.5 support thread](https://discuss.ardupilot.org/t/gremsy-zio-does-not-work-if-set-to-mount-2/116127).

I've informed Gremsy as well.


Hwurzburg:
tnx

PR #5899, opened 2024-04-24
add info on scurve behavior

Hwurzburg:
I think this may be enough info to help people figure out why the defaults for speed and radius , and intervening DO cmds seem to give a non-Scurving path

### DIFF
--- a/copter/source/docs/auto-mode.rst
+++ b/copter/source/docs/auto-mode.rst
@@ -22,6 +22,8 @@ Overview
 
 AUTO mode navigates between each waypoint in a straight line and as it approaches each waypoint follows a smooth curving path towards the next waypoint (S-Curve). It controls the path as it approaches a waypoint to be within the WPNAV_RADIUS of each waypoint. The speed is lowered below :ref:`WPNAV_SPEED<WPNAV_SPEED>`, as necessary, to keep the path within the set corner lateral acceleration limit parameter (:ref:`WPNAV_ACCEL_C<WPNAV_ACCEL_C>`).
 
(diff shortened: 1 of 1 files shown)


copter/source/docs/auto-mode.rst
```diff
@@ -22,6 +22,8 @@ Overview
 
 AUTO mode navigates between each waypoint in a straight line and as it approaches each waypoint follows a smooth curving path towards the next waypoint (S-Curve). It controls the path as it approaches a waypoint to be within the WPNAV_RADIUS of each waypoint. The speed is lowered below :ref:`WPNAV_SPEED<WPNAV_SPEED>`, as necessary, to keep the path within the set corner lateral acceleration limit parameter (:ref:`WPNAV_ACCEL_C<WPNAV_ACCEL_C>`).
 
+.. note:: this smooth curving path as it changes direction to the next waypoint will not occur if the next mission item after the waypoint is anything other than a normal waypoint, including camera commands, speed change commands, etc. In this case the copter will approach the waypoint, stop, turn, and proceed to the next navigation waypoint instead of smoothly curving in front of it and proceeding to that next navigation waypoint. Also, small :ref:`WPNAV_RADIUS<WPNAV_RADIUS>` values will require that the curve be very small, effectively emulating the non-S-Curve behavior explained above.
```
rmackay9, line 25: Thanks for this.  "Do" and "Conditional" commands will not cause the vehicle to stop at the waypoints though.  How about this?

>The vehicle will stop at the waypoint if a delay has been specified or the next navigation command is not a waypoint or spline command.

I suspect that we may instead need a small section on how to increase the corner cutting.

Hwurzburg: In my mission do change speed after a waypoint killed the scurve on that waypt as I recall...will confirm in Sitl when I can

rmackay9: I just tested do-change-speed and it seems OK.  Here's the mission
image

.. and here's the resulting path
image

Hwurzburg: fixed

PR #5912, opened 2024-04-29
Update tecs-total-energy-control-system-for-speed-height-tuning-guide…

timtuxworth:
A couple of "centi-" notes are no longer correct with the new _DEG parameters


Hwurzburg:
nice thanks

PR #5936, opened 2024-05-07
Small fixes in typography and expression

Georacer:


common/source/docs/common-imu-notch-filtering.rst
```diff
@@ -20,7 +20,7 @@ ArduPilot has support for two notch filters whose filter frequency can be linked
...
 
-For the gyro based rate controllers, this reduces their ability to respond to fast disturbances. If the gyro lowpass filter can be set higher, the phase lag induced is lower and the tune can be more aggressive. But this allows more noise and vibration, effectively canceling that gain out. Enabling either one or both of the Harmonic Notch filters allow targeting the noise generated by the motors, allowing a higher frequency of the following low pass to be set and therefore allowing tighter tune.
+For the gyro based rate controllers, this reduces their ability to respond to fast disturbances. If the gyro lowpass filter is set higher, the phase lag induced is lower and the tune can be more aggressive. But this allows more noise and vibration, effectively canceling that gain out. Enabling either one or both of the Harmonic Notch filters gives the ability to target the noise generated by the motors, allowing a higher frequency of the following low pass to be set and therefore allowing tighter tune.
```
Hwurzburg, line 23: I disagree with the first change.....raising the lowpass filter may NOT be possible if the noise cannot be effectively filtered....second change is okay

copter/source/docs/thrust_loss_yaw_imbalance.rst
```diff
@@ -6,13 +6,13 @@ Thrust Loss and Yaw Imbalance Warnings
...
-These warnings are designed to detect hardware failures in the propulsion system, if they begin to occur on a vehicle that has been flying without warnings the propulsion system should be checked.
-Warning are more likely to happen with larger payloads and in strong/gusty winds.
+These warnings are designed to detect hardware failures in the propulsion system. If they begin to occur on a vehicle that has been flying without warnings the propulsion system should be checked.
+Warnings are more likely to occur with larger payloads and in strong/gusty winds.
 
 Potential Thrust Loss
 =====================
 
-If potential thrust loss messages are seen on the GCS or in the data-flash logs investigation should be done to find the cause and remedy. The warning will give a motor number, for example:
+If potential thrust loss messages are seen either the GCS or in the data-flash logs, then an investigation should be done to find the cause and remedy. The warning will give a motor number, for example:
```
Hwurzburg, line 15:
```suggestion
If potential thrust loss messages are seen either on GCS or in the data-flash logs, then an investigation should be done to find the cause and remedy. The warning will give a motor number, for example:
```

Hwurzburg (changes requested):

common/source/docs/common-imu-notch-filtering.rst
Georacer: This phrase is a bit long; a bit hard to wrap my head around it.

With the original wording it reads to me:
1. If the knee frequency can be set higher (depends on what?)
2. then the filter phase lag is lower (agreed)
3. then the tune has more bandwidth to work with (agreed) and can be tuned more aggressively.
4. But then the more aggressive control input will bring on more noise and vibration, cancelling the merits (agreed).

What I tried to do was do away with the unmentioned dependency of #1 to simplify the phrase.

But no worries, I'll adapt as you ask, it's no biggie.

Georacer:
@Hwurzburg I believe I addressed your requests.

PR #5938, opened 2024-05-07
4.6:add NanoRadar MR72

Hwurzburg:


rishabsingh3003:
@Hwurzburg thanks so much for this! Here is a documentation from NanoRadar that has a lot more details incase we want to put some more info. 

[MR72_Open_Source_CAN_Protocol_-_User_manual.pdf](https://github.com/ArduPilot/ardupilot_wiki/files/15286338/MR72_Open_Source_CAN_Protocol_-_User_manual.pdf)

By the way, we backported this driver so it should be included for 4.5 and not a "future" driver.

Hwurzburg:
do you have a link to the English versions on the web of the MR72 (and NRA24) user manuals....their links return gibberish

PR #5959, opened 2024-05-09
Copter: wind estimate video update

rmackay9:
This moves a more easily understood video for users to the top of the page and moves the developer specific video to the bottom.


Hwurzburg:
thanks

PR #5970, opened 2024-05-11

stevenjowens:
Clarified -w option.

Removed --console and --map from sim_vehicle.py example because GUI dependencies are not included in the SITL docker instance.

Added note about using  --no-mavproxy.

### DIFF
--- a/dev/source/docs/setting-up-sitl-on-linux.rst
+++ b/dev/source/docs/setting-up-sitl-on-linux.rst
@@ -37,19 +37,34 @@ For example, for the multicopter code change to **ardupilot/ArduCopter**:
    cd ardupilot/ArduCopter
 
 Then start the simulator using **sim_vehicle.py**. The first time you
-run it you should use the -w option to wipe the virtual EEPROM and load
-the right default parameters for your vehicle.
+run it you should use the ``-w`` option.  This will both wipe the virtual 
+EEPROM, removing any user-changed defaults, and also load the right 
+default parameters for your vehicle.
 
 ::
 
     sim_vehicle.py -w
 
 After the default parameters are loaded you can start the simulator
-normally.  First kill the sim_vehicle.py you are running using Ctrl-C.  Then:
+normally.  First kill the sim_vehicle.py you are running using Ctrl-C.  
+Then run sim_vehicle.py with no parameters:
 
 ::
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
strongly disagree with the changes about removing the console and map options....this page is a native Linux setup which includes installation of ALL necessary components for the GUI elements...not a DOCKER based page...the minor tweaking of the -w wording is fine but really the first run should be : sim_vehicle.py  --map --console -w to avoid unnecessary restarts....feel free to change to this....

stevenjowens:
> this page is a native Linux setup 

That's an excellent point.  I started here:
https://ardupilot.org/dev/docs/sitl-simulator-software-in-the-loop.html
which links to: 
https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html
which says "Please follow the instructions in Setting up the Build Environment (Linux/Ubuntu) to set up the full environment, including SITL." and links to: 
https://ardupilot.org/dev/docs/building-setup-linux.html#building-setup-linux 

I followed the section titled "Setup using Docker", then went back to:
https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html
Which then walks you through starting sim_vehicle.py, and tells you to use the --console and --map parameters, which of course doesn't work in the docker SITL.

So tell me where the information should be and I'll make that edit.  Either setting-up-sitl-on-linux needs two two subsections under "Starting SITL simulator", one for native linux and one for docker, or there needs to be a separate page somewhere for starting SITL under docker.

Ryanf55:
I would add info in the "Setup using Docker" section that docker is not recommended unless you want isolated environments or your host environment is not what you want to develop in. 

Given the wiki hardly mentions docker, and has tons of other things that need exposed to work for a full IDE, and we don't have a VSCode devcontainer, I honestly wouldn't recommend anyone use docker except for CI.

Hwurzburg:
@Ryanf55 I agree....I think we should have:
1. A note on the Docker section saying that it is not recommended unless an isolated environment is needed
2. the startup instructions compressed to a single command as I suggest
3. a warning that the map and console options wont work in a Docker situation
see #5972 and we can close this.... @stevenjowens thanks for bringing the Docker issue up

Hwurzburg:
closing since #5972 now merged

PR #5972, opened 2024-05-12
add warning notes on Docker setups

Hwurzburg:


dev/source/docs/building-setup-linux.rst
```diff
@@ -170,6 +170,8 @@ Then add /usr/lib/ccache to the front of your $PATH
 Setup using Docker
 ==================
 
+.. note:: this is not recommended except for specail cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL.
```
Ryanf55, line 173:
```suggestion
.. note:: this is not recommended except for special cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL unless custom options are added.
```

dev/source/docs/setting-up-sitl-on-linux.rst
```diff
@@ -30,26 +30,27 @@ Start SITL simulator
...
 
 To start the simulator first change directory to the vehicle directory.
-For example, for the multicopter code change to **ardupilot/ArduCopter**:
+For examples below, for the Multicopter code change to **ardupilot/ArduCopter**:
 
 ::
 
    cd ardupilot/ArduCopter
 
+.. note: see option in startup command below to start SITL in the base **ardupilot** directoty if desired.
```
Ryanf55, line 39:
```suggestion
.. note: see the option in the startup command below to start SITL in the base **ardupilot** directoty if desired.
```

Ryanf55 (approved):
Nice additions, just some minor changes to make complete sentences and some little formatting.

dev/source/docs/setting-up-sitl-on-linux.rst
```diff
@@ -30,26 +30,27 @@ Start SITL simulator
...
 
 ::
 
-    sim_vehicle.py -w
-
-After the default parameters are loaded you can start the simulator
-normally.  First kill the sim_vehicle.py you are running using Ctrl-C.  Then:
+   sim_vehicle.py --console --map -w
 
+alternatively if not starting in the ArduCopter directory but at the base **ardupilot** directoty:
```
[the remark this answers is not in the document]
stevenjowens: s/directoty/directory/

dev/source/docs/building-setup-linux.rst
```diff
@@ -170,6 +170,8 @@ Then add /usr/lib/ccache to the front of your $PATH
 Setup using Docker
 ==================
 
+.. note:: this is not recommended except for special cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL unless custom options are added.
```
Ryanf55, line 173:
```suggestion
.. note:: This is not recommended except for special cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL unless custom options are added.
```

Ryanf55 (approved):
Just one more patch

dev/source/docs/building-setup-linux.rst
Hwurzburg: easier to patch after merge

stevenjowens:
Besides CI/CD use of the Docker SITL image, there's also the obvious case of Mac users, who last I looked couldn't use SITL natively.

The change you've suggested still doesn't address the root cause of the confusion.  It does not:

a) make it explicit that sim_vehicle.py runs mavproxy at the end.

b) mention the --console and --map parameters.

c) mention the --no-mavproxy parameter.

I suggest,  in "Start SITL simulator":

https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#start-sitl-simulator

At the end, but before the "Tips", add something like:

> Note that sim_vehicle.py will also run MAVProxy, resulting in a MAVProxy prompt "MAV>". See the section below titled "Learn MAVProxy".

And then in "Learn MAVProxy", add something like:

> Note that if you're using the SITL docker image, the MAVproxy --map and --console features in the example above won't work.  Refer to https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker for details.

And then in "Setup using Docker":

https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker

> .. note:: this is not recommended except for special cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL unless custom options are added.
>
> Specifically, the --console and --map features require GUI support, which is omitted from the docker SITL build.  You may find it useful to run sim_vehicle.py with the --no-mavproxy parameter, and run mavproxy externally from the SITL.  Doing so will also require you to add "-p 5760:5760" to your "docker run" command, to export SITL's mavlink port to your host machine.

Unrelated to the above, the following is a side issue that I only brought up because I was suggesting an edit to the same page.  There's still ambiguity in phrasing of this section:

> you should use the ``-w`` option to wipe the virtual EEPROM of user changed parameters and load the default parameters for your vehicle.

This does not make it clear whether "and load the default parameters" is:

a) an effect of using the -w option (the correct answer) or

b) an instruction to the user (which was how I read it when I first encountered it).

This is why I suggest phrasing it as:

> This will both wipe the virtual EEPROM, removing any user-changed defaults, and also load the right default parameters for your vehicle.

An even more minor issue is that if you follow the instructions as written, and you're using docker, it results in this sequence:

https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#install-steps
https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker
https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#start-sitl-simulator

...it says to cd into ardupilot/ArduCopter, but the instructions at the end of Setup using Docker result in you already being in the ardupilot directory.  This is, of course, not nearly as confusing as the omitted mavproxy information.

stevenjowens:
This still doesn't address the root cause of the confusion.  It does not make it explicit that sim_vehicle.py runs mavproxy at the end, the warning about "will prevent the use of graphical elements in SITL unless custom options are added" is vague, and it still doesn't mention the --no-mavproxy parameter.

Also, while the following is a side issue that I only brought up because I was suggesting an edit to the same page, there's still ambiguity in phrasing of this section:

> you should use the ``-w`` option to wipe the virtual EEPROM of user changed parameters and load the default parameters for your vehicle.

This does not make it clear whether "and load the default parameters" is a) an effect of using the -w option (the correct answer) or b) an instruction to the user (which was how I read it when I first encountered it).  This is why I suggest phrasing it:

> This will both wipe the virtual EEPROM, removing any user-changed defaults, and also load the right default parameters for your vehicle.

Ryanf55:
> Besides CI/CD use of the Docker SITL image, there's also the obvious case of Mac users, who last I looked couldn't use SITL natively.
> 
> The change you've suggested still doesn't address the root cause of the confusion. It does not:
> 
> a) make it explicit that sim_vehicle.py runs mavproxy at the end.
> 
> b) mention the --console and --map parameters.
> 
> c) mention the --no-mavproxy parameter.
> 
> I suggest, in "Start SITL simulator":
> 
> https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#start-sitl-simulator
> 
> At the end, but before the "Tips", add something like:
> 
> > Note that sim_vehicle.py will also run MAVProxy, resulting in a MAVProxy prompt "MAV>". See the section below titled "Learn MAVProxy".
> 
> And then in "Learn MAVProxy", add something like:
> 
> > Note that if you're using the SITL docker image, the MAVproxy --map and --console features in the example above won't work.  Refer to https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker for details.
> 
> And then in "Setup using Docker":
> 
> https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker
> 
> > .. note:: this is not recommended except for special cases where an isolated build environment is required. Using this setup below will prevent the use of graphical elements in SITL unless custom options are added.
> > Specifically, the --console and --map features require GUI support, which is omitted from the docker SITL build.  You may find it useful to run sim_vehicle.py with the --no-mavproxy parameter, and run mavproxy externally from the SITL.  Doing so will also require you to add "-p 5760:5760" to your "docker run" command, to export SITL's mavlink port to your host machine.
> 
> Unrelated to the above, the following is a side issue that I only brought up because I was suggesting an edit to the same page. There's still ambiguity in phrasing of this section:
> 
> > you should use the `-w` option to wipe the virtual EEPROM of user changed parameters and load the default parameters for your vehicle.
> 
> This does not make it clear whether "and load the default parameters" is:
> 
> a) an effect of using the -w option (the correct answer) or
> 
> b) an instruction to the user (which was how I read it when I first encountered it).
> 
> This is why I suggest phrasing it as:
> 
> > This will both wipe the virtual EEPROM, removing any user-changed defaults, and also load the right default parameters for your vehicle.
> 
> An even more minor issue is that if you follow the instructions as written, and you're using docker, it results in this sequence:
> 
> https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#install-steps https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker https://ardupilot.org/dev/docs/setting-up-sitl-on-linux.html#start-sitl-simulator
> 
> ...it says to cd into ardupilot/ArduCopter, but the instructions at the end of Setup using Docker result in you already being in the ardupilot directory. This is, of course, not nearly as confusing as the omitted mavproxy information.

While you bring up some good points, they are all mixed together and separate concerns. This PR was intended to clarify one thing - docker. Let's address the others in follow ups?

PR #5975, opened 2024-05-12
Dev: ROS: Split into ROS 1 and ROS 2 docs

Ryanf55:
Before and After:

image

I want to also recommend using ROS 2. Note the ethernet changes were because it wasn't representing in the TOC correctly. Incorrect headings were used.

### DIFF
--- a/dev/source/docs/ros.rst
+++ b/dev/source/docs/ros.rst
@@ -1,29 +1,46 @@
 .. _ros:
 
-========
-ROS/ROS2
-========
+=============
+ROS 1 / ROS 2
+=============
 
 .. image:: ../images/logos/rosorg_logo.png
     :target: ../_images/logos/rosorg_logo.png
 
 ArduPilot capabilities can be extended with `ROS <http://www.ros.org/>`__ (aka Robot Operating System).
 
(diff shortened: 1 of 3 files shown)


dev/source/docs/ros.rst
```diff
@@ -1,29 +1,35 @@
 .. _ros:
 
-========
-ROS/ROS2
-========
+=========
+ROS 1 / ROS 2
+=========
```
Hwurzburg, line 5: underline/overline error

Hwurzburg (changes requested):

Ryanf55:
What do you think about moving up the `Prerequisites` section? We get a fair number of people trying to build ardupilot in a ros2 workspace before they even know what ArduPilot is or how to use ROS 2.

Hwurzburg:
I like it....first thing before ROS1/2 is even mentioned....

Ryanf55:
> I like it....first thing before ROS1/2 is even mentioned....

Done, it's nearly at the top. I think it's work introducing ROS 1 and ROS 2, otherwise the links in the pre-reqs don't make sense.

PR #5978, opened 2024-05-13
Delete duplicate images/supporters/supporters_liaz.PNG

valeriyvan:
There's same image images/supporters/supporters_liaz.png. On case insensitive file systems, like macOS with default settings, these two files .PNG and .png clash.


Hwurzburg:
wrong version to delete....hence the build errors.....wiki uses the .PNG file

valeriyvan:
> wrong version to delete....hence the build errors.....wiki uses the .PNG file

Fixed reference instead because all other extensions are lowercased.

Hwurzburg:
lgtm..thanks

PR #5984, opened 2024-05-15
Dev: update VSCode launch.json gdb args

robertlong13:
This automatically adds a breakpoint at AP_HAL::panic() when debugging

I normally debug ArduPilot through vscode like this without the extra step of using `sim_vehicle.py`. Recently, I had a panic in SITL that I can't reproduce and couldn't debug at the time because I hadn't enabled this breakpoint in my setup. I came up with this addition for my personal `launch.json` and figured it would be useful for others.

I tested by triggering a panic with the appropriate mav command, and this does indeed cause gdb to break there now.


andyp1per (approved):

IamPete1 (approved):

Hwurzburg:
tnx

PR #5994, opened 2024-05-16
4.6: Add notes on Extended DShot telemetry

mbuzdalov:
This includes:
- mentioning extended DShot telemetry and that it can send voltage, current etc. via bi-directional DShot;
- extended DShot telemery v2 explanation;
- updating the link to the current Bluejay development repo;
- fixing rpm => RPM where appropriate.


Hwurzburg:
thanks....most do not write up something for their code and I have to do it....will be merged into wiki when first 4.6 beta is released...I squashed the commits for you

PR #5995, opened 2024-05-17
Update common-mount-targeting.rst to describe mapping of mounts

timtuxworth:
Mapping of MNT mounts to specific physical devices is a bit tricky. This tries to explain it.

### DIFF
--- a/common/source/docs/common-mount-targeting.rst
+++ b/common/source/docs/common-mount-targeting.rst
@@ -91,3 +91,21 @@ Control during Auto mode missions
 ---------------------------------
 
 Commands to control the gimbal are listed on the  :ref:`Camera Control in Auto Missions <common-camera-control-and-auto-missions-in-mission-planner>`, :ref:`Copter Mission Command List <mission-command-list>` and :ref:`Mission Commands <common-mavlink-mission-command-messages-mav_cmd>` pages
+
+Multiple Gimbal/mounts on a vehicle
+-----------------------------------
+ArduPilot supports one or two gimbals on a vehicle which are addressed as MNT1 and MNT2. So all of the above 
+parameters refering to MNT1 would work for the second gimbal by replacing the 1 with 2. For example Retract Mode for the 2nd gimbal mount would be specified as 
+Retract Mode: the gimbal will move to the roll, pitch and yaw angles held in the :ref:`MNT2_RETRACT_X<MNT2_RETRACT_X>`, :ref:`MNT2_RETRACT_Y<MNT1_RETRACT_Y>`, :ref:`MNT2_RETRACT_Z<MNT2_RETRACT_Z>` parameters respectively
+
+Each gimbal can be connected as either a servo gimbal, controlled by SERVO functions, or to either a UART or a network port and is configured by specifying the relevant 
+protocol as 8: Gimbal. eg. for a gimbal attached to the UART which is SERIAL2 then SERIAL2_PROTOCOL<SERIAL2_PROTOCOL> = 8.
+ArduPilot maps the MNTn gimbals to the configured physical mount points as follows:
+
+1. If there are servo functions specifying "Mount 1" or "Mount 2", then the relevant Mount (MNTn) will be mapped to that specifi servo gimbal.
+2. If there are SERIALn_PROTOCOL values set to "8: Gimbal" then the remaining mounts will be mapped to the relevant SERIAL ports (UARTS) starting with the lowest.
+3. If there are NETn_PROTOCOL values set to "8: Gimbal" then the remaining mounts will be mapped to the relevant NET ports starting with the lowest first.
+
+
+
+


common/source/docs/common-mount-targeting.rst
```diff
@@ -91,3 +91,21 @@ Control during Auto mode missions
...
+protocol as 8: Gimbal. eg. for a gimbal attached to the UART which is SERIAL2 then SERIAL2_PROTOCOL<SERIAL2_PROTOCOL> = 8.
+ArduPilot maps the MNTn gimbals to the configured physical mount points as follows:
+
+1. If there are servo functions specifying "Mount 1" or "Mount 2", then the relevant Mount (MNTn) will be mapped to that specifi servo gimbal.
+2. If there are SERIALn_PROTOCOL values set to "8: Gimbal" then the remaining mounts will be mapped to the relevant SERIAL ports (UARTS) starting with the lowest.
+3. If there are NETn_PROTOCOL values set to "8: Gimbal" then the remaining mounts will be mapped to the relevant NET ports starting with the lowest first.
+
+If only one of the gimbals uses Serial or Ethernet then I’d recommend setting up the serial/ethernet mount as Mount1 and the servo gimbal as Mount2. The reason is that the 2nd serial/ethernet mount will use the 2nd configured serial/ethernet port regardless of what protocol Mount1 uses. So if SERIAL2_PROTOCOL<SERIAL2_PROTOCOL> = 8 and SERIAL4_PROTOCOL<SERIAL4_PROTOCOL> = 8 then Mount2 would use Serial4 even if Mount1 was defined as a servo gimbal which doesn’t use serial.
```
Hwurzburg, line 109: third person, not first is used in the wiki...then it is recommended to set up....

common/source/docs/common-mount-targeting.rst
```diff
@@ -91,3 +91,21 @@ Control during Auto mode missions
...
+-----------------------------------
+ArduPilot supports one or two gimbals on a vehicle which are addressed as MNT1 and MNT2. So all of the above 
+parameters refering to MNT1 would work for the second gimbal by replacing the 1 with 2. For example Retract Mode for the 2nd gimbal mount would be specified as 
+Retract Mode: the gimbal will move to the roll, pitch and yaw angles held in the :ref:`MNT2_RETRACT_X<MNT2_RETRACT_X>`, :ref:`MNT2_RETRACT_Y<MNT1_RETRACT_Y>`, :ref:`MNT2_RETRACT_Z<MNT2_RETRACT_Z>` parameters respectively
+
+Each gimbal can be connected as either a servo gimbal, controlled by SERVO functions, or to either a UART or a network port and is configured by specifying the relevant 
+protocol as 8: Gimbal. eg. for a gimbal attached to the UART which is SERIAL2 then SERIAL2_PROTOCOL<SERIAL2_PROTOCOL> = 8.
+ArduPilot maps the MNTn gimbals to the configured physical mount points as follows:
+
+1. If there are servo functions specifying "Mount 1" or "Mount 2", then the relevant Mount (MNTn) will be mapped to that specifi servo gimbal.
```
Hwurzburg, line 105: sp : specific

Hwurzburg (changes requested):
Overall this is over complicated and incorrect as far as I understand it....the issue arises only with  mounts that are serial based, either with UART or virtual Ethernet....then each mount in turn looks for the first serial port (uarts first then virtual Ethernet) with Gimbal protocol, and snags it...then MNT2 gets the next serial port with protocol 8.....servo mounts really don't play into this....

common/source/docs/common-mount-targeting.rst
Hwurzburg: also, if the mount type is NOT servo, then those outputs are non-functional....the type param sets the mount type, if its servo then the outputs are used and active

rmackay9:
.. maybe we should just fix the code so that it works like rangefinders..

timtuxworth:
> .. maybe we should just fix the code so that it works like rangefinders..

I'm guessing that's a bigger deal, but also a good idea. Assuming that might take a while I'm thinking it probably makes sense to update the wiki in the meantime.

common/source/docs/common-mount-targeting.rst
timtuxworth: I read the code and tried to write more clearly (and correctly) what I think is happening.

common/source/docs/common-mount-targeting.rst
timtuxworth: fixed

common/source/docs/common-mount-targeting.rst
timtuxworth: specific is fixed

common/source/docs/common-mount-targeting.rst
```diff
@@ -98,8 +98,8 @@ ArduPilot supports one or two gimbals on a vehicle which are addressed as MNT1 a
 parameters refering to MNT1 would work for the second gimbal by replacing the 1 with 2. For example Retract Mode for the 2nd gimbal mount would be specified as 
 Retract Mode: the gimbal will move to the roll, pitch and yaw angles held in the :ref:`MNT2_RETRACT_X<MNT2_RETRACT_X>`, :ref:`MNT2_RETRACT_Y<MNT1_RETRACT_Y>`, :ref:`MNT2_RETRACT_Z<MNT2_RETRACT_Z>` parameters respectively
 
-Each gimbal can be connected as either a servo gimbal, controlled by SERVO functions, or to either a UART or a network port. The port is configured by specifying the relevant 
-protocol as 8: Gimbal. 
+Each gimbal can be connected as either a servo gimbal, controlled by SERVO functions, or to either a UART or a network port. Theese ports are configured by specifying the relevant
```
Hwurzburg, line 101: spelling

Hwurzburg (changes requested):

common/source/docs/common-mount-targeting.rst
timtuxworth: fixed

timtuxworth:
Are you sure that multiple serial/net gimbals can be used @rmackay9 - looking at the code in find_serial() -> find_protocol_instance() it seems to me that it might not work, i.e.

You might only be able to have 2 gimbals if they are different types - i.e. a serial gimbal + a servo/gremsy/solo... but not two serial type gimbals like the siyi, alexmos, viewpro ...

Do I have it wrong?

rmackay9:
@timtuxworth,

Ah,, darn.  You are right.  The [line here](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount_Siyi.cpp#L40) should be using the instance number instead of "0"

Thank you very much for finding this.  We will fix it!  Or you could probably fix if you want.  No pressure.

rmackay9:
Hi @timtuxworth,

I've created a flight code PR https://github.com/ArduPilot/ardupilot/pull/27129 that fixes the underlying issue (at least in my light testing) it does.  Hopefully we will be able to get this merged and backported to a beta in the near-ish future.

Hwurzburg:
@timtuxworth can we closed this now?

timtuxworth:
> @timtuxworth can we closed this now?

Yes - thanks @Hwurzburg !

PR #5997, opened 2024-05-17
add PodmanH7 FC board wiki

burgeruser:
Added board wiki files and pictures.


Hwurzburg:
@burgeruser can we get a wiring diagram with English labels instead of Chinese?

PR #5999, opened 2024-05-17
Add ethernet vehicle example

Hwurzburg:
decided to add it into the wiki after all link it several places


rmackay9 (approved):
Looks pretty good, certainly it's important information we need to get out there.

We probably don't need  "ArduPilot" in the title of the page 'cuz everything's ArduPilot really.

It could also be linked from the Copter reference frames page.

I think we should aim to get most of the information that is on this page into other areas of the wiki.  For example, the Siyi A8 networking info should probably be on the Siyi setup page.

Anyway, I think it's good to get this into the wiki as-is for now at least.

PR #6004, opened 2024-05-20
Changes for RPLidar A2M12

metaphaz:
I was trying to set RPLidar A2M12 to Copter for obstacle avoidance but it was never working and shoving "PRX1: No Data" on HUD on Mission Planner. After reading datasheet only A2M12 model got different baud rate from rest of A2 Lidars. 
Screenshot from 2024-05-20 12-00-32

As you can see its working at 256000 bps. after changing SERIAL1_BAUD parameter to "256" everything works fine.


rmackay9:
Hi, thanks for this.  The page has an existing line that tries to clarify the baud rate per model.

- SERIAL1_BAUD = “115” for C1/A2, “256” for S1, if using Serial1

So perhaps we should change this line instead to be:

- SERIAL1_BAUD = “115” for A2 and C1, “256” for A2M12 and S1

While we're at it we might want to change the line slightly above to be, "Example setup below shown for first proximity sensor **assuming it is connected to Serial1**:"

Hwurzburg:
closing...thia content already added i n interim

Issue #5759, opened 2024-02-02
4.6 Wiki (should be in Upcoming features)


Hwurzburg:
commits after 1/28 that need wiki:
- [x] 22513 Param controller applet (@Hwurzburg )
- [ ] 23056 RDSS02UF Rangefinder (@zebulon86) need to verify this is obtainable by users and how
- [x] 24645 Allow more rpm instanes(@magicrub )
- [ ] 24801System ID-Pos Cntrlr(@bnsgeyer )
- [x] 24899 GPS Jamming (@tridge )
- [x] 25273 MatekF765-SE bdshot (@FOGYamato)
- [x] 25442 harmonic notch rework (@tridge )
- [x] 25844 CRSF Extended RF reports (@rmaia3d )*
- [x] 25940 LR-D1rangefinder (@jkronk)
- [x] 25993 Temp sense from AP_Periph (@IamPete1 )
- [x] 26021 AnyleafH7 board (@David-OConnor )*
- [x] 25621 EKF filtering (@andyp1per )
- [x] 25714 PixFlamingoF767 (@subashchandar)*
- [x] 25801 MR72CAN(@rishabsingh3003 )*
- [x] 26030 YJUAVA6Ultra(yunjiuav) *
- [x] 26057 default param extract utility(@amilcarlucas )
- [x] 26129 Chute option(python36)
- [x] 26131 MicroStrain7 orientation (@Ryanf55 )
- [x] 26139 RadiolinkPIX6*
- [x] 26172 Mavlink camera source (@rmackay9 )
- [x] 26197 STAMPH743(@andyp1per )
- [x] 26212 Hi-speed VTOL fixes (@priseborough )
- [x] 26245 Improve AUTOTUNE (@lthall )
- [x] 26260 VTOL takeoff FS (@CallanDaniel )
- [x] 26272 Fence on SD(@tridge )*
- [x] 26265 low drag TECS option (@priseborough )
- [x] 26321DShot V2 telem (@mbuzdalov)
- [x] 26463 ControlZero H7 OWM RevG (@TOTON95)*
- [x] 26568 Relay invert(@LaneaLucy)
- [x] 26317 GPS_DRV_OPTION for full parsing (@tridge )*
- [x] 26345 Heli inverted flight (@IamPete1 )
- [x] 26356 MAVLink RC (@olliw42 )
- [x] 26368 MicoAir405V2 (@Minderring)*
[...]

PR #6010, opened 2024-05-23
Fix Aocoda-RC H743Dual motor 5-8 mis-labeled issue 26812

lida2003:
[26812 Aocoda-RC H743Dual motor 5-8 mis-labeled issue](https://github.com/ArduPilot/ardupilot_wiki/issues/5759)


Hwurzburg:
no need for this PR...already one pending which is more complete and is on merge delay until 4.6 beta per policy

PR #6020, opened 2024-05-29
Update Marine Use-Cases and Vehicle Types

ES-Alexander:
- Skid-steer boats were missing entirely
- The previous Submersible ROV image was for a really old vehicle type, and wasn't a nice image


Hwurzburg:
tnx

PR #6031, opened 2024-06-05
Common: simplify camera servo and camera relay setup instructions

rmackay9:
This simplifies/clarifies the camera setup when using servo or relays.

I've removed a picture and perhaps that means we could remove the pic from the images directory as well, but sometimes we use images across pages.

I've tested it locally and it looks OK to me.

### DIFF
--- a/common/source/docs/common-3d-mapping.rst
+++ b/common/source/docs/common-3d-mapping.rst
@@ -24,13 +24,7 @@ Equipment you will need
 `Plane <https://ardupilot.org/plane/index.html>`_ or
 :ref:`Multicopter <copter:home>`
 
-A digital still camera:
-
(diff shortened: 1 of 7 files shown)


Hwurzburg:
Going to take till the weekend for me to list all the changes required....lots of interdependencies not covered here that will direct people to ARCHIVED pages and pages archived need to be marked as such...and you are right lots of pages use those figures, perhaps incorrectly now and also need changes??

rmackay9:
I just noticed a small spacing issue so I've re-based and re-pushed.

@Hwurzburg, OK, I'll try and do a search for references to the archived pages.

Hwurzburg:
I have updated sections using CHDK, added archive titles, tweaked spelling in servo doc....figure is used elsewhere so cant delete...
if everything is okay, I will merge this...thanks

rmackay9 (commented):
Looks good to me. Thanks!

PR #6036, opened 2024-06-06
Adjust max and min battery voltage suggestions to take into account different battery chemistries

MichelleRos:
The 4.2v and 3.3v numbers apply to standard LiPos. 
They would not be correct for a LiHV battery or a Liion battery, for example.

Not sure if we want to add the numbers for other chemistries, or just let the user look it up themselves?

### DIFF
--- a/copter/source/docs/setting-up-for-tuning.rst
+++ b/copter/source/docs/setting-up-for-tuning.rst
@@ -25,8 +25,8 @@ There are 3 common causes of a non-linear thrust curve.
 Start with setting up the voltage range to cope with voltage sag.
 Parameters used to linearise your motor thrust curve.
 
-- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>`: 4.2v x No. Cells
-- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>`: 3.3v x No. Cells
+- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>` should be set to the maximum voltage of your batteries, e.g. 4.2v x No. Cells for standard LiPos.
+- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>` should be set to the minumum voltage of your batteries, e.g. 3.3v x No. Cells for standard LiPos.
 - :ref:`MOT_OPTIONS<MOT_OPTIONS>` = 0 (default). The default is to use sag compensated voltage for the above and during tuning computations. It can be set to 1 to use raw voltage instead of sag compensated voltage, which may improve tuning results for light vehicles.
 
 Next setup the thrust expo. If you are setting up a professional


copter/source/docs/setting-up-for-tuning.rst
```diff
@@ -25,8 +25,8 @@ There are 3 common causes of a non-linear thrust curve.
 Start with setting up the voltage range to cope with voltage sag.
 Parameters used to linearise your motor thrust curve.
 
-- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>`: 4.2v x No. Cells
-- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>`: 3.3v x No. Cells
+- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>` should be set to the maximum voltage of your batteries, e.g. 4.2v x No. Cells for standard LiPos.
```
Hwurzburg, line 28:
```suggestion
- :ref:`MOT_BAT_VOLT_MAX <MOT_BAT_VOLT_MAX>` should be set to the maximum voltage of your batteries, e.g. 4.2v x No. Cells for standard LiPos, or as appropriate if using a different battery type .
```

copter/source/docs/setting-up-for-tuning.rst
```diff
@@ -25,8 +25,8 @@ There are 3 common causes of a non-linear thrust curve.
...
+- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>` should be set to the minumum voltage of your batteries, e.g. 3.3v x No. Cells for standard LiPos.
```
Hwurzburg, line 29:
```suggestion
- :ref:`MOT_BAT_VOLT_MIN <MOT_BAT_VOLT_MIN>` should be set to the minumum voltage of your batteries, e.g. 3.3v x No. Cells for standard LiPos, or as appropriate if using a different battery type .
```

Hwurzburg (changes requested):
yeah lets generalize a bit...thanks

copter/source/docs/setting-up-for-tuning.rst
MichelleRos: Well, putting the e.g. was meant to mean the 4.2v is just an example.  I've already got "of your batteries" so I'm not sure we need the extra bit as well?
Or should we change the "e.g." to "i.e.", so it makes at least a bit more sense?

copter/source/docs/setting-up-for-tuning.rst
MichelleRos: See comment on the other one.

copter/source/docs/setting-up-for-tuning.rst
MichelleRos: I *do* know the difference between eg and ie, actually...
In fact, I was thinking *precisely* "that is" would fit better than "for example".

If we're giving just one example - like just mentioning LiPos and nothing else as my original PR had, then "eg" would make sense.
But "4.2v x No. Cells for standard LiPos, or as appropriate if using a different battery type" is effectively all possible options, therefore "for example" doesn't make sense anymore and "that is" makes more sense.

Hwurzburg:
I continue to disagree...examples of a statement use eg..restating the statement in simpler terms uses ie

PR #6039, opened 2024-06-08
common: Change the time range specification

muramura:
The range of hours is from 0 to 23.


Hwurzburg:
@rmackay9 review pls

rmackay9 (approved):
sorry for the delay, thanks!

PR #6040, opened 2024-06-09
Common: Update baud rate for M12 variant of RPLiDAR A2

snktshrma:
Other variants of RPLiDAR A2 like M7 or M8 supports 115200 baud but M12 variant supports 256000 baud rate.


Hwurzburg:
LGTM, thanks

PR #6046, opened 2024-06-13
Common: Mount controls page gets QGC section

rmackay9:
Now that QGC-4.4 is released we should explain to users how it can be used to control the gimbal.  Previously we only had Mission Planner instructions.

I've also made a drive-by fix to a broken link and also slightly modified some wording.


common/source/docs/common-mount-targeting.rst
```diff
@@ -81,12 +79,48 @@ To point the gimbal at a particular location (e.g. lat, lon, alt), on the Data s
...
+    :target: ../_images/mount-targeting-qgc-clicktopoint.png
+
+To point the gimbal at a Location
+
+- While the vehicle is flying, maximise the map
+- Click anywhere on the map and select "ROI at location"
+
+.. image:: ../../../images/mount-targeting-qgc-roi.png
+    :target: ../_images/mount-targeting-qgc-roi.png
+
```
Hwurzburg, line 115: should add how to release ROI lock, point home, and click to point actions....I have not tried but click to point does not track like home and ROI, right? probably should mention that also

Hwurzburg (changes requested):

common/source/docs/common-mount-targeting.rst
rmackay9: Hi @Hwurzburg, OK, I've added a line mentioning how to cancel the ROI.  .. also rebased on master

PR #6051, opened 2024-06-15
Common: xacti gets remote download instructions

rmackay9:
This adds instructions for how a companion computer can download images and videos from a Xacti camera gimbal

I've tested this locally and it seems OK to me


Hwurzburg:
thanks

PR #6052, opened 2024-06-15
Update common-underwater-sonars-landingpage.rst

AlksSAV:


AlksSAV:
The list of sensors has been updated
https://discuss.ardupilot.org/t/new-type-sonar-with-nmea-out/119111

PR #6057, opened 2024-06-17
Common: add RToS SparkKit to RTF rovers list

rmackay9:
This adds the RToS SPARK Kit to our list of RTF vehicles.  RToS is one of our newer partners

This has been tested locally and looks OK to me


Hwurzburg:
thanks

PR #6059, opened 2024-06-17
Update loiter-mode.rst

Allister779:
Updated required HDOP to 1.4 from 2.0 per default GPS_HDOP_GOOD value.


rmackay9 (approved):
Thanks for this.  Indeed it appears that @andyp1per changed this way back in 2015
https://github.com/ArduPilot/ardupilot/commit/425e26c258c9af7c4f1ad77c74110da6527bf192

PR #6069, opened 2024-06-25
4.6:add GPS jamming EKF3 option

Hwurzburg:


rmackay9 (approved):
LGTM.  I'm not the expert on the feature but it matches my understanding

PR #6073, opened 2024-06-27
Dev: gazebo installation update

rmackay9:
This includes a few changes that I've separated into different commits so that we can easily drop those that aren't acceptable to @khancyr and @srmainwaring:

1. most importantly the dependencies are updated
2. added a reference to the ardupilot_gazebo plugin's README which had the correct instructions

The next two i feel less strongly about but I thought they were improvements

3. changed the tip at the top to regular text
4. changed the recommended ardupilot_gazebo path be in the home directory which matches the [README](https://github.com/ArduPilot/ardupilot_gazebo/blob/main/README.md)

### DIFF
--- a/dev/source/docs/sitl-with-gazebo.rst
+++ b/dev/source/docs/sitl-with-gazebo.rst
@@ -89,6 +89,7 @@ is hosted on GitHub at: `https://github.com/ArduPilot/ardupilot_gazebo <https://
 
         sudo apt update
         sudo apt install libgz-sim7-dev rapidjson-dev
+        sudo apt install libopencv-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev gstreamer1.0-plugins-bad gstreamer1.0-libav gstreamer1.0-gl
 
       For Gazebo Harmonic
 
@@ -96,6 +97,7 @@ is hosted on GitHub at: `https://github.com/ArduPilot/ardupilot_gazebo <https://
 
         sudo apt update
         sudo apt install libgz-sim8-dev rapidjson-dev
+        sudo apt install libopencv-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev gstreamer1.0-plugins-bad gstreamer1.0-libav gstreamer1.0-gl
 
     macOS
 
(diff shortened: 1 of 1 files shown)


khancyr (approved):
LGTM

Tested and working on wsl2

srmainwaring:
Thanks @rmackay9. The only change that needs more discussion is the last one altering the recommended path for the plugin repo. Here's the reasoning behind it.

- The `ardupilot_gazebo` plugin can be built as documented in the README, but can also be built using the `colcon` build tool as a ROS 2 ament package. This support was added in https://github.com/ArduPilot/ardupilot_gazebo/pull/53.
- This second approach has the advantage of being consistent with the ROS 2 documentation, and allows the plugin to be integrated in ROS 2 / Gazebo / ArduPilot projects. It also automatically configures the `GZ_*` environment variables.
- The downside is that it requires a ROS 2 (base) install.
- For `colcon` builds the standard practice is to create a workspace folder with structure

```bash
..
└── gz_ws
    ├── build
    ├── install
    ├── log
    └── src
```

and the environment is sourced with

```bash
. ~/gz_ws/install/setup.bash
```

The reason for the documented directory structure is to provide some level of consistency between the two approaches, and for users with a ROS 2 install, the plugin can be built and configured using

```bash
cd ~/gz_ws/install/setup.bash
colcon build
. ~/gz_ws/install/setup.bash
```

rmackay9:
Thanks @khancyr and @srmainwaring for the review.

I've removed the last commit which moved the plugin location.

khancyr:
@srmainwaring it would need to amend the repo readme to match the ros2 structure then. That way wiki and readme are matching

rmackay9:
.. I still have the other commit saved so whatever you guys are happy with ..

srmainwaring:
> @srmainwaring it would need to amend the repo readme to match the ros2 structure then. That way wiki and readme are matching

Maybe we run with @rmackay9's proposal for the repo and basic install instructions, and leave the ROS 2 workspace version for the AP / ROS 2 wiki entry. Otherwise the risk is confusing people who just want the standalone version.

srmainwaring:
> .. I still have the other commit saved so whatever you guys are happy with ..

Let's go with this - I think it simplifies the standalone install instructions.

rmackay9:
@srmainwaring, sorry to be pendantic but I think you're saying you're happy with this PR as it is now right?  You don't want me to add back in change to the ardupilot_gazebo install location right?

srmainwaring (approved):

rmackay9:
Thanks!

PR #6077, opened 2024-06-28
common: add ArduX to commercial support page

rmackay9:
This adds ArduXJapan to the commercial support page.  ArduXJapan is a joint venture between my company (Japan Drones) and the tragically similarly named (Drone Japan) which is an AP partner.


Hwurzburg:
tnx

PR #6086, opened 2024-07-01
Added Controlled Ascent consulting

Georacer:
I've added my consulting business to the list.

One thing I'm not sure I did correctly was to upload and reference the logo. It's not ending up in the `_images/...` folder when I build locally.


Hwurzburg:
try putting the image in the base /images directory

Georacer:
I fixed the image. It needed to be added in the references.

PR #6092, opened 2024-07-04
Common: add topotek gimbal

rmackay9:
This adds a setup page for the Topotek gimbals which should be supported in master shortly.  See PR https://github.com/ArduPilot/ardupilot/pull/27053

I have the hardware and have verified the connection and I've also tested this locally and it looks OK to me.


Hwurzburg (approved):

rmackay9:
The flight code PR has gone in now so I think we could merge this

rmackay9:
Thanks!

PR #6093, opened 2024-07-04
Common: viewpro gets stream to QGC explanation

rmackay9:
This adds details on how a ViewPro camera video can be streamed to QGC.  This is in response to a support call I received.


Hwurzburg:
thanks

rmackay9:
Thank you!

PR #6099, opened 2024-07-07
common-lua-scripts: update serial binding docs

tpwrules:
Includes new `readstring` and `writestring` methods, as well as new serial device simulation support.

Completes https://github.com/ArduPilot/ardupilot/pull/27219 .


Hwurzburg:
thanks for documenting your code, most dont

PR #6101, opened 2024-07-08
Updated the instructions for adding a simulated RPM sensor

CallanDaniel:
Changed RPM_TYPE from 1 to 10 for normal SITL as 1 does not work. Added RPM2_TYPE which needs to be used when running flightaxis link for RealFlight

### DIFF
--- a/dev/source/docs/adding_simulated_devices.rst
+++ b/dev/source/docs/adding_simulated_devices.rst
@@ -169,20 +169,35 @@ Go for a flight and see if you get reasonable data.
 Adding an RPM sensor
 ====================
 
+Depending on the frame/vehicle you are simulating, its physics model might provide either the first or second RPM sensor, or both:
+
 You can add an RPM sensor like this:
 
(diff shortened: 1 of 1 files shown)


Hwurzburg (changes requested):
Actually this is very dependent on the physics model used....some use RPM1, some RPM2, some both....in addition, the wiki needs the  SIM_VIB_MOT_MSK explained as its necessary to get the correct ESC associated with each RPM monitor....so needs some work

CallanDaniel:
@Hwurzburg Copy that - perhaps you could give me a rundown on how this works and I can make a more detailed wiki update? I'm still of the opinion that this is an improvement on the extremely vague info that is currently there regarding RPM simulation.

Hwurzburg:
@CallanDaniel it was just easier to write it up...thanks for bringing this to our attention...

CallanDaniel:
Happy to help! @Hwurzburg

Issue #6103, opened 2024-07-09
Copter: add more Guided mode parameter descriptions


rmackay9:
We should add some more parameter descriptions to the [Copter guided mode page](https://ardupilot.org/copter/docs/ac2_guidedmode.html).

We can probably most just take the information from the [Auto mode page](https://ardupilot.org/copter/docs/auto-mode.html) or just refer to the auto mode page.

The parameters most users probably need to know about are:

- WPNAV_SPEED
- WPNAV_SPEED_UP
- WPNAV_SPEED_DN

Maybe they also need to know about would be WPNAV_ACCEL and WPNAV_JERK but I don't think we need this on the Guided mode page really.

This is in response to this support question where the user was not sure which parameters to use to set the vehicle's maximum horizontal speed. https://discuss.ardupilot.org/t/speed-parameters-in-guided-mode/120415

PR #6113, opened 2024-07-14
add GUIDED mode speed info

Hwurzburg:
addresses #6103

### DIFF
--- a/copter/source/docs/ac2_guidedmode.rst
+++ b/copter/source/docs/ac2_guidedmode.rst
@@ -67,6 +67,20 @@ Instructions
    modes as "Guided". This may not be the case for other Ground Control
    Stations.
 
+Speed Control
+=============
+
+The maximum horizontal speed of the copter can be adjusted with the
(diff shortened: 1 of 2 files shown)


copter/source/docs/ac2_guidedmode.rst
```diff
@@ -67,6 +67,20 @@ Instructions
    modes as "Guided". This may not be the case for other Ground Control
    Stations.
 
+Speed Control
+=============
+
+The maximum horizontal speed of the copter can be adjusted with the
+**Speed** (:ref:`WPNAV_SPEED<WPNAV_SPEED>`) parameter from the Mission Planner's
+Config/Tuning >> Copter Pids screen (see blue box above).  The default
+is 500 meaning 5m/s.  A typical copter can reach top speeds of 10m/s ~
```
rmackay9, line 76: Thanks for this.  It looks great with one exception.  The WPNAV_SPEED default is actually 1000 (10m/s).  It was 500 a few versions ago but we've increased it now.

Hwurzburg: fixed.....also checked the entire wiki for this default change....only here

PR #6114, opened 2024-07-15
Copter: warning about do commands being skipped

rmackay9:
This adds a note/warning about do-commands following a takeoff command.

This is created as as result of this support issue https://discuss.ardupilot.org/t/copter-4-2-x-cant-accept-do-change-speed-in-auto-mission/87509/65


rmackay9:
Thanks!

PR #6117, opened 2024-07-16
Update dependency 2024

khancyr:
update theme to 2.0.0. need https://github.com/ArduPilot/sphinx_rtd_theme/pull/22
-> this fix some small html and js errors
-> allow to use python 3.11 and recent docutil to generate html
update youtube plugin to latest version need https://github.com/ArduPilot/sphinxcontrib-youtube/pull/6
-> fix the status_iterator warning
->fix privacy video link
-> don't generate thumbnails unless building for latex or pdf

raise docutil dependency to 0.20
raise sphinx version to 7.4.4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,9 +2,9 @@
 # Docutils version is for correct bullet point rendering. Can be rolled forward after the theme is updated to >=0.5.1
 # See https://stackoverflow.com/a/68685753/2578171
 sphinx==7.1.2; python_version=="3.8"
-sphinx==7.2.6; python_version>"3.8"
+sphinx==7.4.7; python_version>"3.8"
 sphinx-tabs==3.*
-docutils<0.19
+docutils<0.21
 requests>=2.31.0
 
 # lxml for parameter parsing:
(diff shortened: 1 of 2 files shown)


Hwurzburg:
lots of seemingly unrelated changes

khancyr:
yes, it needs https://github.com/ArduPilot/ardupilot_wiki/pull/6118 first as we got some recursive dependences. Current Sphinx build don't catch it but the new version is catching it.

Hwurzburg:
most of the files changed in this PR bear no relation to the title or stated intention in its description...please separate into a different PR

khancyr:
It needs the recursion fix PR otherwise it won't build and the two other PR. So this one should be the latest merged

rmackay9:
@khancyr I think maybe it would be good to remove the commit that is already included in #6118

Hwurzburg:
@khancyr have you tested this thoroughly?

khancyr:
I have build all wikis without issues reported.
I have check manually that each a opening and open random pages : No visibles diff with the current version, images are at same place and video too. 
So that seems pretty safe.

For the build, it correct a bunch of issues and allow to build a bit faster this we can now disable some options that weren't used in our build !

PR #6120, opened 2024-07-16
Fix product links in common-lightware-lw20-lidar.rst

mbuzdalov:
I checked this page today and found that the links are 404, but the products are still available on a new website.


rmackay9:
Looks OK to me, thanks!

PR #6136, opened 2024-07-23
smart-battery-landingpage: Remove limitation

MattKear:
It looks like this is a hangover from a while back.  The limitation that enforced smart batteries to have an address of 0x0B was removed when the address parameter was added in this PR: https://github.com/ArduPilot/ardupilot/pull/17820

Had a report of it confusing a user about to write their own smart battery driver.  Hence removing.


Hwurzburg:
LGTM...thanks

PR #6138, opened 2024-07-24
common-rpm: fix GPIO pin number and clarify motor scaler calculation

MattKear:
Two minor fixups to RPM:
- Correct param value for GPIO pin RPM type.  This was a mistake noticed by a partner.
- Add clarity to the RPM scaler calculation for eRPM.  I added the example originally, but I want to add some clarity that illustrates the difference between number of poles and pole-pairs.


Hwurzburg:
LGTM...thanks

PR #6141, opened 2024-07-28
Update guide-vtail-plane.rst

Bornlove:
Servo Function & Reversal
Update the servo function output from Roll to Yaw


Hwurzburg:
this is incorrect....read the paragraph above:
"See the table for the correct control 
surface response to the movements. In each instance, the plane 
should move its control surfaces to level itself."
this is the  correcting movement in  fbwa

PR #6151, opened 2024-07-31
dev: Improve first time setup

Ryanf55:
# Purpose

* Skip duplicate installation of MicroXRCEGen
* Add context to why we can't call waf directly
* Split long lines up

# Note

I would like to change the `ros2 launch` command on the `ros2-sitl` page to have one instruction for `4.5` and one for everything later. Is there a common way that we communicate differences across versions in the wiki that would be best?

### DIFF
--- a/dev/source/docs/ros2-sitl.rst
+++ b/dev/source/docs/ros2-sitl.rst
@@ -12,7 +12,9 @@ Once ROS2 is correctly :ref:`installed <ros2>`, and SITL is also :ref:`installed
     cd ~/ros2_ws/
     colcon build --packages-up-to ardupilot_sitl
     source ~/ros2_ws/install/setup.bash
+    # ArduPilot 4.5
     ros2 launch ardupilot_sitl sitl_dds_udp.launch.py transport:=udp4 refs:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/dds_xrce_profile.xml synthetic_clock:=True wipe:=False model:=quad speedup:=1 slave:=0 instance:=0 defaults:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/copter.parm,$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/dds_udp.parm sim_address:=127.0.0.1 master:=tcp:127.0.0.1:5760 sitl:=127.0.0.1:5501
+    # ArduPilot 4.6 removed the refs argument
 
 If the ROS 2 topics aren't being published, set the ardupilot parameter DDS_ENABLE to 1 manually and reboot the launch.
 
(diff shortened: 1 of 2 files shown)


dev/source/docs/ros2-sitl.rst
```diff
@@ -12,7 +12,9 @@ Once ROS2 is correctly :ref:`installed <ros2>`, and SITL is also :ref:`installed
     cd ~/ros2_ws/
     colcon build --packages-up-to ardupilot_sitl
     source ~/ros2_ws/install/setup.bash
+    # ArduPilot 4.5
     ros2 launch ardupilot_sitl sitl_dds_udp.launch.py transport:=udp4 refs:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/dds_xrce_profile.xml synthetic_clock:=True wipe:=False model:=quad speedup:=1 slave:=0 instance:=0 defaults:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/copter.parm,$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/dds_udp.parm sim_address:=127.0.0.1 master:=tcp:127.0.0.1:5760 sitl:=127.0.0.1:5501
+    # ArduPilot 4.6 removed the refs argument
```
Ryanf55, line 17: Let me know how to best handle this. 

In ROS, their wiki supports a tab structure like this. can we do that too, or is there another approach.
https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Configuring-ROS2-Environment.html#source-the-setup-files

Georacer:
> Is there a common way that we communicate differences across versions in the wiki that would be best?

Henry told me recently that the wiki is supposed to reflect `stable`, so it gets overwritten every new release.

But if you think it's really necessary, you can mention a "prior to 4.6..."  paragraph in there.

Hwurzburg:
you can mention rev that applies...4.6 things would be 4.6 wiki label pr,  and ref'd in upcoming featues....4,6 wiki prs get merged when 4,6 is released and removed from upcoming features

Georacer:
While we-re at it, @Ryanf55 do you think we could do away with the `vcs` invocation?

Or has ROS pushed the use of vcs files so heavily that we can't do without it?

Ryanf55:
> > Is there a common way that we communicate differences across versions in the wiki that would be best?
> 
> Henry told me recently that the wiki is supposed to reflect `stable`, so it gets overwritten every new release.
> 
> But if you think it's really necessary, you can mention a "prior to 4.6..." paragraph in there.

Ah ok. Tridge recommended we have devs run on master (ROS2 is a developer feature), so the wiki will need to support both. I'll figure it out.

Ryanf55:
> While we 're at it, @Ryanf55 do you think we could do away with the `vcs` invocation?
> 
> Or has ROS pushed the use of vcs files so heavily that we can't do without it?

We can discuss in another issue. Yes, it's possible, but VCS is well suited when you don't have a monorepo compared to something like submodules.

Hwurzburg:
dev only things I can merge immediately.....its normal users that I am protective of

dev/source/docs/ros2-sitl.rst
Ryanf55: We will need to add the spinhx-tabs instructions.
https://sphinx-tabs.readthedocs.io/en/latest/

I'm fighting with the setup scripts to try to install it without dependency conflicts. Can we merge this as-is, and get the sphinx-tabs introduced in a follow-up?

Hwurzburg:
if you are okay and it does break the wiki, I'm happy to merge

PR #6157, opened 2024-08-01
Add tab structure to document different version run commands for DDS

Ryanf55:
We introduced a breaking change after 4.5 to remove the `refs` file which changed the build command. 
I think introducing the tab extension is a nice way to handle the difference between `4.5` and `master`/`4.6`. 

Ideally, ArduPilot wikii would have a version selector at the bottom for the entire site, but this seems like a simpler solution.

FYI @srmainwaring 

image


Hwurzburg:
sorry it took so long before I tested...looks good...thanks

Ryanf55:
> sorry it took so long before I tested...looks good...thanks

No worries everyone needs a vacation.

PR #6165, opened 2024-08-05
Tradheli-provide clarification on setting level when calibrating

bnsgeyer:
Based on user feedback, updating the hover trim section to make it clearer for leveling the IMU using MP or QGC.  Pulled out some information that made it confusing and was unnecessary.


Hwurzburg (approved):
lgtm...tnx

PR #6056, opened 2024-06-15
Ubuntu24.04 setup

timtuxworth:
Better documents setup on Ubuntu 24.04 and also describes new GitHub requirement for cloning using ssh and the creation of a python virtual environment by the install scripts.

PR #6178, opened 2024-08-13
Revert "Ubuntu24.04 setup"

Hwurzburg:
Reverts ArduPilot/ardupilot_wiki#6056


Hwurzburg:
@timtuxworth sorry, when I fully understood this, its not the only way (I use a personal access token)...its not how GitHub clone button works, and many other parts of the git docs may need changes...we can discuss when I return

PR #6180, opened 2024-08-14
Add GPIO current limit warning(s)

yuri-rage:
Included a GPIO current limit warning and referenced it on associated common articles.


common/source/docs/common-buttons.rst
```diff
@@ -4,6 +4,9 @@
 Buttons
 =======
 
+.. warning::
```
Hwurzburg, line 7: buttons are configured as inputs,not outputs so dont need this warning

Hwurzburg (changes requested):
we dont have a gripper that uses a gpio, my mistake....but we do use the output pins many other places like ice ignition/start,etc. but I dont think we need to worry about those since they are logic outputs to those devices...perhaps the buzzer page after line 32?

common/source/docs/common-buttons.rst
yuri-rage: Fixed along with adding same for buzzer. Agree on the other various functions - I took a look at a solid sample of those and came to the same conclusion.

PR #6189, opened 2024-08-16
speedybeef405wing readme, adds MAVLink ELRS howto

msoos:
This add a sentence about how to configure SERIAL1 to have MAVLink telemetry on the radio. Otherwise, while MAVLink will work (through backpack->wifi->QGC for example), the radio will only show ELRS telemetry.


msoos:
OOops no, sorry.

PR #6190, opened 2024-08-16
add warning to fft re option bit 1

Hwurzburg:
needed to update these in prep for adding per mtr thr based addition


common/source/docs/common-esc-telem-based-notch.rst
```diff
@@ -16,9 +16,7 @@ If ESCs with telemetry are used (either via a separate telemety wire or bi-direc
...
+.. warning:: If you set the multi-source option (i.e. a notch per motor) mentioned below, the bandwidth should not be half the frequency. It should be greatly reduced as more notches cause more phase lag (i.e. latency), thus the bandwidth needs to be reduced to maintain a reasonable phase lag, else oscillation and a poorer tune will result despite the more accurate filtering compared to throttle-based filtering. The suggested starting point is :ref:`INS_HNTCH_BW<INS_HNTCH_BW>` / number of motors. For example, with a quadcopter, :ref:`INS_HNTCH_BW<INS_HNTCH_BW>` / 4 is suggested as a starting point.
```
andyp1per, line 19: It should be INS_HNTCH_FREQ/4 we have already proved that F/number of motors is in fact wrong in a bad way.

andyp1per (changes requested):
Needs couple of amendments

common/source/docs/common-esc-telem-based-notch.rst
Hwurzburg: this should be better now, hopefully

andyp1per: I think just take out :ref:`INS_HNTCH_BW<INS_HNTCH_BW>` / number of motors - just say starting point should be :ref:`INS_HNTCH_BW<INS_HNTCH_BW>` / 4 - that's always a good starting point and the former is wrong in many instances

andyp1per (changes requested):
Take out number of motors

common/source/docs/common-esc-telem-based-notch.rst
Hwurzburg: yep...fixed it on esc and forgot to copy it to fft...

andyp1per (approved):

PR #6200, opened 2024-08-18
update common-temperature-sensor.rst

yuri-rage:
Updated to include a note that this feature is not included by default on most boards.

Included amplifying info throughout.

### DIFF
--- a/common/source/docs/common-temperature-sensor.rst
+++ b/common/source/docs/common-temperature-sensor.rst
@@ -5,40 +5,45 @@
 Temperature Sensors
 ===================
 
-ArduPilot supports up to nine individual temperature sensors with I2C interfaces:
+ArduPilot supports up to nine individual temperature sensors:
 
 - `MAX31865 <https://www.analog.com/media/en/technical-documentation/data-sheets/MAX31865.pdf>`__
 - `MCP9600 <https://ww1.microchip.com/downloads/en/DeviceDoc/MCP960X-Data-Sheet-20005426.pdf>`__
 - `TSYS01 <https://www.te.com/usa-en/product-G-NICO-018.datasheet.pdf>`__
 - `TSYS03 <https://www.te.com/usa-en/product-CAT-DTS0001.datasheet.pdf>`__
+- Analog sensors (thermistors, `LM35 <https://www.ti.com/product/LM35>`__, `TMP36 <https://www.analog.com/media/en/technical-documentation/data-sheets/TMP35_36_37.pdf>`__, etc)
 
 .. image:: ../../../images/temperature-sensor.jpg
    :target: ../_images/temperature-sensor.jpg
 
-In addition, analog temperature monitors can be used.
-
-ArduPilot already has numerous possible sources for temperature reports: ESCs, Smart Batteries, Motor EFI and these independent sensors can be used to replace those devices temperature reports which already exist in ArduPilot. They can also just be logged.
+ArduPilot already has numerous possible sources for temperature reports: ESCs, Smart Batteries, Motor EFI and these independent sensors can be used to replace those device's temperature reports that already exist in ArduPilot. They can also just be logged.
 
 I2C Sensor Setup
 ================
 
-Attach sensor to one of the autopilot's I2C ports using SDA and SCL pins. Then set (examples shown for first sensor):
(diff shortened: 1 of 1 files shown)


common/source/docs/common-temperature-sensor.rst
```diff
@@ -5,54 +5,60 @@
...
+.. note:: Each I2C device on any single bus must have a unique address. If more than one of the same type of I2C temperature sensor is used, ensure each is configured with a unique address. Some breakout boards include jumpers for this purpose. Refer to the product datasheet for specific instructions regarding I2C address configuration.
+
+.. note:: The MAX31865 is an SPI device, which uses the same parameters and basic configuration procedure as I2C sensors but must be connected to an SPI port rather than an I2C port. Some autopilots do not expose an external SPI port, in which case the MAX31865 cannot be easily used.
 
 Analog Sensor Setup
 ===================
 
-Atttch the sensor to an analog input pin. The pin number can be determined from the autopilot's wiki page. Then set (examples shown for first sensor):
+Attach the sensor to an analog input pin (an unused battery voltage or current monitor pin, for example). The pin number can be determined from the autopilot's wiki page. Use care not to exceed the pin's rated voltage (typically 3v3) with the sensor output. Then set (examples shown for the first sensor):
```
Hwurzburg, line 41:
```suggestion
Attach the sensor to an analog input pin (an unused rssi input, analog airspeed input, battery voltage or current monitor input pin, for example). The pin number can be determined from the autopilot's wiki page. Use care not to exceed the pin's rated voltage (typically 3v3) with the sensor output. Then set (examples shown for the first sensor):
```

common/source/docs/common-temperature-sensor.rst
```diff
@@ -5,54 +5,60 @@
 Temperature Sensors
 ===================
 
-ArduPilot supports up to nine individual temperature sensors with I2C interfaces:
+.. note:: Released firmware for most autopilots does not include temperature sensor support by default. You may need to use the `Custom Firmware Builder <https://custom.ardupilot.org>`__ and enable ``Sensors/Temperature Sensors`` as well as one or more of ``Sensors/Temp Sensor - <TYPE>`` to use this feature on your autopilot. If using an analog sensor, enable any one (or more) of the I2C devices in the build menu to ensure the backend is properly included.
```
Hwurzburg, line 8: this is incorrect....all bds with 2MB or more have several temp sensors included...and in general I dont like warning about feature inclusion/exclusion except in the firmware limitations page due to maintenance problems that are caused since they are in continuous flux as the code evolves...so I think this should be removed

Hwurzburg (changes requested):

common/source/docs/common-temperature-sensor.rst
MichelleRos: Also, probably best to use "3.3 V" instead of "3v3" which might be confusing to people not used to this short-hand form.

common/source/docs/common-temperature-sensor.rst
yuri-rage: We may have a bigger issue, then (and this was the primary motivation for the PR):

`AP_TEMPERATURE_SENSOR_ENABLED = 2` is what those boards get. It enables dummy drivers, shows enabled in `features.txt`, but does nothing. Tested with 4.5.5 and CubeOrangePlus today. Even to get analog sensor support, I had to either explicitly `define AP_TEMPERATURE_SENSOR_ENABLED 1` in the hwdef or use the Custom Firmware Builder  and enable an I2C sensor in addition to the basic backend in order to get any TEMPx parameter to display.

There are only 2 periph boards that actually include temperature sensor support in the released firmware.

Hwurzburg: okay then the code needs fixing,not the wiki....needs a code bug report....someone assumed having a bool enable set to 2 sets the normal temp sensors backend includes to test true

common/source/docs/common-temperature-sensor.rst
yuri-rage: Fixed both.

Hwurzburg (approved):

common/source/docs/common-temperature-sensor.rst
yuri-rage: I agree. Some Discord discussion with @peterbarker and @IamPete1 drove me to add the note. Will raise the issue. Note removed.

PR #6215, opened 2024-08-24
Dynamically generate motor diagrams (take 2)

yuri-rage:
Complete rework of https://github.com/ArduPilot/ardupilot_wiki/pull/5142. Zero Javascript, all static images.

Diagrams are generated by `build_motor_diagrams.py` based on JSON output from `AP_Motors_test` (see https://github.com/ArduPilot/ardupilot/pull/23821).

This updates the page with cleaner, more intuitive diagrams that have motor test lettering throughout and adds more than a few missing frame types. Example:

image

Thanks to @TunaLobster for answering a few questions today!


yuri-rage:
@IamPete1 suggested this mechanism https://github.com/IamPete1/ardupilot_wiki/commit/bb913c5fd446ef8dddcc2de57acf5c266653e8a8, that would require manually running a script to download the latest JSON data, assuming it were available as an artifact.

I'd like to see an even cleaner approach where the entire process remains as automated as possible. This avoids the scenario where a new frame is added, and then someone has to remember that we have some script laying dormant.

scripts/motor_diagram_data/connect-escs-and-motors.rst
```diff
@@ -19,163 +19,15 @@ Motor order diagrams
...
-
-.. image:: ../images/motororder-tri-a-2d.png
-    :target: ../_images/motororder-tri-a-2d.png
-    :scale: 40%
-
-.. image:: ../images/motororder-tri-b-2d.png
-    :target: ../_images/motororder-tri-b-2d.png
-    :scale: 40%
-
-.. note::
```
Hwurzburg, line 88: how does the diagram generator add this note?

scripts/motor_diagram_data/connect-escs-and-motors.rst
```diff
@@ -19,163 +19,15 @@ Motor order diagrams
...
-
-.. image:: ../images/motororder-octoquad-v-2d.png
-    :target: ../_images/motororder-octoquad-v-2d.png
-    :scale: 35%
-
-.. image:: ../images/motororder-octoquad-h-2d.png
-    :target: ../_images/motororder-octoquad-h-2d.png
-    :scale: 35%
-
-DodecaHexacopter
```
Hwurzburg, line 169: can you show the complete image output generated that will be inserted which includes the section titles and ordering?

Hwurzburg (changes requested):
want to see the finished page...can I just test locally on a test branch of this without changing anything else? also I have been doing the diagrams as we add frames...its not difficult...but me editing a script will be...so I am not really too keen on this and I am usually very happy for automation

scripts/motor_diagram_data/connect-escs-and-motors.rst
yuri-rage: It lives in `scripts/motor_diagram_data/AP_Motors_display.json`

scripts/motor_diagram_data/connect-escs-and-motors.rst
yuri-rage: The short answer is yes to showing page output.

And yes, you can build this branch with absolutely no changes to the build  environment as it stands now. That was my goal, and it was met.

Here's a longer explanation of where I'm headed so you have fewer worries:

I was hard at work this weekend hacking up a way to pull this stuff down from `ArduPilot/ardupilot`, and I'm about to abandon that idea entirely. It's just too fragile, and I don't want to add unnecessary complexity.

Instead, I will update this PR to do the following:
- keep the page tracked and editable in its usual location
- take the image generation calls out of `update.py`
- place the script into `scripts` for manual execution
- add a page under `dev` describing its usage for updating the images and link it at the bottom of the "Connect Motors and ESCs" page.

I think that will satisfy everyone in the most amicable (and non-breaking) way.

Hwurzburg: Having an image file(s) that can be generated from which I could cut and paste as needed when a new motor set is added would be very helpful, as well as leaving the wiki page layout,etc. still edited normally... would be great,thanks

yuri-rage:
Latest commits comply with desires via Discord and discussion above.

`connect-escs-and-motors.rst` remains a tracked file, and all images are tracked as well. The script simply allows for convenient updates and is documented to the best of my ability on its own (added) page.

Even if you never used the scripted features, this PR amends the page with all available frame classes and types, and includes more informative, clean diagrams that scale nicely when clicked.

scripts/generate_motor_json.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+    re_prefix = f"^{prefix}"
+    if prefix is None or prefix == "":
+        re_prefix = ""
+    version_regex = re.compile(rf"{re_prefix}(\d+\.\d+\.\d+|\d+\.\d+)")
+    branches = fetch_branches(owner, repo)
+    version_branches = [
+        (branch, Version(match.group(1)))
+        for branch in branches
+        if (match := version_regex.search(branch["name"]))
```
IamPete1, line 88: I think this is what is upsetting CI (`:=`)

yuri-rage: Indeed. Apparently we are a little behind on our Python operator linting. Fixed. And thank you!

IamPete1 (approved):
Nice improvement! Great work!

yuri-rage:
Latest push:
- modified SVGs with long frame names to use a smaller font
- modified scaling to provide consistent width
- modified helper script to NOT automatically update markdown docs (preview output only)
- revised script doc to remove reference to automatic markdown update

Hwurzburg (approved):
lgtm...tnx...merging

PR #6228, opened 2024-08-30
Update MATLAB-Simulation.rst

dbash99:
Added a note section to make it easier to follow the docs for users who are less experienced with the WSL networking issues that can come-up when trying to establish the connection between SITL and Matlab.


IamPete1 (approved):
Nice, thanks!

Hwurzburg:
fixed formatting error...thanks

PR #6231, opened 2024-08-30
Nanoradar NRA 24 UART connection information update

prathamEndu:
[Nanoradar NRA 24](https://ardupilot.org/copter/docs/common-rangefinder-nra24.html)

Adding Information regarding UART setup of Nanoradar NRA24 UART version, currently only CAN version is there in the documentation and it is written that only CAN is supported. While I have used it in UART configuration by gathering information from discussion forum, hence sharing the same.

The configuration is very similar to [Ainstein US-D1 Radar](https://ardupilot.org/copter/docs/common-aerotenna-usd1.html) UART version.

I have tested the module on TELEM1 and TELEM2 ports though it should be functioning on any serial port.


common/source/docs/common-rangefinder-nra24.rst
```diff
@@ -7,7 +7,7 @@ Nanoradar NRA 24
 .. note::
     Support for this sensor is available in firmware versions 4.5 and later.
 
-The `NRA24 <http://en.nanoradar.cn/Article/detail/id/372.html>`__ is K-band radar commonly used for detecting height above terrain in all-weather conditions. Currently only connection via CAN is supported. Serial versions of this Radar are not officially supported.
+The `NRA24 <http://en.nanoradar.cn/Article/detail/id/372.html>`__ is K-band radar commonly used for detecting height above terrain in all-weather conditions. Currently only connection via CAN is supported for one or more NRA24 modules. Serial (UART) versions of this Radar is only tested with single NRA24 module.
```
Hwurzburg, line 10:
```suggestion
The `NRA24 <http://en.nanoradar.cn/Article/detail/id/372.html>`__ is K-band radar commonly used for detecting height above terrain in all-weather conditions. Currently, only connection via CAN is supported for more than one NRA24 module. Serial (UART) versions of this Radar have only been tested with a single NRA24 module.
```

common/source/docs/common-rangefinder-nra24.rst
```diff
@@ -16,7 +16,7 @@ Connecting via CAN
 Setup the sensor through the NSM Software. In particular set the "ID" to a unique number if planning to use multiple NRA24 on a single CAN bus. You can also set the bitrate from the software.
 
 .. note::
-    ArduPilot does not currently support  same CAN rangefinders connected to different CAN ports. If you would like to use multiple CAN rangefinders together, connect them to the same bus.
+    ArduPilot does not currently support same CAN rangefinders connected to different CAN ports. If you would like to use multiple CAN rangefinders together, connect them to the same bus.
```
Hwurzburg, line 19:
```suggestion
    ArduPilot does not currently support the same type CAN rangefinders connected to different CAN ports. If you would like to use multiple CAN rangefinders together, connect them to the same bus.
```

common/source/docs/common-rangefinder-nra24.rst
```diff
@@ -31,14 +31,30 @@ For a serial connection you can use any spare CAN port. Since this is not a Dron
...
 
 
+Connecting via UART
+===========================
+For a serial connection TELEM1 and TELEM2 ports are tested. 
+
+.. image:: ../../../images/nra24_uart_connections.png
+    :target: ../_images/nra24_uart_connections.png
+
+Set the following parameters (For setup on TELEM1. Edit SERIAL2 for for TELEM2)
```
Hwurzburg, line 49:
```suggestion
Set the following parameters (Example for setup on TELEM1/SERIAL1)
```

Hwurzburg (changes requested):

rishabsingh3003:
Hi @prathamEndu. As I wrote in the original documentation of NRA24 CAN RangeFinder, we do not officially support the serial driver.  With the NRA 24, there some edge cases that we properly handle in the CAN driver. For example, it tends to stop sending distances when there is no relative motion which I am sure isn't handled by the USD-D1 serial driver. 

I would be uncomfortable in using another sensors backend (USD D1 in this case), which does not match the communication protocol of the original sensor (according to the datasheet on the website), so I assume there is a special firmware to get it to output data similar to the USD-D1. 

In my personal opinion, we should either document this "hack" with a warning stating we do not support the sensor, and do not recommend using it till the time the actual driver is implemented and point users towards the CAN version which is properly supported and well tested, or we should not document it at all. I'll leave it to the judgment of @Hwurzburg.

prathamEndu:
Hi @rishabsingh3003 ,

Got it. I did realized that there is not any official library for UART version of NRA24 and while searching for a way I stumbled upon this post https://discuss.cubepilot.org/t/integration-of-nra24-with-cube-orange/12325 where the person `Aniket` who asked the question `Integration of NRA24 (UART version) with cube orange` gave the solution himself stating that he has had a discussion with the manufacturers and they suggested the param that I added in the doc

I thought probably if this is the manufacturer then they might have checked it properly (Here is also a article from Nanoradar on the firmware changes http://en.nanoradar.cn/Article/detail/id/495.html)

If there is still doubts then putting it in a Warning status is the right course forward.

Hwurzburg:
I suggest adding a warning to the UART section:
.. warning:: this is the recommended configuration by Nanoradar, but ArduPilot has not verified that all operational edge cases are satisfied using the USD-1 driver. Caution is advised.

Hwurzburg:
tweaked a little, squashed and rebased.....merging...thanks

PR #4642, opened 2022-09-22
Move python packages to requirements.txt

TunaLobster:
DELAY MERGE. There are dependencies that need to be merged before this can go in. See ArduPilot/sphinxcontrib-youtube#4.

This is a change for easier maintenance between the install scripts. Also works as an added bonus to check that the build dependencies are working on the packages ArduPilot maintains.

CI will fail on this PR. I ran this change using my sphinxcontrib-youtube repo and you can see the passing CI change over here.
https://github.com/TunaLobster/ardupilot_wiki/actions/runs/3106686180
https://github.com/TunaLobster/ardupilot_wiki/actions/runs/3106686182

PR #6235, opened 2024-08-30
Setup: Add venv for Ubuntu Noble (24.04 LTS)

TunaLobster:
This is an update to #5481 for LTS Noble.

Same ideas and problems being solved here.

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -10,23 +10,23 @@ fi
 DISTRIBUTION_ID=$(lsb_release -i -s)
 if [ ${DISTRIBUTION_ID} = 'Ubuntu' ]; then
   DISTRIBUTION_CODENAME=$(lsb_release -c -s)
-  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'lunar' ]; then
+  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'noble' ]; then
     sudo add-apt-repository universe -y
   fi
 fi
 
+sudo apt-get -y update
+sudo apt-get install -y unzip git imagemagick curl wget make python3
+
 # create a Python venv on more recent releases:
-if [ ${DISTRIBUTION_CODENAME} == 'lunar' ]; then
-    sudo apt install python3.11-venv
+if [ ${DISTRIBUTION_CODENAME} = 'noble' ]; then
+    sudo apt install python3-venv
     python3 -m venv $HOME/venv-ardupilot-wiki --upgrade-deps
 
     # activate it:
     SOURCE_LINE="source $HOME/venv-ardupilot-wiki/bin/activate"
     $SOURCE_LINE
 fi
-
-sudo apt-get -y update
-sudo apt-get install -y unzip git imagemagick curl wget make python3
 
 # Install packages release specific
 if [ ${DISTRIBUTION_CODENAME} = 'bionic' ]; then
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -10,9 +10,19 @@ fi
 DISTRIBUTION_ID=$(lsb_release -i -s)
 if [ ${DISTRIBUTION_ID} = 'Ubuntu' ]; then
   DISTRIBUTION_CODENAME=$(lsb_release -c -s)
-  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ]; then
-    sudo add-apt-repository universe
+  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'lunar' ]; then
+    sudo add-apt-repository universe -y
   fi
+fi
+
+# create a Python venv on more recent releases:
+if [ ${DISTRIBUTION_CODENAME} == 'lunar' ]; then
+    sudo apt install python3.11-venv
+    python3 -m venv $HOME/venv-ardupilot-wiki --upgrade-deps
+
+    # activate it:
+    SOURCE_LINE="source $HOME/venv-ardupilot-wiki/bin/activate"
+    $SOURCE_LINE
 fi
 
 sudo apt-get -y update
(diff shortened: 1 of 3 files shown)


Sphinxsetup.sh
```diff
@@ -45,27 +55,26 @@ else
...
-
 # and youtube and video plugins:
 # This command might require a --force option if you have and older extension installed
 # Rerun Sphinxsetup.sh after doing that
-python3 -m pip install --user --upgrade git+https://github.com/ArduPilot/sphinxcontrib-youtube.git
-
+# and a parser to use getting posts from Discourse (forum) and insert in FrontEnd
 # Install flake8
-python3 -m pip install --user --upgrade flake8==3.7.9
+python3 -m pip install $PIP_USER_ARGUMENT --upgrade sphinx==${SPHINX_VERSION} "docutils<0.19" "requests>=2.31.0"  sphinx-tabs==3.* lxml git+https://github.com/ArduPilot/sphinx_rtd_theme.git git+https://github.com/ArduPilot/sphinxcontrib-youtube.git beautifulsoup4 flake8==3.7.9
```
peterbarker, line 77: Please leave these as separate lines.

The efficiency is in no way important for this script, and it's easier to tell what's gone wrong with the separate lines.

It's also easier to tell exactly what's going on with the separate lines

update.sh
```diff
@@ -31,6 +31,11 @@ lock_file build.lck || {
     echo "$(date +%s) already locked" >>build.lck.log
     exit 1
 }
+# activate Python venv on more recent releases:
+if [ ${DISTRIBUTION_CODENAME} = 'mantic' ]; then
```
peterbarker, line 35: Weren't we talking noble?!

Sphinxsetup.sh
```diff
@@ -10,14 +10,24 @@ fi
 DISTRIBUTION_ID=$(lsb_release -i -s)
 if [ ${DISTRIBUTION_ID} = 'Ubuntu' ]; then
   DISTRIBUTION_CODENAME=$(lsb_release -c -s)
-  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ]; then
-    sudo add-apt-repository universe
+  if [ ${DISTRIBUTION_CODENAME} = 'focal' ] || [ ${DISTRIBUTION_CODENAME} = 'bionic' ] || [ ${DISTRIBUTION_CODENAME} = 'noble' ]; then
```
peterbarker, line 13: Do we need universe for noble?

peterbarker (changes requested):

Sphinxsetup.sh
TunaLobster: I put them together as pip has a much improved dependency resolver that can handle version conflicts when given all of the information. I can separate them out. I'm going to update #4642 next to move the list of packages to something common between setup scripts.

update.sh
TunaLobster: Sorry. Did a push without commiting from the old PR. Should be good now.

Sphinxsetup.sh
TunaLobster: I'm not sure I follow. With the current --upgrade option doing this on separate lines could result in different behavior as follow on pip install commands do not take into account the previously commanded pinning. Only when given all packages at once can the dependency resolver correctly do both updating and pinning.

khancyr: I would say that isn't a bad idea to regroup things together

Hwurzburg:
@peterbarker @khancyr awaiting approvals

peterbarker:
> @peterbarker @khancyr awaiting approvals

... want to know the impact of `--upgrade XYZZY` on packages which are *not* XYZZY and have been pinned

peterbarker:
> ```
> ╭─khancyr@MAN-LP-0101 /mnt/c/Documents and Settings/PierreKancir/Downloads/logs
> ╰─$ sudo pip install -U pymavlink==2.4.13
> Requirement already satisfied: pymavlink==2.4.13 in /usr/local/lib/python3.10/dist-packages (2.4.13)
> Requirement already satisfied: future in /usr/local/lib/python3.10/dist-packages (from pymavlink==2.4.13) (0.18.3)
> Requirement already satisfied: lxml in /usr/lib/python3/dist-packages (from pymavlink==2.4.13) (4.8.0)
> WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
> ╭─khancyr@MAN-LP-0101 /mnt/c/Documents and Settings/PierreKancir/Downloads/logs
> ╰─$ sudo pip install -U mavproxy
> Collecting mavproxy
>   Using cached MAVProxy-1.8.71-py3-none-any.whl (7.2 MB)
> Requirement already satisfied: pyserial>=3.0 in /usr/lib/python3/dist-packages (from mavproxy) (3.5)
> Collecting pymavlink>=2.4.14
>   Using cached pymavlink-2.4.41-py3-none-any.whl (11.6 MB)
> Requirement already satisfied: pynmeagps in /usr/local/lib/python3.10/dist-packages (from mavproxy) (1.0.39)
> Requirement already satisfied: numpy in /usr/lib/python3/dist-packages (from mavproxy) (1.21.5)
> Requirement already satisfied: future in /usr/local/lib/python3.10/dist-packages (from pymavlink>=2.4.14->mavproxy) (0.18.3)
> Requirement already satisfied: lxml in /usr/lib/python3/dist-packages (from pymavlink>=2.4.14->mavproxy) (4.8.0)
> Installing collected packages: pymavlink, mavproxy
>   Attempting uninstall: pymavlink
>     Found existing installation: pymavlink 2.4.13
>     Uninstalling pymavlink-2.4.13:
>       Successfully uninstalled pymavlink-2.4.13
> Successfully installed mavproxy-1.8.71 pymavlink-2.4.41
> WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
> ```
> 
> Pining only works for the current cmdline, so any further pip install -U will remove the pining to satisfy the dependency. If you try `sudo pip install -U pymavlink==2.4.13 mavproxy ` it will look for the mavproxy version that is compatibles with the pinned pymavlink

OK, thanks for checking.

I guess I have to withdraw my objections to the one-line thing, 'though it makes it a PITA to work out what the problems are.  Could we at least use `\` to separate the packages onto separate lines to make it easie to see them all at a glance?

TunaLobster:
This should be good as is. I think a better way would be to check if the year of the Ubuntu release is greater than 24. That would keep it working on the non-LTS releases too.

Hwurzburg:
Thanks for this @TunaLobster, and sorry it sat so long. Closing it, as most of it has been overtaken and one part of it turns out not to work.

**Already on master, independently of this PR.** The `distutils` removal and `pkg_resources` guard landed in 681da9f83 ("update.py: remove distutils.dir_util.mkpath for os.makedirs"), using the same `os.makedirs(..., exist_ok=True)` change proposed here. `update.py` no longer references either module. This also makes @peterbarker's question above about `--upgrade` on pinned packages moot for this diff.

**The `update.sh` hunk does not work.** It guards on `${DISTRIBUTION_CODENAME}`, which is never set in `update.sh` - that script has no `lsb_release` call. The test expands to `[ == 'noble' ]`:

```
+ '[' == noble ']'
bash: line 1: [: ==: unary operator expected
```

It does not abort the build (a failing `if` condition is exempt from `set -e`), so the effect is an error on stderr on every production build run while the venv silently never activates. There is also a name mismatch: the venv is created as `venv-ardupilot-wiki` but the message in `update.py` refers to `arudpilot-wiki-venv`.

**The `--user`-in-a-venv problem is handled by #7907**, which skips `--user` when `$VIRTUAL_ENV` is set. That takes the opposite approach to this PR - it reacts to a venv the user has already created rather than creating and activating one - which avoids needing the activation to survive from the setup script into the user's shell.

**What that leaves.** PEP 668 on noble is now answered by "create your own venv", with #7907 making that work. If anyone wants to carry one idea forward from here, it is the friendly guard around the sphinx import - `update.py:58` is still a bare `from sphinx.application import Sphinx`, so a missing install gives a traceback rather than "run Sphinxsetup.sh". Worth a small separate PR if someone feels like it.

Happy to reopen if you think something here still needs doing.

PR #6249, opened 2024-09-06
Stop testing ROS2 in parallel

Ryanf55:
* It causes issues with multiple SITL ports and DDS topics overlapping

Follow up to upstream: https://github.com/ArduPilot/ardupilot/pull/28024


Hwurzburg:
lgtm...thanks

PR #6262, opened 2024-09-10
Fix user/group

Ryanf55:
* It's needed for both docker environments
* It's linux specific

image

### DIFF
--- a/common/source/docs/common-wiki-editing-setup.rst
+++ b/common/source/docs/common-wiki-editing-setup.rst
@@ -109,10 +109,22 @@ Setup with Docker
 
 #. Open a command prompt in the root of the ardupilot_wiki repo and build the docker container for the wiki:
 
-   .. code-block:: bash
+.. tabs::
 
-       cd ardupilot_wiki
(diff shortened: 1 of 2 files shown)


Ryanf55:
@yuri-rage Test this please a follow up to https://github.com/ArduPilot/ardupilot_wiki/pull/6256

yuri-rage:
Looks like the real change (other than differentiating Powershell, which is a good call) is use of  `pwd` vice the `$PWD` environment variable. Have not tested thoroughly, but quick run of the following wiki build worked without issue on an Arch machine and Ubuntu 22.04 (WSL2) environment.

`docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 update.py --fast --site copter`

Might be wise to update the utility script to use the same to ensure compatibility where the env variable isn't guaranteed. Wouldn't hurt to make that change in this PR.

Hwurzburg:
since your both agree this is good to go, merging

PR #6266, opened 2024-09-10
Plane SITL dev docs update

YuxinPan:
**Motivation**: Had a bit confusion myself when setting up plane SITL when following this page

1. Updated command path to forward slash given ArduPilot SITL now supports `WSL` and `Linux` (and no longer recommends Windows). This will also make it consistent with the `SITL Copter` docs.
2. Update the screenshot in the doc to be plane specific (instead of copter).

@Hwurzburg


YuxinPan:
The screenshot is reproduced with the same format as the original images, with the only difference being shifting to plane model rather than a copter model.

Hwurzburg:
LGTM

PR #6267, opened 2024-09-11
Add Holybro Pixhawk 6X Pro to common-autopilots.rst

vincentpoont2:
Add Holybro Pixhawk 6X Pro


vincentpoont2:
@Hwurzburg Can I add the product link here first?

Hwurzburg:
Yes...needs to be called "pro" in label...also does this use the existing 6X firmware?

Hwurzburg (changes requested):

vincentpoont2:
> Yes...needs to be called "pro" in label...also does this use the existing 6X firmware?

Thanks, forgot the "Pro" part. It does use the same target as the Pixhawk 6X.
I put the note here: https://docs.holybro.com/autopilot/pixhawk-6x-pro/supported-firmware

Hwurzburg:
shouldnt we update this page instead?I can add this link in the interim

PR #6268, opened 2024-09-11
Revert "update.py: split procs per tasks and max it out"

peterbarker:
This reverts commit 9238723f7a993428032a4820a0a15002d01f4bb4.

This increased parallelisation on the Wiki server which causes it to run out of memory.

There does not seem to be an command-line option to specify the parallelisation  - just remove it.

We are memory-limited in our build, not CPU-limited.

ping @khancyr

### DIFF
--- a/update.py
+++ b/update.py
@@ -232,7 +232,7 @@ def fetch_ardupilot_generated_data(site_mapping: Dict, base_url: str, sub_url: s
         executor.map(fetch_and_rename, urls, targetfiles, names, timeout=5*60)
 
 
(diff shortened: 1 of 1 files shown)


khancyr:
No problem to revert if it fix the build currently.

I will give a try to profile the build sequencially but with max cpu against our current solution.

Problem here is that we multiply the sphinx build call and the param versionning is nuking the ram, so even in reducing the cpu count on sphinx build we could overflow the ram depending on how much wiki build reach the parameters parsing part at the same time ...

PR #6274, opened 2024-09-13
Dev: update dev conference wiki page

rmackay9:
This updates our developer conference wiki page to show info about the 2024 conference.

It's not very visible so we may need to add a link from somewhere else (the community menu at the top of ArduPilot.org?) but this is a start


Hwurzburg:
Those menus are in the "concrete" I think and I dont know how to change them or cat remember, and not sure a transient event belongs there...we could put a bullet heading here:
image

rmackay9:
Yes, thanks for that.

Maybe we can just merge this as-is and I'll do another blog post in the near future to keep it visible in the blog sections of ardupilot.org.

Issue #6273, opened 2024-09-12
Serial Passthrough documentation needs correction and improvement


brad112358:
https://ardupilot.org/copter/docs/common-serial-passthrough.html incorrectly states that SERIALx_PROTOCOL should be set to -1("None").  

This would seem to be bad advice since this disables Rx and Tx pins at the driver level.  Instead, we should note that when SERIAL_PASS2 is set to non-zero, both serial ports are locked, temporarily blocking their normal use.

In addition, it would be nice to add a note that documents the special handling of USB ACM ports which will automatically set the baud rate and other settings of the SERIAL_PASS2 port to match the USB ACM host serial settings.  This means, for example, that the user does not need to modify their normal settings for SERIALx_BAUD in the common use case where SERIAL_PASS1 is a USB port.

It might also be helpful to note at the bottom that setting SERIAL_PASSTIMO to 0 avoids the passthrough timeout entirely, allowing passthrough operation until the Flight Controller is rebooted.  In fact, recommending this rather than 30 to 60 seconds may help avoid user support issues caused by the timeout expiring before the host serial application is started by the user.

peterbarker:
@brad112358 did you want to put a PR forward with your suggested changes?

brad112358:
> @brad112358 did you want to put a PR forward with your suggested changes?

Sure, I can do that.  Do all three changes I proposed sound OK?

Hwurzburg:
I thought it needed some other updates so I just pushed a PR rather than typing a lot of words here (currently one armed)

PR #6275, opened 2024-09-13
Update/correct serial passthru info

Hwurzburg:
addresses #6273

### DIFF
--- a/common/source/docs/common-serial-passthrough.rst
+++ b/common/source/docs/common-serial-passthrough.rst
@@ -8,10 +8,6 @@ Serial Passthrough
 
 "Serial Passthrough" allows the autopilot to route serial protocol messages between serial ports on the autopilot including the USB port.  This can be used to configure a sensors or devices (which have a serial interface) without disconnecting it from the autopilot using a USB connection. It is also possible to pass data from any serial port to another serial port.
 
-.. note::
-
-     This feature is available in Copter-4.0 (and higher), Plane-3.10, Rover-3.6 (and higher)
-
 Setup
 -----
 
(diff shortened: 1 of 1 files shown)


common/source/docs/common-serial-passthrough.rst
```diff
@@ -20,19 +16,18 @@ Setup
...
 
 .. figure:: ../../../images/serial-passthrough-gps-baud.png
    :target: ../_images/serial-passthrough-gps-baud.png
    :width: 300px
 
 - The default for :ref:`SERIAL_PASS1<SERIAL_PASS1>` (the first serial port to be interconnected) defaults to SERIAL0, the primary USB port.
-- Set :ref:`SERIAL_PASSTIMO <SERIAL_PASSTIMO>` to a length of time (in seconds) that gives you enough time to connect with the sensor's configuration software.  30 to 60 seconds is a good choice
+- Set :ref:`SERIAL_PASSTIMO <SERIAL_PASSTIMO>` timeout to a length of time (in seconds) that gives you enough time to connect with the sensor's configuration software. 30 to 60 seconds is a good choice but setting it to zero will remain in passthrough until reboot occurs.
 - Set :ref:`SERIAL_PASS2 <SERIAL_PASS2>` to the number of the serial port connected to the sensor.  I.e. "2" if the sensor is connected to Telem2/Serial2.
 - Press the "Disconnect" button on the ground station but leave the USB cable from the PC to the autopilot connected.
```
brad112358, line 29: Some GCS (QGC for example) don't have an obvious "Disconnect" button.  It's probably best to advise the user to close the GCS application to ensure the port is available for use by the configuration software or tool.

brad112358:
I'm guessing that setting SERIALx_PROTOCOL to 0 ("None") didn't originally disable the serial port at the hardware/driver level, but just prevented it's use by the higher layers.  I ran into a case just today where this capability is still required and where the serial port locking is not good enough.  

I have an AIO flight controller with a built-in serial ELRS receiver and I was attempting to use passthrough to flash a custom firmware image using esptool.  This requires putting the receiver into bootloader mode by holding down the boot button when plugging in the USB which powers it up.  But, most of the time I couldn't get it to flash after setting up passthrough.  

It turns out that the bootloader seems to get confused by the MAVLink messages which Ardupilot was sending before I could use QGC to setup passthrough and it becomes unresponsive.  In the past, when the MLRS receiver was external to the FC, I have arranged to power the Rx only when the battery is connected and connect the battery while holding down the boot button only after starting passthrough which avoids this problem. But this solution is not available for the integrated receiver.

The workaround I found today is to set SERIALx_PROTOCOL to 28 ("Scripting") which seems a safe setting since I have no scripts so nothing would be using UART.write() to output on this port.  This completely solved the problem, but is less than ideal from a documentation point of view.  I think the ideal solution is to introduce a "unused" or "passthrough only" setting for SERIALx_PROTOCOL which ensures no output prior to enabling passthrough without disabling the serial port at the hardware/driver level.

Until this can be discussed with developers and a better solution implemented, I propose documenting the possible problem and a workaround here.  "Scripting" was the first thing I tried.  Is there a better existing setting for SERIALx_PROTOCOL which never outputs anything on the serial port.

My current workaround is documented in the Flashing section [here:](https://github.com/brad112358/mLRS-docu/blob/esp_passthrough_proto/docs/ELRS_RECEIVERS.md)

common/source/docs/common-serial-passthrough.rst
Hwurzburg: done

Hwurzburg:
I have opened a feature enhancement request #28105 for a change to protocol =0 to cover this...but don't hold your breath, the devs capable of doing this are swamped...thanks for the inputs...

PR #6276, opened 2024-09-14
Common: Siyi gets download script link

rmackay9:
The Siyi wiki page gets a link to a python script that allows downloading images and videos from the camera


Hwurzburg:
thanks

PR #6290, opened 2024-09-20
Update terrain-following.rst : Mission Planner DAT file creating

EosBandi:
New plugin is MP, supports terrain dat file creation based on MP's terrain data (including DEM files)


Hwurzburg (commented):
the ArduPilot wiki recommends 100m spacing as anything less is unnecessary and wastes SD card space....shouldn't the default reflect that? unless you are using a server with better resolution than ArduPilot's SRTM server? and Plane wouldn't benefit anyway to more resolution.... having tiles 10X larger than current for plane seems unnecessary....even if Copter could use them, I think we think to elaborate or change the default.

Hwurzburg (approved):

PR #6291, opened 2024-09-21
modify notes on commit squashing

Hwurzburg:


rmackay9 (approved):
It looks OK although it's a bit redundant because of this section higher up in the instructions

`- Commits should be small and do just one thing. If a change touches multiple libraries then there should be a separate commit per library, and a separate commit per vehicle directory. This is true even if it means that intermediate commits break the build.`

Anyway, fine to merge from my point of view

PR #6299, opened 2024-09-24
Update common-crsf-telemetry.rst

ot0tot:
Update info on ELRS telemetry, as this has not been accurate for several years.


Hwurzburg:
thanks

PR #6300, opened 2024-09-25
Update common-tbs-rc.rst

ot0tot:
Update ELRS telemetry info.

### DIFF
--- a/common/source/docs/common-tbs-rc.rst
+++ b/common/source/docs/common-tbs-rc.rst
@@ -6,14 +6,14 @@ Crossfire and ELRS RC Systems
 
 Any Crossfire/ELRS compatible receiver can be used with ArduPilot. 
 
-.. note::  ELRS (ExpressLRS) RC systems use the Crossfire protocol and are connected in a similar manner. However, ERLS telemetry does not provide bi-directional telemetry like Crossfire. Also, on F4/F7 based autopilots the UART MUST be DMA enabled for reliable operation. Consult :ref:`common-autopilots`.
+.. note::  ELRS (ExpressLRS) RC systems use the CRSF protocol and are connected in a similar manner as Crossfire receivers. Also, on F4/F7 based autopilots the UART MUST be DMA enabled for reliable operation. Consult :ref:`common-autopilots`.
 
-If you do not wish to use telemetry then a **TBS Crossfire** receiver can be connected to the **RCIN** port using :ref:`SBUS <common-rc-systems>`. You must configure the Receiver to output SBUS, of course.
+If you do not wish to use telemetry then the receiver can be connected to the **RCIN** port using :ref:`SBUS <common-rc-systems>`. You must configure the Receiver to output SBUS, of course.
 
 CRSF Receivers 
 ==============
 
-If you wish to use telemetry then a TBS receiver can be connected to a UART utilizing the `CRSF <https://www.team-blacksheep.com/products/prod:crossfire_tx>`__ protocol.
+If you wish to use telemetry then a receiver can be connected to a UART utilizing the `CRSF <https://www.team-blacksheep.com/products/prod:crossfire_tx>`__ protocol.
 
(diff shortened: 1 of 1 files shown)


Hwurzburg (commented):
before I merge more on the ELRS MAVLink, I think we need to determine if this is using RC overrides to transport the RC or the embedded RC protocol .....because autopilot setup is different....if only overrides are used it must be setup for GCS only operation and failsafe vs using embedded RC as an RC system protocol, with different failsafe...can you determine which is being used...the ELRS site does not specify

MUSTARDTIGERFPV:
> before I merge more on the ELRS MAVLink, I think we need to determine if this is using RC overrides to transport the RC or the embedded RC protocol .....because autopilot setup is different....if only overrides are used it must be setup for GCS only operation and failsafe vs using embedded RC as an RC system protocol, with different failsafe...can you determine which is being used...the ELRS site does not specify

Presently ExpressLRS MAVLink sends RC_OVERRIDE frames. In copious testing, this has been 100% sufficient on default configurations. Fail-safes and safety checks pass without issue, and all behavior is as expected. 

When proper MAVLink RC support is merged in the next few weeks/months, ExpressLRS will be happy to support it.

Hwurzburg:
the setup of failsafe in the autopilot is different if using gcs only (ie RC overrides)....we need to point to this https://ardupilot.org/plane/docs/common-gcs-only-operation.html in each place where we say you can use ELRS MAVlink for the control link

MUSTARDTIGERFPV:
@Hwurzburg that's incorrect; this kind of failsafe relates to a loss of heartbeat messages from the GCS which is not relevant. The loss of RC_OVERRIDE is enough to trigger failsafe.

Hwurzburg:
you are right....I checked with the lead dev on that code and he even wasnt sure (he had reviewed the gcs only page last year), so we setup a CubeOrange and tested....if THR_FAILSAFE is not set to zero (like normal RC) then losing RC Overrides from the GCS will do a normal short and long failsafe..
so I removed links to that page (I need to expand it a bit in light of this). and tweaked/expanded the info on your two PRs and pushed it to them...will merge tomorrow AM after CI pass

common/source/docs/common-tbs-rc.rst
```diff
@@ -34,7 +34,18 @@ See :ref:`common-crsf-telemetry` for information about telemetry data sent, disp
...
-ELRS can be setup in the same manner as CRSF above , however, bit 13 of :ref:`RC_OPTIONS<RC_OPTIONS>` should be set to alter the baudrate from 416KBaud that CRSF uses, to 420KBaud that ELRS uses. As mentioned above, the UART selected must have DMA capability.
+ELRS can be setup in the same manner as CRSF above , however, bit 13 of :ref:`RC_OPTIONS<RC_OPTIONS>` should be set to alter the baudrate from 416KBaud that CRSF uses, to 420KBaud that ELRS uses. As mentioned above, the UART selected must have DMA capability. ELRS can also be configured to output MAVLink, enabling full bidirectional Mavlink telemetry and RC control over the same link. See :ref:`ELRS <common-rc-systems>` for more information.
+
+ELRS MAVLink Configuration
+--------------------------
+
+Instead of CRSF protocol, MAVLink protocol can be used. In this case, using SERIAL 4 for example:
+
+- Set :ref:`SERIAL4_PROTOCOL <SERIAL4_PROTOCOL>` = 23
+- Set :ref:SERIAL4_BBAUD <SERIAL4_BAUDL>` = 460
```
ot0tot, line 45: Shouldn't this be set to 2 for Mavlink 2?
```suggestion
- Set :ref:`SERIAL4_PROTOCOL <SERIAL4_PROTOCOL>` = 2
- Set :ref:SERIAL4_BAUD <SERIAL4_BAUDL>` = 460
```

Hwurzburg: yep..copy/paste error....thanks

PR #6305, opened 2024-09-28
initvagrant: change directories before running Sphinxsetup

peterbarker:
recent proposed change assumes you were in this cwd, so adjust for that

### DIFF
--- a/scripts/initvagrant.sh
+++ b/scripts/initvagrant.sh
@@ -14,7 +14,9 @@ if [ ${DISTRIBUTION_ID} == 'Ubuntu' ]; then
   fi
 fi
 
-sudo -H -u vagrant /vagrant/Sphinxsetup.sh
+pushd /vagrant
+sudo -H -u vagrant ./Sphinxsetup.sh
+popd
 OUTDIR=/var/sites/wiki/web
 OUTDIR_OLD=/var/sites/wiki/web/old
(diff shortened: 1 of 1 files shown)


scripts/initvagrant.sh
```diff
@@ -14,7 +14,9 @@ if [ ${DISTRIBUTION_ID} == 'Ubuntu' ]; then
   fi
 fi
 
-sudo -H -u vagrant /vagrant/Sphinxsetup.sh
+pushd /vagrant
```
NDevDrone, line 17: Maybe could add comment along the lines of "Forces shell to /vagrant so Sphinxsetup.sh always runs in the correct directory"

NDevDrone (approved):
Forces shell's working directory to be always correct to ensure relative paths defined in script operate as expected, and then returns to CWD once script is complete (or exits).

scripts/initvagrant.sh
NDevDrone: Any interest in hiding the directory‐stack printout
`pushd /vagrant > /dev/null`

peterbarker: > Any interest in hiding the directory‐stack printout `pushd /vagrant > /dev/null`

No - this is an install script.  It's *allowed* to be noisy and verbose because you run it rarely, and if something goes wrong you want to know exactly what was going on.

PR #6306, opened 2024-09-28
Relay Page update

jlpoltrack:
Mission Planner now seems to allow all 16 relays to be modified and also no longer seems to use 0-index values.

Mission Planner version installed for me is 1.3.82 build 1.3.8979.17128


Hwurzburg:
lgtm...thanks

PR #6330, opened 2024-10-16
Update common-esp8266-telemetry.rst

vcrann:
Fixed a broken link for a product linked on this page


Hwurzburg:
The existing link is correct...

vcrann:
Yep, it was broken but it's been changed on the other end now so I'll close

Issue #6311, opened 2024-10-01
ROS 2: microxrceddsgen version: null


AssaX0:
Hello,

currently I am trying to get a Pixhawk6C to raspberrypi4B connection Working.
I am using Ubuntu 22.04 with ROS2 Humble and try to follow the Tutorial:
[ROS 2](https://ardupilot.org/dev/docs/ros2.html) 
I followed everything (twice - removed everything and tried again after it did not work) up to the point where you are supposed to check the
"microxrceddsgen -version"
wich returns
"microxrceddsgenversion: null"
I tried fixing it myself, but I am no Linux Main so I asked chat gpt who found a probable typo (?) in:
nano /home/pi/ardu_ws/Micro-XRCE-DDS-Gen/scripts/microxrceddsgen
one Line says
"if ! hash ${jave_exec}; then" what should probably be a:
"if ! hash ${java_exec}; then" ... but after fixing I stil got a 
"microxrceddsgen version: null" and chatgpt is not helpfull anymore.

I hope someone can help me fixing this!

BasharAmeen:
+1

Ryanf55:
Ignore the version. We can't do a formal release.

Use `microxrceddsgen -h` to check it works.

AssaX0:
Well i guess it doesnt?
![image](https://github.com/user-attachments/assets/248ec8d2-0c6a-45a3-832c-241738bb9861)

And If I would like to continue the dds tests start to fail... I guess thats the consequence?
![image](https://github.com/user-attachments/assets/19fd149c-a2c4-4624-92d3-555f0b6627cc)

Ryanf55:
[...]

PR #6340, opened 2024-10-18
Use help instead of version to verify uxrcegen

Ryanf55:
Solves #6311


Hwurzburg:
u da man on DDS , so merging

PR #6341, opened 2024-10-19
Fix dockerfile missing requirements

Ryanf55:
# Problem

The requirements file was not copied into the dockerfile.

image

# Solution

I switch to bind mount so you don't have to delete the file, and this is better for cache.

# Test 

```bash
docker build . -t ardupilot_wiki --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) --progress=plain
```


Ryanf55:
Caused by https://github.com/ArduPilot/ardupilot_wiki/pull/4642

Hwurzburg:
ditto...thanks

PR #6347, opened 2024-10-24
add new model DATAGNSS NANO RTK Receiver

datagnss:
- add new model
- add datagnss into the store list


Hwurzburg:
LGTM...thanks

PR #6348, opened 2024-10-24
Wiki: replace old command

lejan:
Discouraged older git protocol, use new one wich is more secure.


Hwurzburg:
lgtm , especially since the git: address is no longer valid

PR #6350, opened 2024-10-26
Update ESP32 Telemetry

seeul8er:
Added links to new official Wiki and website for always up-to-date information. Restructured to be more readable.
Added information about the new ESP-NOW mode that allows for significantly more range using a non WiFi-based protocol.


Hwurzburg:
a little editing for title levels and to remove repetition .....merging

PR #6359, opened 2024-10-30
ROS: update rosdep command

srmainwaring:
Add `-y -r` flags to continue on error and accept all changes.

### DIFF
--- a/dev/source/docs/ros2.rst
+++ b/dev/source/docs/ros2.rst
@@ -53,7 +53,7 @@ Now update all dependencies:
     sudo apt update
     rosdep update
     source /opt/ros/humble/setup.bash
-    rosdep install --from-paths src --ignore-src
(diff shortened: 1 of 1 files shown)


Ryanf55:
Which package do we need the `-r` flag for?

srmainwaring:
Gazebo harmonic packages are not found using `rosdep` on jammy. Could use skip keys or the alternative rosdep sources for non-standard distros as is done in the ardupilot_ros CI workflow definition.

rosdep-missing

Ryanf55:
Yea, perhaps it's time we start recommending this workflow to our users. It works well.
https://github.com/ArduPilot/ardupilot_ros/blob/940efb9a41d80eb6d4ae5a0f66cdab412f2a1e70/.github/workflows/colcon.yml#L44

Hwurzburg:
going to merge as is, since it appears okay...further changes can be in a future PR

PR #6361, opened 2024-11-01
Add XBee SX 868 Config

rodeo-sam:
- added XBee SX 868 example config
- reformatted XBee ZB table for better readability (more spaces)
- some consistency (80 char line length, line breaks with new sentences, ...)
- typo fixes


Hwurzburg:
thanks....

PR #6362, opened 2024-11-01
images: fix misplaced image file extensions

rsaxvc:
Rename file extensions to match file formats and update references accordingly.


Hwurzburg:
no impact to wiki rendering....I was hard pressed to find why this makes a difference, every graphics editing or conversion program I have simply ignores the file extension and uses the file's internal data header....what program  does this make a difference to, out of curiosity? since it has no impact on the wiki and is technically correct I am merging...

rsaxvc:
Gimp - depends how file is specified.

Eye of Mate - "Error interpreting JPEG image file (Not a JPEG file: starts with 0x89 0x50)", but Eye of Gnome works.

Android Photos app - saved files from wiki render correctly but shown with "0.0MP"

`find . -iname "*.jpeg" -exec jpegoptim {} \+` and `find . -iname "*.png" -exec pngcrush -ow {} \;
` now works as expected.

PR #6367, opened 2024-11-03
Update common-radio-control-calibration.rst

HuguesD:
Added notes on throttle and radio transmitter throttle stick configuration. Added a note at the end of the calibration procedure to advise to check RC3_TRIM is throttle behaviour is not as expected


Hwurzburg:
thanks for this...I think it needs some work and I will try to modify it this weekend, or ASAP:
reverse applies to plane throttles also and it has a wiki page that needs referencing
the PWM ranges depend on params, not fixed values
plane operation with a single direction throttle and an AUX switch for reverse action is possible, not just a centered throttle
and mentioned param needs to be linked

Hwurzburg:
made some tweaks...thanks..merging

PR #6370, opened 2024-11-05
Ark docs

patrikpordi:
I have added the ardupilot compatible hardware components from ARK Electronics


common/source/docs/common-non-gps-navigation-landing-page.rst
```diff
@@ -14,6 +14,8 @@ These are the available options that allow a vehicle to estimate its position wi
 .. toctree::
     :maxdepth: 1
 
+    ARK ADIS16507 <common-ark-adis16507>
```
Hwurzburg, line 17: I am not sure we should be putting IMUs, gyros, etc. in the AP wiki....these are board/subsystem components and definitely do not provide non-gps-navigation in and of themselves....they really do not fit anywhere I see in the AP wiki
I would suggest that ARK Elctronics be added here, however:https://ardupilot.org/ardupilot/docs/common-stores.html

Hwurzburg (changes requested):

common/source/docs/common-non-gps-navigation-landing-page.rst
patrikpordi: Maybe if I would add it below the note I have attached?
Screenshot 2024-11-07 at 10-27-05 Non-GPS Navigation — Copter documentation

If not I'll remove it. I'll add ARK as a shop as well after your feedback!

Hwurzburg: I think it should just be removed....thanks

patrikpordi:
@Hwurzburg I have removed the IMUs and added ARK to the stores!
Thanks in advance for the merge!

Hwurzburg:
LGTM....rebased and merging

PR #6371, opened 2024-11-06
Document interactive development of mavproxy

Ryanf55:
* Use a venv and develop mode from setup.py

Relates to https://github.com/ArduPilot/MAVProxy/issues/1215#issuecomment-2453643372


mavproxy/source/docs/development/mavdevenvlinux.rst
```diff
@@ -13,20 +13,43 @@ The pip-installed MAVProxy will need to uninstalled (if already installed) to pr
...
+
+   .. group-tab:: User Python Environment
+
+    After making the desired changes, MAVProxy is required to be installed
+    (any change to the modules won't work otherwise). This needs to happen after any
+    changes to the source code. This can be done by:
+
+      .. code-block:: bash
+
+        python3 setup.py build install --user
```
peterbarker, line 36: Both tridge and myself have switched to `python3 -m pip install .` for this step.

mavproxy/source/docs/development/mavdevenvlinux.rst
```diff
@@ -13,20 +13,43 @@ The pip-installed MAVProxy will need to uninstalled (if already installed) to pr
...
+   .. group-tab:: Virtual Python Environment
+
+    Using a virtual environment allows quicker editing.
+    If you make code changes, just re-run MAVProxy.
+
+      .. code-block:: bash
+
+        python3 -m venv .venv
+        source .venv/bin/activate
+        python3 setup.py develop
```
peterbarker, line 48: My understanding is that running `setup.py` directly is very much deprecated.

peterbarker (commented):
Wouldn't a user ordinarily be in the ArduPilot venv here?  It seems you're creating a new venv as part of these instructions...

mavproxy/source/docs/development/mavdevenvlinux.rst
Ryanf55: Good info. Seems like `wx` is not available in the virtual environment, but that host workflow works.

mavproxy/source/docs/development/mavdevenvlinux.rst
Ryanf55: Yep, https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead

Ryanf55:
> Wouldn't a user ordinarily be in the ArduPilot venv here? It seems you're creating a new venv as part of these instructions...

These instructions are for an  isolated env. If the instructions are making an assumption you use the ardupilot rep's venv, then maybe it's worth adding that as a prerequisite. I've found it's not required to share the same environment for both repos. 

Also, I saw that MAVProxy would be installed twice. 
https://github.com/ArduPilot/ardupilot/blob/c0ee3b22163da1b77fcc7362aa6e0e3a8c6a9f8d/Tools/environment_install/install-prereqs-ubuntu.sh#L166

Ryanf55:
How  are we supposed to install the wsproto dependency of pymavlink?

https://github.com/ArduPilot/pymavlink/pull/967/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R10

* ardupilot/ools/scripts/install-prereqs-ubuntu.sh doesn't have it
* The setup.py file of pymavlink doesn't have it, only the requirements.txt
* We don't ever tell devs to install pymavlink's `requirements.txt`

Ryanf55:
@peterbarker  Can you explain why the ardupilot virtual environment is necessary? The docs don't seem to agree. Here's why:

 The ardupilot setup is explained as the setup for building the Ardupilot code with waf:
image

Which prompts you on adding the venv to your bashrc:
https://github.com/ArduPilot/ardupilot/blob/4d75b4477550e3fccb6a1917dd8a49594d3d2c9b/Tools/environment_install/install-prereqs-ubuntu.sh#L394
And is not recommended in the wiki which option to chose

If you just start with MAVProxy, which does not even reference the Ardupilot environment install
image

Then, why would a developer know that the have to install MAVProxy INSIDE of the ardupilot repo's virtual environment? 

If this is actually true, then shouldn't it be documented that this is the only supported workflow ?

peterbarker:
OK, sorry, I wasn't away you'd linked the MAVProxy Wiki there.

So someone wants to develop ArduPilot will get MAVProxy as part of the environment install script - but not give them a development environment.

You're trying to add words around developing MAVProxy inside the ArduPilot Wiki here - and they are conflicting with words over in the ArduPilot docco.... perhaps the MAVProxy Wiki should acknowledge you may want to work within the ArduPilot venv.  The MAVProxy docs you link to need updating anyway - they won't work  well without a venv now, and that's not called out.

Hwurzburg:
@Ryanf55 @peterbarker still awaiting you and PeterB to resolve and he approve...

PR #6380, opened 2024-11-10
removed broken roadmap links

RamReso:
I removed the links to the Roadmap from the navigation menu, because they no longer worked.

One could replace it with the [project overviews page](https://github.com/orgs/ArduPilot/projects), but I wasn't sure what is the right choice there.


rmackay9:
yes, I noticed the link was broken as well and I can't find where the older project has gone to.

Hwurzburg:
> I removed the links to the Roadmap from the navigation menu, because they no longer worked.
> 
> One could replace it with the [project overviews page](https://github.com/orgs/ArduPilot/projects), but I wasn't sure what is the right choice there.

I think this is best for now....merging

PR #6387, opened 2024-11-14
Add Copter and Rover records pages

peterbarker:


Hwurzburg:
these need to be linked somewhere (I suggest in the Appendix) and have TOClist entries....this is a build error wiki CI does not flag!

peterbarker:
> these need to be linked somewhere (I suggest in the Appendix) and have TOClist entries....this is a build error wiki CI does not flag!

Yep, I do realise these are unlinked - just looking for guidance on that...

Hwurzburg:
I suggest adding a Records topic here https://ardupilot.org/copter/docs/common-appendix.html... have plane,copter,and rover links to those pages on it under a toclist header killing two birds with one stone

peterbarker:
> I suggest adding a Records topic here https://ardupilot.org/copter/docs/common-appendix.html... have plane,copter,and rover links to those pages on it under a toclist header killing two birds with one stone

I've added it to the toctree for each of Rover, Plane and Copter.

I haven't put it into the appendix - doesn't look like we have precedent for vehicle-specific stuff in there.

Hwurzburg:
@peterbarker I really want to keep only required info in left menus, so to avoid list-creep I moved the records info to the appendix
unless you object, I will merge tomorrow after CI and you get a chance to look tonight

PR #6395, opened 2024-11-18
Update Partner logos

rishabsingh3003:


Hwurzburg:
why is botlab added at top of list instead at bottom like other new partners? or was this indeed to indicate chronology of joining?

rishabsingh3003:
@Hwurzburg, botlab left the Partners Program for a couple of months and then joined back. I treated this as a continuation of their membership rather than starting a new application. The logo was added where it previously belonged.

PR #6402, opened 2024-11-22
Common: add voxl2 page

rmackay9:
This adds details on how to setup the VOXL 2 for use with ArduPilot.  There is some copy-and-paste from VOXL 1 and I'm sure some details are incorrect but it's a good start


Hwurzburg:
lgtm...thanks

rmackay9:
Thanks very much!

PR #6404, opened 2024-11-22
Update common-stores.rst：Add Hardware Vendor Pogo-Tech.

dmokel:
Add Hardware Vendor Pogo-Tech.


rmackay9:
Sorry about this but we actually only add links for Partner companies.  https://ardupilot.org/copter/docs/common-partners-program.html

dmokel:
> Sorry about this but we actually only add links for Partner companies. https://ardupilot.org/copter/docs/common-partners-program.html
> 
> @Hwurzburg could you revert this for us?

All right, so thanks for your reply, maybe we can apply to become a partner of ardupilot in the near future.

PR #6416, opened 2024-12-05
Common: update Siyi and Topotek with real-time video via ethernet info

rmackay9:
This primarily adds info on the RTSP streams users can use with Siyi and Topotek camera gimbals to view the live video.

I've also made some drive-by changes including:

1. re-ordering the Siyi models so A8 is at the front and the ZT30 is at the end
2. added links to the camera control wiki page from the Siyi and Topotek pages
3. minor formatting changes to the common-ethernet-vehicle page


Hwurzburg:
LGTM

rmackay9:
Thanks!

PR #6417, opened 2024-12-06
add holybro GPS & update links

vincentpoont2:
add holybro GPS, PM, update links & photos


Hwurzburg:
thanks...squashed and rebased..merging

PR #6422, opened 2024-12-08
Copter: slung-payload page gets ready-to-use device link

rmackay9:
This adds a link to the Copter slung-payload wiki page to the JapanDrones device shown in the videos.

BTW, Japan Drones is my company so I have a slight conflict of interest here in that I will profit from customers who buy the device.  Still, if competitors appear we can certainly add them too and there's already some details on the page about what is required for individuals to build one of their own:

`This payload autopilot must transmit the GLOBAL_POSITION_INT mavlink message to the autopilot at 10hz using a system id that is different from the autopilot’s.`


Hwurzburg:
LGTM...thanks

PR #6432, opened 2024-12-09
4.7:Script heap auto resizing

Hwurzburg:


tridge (approved):

common/source/docs/common-lua-scripts.rst
```diff
@@ -43,6 +43,8 @@ Getting Started
...
 
+.. note:: upon script load, if insufficient memory has been allocated, an error message will be sent to the GCS. Unfortunately, eliminating this by changing the :ref:`SCR_HEAP_SIZE <SCR_HEAP_SIZE>` does not guarantee that while running after aming, the LUA script could require even more and fail. In order to mitigate this, ArduPilot in versions 4.7 and later, will automatically try to increase the allocated memory and print a warning message to the GCS to allow the user to resize before the next arm, but if the additional memory is NOT available, the script will still fail.
```
tpwrules, line 46:
```suggestion
.. note:: upon script load, if insufficient memory has been allocated, an error message will be sent to the GCS. Unfortunately, eliminating this by changing the :ref:`SCR_HEAP_SIZE <SCR_HEAP_SIZE>` does not guarantee that while running after arming, the LUA script could require even more and fail. In order to mitigate this, ArduPilot in versions 4.6 and later, will automatically try to increase the allocated memory if the heap size is insufficient and print a warning message to the GCS to allow the user to resize before the next arm, but if the additional memory is NOT available, the script will still fail.
```

The change has been backported.

PR #6438, opened 2024-12-10
Common: benewake link fixes

rmackay9:
This fixes some of the links on the two Benewake lidar pages and also removes the out-of-date reference to the version that supports them


Hwurzburg:
LGTM...tnx

PR #6439, opened 2024-12-10
4.7:Add Quadplane QwikTune

Hwurzburg:
@MichelleRos I created shell for the QwikTune PR and highlighted in bold sections that need your input...if you could squash your added commits to this to a single commit, it would save me a step later...thanks

### DIFF
--- a/plane/source/docs/quicktune.rst
+++ b/plane/source/docs/quicktune.rst
@@ -1,0 +1,60 @@
+.. _quicktune:
+
+==============
+VTOL QUICKTUNE
+==============
+
+.. note: QUICKTUNE can only be used in QHOVER,QLOITER, or GUIDED VTOL modes.
+
+ArduPlane provides a quick means of obtaining a good tune for Quadplanes in VTOL modes. The process slowly increases the relevant gains until it detects an oscillation.  It then reduces the gains by 60% and moves onto the next gain. Once all the gains have been tuned the tune completes and the user can decide to save or discard the new gains.
+
+.. note:: be sure that you have prepared the vehicle for tuning by setting up parameters discussed here: :ref:`quadplane-vtol-tuning-process` Steps 1 to 11 and then use this mechanism in Step 12. Also to assure the best tune, setup the noise notch filtering, see :ref:`common-imu-notch-filtering`. You may run QUICKTUNE without this step to obtain initial fine tune, see :ref:`ac_rollpitchtuning` if the vehicle cant do an initial hover stably. Then setup the filters and retune for best results.
+
+The script attempts to automatically tune the P/I and D gain of each axis enabled for tuning when an two or three position RC switch, whose ``RCx_OPTION`` is "181", is moved to the "tune" position. It will, by default, also adjust some PID loop filter values based on the :ref:`INS_GYRO_FILTER<INS_GYRO_FILTER>` setting.
+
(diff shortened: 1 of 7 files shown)


MichelleRos (changes requested):
This seems to be adding a new page called "qwiktune" to describe the usage of the quicktune c++ in Plane. I don't think this is the right way to go. We don't want to be maintaining two separate pages.

What I was suggesting on Discord was that we make the current quicktune page in Copter be available under Plane.

Hwurzburg:
Copter uses the LUA script not the C++ code...changing the C++ code does not impact the Copter page and vice versa...IF Copter ever adopts QwikTune then it might be worthwhile...

however, if you want to create a common page (common-quicktune.rst) the differences in text (which will be significant as will the other link changes throughout two wiki's tuning sections) can be isolated to a particular wiki with [site wiki="plane"] [/site]  or [site wiki="copter"] [/site] delimiters around wiki site specific wording, params, links, etc.

Hwurzburg:
@MichelleRos since this is back ported to 4.6, there is some urgency to complete and merge this.....can you contribute?

MichelleRos:
@Hwurzburg The intent AFAIK was to keep the Lua and C++ implementations the same (well, actually the intent was to get rid of the Lua implementation and just have the C++ implementation for both copter and plane instead of doubling our maintenance workload, but certain people refused to allow that).

As for the documentation, I don't really know whether there would be enough differences between the two to make it easier to do two separate pages, but that'd really be a question for @tridge as quicktune is really his thing.

Hwurzburg:
I have updated and will merge once Tridge approves

plane/source/docs/quicktune.rst
```diff
@@ -0,0 +1,60 @@
...
+.. note:: RC AUX switches only operate on transitions of the switch. Their position at boot does nothing, so if a switch is "high" at boot, in order to get the "high" function, it must be moved to another position and then back "high".
+
+With a two position switch, "high" is the "tune" position, and "low" stops/reverts tuning. In order to save parameters, you must have set the :ref:`QWIK_AUTO_SAVE<QWIK_AUTO_SAVE>` parameter or use a GCS AUX switch facility.
+
+Tuning Flight
+=============
+- Wait for a calm day and go to an open area with good GPS reception
+- Use an OSD, TX OSD (using :ref:`Yaapu <common-frsky-yaapu>`), or GCS. This is where output from the tune will appear
+- Move the RC switch to the low position OR push MP's Aux Function's "Low" button
+- Arm and takeoff in Loiter mode and climb to a height of about 3m
```
tridge, line 39: QLOITER not Loiter

plane/source/docs/quicktune.rst
```diff
@@ -0,0 +1,60 @@
...
+- Begin the tune by moving the RC switch to the "tune" position OR push MP's Aux Function's "Mid" button
+- Monitor the progress of the tune using the GCS's Messages tab
+- If necessary reposition the vehicle using the RC transmitter.  This will temporarily pause tuning and restore the original gains.  Tuning will resume a few seconds after the RC sticks are returned to their center position
```
tridge, line 42: it does not restore the gains while repositioning

plane/source/docs/quicktune.rst
```diff
@@ -0,0 +1,60 @@
...
+- If the vehicle begins oscillating violently cancel the tune by moving the RC switch to the low position OR push MP's Aux Function's "Low" button
```
tridge, line 43: the tune will automatically abort if an angle error over QWIK_ANGLE_MAX is detected

tridge (changes requested):

PR #6441, opened 2024-12-11
Common: CADDXFPV gimbal setup

rmackay9:
This is slightly premature considering we don't have [the driver](https://github.com/ArduPilot/ardupilot/pull/28373) completely working yet but I would like to know if @Hwurzburg thinks the wiring is correct or do I have the RX and TX lines backwards?


Hwurzburg:
I think that they are backwards

rmackay9:
@Hwurzburg, or make use of the SERIALx_OPTIONS to flip them in software.  I'll continue working on this today as well so surely we will get to the bottom of it

rmackay9:
@Hwurzburg,

I think the diagram included in this PR has the TX and RX mapped correctly.  I've asked CADDX so hopefully they will tell us directly but we have this picture in the manual.

image

Then when it discusses the MAVLink interface (which we're not using but surely the pin mapping is the same) it says UART_RX and UART_TX should be mapped to the autopilot's TX and RX **respectively**.

image

Hwurzburg:
I just used the pin marked RX on the gimbal to the autopilot TX since its one way for control....so I had that correct, but still no communication

rmackay9:
Hi @Hwurzburg,

How do you feel about merging this?  I'd like to get it into the wiki even though 4.7 is not out because it gives alpha/beta testers the info they need to test and use it.

Hwurzburg:
fixed up and corrected a build error (CI does not catch TOC errors)....merging....you can refer to the wiki HTML page in your blog post once it builds

rmackay9:
@Hwurzburg thanks so much!

PR #6448, opened 2024-12-12
Update the Partners Program wiki

rishabsingh3003:


common/source/docs/common-partners-program.rst
```diff
@@ -70,11 +70,15 @@ The first step in becoming a partner is to email the partners program
 coordinator at `partners@ardupilot.org
 <mailto:partners@ardupilot.org>`__ to start the application process.
 
+.. note::
```
rmackay9, line 73: It's up to you but I would have left the existing line in place instead of making it a special note.  And then I'd have added a line requiring that they provide a link to their website.  The implication being that if they don't have one they can't be partners ('cuz it's a requirement).

PR #6449, opened 2024-12-12
Common: Siyi ethernet control details added

rmackay9:
This makes a few mostly unrelated changes:

1. Adds a missing MUPilot image
2. Ethernet Vehicles page gets links and the white space around images is reduced
3. Network Interface page's title is shortened to remove "ArduPilot".  I understand that having the word "ArduPilot" there conveys that we're talking about the architecture of the whole system but I don't think we really need the word anyway
4. The "Ethernet Adapters" page's build-server-PPP image shows the "PPP networking" as checked.  E.g how it should be rather than how it might be before the user checks it
5. Added some details to the Siyi page to clarify how to setup control over ethernet.  There's definitely some duplication compared to the "Ethernet Vehicles" page but I think it's justified to have duplicate if it makes it easier for the user by reducing the need for them to jump around the wiki.  I've missed some details however like setting the gateway address but maybe this is OK?

I've tested locally and it looks OK


common/source/docs/common-siyi-zr10-gimbal.rst
```diff
@@ -81,7 +81,20 @@ Ethernet Connectivity
 
 Instead of using a UART serial connection to control the gimbal, an Ethernet virtual serial connection may also be used, as well as transmitting a video stream from the camera.
 
-An example of the setup of the gimbal in a networked ArduPilot vehicle system is detailed in :ref:`common-ethernet-vehicle`
+An example of the setup of the gimbal in a networked ArduPilot vehicle system is detailed in :ref:`common-ethernet-vehicle` but in short, to control the gimbal over Ethernet
```
Hwurzburg, line 84:
```suggestion
An example of the setup of the gimbal in a networked ArduPilot vehicle system is detailed in :ref:`common-ethernet-vehicle` but in short, to control the gimbal over Ethernet:
```

Hwurzburg (changes requested):

common/source/docs/common-siyi-zr10-gimbal.rst
Hwurzburg: such a nit, I am merging anyway

rmackay9: Thanks, I actually debated with myself for a moment whether to put a colon on the end or not.  :-)

PR #6453, opened 2024-12-13
common: restore pitch reversal language

tpwrules:
Primarily restores the pitch reversal language introduced in #5943 that accidentally got deleted by #6367 .

Also smooth out the notes about only reversing using the transmitter. Talked with Randy and it should be safe these days; I've done it myself for years. It's also how QGroundControl does it using the auto calibration routine.

There are still some reasons to do it in the transmitter. See also [this discussion](https://github.com/ArduPilot/ardupilot_wiki/pull/5943#discussion_r1594145357).


Hwurzburg:
LGTM

rmackay9 (approved):
Looks good to me.

PR #6455, opened 2024-12-14
Plane: Documented TKOFF_THR_IDLE

Georacer:
... according to the changes in #28351 

Note that the original description mentioning an options bit is no longer valid.


Hwurzburg:
lgtm...thanks

PR #6460, opened 2024-12-16
4.7:do-return-path-start

Hwurzburg:


IamPete1 (approved):
Looks good, Thanks!

common/source/docs/common-do-return-path-start.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. image:: ../../../images/do-return-path-start.jpg
+    :target: ../_images/do-return-path-start.jpg
+
+Imagine above that the vehicle takes off, flies to WP2 and 3, and if one were to fly up and look at the lake but then have a failsafe, the vehicle would fly directly back to the DO_LAND_START tag, crashing thru the tree, exclusion fence, etc. But if a DO_RETURN_PATH_START mission item is inserted after WP 3, then the vehicle would return to the mission path, proceed to the waypoint after the DO_LAND_START tag and LAND safely.
+
+Multiple DO_RETURN_PATH_START items can be placed in the mission sequence to tailor the return. It is not necessary that the sequence include a :ref:`DO_LAND_START <common-do-land-start>`, but if it does, the return to the mission will occur after the DO_RETURN_PATH_START mission command but before the :ref:`DO_LAND_START <common-do-land-start>`.
```
timtuxworth, line 16: It's not clear what this is saying. Rather than just tacking "but if it does" on the previous sentence, perhaps a new sentence that explains better what happens if there is a DO_LAND_START and DO_RETURN_PATH_START. I'd try to write something, but I really don't understand what would happen.

common/source/docs/common-do-return-path-start.rst
```diff
@@ -0,0 +1,25 @@
...
+
+[site wiki="plane"]
+In Plane, set :ref:`RTL_AUTOLAND<RTL_AUTOLAND>` to 4 to use this.
+[/site]
+[site wiki="copter"]
+In Copter, set any failsafe action to DO_LAND_START, whether or not one exists in the mission.
```
timtuxworth, line 22: So if you do this, does that mean the DO_RETURN_PATH_START would be triggered if the pilot manually switches to RTL?

common/source/docs/common-do-return-path-start.rst
```diff
@@ -0,0 +1,25 @@
...
+
+The DO_RETURN_PATH_START mission command can be used to mark a place or places in the mission sequence, before the :ref:`DO_LAND_START <common-do-land-start>`, that the vehicle should directly rejoin the mission path at the closet point on the mission leading toward the :ref:`DO_LAND_START <common-do-land-start>` sequence, rather than directly to the :ref:`DO_LAND_START <common-do-land-start>`. As shown below:
+
+.. image:: ../../../images/do-return-path-start.jpg
+    :target: ../_images/do-return-path-start.jpg
+
+Imagine above that the vehicle takes off, flies to WP2 and 3, and if one were to fly up and look at the lake but then have a failsafe, the vehicle would fly directly back to the DO_LAND_START tag, crashing thru the tree, exclusion fence, etc. But if a DO_RETURN_PATH_START mission item is inserted after WP 3, then the vehicle would return to the mission path, proceed to the waypoint after the DO_LAND_START tag and LAND safely.
```
timtuxworth, line 14: Will it return to the mission path by going back to WP3? By going to the DO_LAND_START (can that have a location lat/long?) or will it go to WP4? Or will it try to join the mission path at the nearest point between WP3 and WP4?

common/source/docs/common-do-return-path-start.rst
timtuxworth: What if there are multiple DO_LAND_PATH_START tags and a DO_LAND_START, does it pick the closest one like it would do if there are multiple DO_LAND_START tags?

timtuxworth (changes requested):
I have a real use case with a partner who needs to use this, so my questions are specifically to clarify some things I need to try to explain to them.

images/do-return-path-start.jpg
timtuxworth, line 1: Typo in "DO_RETUNR_PATH_START"

common/source/docs/common-do-return-path-start.rst
IamPete1: No, the logic remains the same, it re joins at the closest path. Just there would be multiple paths that it will check.

common/source/docs/common-do-return-path-start.rst
IamPete1: In plane it will, not on copter.

common/source/docs/common-do-return-path-start.rst
IamPete1: It will rejoin at the closest leg between the return path start and the do land start. So if it failsafed at wp3 it would head to wp4.

PR #6465, opened 2024-12-18
Dev: mavlink interface section gets nongps position estimation page

rmackay9:
This adds info for developers on how to get non-gps position estimation (aka non-gps navigation) working

I only describe the preferred message, ODOMETRY.  We could add the rest but this is the most important.


Hwurzburg:
tnx

PR #6468, opened 2024-12-18
fix CAN_D1_UC_SRV_BM naming and example value

Huibean:
fix CAN_D1_UC_SRV_BM naming and example value
from CAN_D1_UC_ESC_SRV_BM to CAN_D1_UC_SRV_BM
https://ardupilot.org/copter/docs/parameters.html#can-d1-uc-srv-bm


Hwurzburg:
tnx...LGTM

PR #6469, opened 2024-12-18
Common: Xacti camera model name and link fixes

rmackay9:
Just some minor fixes I found while testing out new firmware from the manufacturer.

I've tested locally and it looks OK


Hwurzburg:
tnx...

PR #6474, opened 2024-12-19
Copter: fix ground oscillation formatting

rmackay9:
This fixes the bullet points on the ground oscillation page


Hwurzburg:
thanks...probably more of these...think the latest Sphinx changed some things
maybe Peter can help me with a REGX search

PR #6475, opened 2024-12-19
Update common-marvelmind.rst

smoker771:
We propose these changes, because previous version of the article is not actual, the recommendations were obsolete and don't work correctly with last versions of the Ardupilot and Marvelmind software. The proposed version includes link to tested sets of settings which we will continue update in future for new versions of software

### DIFF
--- a/common/source/docs/common-marvelmind.rst
+++ b/common/source/docs/common-marvelmind.rst
@@ -17,43 +17,29 @@ See the `user manual <https://marvelmind.com/pics/marvelmind_navigation_system_m
 Required Hardware
 =================
 
-* `Marvelmind Starter Set <https://marvelmind.com/product/starter-set-hw-v4-9-plastic-housing/>`__
+* `Marvelmind Starter Set <https://marvelmind.com/product/starter-set-super-mp-3d/>`__
 
 Marvelmind Settings
 =====================
 
-- Ensure the Marvelmind is firmware version 5.77 (or higher) and enable the 'Raw Distances Data' in the Dashboard
-- Set the stationary beacon with the lowest address as the origin in your dashboard.
+- Ensure the Marvelmind is firmware version 7.910 (or higher)
 - The height of every stationary beacon, from the ground, must be set on the dashboard.
-- The hedge should be set to output Marvelmind Protocol at baud rate of 115200.
+- The hedge should be set to output NMEA 0183 Protocol at baud rate of 115200.
 
 Connecting to an autopilot
 ==========================
 
-- Set :ref:`EK3_SRC1_POSXY<EK3_SRC1_POSXY>` to 4 (Beacon)
-- Set :ref:`EK3_SRC1_VELXY<EK3_SRC1_VELXY>` to 0 (None)
-- Set :ref:`EK3_SRC1_POSZ<EK3_SRC1_POSZ>` to 4 (Beacon)
-- Set :ref:`EK3_SRC1_VELZ<EK3_SRC1_VELZ>` to 0 (None)
-- Set :ref:`EK3_SRC1_YAW<EK3_SRC1_YAW>` to 1 (Compass)
-- Set :ref:`BCN_TYPE <BCN_TYPE>` to 2 (means using Marvelmind system)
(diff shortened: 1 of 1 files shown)


common/source/docs/common-marvelmind.rst
```diff
@@ -17,43 +17,29 @@ See the `user manual <https://marvelmind.com/pics/marvelmind_navigation_system_m
...
 - Connect the Autopilot to a ground station
 - Click on 'freeze map' on the dashboard
-- After about 1 minute the vehicle's position should jump to the latitude, longitude you input during the configuration step (above)
 - Check that the vehicle's position is relatively stable (i.e. moving around less than one meter)
 - Walk the vehicle around between the beacons and ensure that its position on the map updates correctly
 
-DataFlash logging
-=================
 
-The distance to the beacons can be found in the dataflash log's BCN message's D0, D1, D2, D3 fields.
```
Hwurzburg, line 59: why doesnt the beacon logging work like it did before?

smoker771: New version of the article recommends to setup NMEA0183 protocol mode. Distances to the beacons are not available in this mode. Previous version was based on Marvelmind proprietary protocol, this protocol supports distances. But Marvelmind protocol seems to be not working with Ardupilot in several last versions of software, this was a reason for the update of the article

Hwurzburg: you should raise an issue about that so that it gets fixed

rmackay9:
This is a very different setup and means that the marvelmind acts like a GPS which is OK but removes the ability to do GPS-to-NonGPS transitions.

I thought @amilcarlucas knew the most about the marvelmind but I see that @peterbarker actually [wrote the driver here](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Beacon/AP_Beacon_Marvelmind.h)

@smoker771 do you work at Marvelmind?

smoker771:
Yes, I'm working in Marvelmind. 
We really appreciate this article. This article was valid some time ago, but now it is not valid for some reasons. Many Ardupilot users try to use it without success, ask us to help and we recommend switching to NMEA mode. Reason of changes to the article was to help Ardupilot users integrate with Marvelmind system  successfully. 
Thank you

rmackay9 (approved):

common/source/docs/common-marvelmind.rst
```diff
@@ -17,43 +17,29 @@ See the `user manual <https://marvelmind.com/pics/marvelmind_navigation_system_m
...
-- Set :ref:`BCN_ORIENT_YAW <BCN_ORIENT_YAW>` to the heading from the origin beacon to the 2nd beacon.  One way to capture this value is to stand at the origin holding the vehicle so that it's nose points towards the second beacon.  Read the vehicle's heading from the HUD and enter this value into :ref:`BCN_ORIENT_YAW <BCN_ORIENT_YAW>`
-- Set :ref:`GPS1_TYPE <GPS1_TYPE>` to 0 to disable the GPS
-- Set :ref:`ARMING_CHECK <ARMING_CHECK>` to -9 to disable the GPS arming check
-- Set :ref:`SERIAL1_BAUD <SERIAL1_BAUD>` to 115 to set telemetry1's baud rate to 115200 (if using Telem2 set :ref:`SERIAL2_BAUD <SERIAL2_BAUD>` instead)
-- Set :ref:`SERIAL1_PROTOCOL <SERIAL1_PROTOCOL>` to 13 to enable receiving data from the marvelmind (If using Telem2 set :ref:`SERIAL2_PROTOCOL <SERIAL2_PROTOCOL>` instead)
-- Set :ref:`BRD_SER1_RTSCTS <BRD_SER1_RTSCTS>` to 0 to ensure telem1 does not use flow control (If using Telem2 set :ref:`BRD_SER2_RTSCTS <BRD_SER2_RTSCTS>` instead, not required if using Serial4/5)
+- Here is a package with tested param files for the Autopilot and Marvelmind hedge: <https://marvelmind.com/pics/marvelmind_ardupilot_settings.zip>
```
rmackay9, line 32: I've approved but this line looks a little odd.  I think there should be some single quotes around it.

I also think that we probably need to set the [GPS1_TYPE](https://ardupilot.org/copter/docs/parameters.html#gps1-type-gps-type) = 5 (NMEA)

smoker771: Thank you for approving. 
We included all settings into the "param" files available by the link and "GPS_TYPE = 5" is also set there.
I will make an additional pull request to fix this odd line and some small changes in the design.

PR #6478, opened 2024-12-20
Common: loading firmware via SD card made easier to find

rmackay9:
This makes the page describing how to load firmware via the SD card slightly easier to find.  This also includes a drive-by change to the page's title.


Hwurzburg:
not sure why its hard to find...its in a bold note right at the top of the page?

rmackay9:
@Hwurzburg,

Ah, right. I didn't see that.  I think a section on that page is actually better than a note though isn't it?

Hwurzburg:
meh, actually more likely to be missed....having both is fine by me

PR #6481, opened 2024-12-21
Common: add ADTi setup page

rmackay9:
This adds a basic setup page for the ADTi cameras and gimbals.

There are likely improvements we can make to the integration but this is a start.


Hwurzburg:
lgtm...tnx

PR #6484, opened 2024-12-24
dev: Document missing plane guided commands

Ryanf55:
Document which guided commands are supported: https://github.com/ArduPilot/ardupilot/blob/5a8bc161709aba97a8b7a6da3ce0ec1a49df2718/ArduPlane/GCS_Mavlink.cpp#L897

The wiki is missing a few.

Guided takeoff is not yet supported.


Hwurzburg:
lgtm...tnx

Issue #6490, opened 2024-12-27
Continuing a Mission After Landing (RTL updates)


prathamEndu:
In the video it is told that while continuing, you can also set RTL, but if you do, it will just go back to the position where last arming was done.

[Continuing a Mission After Landing](https://ardupilot.org/copter/docs/common-continue-mission.html) - Describe the issue/suggestion and improve the title. Please keep a link to the original article if relevant.

rmackay9:
Hi @prathamEndu,

Thanks for the note.  If the user sets the home position then the home position won't be reset when the vehicle disarms and re-arms so we probably need to add a comment on the page explaining this if it's not already explained in the video

PR #6494, opened 2024-12-28
add note to mission continue page

Hwurzburg:
addresses #6490


rmackay9:
Thanks for this although I think it's more likely that a user would set the home position from the GCS rather than adding a set-home command to the mission.  Mission Planner doesn't even display the do-set-home mission command.  Anyway, maybe I'll add a PR to fixup the wording and perhaps provide a link back to a user wiki page which shows how to set home

Hwurzburg:
Actually Mission Planner does indeed display DO_SET_HOME as a mission command, at least in Plane. But explaining that home can be set persistently thru disarms, either by right click context menu or mission commands would be a good addition under Mission Planning tab

PR #6495, opened 2024-12-29
Update index.html , included preload image in the head so that loadin…

Knowledgelocater:
…g image should pe optimised

Added Preload images in the head so that loading of image should be optimised.


Hwurzburg:
@khancyr ping

khancyr (changes requested):
Images are loaded only on need, so the only ones we should try to preload are 
slides_01_Ardupilot42.jpg, slides_01_Ardupilot42.webp, and ArduPilot-Motto.png as only those are visible on opening.

khancyr:
Need rebase and squash to have only 1 commit and it will be good !

PR #6496, opened 2024-12-30
dev: Document position target and attitude target support in Plane

Ryanf55:
image

Document the following undocumented interfaces:
* SET_POSITION_TARGET_GLOBAL_INT
* SET_POSITION_TARGET_LOCAL_NED
* SET_ATTITUDE_TARGET

The support in Plane is VERY different than copter. It would be a misnomer to say that plane supports the intent of these packets, hence the very specific notes on what's supported and how to use it. As an autonomy developer, this is one of the most frustrating things to have to handle to develop cross-platform autonomy.

As the support for Copter and Plane become more similar, then we can instead move this to a common wiki that's not vehicle-specific.


Hwurzburg:
tnx...

PR #6498, opened 2025-01-01
Added skyhub and voidray from Idrobots as companion computers

Karthikeyan-Idrobots:


yalexx:
Thank for opening the PR Karthink ! Lets build some robots :) https://idrobots.com/

rmackay9:
Hi @Karthikeyan-Idrobots, @yalexx,

the void-ray product doesn't seem to be available anymore so I think I will remove it from the list of companion computer hardware section.  Just a few lines below there is a link to the idrobotics SkyHub which is still valid.

Hope that's OK. Heres' the PR: https://github.com/ArduPilot/ardupilot_wiki/pull/7778

PR #6508, opened 2025-01-04
Plane: Document setup without GPS or airspeed sensor

Ryanf55:
Document how to set up ArduPlane to fly in FBWA without GPS or airspeed sensor. Document some known limitations.
This is for a minimal plane build to just use AP as a stabilizer.

### DIFF
--- a/common/source/docs/common-advanced-configuration.rst
+++ b/common/source/docs/common-advanced-configuration.rst
@@ -31,6 +31,7 @@ tuning options for the vehicle.
     Auxiliary Functions <common-auxiliary-functions>
     Battery Voltage Compensation <battery-voltage-compensation>
     Bootloader Update <common-bootloader-update>
+    Flying without GPS or Airspeed <no-gps-airspeed>
 [/site]
     BLHeli ESCs <common-blheli32-passthru>
     CAN Bus Setup <common-canbus-setup-advanced>
--- a/common/source/docs/common-non-gps-to-gps.rst
+++ b/common/source/docs/common-non-gps-to-gps.rst
@@ -11,6 +11,11 @@ ArduPilot 4.1 (and higher) support in-flight transitions between GPS and Non-GPS
 
 .. note:: Non-GPS navigation is available for all vehicles. However, it is not applicable to fast or high flying vehicles such as conventional Planes. QuadPlanes can utilize this when in VTOL operation and close to the ground, as when docking inside a hangar using 3D cameras or beacons.
 
+[site wiki="plane"]
+If you want to fly a Plane without a GPS or airspeed sensor installed, see :ref:`Flying without GPS or Airspeed <no-gps-or-airspeed>`.
+This wiki page covers flying with the sensor(s) installed and allowing the EKF to gracefully handle when GPS is lost.
+[/site]
+
 Setup
 -----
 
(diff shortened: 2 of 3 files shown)


Hwurzburg (changes requested):
an RC failsafe results in crash or fly-away

put a warning immediately after the header

Ryanf55:
> an RC failsafe results in crash or fly-away
> 
> put a warning immediately after the header

What do you think about recommending CIRCLE as a failsafe action. That won't fly away, and over flat area, the vehicle will likely land just fine.

Hwurzburg:
CIRCLE is not an option...only GLIDE (FBWA no throttle) is useable with no GPS

Ryanf55:
> an RC failsafe results in crash or fly-away



> CIRCLE is not an option...only GLIDE (FBWA no throttle) is useable with no GPS

From the wiki: `Circle mode is deliberately a very conservative mode, and doesn’t rely on GPS positioning as it is used when GPS fails`

And, the code:
```c++
#include "mode.h"
#include "Plane.h"

bool ModeCircle::_enter()
{
    // the altitude to circle at is taken from the current altitude
    plane.next_WP_loc.alt = plane.current_loc.alt;

    return true;
}

void ModeCircle::update()
{
    // we have no GPS installed and have lost radio contact
    // or we just want to fly around in a gentle circle w/o GPS,
    // holding altitude at the altitude we set when we
    // switched into the mode
    plane.nav_roll_cd  = plane.roll_limit_cd / 3;
    plane.update_load_factor();
    plane.calc_nav_pitch();
    plane.calc_throttle();
}
```

The default for FS_SHORT_ACTN is FS_ACTION_SHORT_BESTGUESS, which means circle if I was in FBWA. You can't be in Auto , Guided, or Loiter if you don't have a GPS or airspeed sensor. Users can also select `CIRCLE` as their FS_SHORT_ACTN.

For FS_LONG_ACTION, we could recommend FS_ACTION_LONG_GLIDE as that is the most likely to not total the aircraft if it doesn't know where it is.

Hwurzburg:
we should sitl the non gps case with defaults for FS actions (CIRCLE/RTL) with RC failure to be definitive on warning as to what will happen

Ryanf55:
```bash
./Tools/autotest/sim_vehicle.py -v Plane --console --map -w
param set STALL_PREVENTION 0
param set GPS1_TYPE 0
param set ARSPD_USE 0
param set AHRS_EKF_TYPE 0
param set ARMING_CHECK 0
map set showsimpos 1
# Move the map to follow the sim position.
mode fbwa
rc 2 2000
rc 3 1500
arm throttle
```

Wait till it gets to ~100m altitude, then trigger a failsafe.
```
rc 3 900
```

It performs the short failsafe CIRCLE, then goes to RTL and does a flyaway. Now, repeat the test with `param set FS_LONG_ACTN 2` (glide).

With `graph SERVO_OUTPUT_RAW.servo3_raw` you can see that it's commanding no throttle. Eventually it will land.

Do you agree that GLIDE is better behavior than a flyaway?

Hwurzburg:
yes....better

Ryanf55:
Alright, I added that and an obvious warning about flyaway risk. 

Before this wiki goes in, we either need to merge this:
https://github.com/ArduPilot/ardupilot/pull/29012

OR, tell people to use the Set Home in mission planner.

Ryanf55:
Yea, I'll document you need to disable arming checks to fly.

PR #6516, opened 2025-01-07
Setting up sitl on linux: Fix FlightGear example

phuesler:
In order for `sim_vehicle.py` to send data to FlightGear, one has to set the `enable-fgview` flag.


peterbarker:
Merged, thanks!

PR #6518, opened 2025-01-08
Common: Ethernet/Networking page and Ethernet adapter page minor improvements

rmackay9:
This is a collection of small changes to the Ethernet Setup and Ethernet Adapter pages.  I've broken them up into individual commits so that we can remove them if they're not to @Hwurzburg's fancy

Perhaps the most contentious change is the renaming of the "Network/Ethernet Setup" page to "Ethernet/Network Setup".  I thought we should stick with the "Ethernet" word first because that's what we generally call it (instead of the more general "Network" term).  This also makes the title match the landing page link

I've built these locally and they seem fine to me


common/source/docs/common-ethernet-adapters.rst
```diff
@@ -47,18 +47,8 @@ In addition, it may implement its own Web Browser interface for status and confi
 .. image:: ../../../images/PPP_web_server.jpg
     :target: ../_images/PPP_web_server.jpg
 
-Adapters
```
Hwurzburg, line 50: why is this being removed?

common/source/docs/common-advanced-configuration.rst
```diff
@@ -46,6 +46,7 @@ tuning options for the vehicle.
     EKF (Extended Kalman Filter) <common-apm-navigation-extended-kalman-filter-overview>
     EKF Affinity & Lane Switching <common-ek3-affinity-lane-switching>
     EKF Sources and Selection <common-ekf-sources>
+    Ethernet/Network Setup <common-network>
```
Hwurzburg, line 49: I am a bit fussed over this....the capability is not restricted to Enet...it includes PPP and could be easily extended to other physical media access types in the future using the same IP protocol second layer that exists

rmackay9: I figured but I think the thing is that users will look for "Ethernet" more than "Network".  If you really want it to remain though as Network we can remove the commit.

Hwurzburg: why dont we let @tridge decide ....

rmackay9: Hi @Hwurzburg, yes, sounds good to me.  He did write a large portion of this and I fully admit that I know little about networking. It's actually been a good learning experience to go through all this

rmackay9: @Hwurzburg,

just to try and convince you one more time I'd like to press my point a bit more..

.. the new name "Ethernet/Network Setup" has all the same words as the old name, "Network/Ethernet Setup" so it's really just a question of the order.  When we talk about the feature we nearly always say ".. Ethernet..", we rarely say "Network" or "Networking" so I think "Ethernet" is what people are going to look for.  I'm pretty confident that if we could secretly monitor how long it takes people to find the page we'd find that with it named "Ethernet.." many hundreds of people will find it a few seconds faster.  It also puts it higher on the page and matches the naming of the "Ethernet Adapters" peripherals page.  Within the camera setup pages the section is called "Ethernet Connectivity".

Anyway, in the end you're the wiki maintainer so if you want to stick with "Network/Ethernet Setup" we can go with that.  I'd prefer we make a decision and get the other fixes in sooner so I can move on with additional changes

rmackay9:
@Hwurzburg, you are so kind.  thank you!!

PR #6520, opened 2025-01-09
plane: Use  RCx_OPTION to enable soaring on a switch

Ryanf55:
Wiki for https://github.com/ArduPilot/ardupilot/pull/14815 that got lost along the way


Hwurzburg:
thanks

PR #6517, opened 2025-01-08
Common: topotek ethernet setup clarified

rmackay9:
This adds info on how to connect a Topotek gimbal via ethernet

I've tested this locally and it looks OK to me

PR #6534, opened 2025-01-13
Ethernet wiring fixes

rmackay9:
This fixes the Ethernet Connectivity section wiring diagrams for the Siyi and Topotek gimbal pages.

The original diagrams were added as part of https://github.com/ArduPilot/ardupilot_wiki/pull/6479 and https://github.com/ArduPilot/ardupilot_wiki/pull/6517

The original wiring worked but I now understand that the swapping of TX/RX pins is only required when we are directly connecting two devices.  When all the devices are connected via an Ethernet switch (as they are in the two diagrams) then we should not swap TX and RX.

I've re-tested the Topotek with the corrected wiring and it works fine. I haven't tested the Siyi wiring but from inspection I think it should be fine

I've build these changes locally and they look OK to me

BTW, I think this means we should also update the [ethernet-vehicle page's house-of-babel section](https://ardupilot.org/copter/docs/common-ethernet-vehicle.html#the-tower-of-babel-physical-wiring)


Hwurzburg:
we can discuss....not sure about the siyi...I think I had issues....because the SIYI mac did not have autoMDIX

PR #6537, opened 2025-01-14
Commont: Herelink Ethernet Connectivity

rmackay9:
This PR adds an Ethernet Connectivity section to the Herelink page mostly based off the [Ethernet Connected Vehicle Example](https://ardupilot.org/copter/docs/common-ethernet-vehicle.html) page with some changes including adding a wiring diagram

Also included is a drive-by change to the Siyi page so that the autopilot's default IP address is used which should slightly ease the setup for the user


Hwurzburg:
thanks

PR #6538, opened 2025-01-14
Use python3 -m pip install rather than pip3 install

peterbarker:
I have not addressed the use of "sudo" - that requires looking more closely at which sort of installation is being used.

This is also going to be out-of-date vis-a-vis venvs.

But pip3 vs python3 -m pip is an *important* time-saver.


Ryanf55 (approved):
This is a good improvement.

PR #6545, opened 2025-01-20
Common: topotek gets firmware update and image download instructions

rmackay9:
This adds instructions on how to update the firmware on a Topotek camera gimbal and also how images and videos can be downloaded.

I've tested this locally and after building the page looked OK to me


Hwurzburg:
thanks

PR #6562, opened 2025-01-23
common: specify sim_vehicle.py command-line arguments rather than SIT…

peterbarker:
…L command-line arguments

the latter might be useful in MissionPlanner, but most people will be using sim_vehicle.py for this


Hwurzburg:
LGTM

PR #6566, opened 2025-01-23
4.7:add WSL2 mirrored netwk SITL option

Hwurzburg:


dev/source/docs/sitl-on-windows-wsl.rst
```diff
@@ -28,6 +28,13 @@ Running SITL
       :target: ../_images/sitl-on-windows-wsl-vscode.png
       :width: 450px
 
+.. note:: if you have configured WSL2 for mirrored networking (ie using the``set networkingMode=mirrored`` command), then use
```
timtuxworth, line 31: Could we refer to this https://learn.microsoft.com/en-us/windows/wsl/wsl-config#main-wsl-settings @Hwurzburg ? It's not a command, it's a setting in the .wslconfig file in your Windows home directory (i.e. not in your WSL home directory).

andyp1per: Yes agreed

andyp1per (approved):
LGTM, but I think we should modify as per @timtuxworth

dev/source/docs/sitl-on-windows-wsl.rst
Hwurzburg: iti s a command, but can be made persistent by writing the config (not a newbee task), if we eventually agree its something anyone with WSL2 (at least a recent version since this isn't available in earlier releases....means people might be FORCED to update their WSL2...not sure that's good) should be doing....given the above I have my doubts

PR #6570, opened 2025-01-24
Copter: update auto mode and add advice re common problems

rmackay9:
This updates the Copter Auto mode page including adding some advice on common problems that users face.  One example of this kind of issue is [here in Discuss](https://discuss.ardupilot.org/t/scruve-logic-in-auto-mission-issue/128974)


Hwurzburg:
thanks

PR #6573, opened 2025-01-25
Dev: mavlink get-set-param page gets json link

rmackay9:
Continuing my quest to improve documentation for GCS developers.  This adds links to the parameter definition in json format

I've tested this locally and it looks OK.  This is a small change and I'm done updating the PR


Hwurzburg:
Thanks

PR #6576, opened 2025-01-26
Stop parameters page in Wiki from just disappearing

peterbarker:
From what I've gleaned from the logs on the Wiki server, it is possible for build_parameters.py to parse out 0 versions to build from its scanning of the firmware server's directory listing pages.

This leads us to removing the parameters.rst file but not copying in a special "versioning" replacement for it based on the master branch.

So.  Run update.sh with `-e` so that it doesn't just rumble on if there's an error.

Change build_parameters.py to exit with a non-zero exit code if there's ever an error in it.

Require at least one version/commit from each directory listing parsed from the firmware server.

Future enhancements here:
 - make build_parameters.py a library, call it from update.py
 - stop parsing directory listing of firmware server in build_binaries.py, grab the manifest file instead and interpret it
    - generate parameter metadata into (eg.) https://firmware.ardupilot.org/Blimp/latest/3DRControlZeroG/, download that in build_parameters.py rather than doing the checkout-each-version thing; manifest would reference that .json / .rst

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -305,6 +305,8 @@ def get_commit_dict(releases_parsed):
     for i in commits_and_codes:
         if commits_and_codes[i][0] != 'error':
             commite_and_codes_cleanned[i] = commits_and_codes[i]
+    if len(commite_and_codes_cleanned) == 0:
+        error("Expected at least one commit")
     return commite_and_codes_cleanned
 
 
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -509,4 +509,4 @@ generate_rst_files(commits_to_checkout_and_parse)
 generate_json(VEHICLES)
 move_results(VEHICLES)
 
-# sys.exit(error_count)
+sys.exit(error_count)
--- a/update.sh
+++ b/update.sh
@@ -132,20 +132,14 @@ find -name "parameters*rst" -delete # Clean possible built and cached parameters
 
 END_UPDATES=$(date +%s)
 
-DO_PARAM_VERSIONING=true
-if $DO_PARAM_VERSIONING; then
-    progress "Starting to build multiple parameters pages"
-    python3 build_parameters.py || {
-	progress "build_parameters.py failed"
-	exit 1
-    }
-    END_BUILD_MPARAMS=$(date +%s)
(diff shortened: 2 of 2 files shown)


build_parameters.py
```diff
@@ -305,6 +305,8 @@ def fetch_commit_hash(version_link, board, file):
     for i in commits_and_codes:
         if commits_and_codes[i][0] != 'error':
             commite_and_codes_cleanned[i] = commits_and_codes[i]
+    if len(commite_and_codes_cleanned) == 0:
```
timtuxworth, line 308: cleaned (spelling) - ok I see it's a pre-existing wierdness, but an opportunity to fix it? (and commite)

update.sh
```diff
@@ -1,9 +1,12 @@
 #!/bin/bash
 # check for changes in docs and run sphinx
 
+set -e
+set -x
+
```
timtuxworth, line 6: debug?

update.sh
```diff
@@ -68,30 +86,29 @@ test -n "$FORCEBUILD" || {
...
-            echo "$site.rst has changed"
+            progress "$site.rst has changed"
             cp new_logmessages/$site.rst old_logmessages/$site.rst
             changed=1
         fi
     done
 
-    [ $changed = 1 ] || exit 0
+    [ $changed = 1 ] || {
+	progress "Nothing changed; no rebuild required, exitting"
```
timtuxworth, line 101: exiting (spelling)

timtuxworth (changes requested):
Some spelling and minor cleanup.

build_parameters.py
peterbarker: Not really a good time to fix this sort of thing.  The variable needs a completely new name, not just a spelling fix :-)

... and I don't even know if "commite" comes from a language-other-than-English or is just a typo...

update.sh
peterbarker: No, I'm leaving it like this deliberately.  One of these is even called out in the commit messages...

update.sh
peterbarker: Fixed, thanks.

PR #6577, opened 2025-01-27
Add note on variable naming

Hwurzburg:


dev/source/docs/style-guide.rst
```diff
@@ -402,22 +402,23 @@ Functions that return a single physical value or variables that represent a phys
...
 
 **Right:**
 
 ::
 
-    uint16 get_distance_ned_cm() { ... };
-    uint16 get_distance_enu_m() { ... };
+    uint16_t get_distance_ned_cm() { ... };
     float position_neu_mm;
+    unit16_t course; (in degrees)
```
rmackay9, line 413: looking pretty good.  This "(in degrees)" part wouldn't compile I think so maybe this should be modified to be a comment?

dev/source/docs/style-guide.rst
```diff
@@ -402,22 +402,23 @@ Functions that return a single physical value or variables that represent a phys
     uint16 get_angle() { ... };
     float distance;
 
-Functions or variables that represent a value relative to a frame should be suffixed with the frame first, then with the physical unit (if any).
+Functions or variables that represent a value relative to a frame should be suffixed with the frame first, then with the physical unit (if different than SI units). SI suffixes are occasionally used to avoid confusion in modules with similar variables of different units,but only if necessary to avoid confusion.
```
peterbarker, line 405:
```suggestion
Functions or variables that represent a value relative to a frame should be suffixed with the frame first, then with the physical unit (if different than SI units). SI suffixes are occasionally used to avoid confusion in modules with similar variables of different units, but only if necessary to avoid confusion.
```

(space added after comma)

dev/source/docs/style-guide.rst
```diff
@@ -402,22 +402,23 @@ Functions that return a single physical value or variables that represent a phys
...
 
 **Right:**
 
 ::
 
-    uint16 get_distance_ned_cm() { ... };
-    uint16 get_distance_enu_m() { ... };
+    uint16_t get_distance_ned_cm() { ... };
     float position_neu_mm;
+    unit16_t course; //if in degrees
```
peterbarker, line 413: maybe `float`?

peterbarker (approved):

dev/source/docs/style-guide.rst
Hwurzburg: fixed

dev/source/docs/style-guide.rst
Hwurzburg: changed

PR #6588, opened 2025-01-30
dev: mavlink interface gets MAVFTP page

rmackay9:
This is a very bare bones page but I'd like to get something on the Dev wiki to at least help GCS developers get started.

If @peterbarker or other core devs have suggestions on what info we should give GCS developers that would be greatly appreciated!


Hwurzburg:
thanks

PR #6589, opened 2025-01-30
Common: benewake tfmini I2C clarified

rmackay9:
This clarifies how to connect the TFMini (and TFminiPlus) using I2C.  The information was already there to some extent but this adds a connection diagram and also separates the serial setup from the I2C which makes it easier for the reader (I think).


Hwurzburg:
thanks...was just waiting on CI

rmackay9:
Hi @Hwurzburg, thanks so much! sorry, I didn't mean to be pushy.  Thanks again!

PR #6590, opened 2025-01-31
Update ros2-over-ethernet.rst to fix issue with ip adresses of companion and ardupilot

nido009:
changed IP-Adresse of Ardupilot and Companion-Computer to match the config-parameters in the table below. stumbled upon this when trying to find out if the DDS_IP-Parameters shall have the IP of the Companion or the Ardupilot itself.

### DIFF
--- a/dev/source/docs/ros2-over-ethernet.rst
+++ b/dev/source/docs/ros2-over-ethernet.rst
@@ -36,8 +36,8 @@ Necessary parameters for static configuration
 This list of parameters is given as an example.
 If you had the following static IP addresses:
 
-* An autopilot has the IP address ``192.168.1.6``
-* A Computer running the MicroROS agent has the IP address ``192.168.1.5``
+* An autopilot has the IP address ``192.168.1.5``
+* A Computer running the MicroROS agent has the IP address ``192.168.1.6``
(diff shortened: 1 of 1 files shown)


rmackay9:
I'm probably misunderstanding but by default, AP autopilots should have the IP address of 192.168.144.14 I think.  In general if we can stick with the defaults that should make it easier for users to setup the system

Ryanf55:
Question - the defaults look to be set here in AP_Networking_config.h. 

The values supplied in the wiki was just a hypothetical from the addresses my DHCP server gave them.
```
// Default Static IP Address: 192.168.144.14
#ifndef AP_NETWORKING_DEFAULT_STATIC_IP_ADDR
#define AP_NETWORKING_DEFAULT_STATIC_IP_ADDR "192.168.144.14"
#endif
```

Maybe we should say :
```
* An autopilot has the IP address ``192.168.144.14``
* A Computer running the MicroROS agent has the IP address ``192.168.144.6``
* The MicroROS agent is running on port ``2019``
```
This would match the rest of the ethernet wikis that have now been written.

nido009:
closed, will create a new one matching ip addresses to documentation

PR #6596, opened 2025-02-04
Update ros2-over-ethernet.rst

nido009:
the text did not match the corresponding config parameters. this changes to ip-adresses to match the defaults look to be set here in AP_Networking_config.h.


rmackay9:
great, thanks for this.  We changed AP's default IP Address for 4.6 so this looks like a good idea to me

Ryanf55 (approved):

PR #6606, opened 2025-02-08
Updated TBS Lucid Pro graphic+ descripotion

kamik2:
- New graphic
- removed PC2/3 undefined note
- added the correct label for PC2/3 into the graphic


Hwurzburg:
this pinout on new image does not agree with the hardware definition file....esc telem is RX5, SBUS in is RX2, there is no UART7, GPS is on UART6

kamik2:
I'll check that tomorrow with the dev

kamik2:
got it - mine was for at32, yours is for stm chipset

PR #6607, opened 2025-02-10
Dev: add missing GSoC 2024 completed project

rmackay9:
We were missing one project from the list.  I've tested this locally and it looks OK to me


Hwurzburg:
thanks

PR #6608, opened 2025-02-11
Common: am32 dronecan gets firmware update instructions

rmackay9:
This adds an ESC firmware Updates section to the AM32 dronecan page.

I've tested the instructions and also build the page locally and it looks OK to me


rmackay9:
I've updated this to also include a simple wiring diagram

Hwurzburg:
@rmackay9 still needs param changed I think

rmackay9:
@Hwurzburg, thanks for that, I've fixed the incorrect parameter reference now

PR #6609, opened 2025-02-11
Dev: debugging on stm32 gets vimdrones debugger

rmackay9:
This fixes up the STM32 debugger hardware list including adding the VimDrones VM Probe.

I've also made a couple of non-critical drive-by changes to wording.

FYI @Huibean


Hwurzburg:
thanks...

PR #6618, opened 2025-02-14
dev: Fix incorrect node name

Ryanf55:
Fixes user question here: https://discord.com/channels/674039678562861068/850924523610963988/1339938011307053137


AnnaXar (commented):
That's correct change

PR #6628, opened 2025-02-18
Common: add hexsoon radar

rmackay9:
This adds a setup page for the Hexsoon WWM radar which were merged to master with PR https://github.com/ArduPilot/ardupilot/pull/29235

I've built this locally and it looks OK to me.  For now it is just in the "upcoming features" section because we haven't released it yet.  I hope to backport it to 4.6 eventually but for now it's a 4.7 feature


rmackay9:
Ah, I just found a link to the configuration tool so I've added that and re-pushed.  I think this is ready now, txs!

Hwurzburg:
LGTM

PR #6642, opened 2025-02-22
Docs: ARKV6X Update Overview with Serial Parameter Number Ordering

hendjoshsr71:
Removes the PX4 info about /dev/tty*  instead to have the ArduPilot Serial Number for each port that is used in the parameter tree

@AlexKlimaj

### DIFF
--- a/common/source/docs/common-ark-v6x-overview.rst
+++ b/common/source/docs/common-ark-v6x-overview.rst
@@ -57,37 +57,37 @@ Specifications
 Serial Port Mapping
 ===================
 
-.. list-table:: Serial Port Mapping
+.. list-table:: Serial Port Mapping for Default Firmware with No IOMCU
    :widths: 15 25 15
    :header-rows: 1
 
    * - UART
-     - Device
+     - Serial Number
      - Port
-   * - USART1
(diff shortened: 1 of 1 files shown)


common/source/docs/common-ark-v6x-overview.rst
```diff
@@ -57,37 +57,37 @@ Specifications
...
      - PX4IO/RC
-   * - UART7
-     - /dev/ttyS6
-     - TELEM1
-   * - UART8
-     - /dev/ttyS7
-     - GPS2
 
 
 More Information
```
Hwurzburg, line 93:
```suggestion
.. note:: Serial ports 5 - 8 by default have no protocol assigned. User must set the protocol to the shown port protocol themselves in order to use the port in the manner listed.

More Information
```

Hwurzburg (changes requested):
might as well put this warning in also since the hwdef does not set those protocols...since you did this using the web editor, I think, I can fix it for you if its easier

Hwurzburg:
@hendjoshsr71 ping

AlexKlimaj (approved):

PR #6650, opened 2025-02-28
Correct typo in common-dronecan-serial.rst

DaGigabyte:
Correct typo in common-dronecan-serial.rst
DrooneCAN -> DroneCAN


Hwurzburg:
thanks

PR #6649, opened 2025-02-28
Sub: start the transfer process from ardusub.com

ES-Alexander:
[ArduSub.com](https://ardusub.com) has been gradually becoming redundant and outdated as we've split out and replicated/updated the control station software and onboard computer documentation elsewhere.

Transferring the remaining firmware-focused content here will allow us to
1. dismantle the old site
2. reduce fragmentation between the sub-related documentation sources
3. continue reducing the rift between ArduSub and the more general ArduPilot project, and
4. encourage Blue Robotics to contribute to the ArduPilot docs upkeep (instead of just doing our own thing)

This is the starting point for that.

I'm intending to make substantial additions to this in subsequent PRs, but it seems worthwhile for the initial PR to be relatively small, and focused on infrastructure, to allow for a more straightforward review process, and to make sure things are set up correctly before we go too deep.

Screenshot 2025-02-28 at 9 52 53 pm

PR #6653, opened 2025-02-28
common: ready-to-fly: avoid unwanted duplication

ES-Alexander:
Only a subset of the listed vehicles are relevant to each vehicle type.

This PR
- splits those out (removing the unwanted ones from the copy action), while also
- removing blimp entirely (it has no pre-made purchaseable vehicles), and
- adding Sub as a target
    - ~cannot be merged before #6649 - hence draft PR for now (will need to be rebased over master once that's merged in)~

### DIFF
--- a/common/source/docs/common-rtf.rst
+++ b/common/source/docs/common-rtf.rst
@@ -12,6 +12,7 @@ Please check with the vendor prior to purchase.
 
    We recommend, when possible, to purchase from our `Partners <https://ardupilot.org/about/Partners>`__ because they directly support ArduPilot and we are more confident that users can upgrade the firmware on the vehicle.
 
+[site wiki="copter"]
 Copters from Partners
 =====================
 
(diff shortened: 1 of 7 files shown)


Hwurzburg (changes requested):
you need to split into two PRs
1.infrastructure
2.wiki content

ES-Alexander:
I intentionally left this as a draft PR, because the first two commits are in a different PR already (this one wasn’t intended to be reviewed until after that’s merged - only the last commit matters here, which I believe is just infrastructure-related)

common/source/docs/common-rtf.rst
```diff
@@ -80,34 +85,47 @@ Boats from Partners
...
 
+[site wiki="copter,plane,rover"]
 Vehicles from Non-Partners
 ==========================
+[/site]
 
+[site wiki="copter"]
 * `Aton <https://traxxas.com/products/models/heli/Aton-Plus>`__ and `Aton-Plus from traxxas <https://traxxas.com/products/models/heli/Aton-Plus>`__ (firmware loaded using an SD Card)
-* `BathyDrone mapping boats <https://www.bathydrone-usv.com/>`__
 * `DRONEE Easy to Use Mapping Plane <https://dronee.aero/pages/droneeplane>`__
```
Hwurzburg, line 104: this is a plane not a copter

Hwurzburg (changes requested):

common/source/docs/common-rtf.rst
ES-Alexander: fixed, thanks :-)

Hwurzburg (approved):

PR #6654, opened 2025-03-01
Update common-proximity-landingpage.rst

geofrancis:
I dont think anyone knows this feature exists to change the range of the proximity display. you could add this image also .
6c0ae83eac04181188d85d7500b300dd5734289b


Hwurzburg:
added image...merging...thanks

geofrancis:
> added image...merging...thanks

I thin there is an issue 

image

Hwurzburg:
fixed...one of those things that builds properly locally but not if you don't git add the image to the commit...

rmackay9:
Thanks, even I didn't know that you could change the range!

PR #6658, opened 2025-03-02
Spell out default

Ryanf55:
`Def` doesn't mean default to everyone. I don't think we should abbreviate this in the wiki,


Hwurzburg (changes requested):
you will need to reformat each table...all columns must be same width

Ryanf55:
> you will need to reformat each table...all columns must be same width

Fixed!

Hwurzburg:
thanks

PR #6674, opened 2025-03-09
Dev: Updated the intel realsense model name in gsoc project

snktshrma:
It seems like some students had confusion with the unavailable Intel D465 model. Replaced it with Intel D455.


Hwurzburg:
thanks

PR #6676, opened 2025-03-10
Dev: copter and plane commands for loiter, rtl, land

rmackay9:
This is a follow up to https://github.com/ArduPilot/ardupilot_wiki/pull/6671

This further clarifies what the MAV_CMD_**NAV_LOITER_UNLIM**, MAV_CMD_NAV_**RETURN_TO_LAUNCH** and MAV_CMD_**NAV_LAND** mav commands do when executed while Copter and/or Plane are in Guided mode which is to actually change the flight mode.

Previously these lines referred the user to the corresponding mission commands which made one community developer at least think that the behaviour would be Loiter-within-Guided and they were thus surprised when the pilot's throttle was suddenly consumed and the vehicle descended quickly.

I haven't added anything to the Rover pages because they simply don't mention these commands.  The description for these commands is there (for all vehicles) in the Get and Set FlightMode page


Hwurzburg:
thanks

PR #6679, opened 2025-03-11
Copter: add links to ChrisR's setup videos

rmackay9:
This adds a link to Chris Rosser's ArduPilot tuning video playlist

I originally wanted to add the link so that a preview was displayed and users watch the video within the AP web site but I couldn't figure out how to make that work


Hwurzburg:
only single videos with the youtub sphinx plugin

rmackay9:
@Hwurzburg,

OK thanks.  I think ChrisR is planning to add more videos to the play list so maybe we just stick with this PR as-is.

rmackay9:
Hi @IamPete1,

Thanks for reviewing the videos.  Sorry, I really should have done that myself instead of letting my enthusiasm get the best of me.

PR #6680, opened 2025-03-11
Common: BotBlox DroneNet setup page

rmackay9:
This adds a dedicated page for the BotBlox DroneNet setup.

I think the setup is difficult enough that a separate page will help users get it up and running more quickly.

I also plan to add a setup page for the CubeNode ETH and then remove some bits here and there on the wiki that talk about the setup of PPP.

I've built this locally and I think it looks OK


rmackay9:
sorry, I've found a link error that I will fix in a moment

Hwurzburg:
thanks

PR #6697, opened 2025-03-18
Common: holybro pixhawk 5x ethernet warning

rmackay9:
This adds a warning to the Holybro Pixhawk 5x to warn users that AP doesn't support networking on F7 boards (including this board)


Hwurzburg:
LGTM

PR #6702, opened 2025-03-20
4.7:StellarF4

Hwurzburg:


common/source/docs/common-stellarf4.rst
```diff
@@ -0,0 +1,110 @@
...
+
+
+* 1, 2, 3, 4  are Group 1;
+* 5, 6, 7, 8  are Group 2;
+* 9           is Group 3;
+* 10           is Group 4;
+
+Battery Monitoring
+==================
+The board has 1 built-in voltage divider on an ADC input and 1x current ADC input and supports and external 3.3V based current sensor. The voltage input is compatible with 2~6S LiPo batteries.
```
nkruzan, line 85:
```suggestion
The board has 1 built-in voltage divider on an ADC input and 1x current ADC input and supports an external 3.3V based current sensor. The voltage input is compatible with 2~6S LiPo batteries.
```

nkruzan:
I would like to see it always referenced as "Stellar F4V2" especially if there is a V1 or if is expected to have a V3+.

common/source/docs/common-stellarf4.rst
Hwurzburg: doone

PR #6704, opened 2025-03-20
Sub: add sidebar link to MAVLink support page

ES-Alexander:
### Preview

Screenshot 2025-03-21 at 1 03 41 am

### Context
I wanted this to be [anchor-based](https://ardupilot.org/ardupilot/docs/common-wiki-editing-style-guide.html#internal-links), but that seemingly doesn't work for the sidebar, so I followed the approach used in the `ardupilot` wiki for linking to the other wikis (which is just a direct site URL).

This is a Sub-only change (@Williangalvani).
- [x] I've built the full wiki locally, with no warnings or errors.

### DIFF
--- a/sub/source/index.rst
+++ b/sub/source/index.rst
@@ -26,4 +26,5 @@ Getting more info
    
    Complete parameter list <docs/parameters>
    Complete log message list <docs/logmessages>
+   MAVLink message support <https://ardupilot.org/dev/docs/ArduSub_MAVLink_Messages.html>
    Board Feature List <docs/binary-features>


Hwurzburg:
Do not merge this...I can fix

Hwurzburg:
fixed

ES-Alexander:
@Hwurzburg I did initially consider using `copywiki` for this, but given there are no site-specific changes it just duplicates the content to two URLs, which (as I understand it) reduces the rankings of both pages in search engines (because links to them are generally distributed between both) unless one of them is marked as [canonical](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls), which I’m currently unsure how to do with rst source files.

It also seems potentially inconsistent to be using `copywiki` for a page that’s not “common”, although I’m unsure of ardupilot_wiki’s conventions for this kind of scenario, given the page is clearly not common to other vehicle wikis 🤷🏼‍♂️

Curious on your thoughts here as to the preferred approach :-)

---

As some broader context, I think the MAVLink message support docs should end up with versions like the parameters, so should likely end up hosted on the individual vehicle sites and then perhaps just linked to from the dev site, but that seems out of scope for this PR.

Hwurzburg:
well apparently it does not work.....wonder why it did when I checked locally for sub....so this will need some more work...

PR #6706, opened 2025-03-20
4.7:Airspeed skip cal option

Hwurzburg:


nkruzan:
"will force you to manually recalibrate the offset in order to arm, each boot in order to remind users that need to allow the sensor to settle to ambient temperature."

needs reworded

Hwurzburg:
reworded

PR #6708, opened 2025-03-20
Move MAVLink message pages to vehicle wikis

Hwurzburg:


Hwurzburg:
This moves MAVLink message pages to each vehicle....you can change the link on the sub index page from URL (which will be dead) to ArduSub_MACVLink_messages after this goes in....

I used the GCS page as the TOCLIST home for these pages...

common/source/docs/common-choosing-a-ground-station.rst
```diff
@@ -1,5 +1,5 @@
 .. _common-choosing-a-ground-station:
-[copywiki destination="plane,copter,rover,blimp"]
+[copywiki destination="plane,copter,rover,blimp,sub"]
```
ES-Alexander, line 2: I'm reluctant to add this page to Sub without some site-specific recommendations (or filtering) based on options that have actually been tested and confirmed to work with ArduSub vehicles.

There have been quite a few modifications made to QGroundControl over the years to make it properly compatible, and I expect several of the other control station software options will expect a radio connection (which Sub doesn't support), and may not have support for things like Sub's joystick button functions via `MANUAL_CONTROL` messages.

I can add an initial version of that in a separate PR next week, which this one can get rebased over, if that works for you?

common/source/docs/common-choosing-a-ground-station.rst
```diff
@@ -295,3 +295,36 @@ For Desktop devices you will need:
 
 -  A USB cable, BT module, WIFI adapter, or other connectivity means as listed here (:ref:`common-telemetry-landingpage`)
```
ES-Alexander, line 296: This page does not exist in Sub, so I expect that link would be broken.

sub/source/index.rst
```diff
@@ -24,6 +24,7 @@ Getting more info
 .. toctree::
    :hidden:
    
+   Choosing a Ground Control Stations <docs/common-choosing-a-ground-station>
```
ES-Alexander, line 27:
```suggestion
   Choosing a Control Station Software <docs/common-choosing-a-ground-station>
```
As I understand it the actual control station is the hardware, whereas this page is about choosing software to run on the control station hardware, so if it's getting included then I think this label is more intuitive.

ES-Alexander (commented):
Thanks for working on this move :-)

common/source/docs/common-choosing-a-ground-station.rst
Hwurzburg: we can restrict on only QGC for sub easily

sub/source/index.rst
Hwurzburg: we use GCS to refer to the hardware/software combination...

common/source/docs/common-choosing-a-ground-station.rst
Hwurzburg: I suspect that it actually may not be....locally, it will redirect to any wiki that has that tag...not sure that it will on the main server, but I am going to merge for a test....I can revert if it does not

PR #6710, opened 2025-03-21
mavproxy/dev: update outdated and broken links

ES-Alexander:
- Change a bunch of cross-wiki links to `:ref:` style, so they can be tracked if/when they get broken
- Update remaining `ardusub.com` references to the new ArduPilot wiki site
- Fix some miscellaneous broken links in the process


Hwurzburg:
LGTM

PR #6722, opened 2025-03-26
Common: custom build server page gets pics

rmackay9:
This adds some pictures to the custom build server page and adds an extra step saying how to upload the .apj file to the autopilot.

More generally I think we should consider promoting this page one level higher but we can do that at a later date if we choose to


Hwurzburg (dismissed):
We should add a note that DESLECTING a feature, does not remove any dependencies that it added. (Currently this can cause a build failure for CAN stuff, but it always wastes space)

Hwurzburg:
thanks

PR #6724, opened 2025-03-27
dev: avoid using gz tf in cartographer exemple

pedro-fuoco:
# Purpose
The ROS 2 cartographer example is being updated. To disambiguate the `TF` trees the user must declare `use_gz_tf:=false` when launching the `gazebo` simulation. This guarantees that the odometry present at `/ap/tf` comes from `Google Cartographer`, as the example intends.

# Dependencies
- [ ] 

These PRs solve ArduPilot/ardupilot_ros#24


Hwurzburg:
this is getting old...any progress?

Issue #6670, opened 2025-03-07
Dev: add instructions on how to pause missions


rmackay9:
Related to issue https://github.com/ArduPilot/ardupilot_wiki/issues/6669 we should add details on how to pause a mission to the Developer wiki's MAVLink interface section.  The most appropriate page is probably the [Mission Upload/Download page](https://ardupilot.org/dev/docs/mavlink-mission-upload-download.html).

A mission can be paused using an auxiliary function (which is demonstrated on the MAVLink Interface RC page) but there may also be a specific mavlink command

rmackay9:
This is resolved now, thanks!

PR #6726, opened 2025-03-28
Dev: mavlink interface mission page gets pause mission explanation

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot_wiki/issues/6670 by adding a section to our mavlink interface section's Mission page to explain how to pause/resume a mission

I've tested the commands listed and also built locally and it seems OK to me


rmackay9:
@Hwurzburg,

Txs for the review and bringing up that this command is also supported in copter missions but not on [this wiki page](https://ardupilot.org/copter/docs/mission-command-list.html).  That really should have been done by the author of https://github.com/ArduPilot/ardupilot/pull/19317 and they also should have added it to the MP command list.  Without doing all that I'm sure nobody's actually using it.

I guess I'll do all though but please let me grumble a bit about it

rmackay9:
Hi @Hwurzburg,

I've tested and the MAV_CMD_DO_PAUSE_CONTINUE command can't be used within a mission.  It can only be used as an immediate command.  https://github.com/ArduPilot/ardupilot/pull/19317 adds support for saving/restoring this command as part of a mission but the Copter mission command handler was never added so it doesn't actually work.

I think we should merge this as-is because it covers what works and it's the most likely use case.

Hwurzburg:
done

PR #6732, opened 2025-03-31
update NANO and GEM1305 specifications

datagnss:
- update the pictures
- modify the description of product, with new compass support


Hwurzburg:
thanks

PR #6733, opened 2025-03-31
Planner: live video setup instructions

rmackay9:
This PR makes a few changes related to camera gimbals:

1. Mission Planner gets a "live video" setup page which describes how to get video appearing on the HUD or in the Map area.  These instructions may not be complete but they are certainly better than what we have now
2. Camera gimbal pages for Siyi, Topotek, XFRobot and ViewPro get links to the above live-video page
3. Viewpro loses the link at the very top to the old Lua script.  4.5 (and above) use the C++ driver so we shouldn't mislead users into think they need to install a Lua script
4. Herelink page's top image uses an existing image without the "Hex" logo.  Links back to CubePilot's pages are slightly improved

I've tested this locally and it looks OK to me with one exception:

- the vehicle links back to the planner:live-video page don't seem to work.


Hwurzburg:
probably an intersphinx inventory issue since you are trying to access cross wiki tags (if live video had been a common file, instead of a specific site doc file, it wouldn't need the intersphinx inventory that the wiki server creates.....I think....

I am merging, and will deal with any issues tomorrow after the wiki rebuilds...

PR #6734, opened 2025-04-01
Common: XFRobot gets custom-build-server explanation

rmackay9:
This adds a section to the XFRobot page to explain how to build using the custom build server.  One tester struggled for a while until we finally figured out that this step was skipped.


Hwurzburg:
thanks

PR #6736, opened 2025-04-01
Added section: Mount Targeting – Making a Gimbal Follow a Vehicle

Ashishbsharma:
This update adds a new section to common-mount-targeting.rst explaining how to configure a camera mount or gimbal to follow a moving vehicle using MAVLink MOUNT_TARGET messages. It includes:

Use case overview based on [@timtuxworth's video](https://www.youtube.com/watch?v=FX60mqi9WMQ)

Required parameter setup

Python code example using pymavlink

Explanation of coordinate frames and MAVLink behavior

Troubleshooting tips and related links

This aims to make it easier for new users to replicate this commonly requested functionality with gimbal tracking.

### DIFF
--- a/common/source/docs/common-mount-targeting.rst
+++ b/common/source/docs/common-mount-targeting.rst
@@ -125,6 +125,95 @@ MAVLink mount commands can also be sent from other sources, such as companion co
 Cameras may also be controlled via MAVLink commands from a companion computer or other source.
 See :ref:`dev:mavlink-camera` documentation.
 
+
+Mount Targeting – Making a Gimbal Follow a Vehicle
+--------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks for this but I wonder where the information that went into this section came from?  It's plausible to someone who doesn't know AP that well but when we look at the details it's incorrect.  I tend to think this was mostly generated by an AI.

I think I'll go ahead and close this but I'll see replies

Ashishbsharma:
Thanks for the feedback! I did use AI to help structure the content, but it was based on real discussions like @timtuxworth’s video and user questions. My goal was to contribute something helpful for new users, especially those trying to replicate this behaviour without deep internal knowledge of AP.

That said, I completely understand if some of the details were off, I’m still learning the system myself. I’d be happy to revise the section based on your corrections or suggestions if you think it’s worth adjusting. Thanks again for the opportunity!

PR #6737, opened 2025-04-02
Dev: add ochin carrier board to list of companion computers

rmackay9:
This adds a link to the Ochin carrier board which I've used successfully.  It's very small and relatively inexpensive.


Hwurzburg:
thanks

PR #6739, opened 2025-04-03
Common: CubeNodeETH firmware update process

rmackay9:
This adds a section to the CubeNodeETH page on how to update the firmware.

Once [this MP issue is resolved](https://github.com/ArduPilot/MissionPlanner/issues/3516) we may update the page again to recommend using MP instead of the DroneCAN GUI Tool.


Hwurzburg:
LGTM...thanks

rmackay9:
Thanks very much!

PR #6741, opened 2025-04-03
Removes TERRAIN_FOLLOW parameter from terrain-following.rst

ccackam:
Per this discussion: https://discuss.ardupilot.org/t/prearm-waiting-for-terrain-data-i-couldnt-get-rid-of-the-error/95415

TERRAIN_FOLLOW has been removed as a parameter.

### DIFF
--- a/copter/source/docs/terrain-following.rst
+++ b/copter/source/docs/terrain-following.rst
@@ -70,7 +70,7 @@ Terrain data files are always created as one file per one-by-one degree area, th
 
 Using Terrain Altitude during RTL and Land
 ==========================================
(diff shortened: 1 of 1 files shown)


Hwurzburg:
@rmackay9 this page also noticably omits the need to set TERRAIN_ENABLE=1 to use and obtain terrain data

rmackay9 (approved):
yes this looks correct to me, thanks!

rmackay9:
@Hwurzburg, yes you're right that TERRAIN_ENABLE is also not mentioned although it's set by default to be 1 I think

Hwurzburg:
merging and will add a note to self to add that to the page...

PR #6749, opened 2025-04-08
Siyi: add product specific download links

rmackay9:
The Siyi page's download page link has changed so I've taken the opportunity to add download links per product.  This will hopefully help users more quickly find the Assistant (although this doesn't depend upon the product) and the user manual and latest firmware.

Thanks to @andyp1per for finding this issue!


Hwurzburg:
thanks

PR #6750, opened 2025-04-08
Add files via upload

TheOracle2212:
showing them leans


Hwurzburg:
Thanks for contributing. Unfortunately, the image does not show what you are trying to show. There is no reference, and it's cluttered. You need to show a graph with AHRS2.Yaw, XKF1[0].Yaw and XKF1[1].Yaw, juxtaposing against the image above 

should really be a roll plot image...ie, AHRS2.Roll, XKF1[0].Roll and XKF1[1].Roll showing the vibration impact...perhaps with roll axis VIBE plot...

you can use the web based editor, but it leads to issues like you had, making adding things to the PR difficult...and this PR is off your master branch, instead of a local working branch, which is not recommended...

if you have an appropriate log of a vehicle with excess vibration causing the "leans", attach it here and I can help get a PR that could be merged....

Issue #6760, opened 2025-04-12
Clarify infoodr flying guidelines is for non-GPS flying


rmackay9:
I've seen at least two users ([see here](https://discuss.ardupilot.org/t/takeoff-failed-with-optitrack-for-non-gps-navigation/132852/6)) become confused when trying to setup Non-GPS navigation because they've simultaneously tried to follow our [Indoor Flying Guidelines](https://ardupilot.org/copter/docs/indoor-flying.html).  We should clarify that the indoor flying guidelines is only valid when no position or velocity sensors (e.g. GPS, optical flow) are used.

PR #6762, opened 2025-04-12
Copter: clarify indoor flight guidelines usecase

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot_wiki/issues/6760 by clarifying the use-case that applies for the "Indoor Flight Guidelines" wiki page

I've also added a few links and cleared up some wording and corrected how the GPS should be disabled.  I don't think setting AHRS_GPS_USE was correct because (at least according to the parameter descriptions this only applies to DCM).  I think the two possible ways to disable the GPS are:

1. set EK3_SRC1_POSXY, EK3_SRC1_VELXY = 0
2. set GPS1_TYPE = 0

I've specified the 2nd because it's easier and I've also testesd in SITL that I can arm and takeoff in stabilize mode successfully

### DIFF
--- a/copter/source/docs/indoor-flying.rst
+++ b/copter/source/docs/indoor-flying.rst
@@ -4,14 +4,14 @@
 Indoor Flying Guidelines
 ========================
 
-This article provides important guidelines for flying your multicopter
-inside.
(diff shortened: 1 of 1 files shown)


Hwurzburg:
your the man on this...merging

rmackay9:
Thanks very much for the merge!

PR #6767, opened 2025-04-18
Common: Benewake TF02/TF03/TFLuna fixes

rmackay9:
This includes a few small fixes to the Benewake rangefinder TF02/TF03/TFLuna pages

- removed instructions on very old version
- simplify rngfnd1-min to remove mention of an unrelated model
- unify naming of TF02-Pro
- fix broken link for DroneCAN model


Hwurzburg:
thanks

PR #6768, opened 2025-04-18
Common: pre-arm failure to create log directory

rmackay9:
This adds a short explanation of what to do if users see the "Failed to create log directory /APM/LOGS : ENOMEM" pre-arm failure.

This came up during 4.6 beta testing in [this discussion](https://discuss.ardupilot.org/t/logging-failed-failed-to-create-log-directory-apm-logs-enomem/132917).


Hwurzburg:
thanks

PR #6774, opened 2025-04-19
Update common-pxfmini.rst

geofrancis:
Site is dead and leads to malicious website.


Hwurzburg:
this is an archived topic which are never updated or changed...thanks anyway (PS every link in this doc is dead)

PR #6779, opened 2025-04-21
Planner: trim live video images

rmackay9:
trim live video images to remove edges and location


Hwurzburg:
LGTM

Issue #6783, opened 2025-04-22
Rover's QuikTune links to Plane's circle mode


rmackay9:
Rover's QuikTune page https://ardupilot.org/rover/docs/quiktune.html, has a link to Circle mode but it points to Plane's circle mode instead of Rover's https://ardupilot.org/rover/docs/circle-mode.html

PR #6784, opened 2025-04-22
Rover: fix circle mode reference label

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot_wiki/issues/6783

I was strange because the reference didn't work from the [QuikTune page](https://ardupilot.org/rover/docs/quiktune.html) but it did work from the [Control Modes page.
](https://ardupilot.org/rover/docs/rover-control-modes.html)

### DIFF
--- a/rover/source/docs/circle-mode.rst
+++ b/rover/source/docs/circle-mode.rst
@@ -1,4 +1,4 @@
-.. circle-mode:
+.. _circle-mode:
 
 ===========
 Circle Mode


Hwurzburg:
Sphinx always tries to find the reference...scans its local wiki first, then ALL other tags in all other wikis...same for param links...this was apparently changed in Sphinx at some point because to get to other refs in other wikis you used to have to preface with "plane:circle-mode"....I guess that now only targets a specific wiki's tag to avoid the search...

that link apparently has not been clicked in two years to catch my mistake on the tag

PR #6788, opened 2025-04-22
Common: update siyi product links

rmackay9:
This fixes a few of the link at the very top of the Siyi page.

ZT6 is no longer for sale so we're using their product page which is a little different than the store page


Hwurzburg:
thanks

rmackay9:
Thanks very much!

PR #6801, opened 2025-04-24
add some info on default sub config

Hwurzburg:
as time permits, I am adding some basic info to the sub wiki


sub/source/docs/sub-frames.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Several bi-directional motor configurations/orientations are available in Sub:
+
+Key to axis: R = Roll, P = Pitch, Y = Yaw, Z= Depth, F = Forward, L = Lateral
+
+=============================   =======  ================  ==================
+:ref:`FRAME_TYPE<FRAME_TYPE>`   #MOTORS  NAME              Controllable Axes
+=============================   =======  ================  ==================
+0                                  6     BlueROV1          R/P/Y/Z/F/L
+1                                  6     Vectored          R/P/Y/Z/F/L
```
Williangalvani, line 15:
```suggestion
1                                  6     Vectored          R/Y/Z/F/L
```

Hwurzburg: ooops....will change...thanks

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+- A depth sensor (underwater barometer) is required.
+- Frame configuration is defaulted to Vectored with 6 thrusters.
+
+.. _sub-defaults:
+
+Sub Defaults
+============
+
+* Default :ref:`frame type <FRAME_TYPE>` is VECTORED
+* RC is disabled via :ref:`RC_PROTOCOLS<RC_PROTOCOLS>` = 0 and :ref:`FLTMODE_CH<FLTMODE_CH>` = 0.
```
ES-Alexander, line 20:
```suggestion
* Radio control is disabled via :ref:`RC_PROTOCOLS<RC_PROTOCOLS>` = 0 and :ref:`FLTMODE_CH<FLTMODE_CH>` = 0. Water absorbs radio waves, so RC is generally impractical unless the vehicle is at the surface, or the signal is converted to some other medium in between (e.g. using a tether cable, or acoustic modems).
```

- People starting out with a Sub vehicle don't necessarily know what RC is.
- People gaining first exposure to underwater vehicles may not be aware of its electromagnetic absorption spectrum
    - Longer term I'm considering including a broader "challenges of the underwater environment" section to explain things like this, which this could link to, but it seems alright to include a relevant note here for now.

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+
+MAVLink Overrides
+-----------------
+Since pilot directional control, normally provided by RC in other vehicles is not enabled by default. Sub uses RC overrides via MAVLink, and the channel mapping is fixed:
```
ES-Alexander, line 24:
```suggestion
ArduPilot vehicles normally use radio control (RC) to provide pilot directional control. Since Sub has RC disabled by default, it instead provides directional control with MAVLink RC overrides, and the channel mapping is fixed:
```

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+Pilot control    RC channel override
+===============  ===================
+Roll             1
+Pitch            0
+Yaw              3
+Throttle(Depth)  2
+Forward          4
+Lateral          5
+Camera Pan       6 (via AUX function on RC7)
+Camera Tilt      7 (via AUX function on RC8)
```
ES-Alexander, line 36: Not certain how best to incorporate this, but the old [ArduSub RC docs](https://www.ardusub.com/developers/rc-input-and-output.html) clarify that

> *Note: Camera Tilt control is not standard servo-style control. Specifying a 'PWM' value sets the speed for a sweep from the current tilt angle to the min/max camera angle. The input difference from the midpoint (1500) sets the speed, e.g. a 1900 input sweeps to the max camera angle twice as fast as 1700.

which seems sufficiently unobvious that it's worth including a note about.

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+
+===============  ===================
+Pilot control    RC channel override
+===============  ===================
+Roll             1
+Pitch            0
+Yaw              3
+Throttle(Depth)  2
+Forward          4
+Lateral          5
```
ES-Alexander, line 34:
```suggestion
Yaw (Turn)       3
Heave (Vertical) 2
Surge (Forward)  4
Sway (Lateral)   5
```
Adjustments to better match [the firmware naming](https://github.com/ArduPilot/ardupilot/blob/master/libraries/SRV_Channel/SRV_Channel.cpp#L114), which links to [ship movement terminology](https://en.wikipedia.org/wiki/Ship_motions).

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+======       ========
+Output       Function
+======       ========
+1            Motor1
+2            Motor2
+3            Motor3
+4            Motor4
+5            Motor5
+6            Motor6
+9            RCin9
```
ES-Alexander, line 55: This recently got renamed to `RCIN9/Lights1Level` in the firmware, but I've just checked and there are seemingly dedicated outputs for lights now instead.

@Williangalvani can you confirm whether this should be changed to `Lights1`, or remain as `RCIN9`?

sub/source/docs/sub-hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+5           Shift Function (when pressed buttons provide alternate second function)
+6           Arm
+7           Center Mount1
+8           Joystick Input Hold Set
+9           Tilt Mount1 Down/Pan Mount Left
+10          Tilt Mount1 Up/Pan Mount Right
+11          Joystick Sensitivity Gain Increase/ Trim Pitch Decrease
+12          Joystick Sensitivity Gain Decrease/ Trim Pitch Increase
+13          Lights1 Dimmer/ Trim Roll Decrease
+14          Lights1 Brighter/ Trim Roll Increase
```
ES-Alexander, line 80:
```suggestion
9           Tilt Mount1 Down / Pan Mount1 Left
10          Tilt Mount1 Up / Pan Mount1 Right
11          Joystick Sensitivity Gain Increase / Trim Pitch Decrease
12          Joystick Sensitivity Gain Decrease / Trim Pitch Increase
13          Lights1 Dimmer / Trim Roll Decrease
14          Lights1 Brighter / Trim Roll Increase
```

ES-Alexander (changes requested):
Thanks for the contribution @Hwurzburg!

Some comments/thoughts on clarification and consistency with the codebase, and the assumptions of what contexts Sub users may come from :-)

sub/source/docs/sub-hardware.rst
Hwurzburg: I like it

sub/source/docs/sub-hardware.rst
Hwurzburg: thanks

sub/source/docs/sub-hardware.rst
Hwurzburg: maybe put the nautical after the conventional...will change

sub/source/docs/sub-hardware.rst
Hwurzburg: actually we are both wrong...there should be no white space around the slash when it indicates an alternative....will correct

sub/source/docs/sub-hardware.rst
Hwurzburg: this matches what the GCS reports

sub/source/docs/sub-hardware.rst
ES-Alexander: Huh, I learnt something new - thanks! :-)

sub/source/docs/sub-hardware.rst
ES-Alexander: Is that an ArduPilot convention?

I read spacing around slashes as "word/alternative" / "option alternative".

Hwurzburg: its a general grammar/punctuation rule...Goggle "whitespace around slashes":

"Generally, whitespace is not used around slashes (/) when they separate single words or letters. However, a space is typically used before and after the slash when it separates phrases, sentences, or a single word from a phrase, especially to improve readability. 
"

ES-Alexander: That's exactly what I was trying to suggest 😅 

As an example, it should be "Tilt Mount1 Down" / "Pan Mount1 Left" (i.e. two phrases being separated). If it's separating just the words in the middle (e.g. "Tilt Mount1 Down/Pan Mount1 Left") then it comes across as those words being the differentiator of two broader options (e.g. "Tilt Mount1 Down Mount1 Left" / "Tilt Mount1 Pan Mount1 Left").

As a stripped down example, "Lights 1/2" makes sense, and "expands" to "Lights 1 / Lights 2", but "Lights 1/Lights 2" is hard to make sense of.

Hwurzburg: yep, I see your point...will fix

PR #6813, opened 2025-04-26
Add customization guide for SITL location and battery settings

SurafelAnshebo:


SurafelAnshebo:
This PR adds a simple guide to customize SITL GPS launch locations and battery settings.

Location customization:
- Shows how to start SITL at a custom location
- Explains how to define a named location using `location.txt`
- Shows Blacksburg, VA coordinates as an example

Battery customization:
- Explains how to set battery monitor and capacity using MAVProxy
- Shows how to update default battery capacity in firmware for SITL

Hwurzburg:
thanks for this, but I have a few issues:
1. the location setup duplicates the basic Using SITL information that precedes these vehicle examples...
2. changing params in SITL to get different behavior is basic to the use of SITL, I don't think it needs explanation, If so, EVERY thing one might change in a sim would need to be explained...
3. the defaults change you suggest is not permanent...only until you switch a branch...and using a param file as suggested in the preceding Using SITL instructions is preferred since messing with the code itself is dangerous to inexperienced users...you don't need to know how to build code to use SITL..

SurafelAnshebo:
Thank u very much for the detailed feedback and reviewing my contribution! @Hwurzburg 

So I am working on revising the PR to match what you outlined.  
But before I proceed, I just wanted to kindly ask for a bit of clarification to ensure I align correctly and can finalize the update without introducing more issues:

- For the GPS location setup: Should I completely remove the explanation and example I added, and instead simply mention that users can set custom locations using `-l`, while referring them to the "Using SITL" section for full details?
- Adn as for the battery settings: Would it be better if I avoid explaining specific parameter changes (like `BATT_MONITOR`) and instead just mention that users can modify parameters generally, recommending the use of `.param` files for consistent behavior during simulation?

My goal is to revise this contribution promptly and correctly so it can be completed without leaving it idle or open too long.  
I appreciate your guidance and your time thanks again for helping me improve my contribution!

Hwurzburg:
yes, to both of your proposals

SurafelAnshebo:
Thanks @Hwurzburg, finished the revision.

Hwurzburg:
not exactly what I meant...I modified it and rebased, squashed....merging...thanks

PR #6824, opened 2025-04-28
Plane-architecture: Change the task method name

muramura:
In Japan, there is an ArduPilot Japan community on Facebook.
This community includes participants such as Professor Miwa from Tokushima University.
Professor Miwa pointed out that for Plane, there is no update_flight_mode, but rather update_control_mode.
I investigated the source code and agreed with Professor Miwa.

If this is correct, please modify the image.
I was not able to modify the image.

plane_architecture


peterbarker (approved):
Mmmm... I think we wanted to be consistent across our vehicles, so while this might be correct now it may not be in future.

Rover has `update_current_mode` - that's probably want we will end up going with across vehicles to try to make things familiar across different vehicle types.

PR #6826, opened 2025-04-29
Dev: document unit abbreviations for c++

timtuxworth:
Document standard c++ suffix/abbreviations for units. 

Updated based on the DevCall 13 May 2025.

### DIFF
--- a/dev/source/docs/style-guide.rst
+++ b/dev/source/docs/style-guide.rst
@@ -421,6 +421,137 @@ Functions or variables that represent a value relative to a frame should be suff
     float course_deg;
 
 
+Preferred Abbreviations for Units
+---------------------------------
+
+Variables that reference units of measure should normally be suffixed so that it's easier to 
+understand the values when reading the code. No suffix is required when a variable is in seconds, meters, or (angular) degrees or degrees-Celsius.
+
+.. list-table:: Standard name suffixes for Units
+   :widths: 25 25 25 25 75
+   :header-rows: 1
+
+   * - Measure
+     - Unit
+     - Preferred Suffix
+     - Deprecated Suffix
+     - Comments
+   * - Acceleration
+     - meters/second/second
+     - _mss
+     - 
+     - Acceleration
+   * - Acceleration
+     - centimeters/second/second
+     - _cmss
+     - 
+     - 
+   * - Angle
+     - degrees 
+     - _deg
+     - _d
+     - Angular Measurement
+   * - Angle
+     - centi-degrees 
+     - _cd
+     - _cdeg
+     - Angular Measurement - often used internally
+   * - Angle 
+     - radians
+     - _rad
+     - 
+     - Angular Measurement, preferred for internal variables
+   * - Angular acceleration
+     - degrees/second/second
+     - _degss
+     - _dss
+     - Angular acceleration
+   * - Angular acceleration
+     - radians/second/second
+     - _radss
+     - 
+     - Angular acceleration
+   * - Angluar Velocity
+     - degrees/second
+     - _degs
+     - 
+     - Angular change/second
+   * - Angular Velocity
+     - radians/second
+     - _rads
+     - 
+     - Angular Change/Second
+   * - Force
+     - Newtons
+     - _n
+     - 
+     - 
+   * - Current
+     - Ampere
+     - _amp
+     - 
+     - 
+   * - Distance
+     - meter
+     - _m or none
+     - 
+     - 
+   * - Distance 
+     - centimeter
+     - _cm
+     - 
+     - 
+   * - Frequency
+     - Hertz
+     - _hz
+     - 
+     - 
+   * - Jerk
+     - degrees/second/second/second
+     - _degsss
+     - 
+     - 
+   * - Pressure
+     - Pascal
+     - _p
+     - 
+     - 
+   * - Temperature
+     - degrees Celcius
+     - _degc
+     - 
+     - 
+   * - Time
+     - seconds
+     - _s
+     - 
+     - 
+   * - Time
+     - milliseconds
+     - _ms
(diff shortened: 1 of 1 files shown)


dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,135 @@ Functions or variables that represent a value relative to a frame should be suff
...
+   * - Measure
+     - Unit
+     - Preferred Suffix
+     - Deprecated Suffix
+     - Comments
+   * - Distance
+     - meter
+     - _m or none
+     - 
+     - No suffix is preferred for SI units, but _m is allowed for safety/clarity
```
Hwurzburg, line 446:
```suggestion
     - Using no suffix is preferred for SI units, but _m is allowed for safety/clarity
```

dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,135 @@ Functions or variables that represent a value relative to a frame should be suff
     float course_deg;
 
 
+Abbreviations
+-------------
+
+Variables that reference units of measure should normally be suffixed so that it's easier to 
+understand the values when reading the code.
+
+In general no suffix is required when a variable is in SI units such as seconds, meters, but its
```
Hwurzburg, line 430:
```suggestion
In general no suffix is required when a variable is in `SI units <https://en.wikipedia.org/wiki/International_System_of_Units>`__ such as seconds, meters, etc. but its 
```

Hwurzburg (changes requested):
a few tweaks

dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,135 @@ Functions or variables that represent a value relative to a frame should be suff
...
+     - 
+     - Angular Measurement, preferred for internal variables
+   * - Angluar Velocity
+     - degrees/second
+     - _degs
+     - 
+     - Angular Change/Second
+   * - Angular acceleration
+     - degrees/second/second
+     - _dss
```
lthall, line 509: I have been using
acceleration : degss
jerk: degsss

Hwurzburg (commented):
we probably should stress that outward facing variables(params) should always be SI

timtuxworth:
> I'd like something in there about not removing `_m` from methods to avoid rejecting PRs of people trying to move things towards this style guide where we hjave a good reason to not.

"_m is allowed for safety/clarity" - is that enough?

timtuxworth:
> we probably should stress that outward facing variables(params) should always be SI

Is that true? deg is preferred for parameters but radians are SI. (I have a feeling there is another example).

dev/source/docs/style-guide.rst
timtuxworth: There are only two examples of _degss, but a much larger number of _dss

dev/source/docs/style-guide.rst
timtuxworth: done

dev/source/docs/style-guide.rst
timtuxworth: done

timtuxworth:
Is this ok to go in @Hwurzburg ? Alternatively if you think we should discuss it, could I ask you to tag for the dev call, I don't seem to be able to do that here in Wiki

Hwurzburg:
need Tridges and Randys approval
we need to state what is expected on outward facing parameter units...even if deg is noted as an exception

rmackay9 (approved):
great, could we put these in alphabetical order?

dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,141 @@ Functions or variables that represent a value relative to a frame should be suff
...
+   * - Force
+     - Newtons
+     - _n
+     - 
+     - 
+   * - Torque
+     - Newton meters
+     - _nm
+     - 
+     - SI uses Newton Meters for torque
```
rmackay9, line 537: let's not list units that we don't use

dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,141 @@ Functions or variables that represent a value relative to a frame should be suff
     float course_deg;
 
 
+Abbreviations
```
rmackay9, line 424: Let's make the title something like, "Preferred Units" or "variable suffix for units"

dev/source/docs/style-guide.rst
```diff
@@ -421,6 +421,141 @@ Functions or variables that represent a value relative to a frame should be suff
...
+-------------
+
+Variables that reference units of measure should normally be suffixed so that it's easier to 
+understand the values when reading the code.
+
+In general no suffix is required when a variable is in `SI units <https://en.wikipedia.org/wiki/International_System_of_Units>`__ 
+such as seconds, meters, but its not wrong to include it for _m, _s. Otherwise these suffixes listed below
```
peterbarker, line 431:
```suggestion
No suffix is required when a variable is in seconds, meters, or (angular degrees), degrees-Celsius.
```

dev/source/docs/style-guide.rst
timtuxworth: removed

dev/source/docs/style-guide.rst
timtuxworth: done

dev/source/docs/style-guide.rst
timtuxworth: Preferred Abbreviations for Units

timtuxworth:
> we probably should stress that outward facing variables(params) should always be SI

There is a note in the section below about Parameters that gives guidance for units in parameter names. Given the discussion about SI Units on the dev call, I think maybe what it says is sufficient.

rmackay9 (approved):

PR #6827, opened 2025-04-29
common: bluerobotics-ping: update with latest information

ES-Alexander:
- newer device version has some improved technical details
- there is now an open source firmware option


Hwurzburg:
thanks...fixed build error...merging

PR #6835, opened 2025-05-01
Common: more ModalAI voxl2 setup explanation fixes

rmackay9:
This improves our voxl2 setup instructions a bit more based on my experiences setting up a Starling2 Max

Enhancements include:

- Added instructions on how to disable the "figure_eight" commands during Guided mode.  I think this default behaviour is surprising and even dangerous for new users.  By default, if the vehicle is switched to Guided mode it immediately flies home and starts doing a figure of eight.
- Fixed up Serial parameter settings to allow disabling the ODOMETRY messages from reaching the GCS.  This greatly improves the Wifi performance
- VISO_POS_XYZ parameter value are fixed (they should be 0)
- Rangefinder and Proximity sensor values are added

I've built this locally and it all looks OK to me


Hwurzburg:
thanks

PR #6842, opened 2025-05-06
Copter: update vibration failsafe for EKF3

rmackay9:
This updates the EKF fields to look at now that EKF3 is the default (the old values were for the EKF2)

I've also removed mention of a very old version of AP.

The vibration failsafe came up in [this 4.6 beta testing discussion](https://discuss.ardupilot.org/t/jumps-with-gaining-altitude-when-changed-mode-from-rtl-to-loiter/133889)


Hwurzburg:
thanks

PR #6839, opened 2025-05-05
Sub:add arm/disarm and tuning pages

Hwurzburg:

PR #6861, opened 2025-05-15
sub: Rename arming_the_motors.rst to arming-the-motors.rst

ES-Alexander:
#6839 added the page, but links to it with dashes in the name, so it's not appearing in the sidebar.

Dashes also seems more consistent with most of the other filenames, so I opted to rename the (recently added) file rather than changing the link.


Hwurzburg:
I had fixed that earlier today on another pr

PR #6864, opened 2025-05-16
update gpio and wheel encoder pages

Hwurzburg:


rmackay9:
thanks so much!  nice to get rid of all the out-of-date stuff as well

PR #6883, opened 2025-05-23
dev: adding-sim-devices gets ld06 proximity sensor

rmackay9:
This updates the [adding-simulated-devices](https://ardupilot.org/dev/docs/adding_simulated_devices.html) page to include the LD06

This simulator was added as part of PR: https://github.com/ArduPilot/ardupilot/pull/27045


Hwurzburg:
thanks...I try to look at every new PR and tag them, but some escape...there are a LOT of PRs

PR #6884, opened 2025-05-23
Common: update rangefinder testing image and replace sonarrange text

rmackay9:
This update the image that appears at the bottom of almost every rangefinder page and shows users how to check what distance (if any) the rangefinder is sending.  I've also gone through the wiki and replaced all instances of "sonarrange" with "rangefinder1".  The only place this word remains is on an archived page.


Hwurzburg:
thanks for this!

PR #6885, opened 2025-05-23
Common: external ahrs gets custom build server info

rmackay9:
This adds a section to the external AHRS page to inform users when they need to use the custom build server to create a firmware


Hwurzburg (changes requested):
think we need more notes on the image....you circle stuff only for MicroStrain and the VectorNAV boxes are not mentioned...

PR #6920, opened 2025-06-05
expand custom firmware information page

Hwurzburg:


common/source/docs/common-custom-firmware.rst
```diff
@@ -41,8 +41,348 @@ To submit a build request, follow these steps:
...
-custom-beta
------------
+Versions
+========
+
+Latest
+------
+This is the current development branch of ArduPilot. It includes new features, enhancements, clean-ups, etc. but has not been throught a release cycle of beta testing. It has been alpha tested and thoroughly tested in ArduPilot's Continuous Integration auto-test suite. Many experienced users select this to get the latest features but it has the highest risk, in terms of potential bugs.
+
+Stable
```
rmackay9: the descriptions look great.  Maybe we should put them in our recommended order?  E.g. Stable, Beta, Latest?

I suspect you picked this order because it's the order that they appear in the drop-down but we should fix that ordering too (I'll raise an issue)

common/source/docs/common-custom-firmware.rst
```diff
@@ -41,8 +41,348 @@ To submit a build request, follow these steps:
...
+
+Beta
+----
+This is a release of the development branch used for widespread testing and feedback. While the risk of major issues is low, it is expected that minor behavioral deviations may pop up. These are normally addressed in later beta releases of a major new release before declaring the code as a stable release.
+
+Guide to Size Impacts
+=====================
+
+Often, the custom build server is used to create firmware with features that have been removed in the standard firmware to conserve flash space, but desired by the user. If a build fails to complete due to lack of flash, the user may remove features/sensors that he does not need in order to make room for those he wants to add. In order to facilitate those choices, the table below shows what impact a feature's removal has on flash size. Note that a removed feature in this list automatically removed all features dependent upon it.
```
rmackay9: "that he does not need" could be replaced with, "that are not needed"

rmackay9 (approved):
A couple of suggestions but in general, looks great!

common/source/docs/common-custom-firmware.rst
Hwurzburg: the list is very long ....putting latest and beta  at the end would not be good, I think...

common/source/docs/common-custom-firmware.rst
Hwurzburg: reworded..thanks

PR #6941, opened 2025-06-17
Common: AM32 dronecan ESC wiring gets terminator

rmackay9:
This adds a 120 ohm resistor to the AM32 DroneCAN wiring diagram.  I've also added a corresponding comment.  TBH, I'm unsure if this is exactly the correct way to do it.

I've built this locally and it looks OK.

@Huibean does this look OK?


Huibean:
thanks Randy, this looks good at me

common/source/docs/common-am32-escs.rst
```diff
@@ -23,7 +23,7 @@ More ESCs may be found from the `AM32 targets.h <https://github.com/am32-firmwar
 ArduPilot Connection and Configuration
 ======================================
 
-The ESCs should be connected to the autopilot's CAN port.  Below shows the ESCs connected to the autopilot's CAN1 port but CAN2 may also be used.
+The ESCs should be connected to the autopilot's CAN port.  Below shows the ESCs connected to the autopilot's CAN1 port but CAN2 may also be used.  The final ESC should have a CAN bus terminator connecting the CAN_L and CAN_H pins.
```
Hwurzburg:
```suggestion
The ESCs should be connected to the autopilot's CAN port.  Below shows the ESCs connected to the autopilot's CAN1 port but CAN2 may also be used.  The  ESC furthest from the autopilot's CAN port should have a CAN bus terminator connecting the CAN_L and CAN_H pins.
```

rmackay9: Hi @Hwurzburg, thanks for the review, I've updated the wording

Huibean:
besides the very early Vimdrones ESC nano, now the ESC nano and S50 both has a 120Ω jumper on it, but when connect multiple nodes on the bus still need a GH1.25 splitter to manage, so I also made one splitter with a 120Ω jumper, so it   need to enable the last node or the splitter in the last position

Ryanf55:
The wiki could make it clearer that some ESC's have an option for built-in jumper. Not everyone has 120 ohm resistors lying around.

PR #6943, opened 2025-06-18
Common: CubeNode-ETH firmware links updated

rmackay9:
This updates the firmware links for the CubeNode-ETH page.

This is required because we've recently released AP_Periph ver 1.8 and the device was renamed since the previous stable release

FYI @bugobliterator


Hwurzburg:
thanks

PR #6945, opened 2025-06-18
4.7/land alt

Hwurzburg:
This assumes #6944 is merged


rmackay9:
Ah, I somehow missed this.  I don't think this is correct actually.  The land altitude is not used to determine the switch over speed.

"Unlike the LAND mode, the switch point altitude is set by the mission altitude parameter, will be 10m above that altitude in whatever reference frame being used for the altitude parameter"

PR #6957, opened 2025-06-28
APM Planner 2: remove due to dormancy

yuri-rage:
APM Planner 2 has seen zero maintenance/update activity in a couple of years, and users are beginning to report broken features/incompatibilities.

I mean no slight to those who worked on the project, and I know this is a rather big ax-cut to the documentation, but I think it is warranted. If there's a better way, I'm happy to close this PR and use a different approach.


rmackay9 (approved):
yes, I agree..

billbonney:
Can you not put it in an Archive location?

yuri-rage:
> Seems to be working for with ArduRover 4.4.0. I'll have a look at why calibration of plane is not looking
It SHOULD work with Rover 4.4 because that's as old as the most recent commit/release of APM Planner.

I'm not concerned with combing through the wiki to find other unmaintained projects. This one caught my attention due to a few recent user complaints, so I took action. Like I said, I'm open to suggestions. Will await further review.

billbonney:
Here's APM Planner working with ArduPlane 4.6.0 (latest release). Mission loaded and executed from APM Planner 2.

I think deleting the documentation is a bit to drastic. I'll create a PR to stress the development may no include newer features. And re-order the list of GCS in the sidebar.

image

Hwurzburg (changes requested):
the main planner page and its images should not be deleted....the page should be marked archived and added to the archived list and its images retained...IF we decide to archive the topic at Devcall..

common/source/docs/common-choosing-a-ground-station.rst
```diff
@@ -59,24 +59,6 @@ Full featured and widely used GCS.
...
-   (GPLv3)
-
-   .. image:: ../../../images/planner2_choose_agcs.jpg
-    :target: ../_images/planner2_choose_agcs.jpg
-
-- :ref:`Wiki <planner2:home>`
-- `Support Forum <https://discuss.ardupilot.org/c/ground-control-software/apm-planner-2-0>`__
-- `Issues List <https://github.com/ArduPilot/apm_planner/issues>`__
-- `Source Code <https://github.com/ArduPilot/apm_planner>`__
-
```
billbonney, line 79: Undo deletion

ardupilot/source/index.rst
```diff
@@ -236,7 +236,6 @@ Features
    :caption: Ground Control Stations
 
    Mission Planner <https://ardupilot.org/planner/index.html>
-   APM Planner 2 <https://ardupilot.org/planner2/index.html>
```
billbonney, line 239: suggestion: re order links and add Other GCSs Link

tridge:
@billbonney can you check that mavlink and bin log message instances work? We use that a lot now, and will be using it more and more over time

PR #6959, opened 2025-06-28
Common: add vimdrones to stores list

rmackay9:
This adds VimDrones (run by Huibean) to our list of stores (VimDrones is a Partner)

BTW, 3DXR's position in the list was not in alphabetical order, it's unclear whether "3" means it should be at the top or the bottom.  I've put it at the bottom but I could move it to the top if people prefer.


Hwurzburg (changes requested):
numericals are usually listed first in alphabetizations, I think

rmackay9:
thanks!

Issue #6351, opened 2024-10-28
L298N Motor Driver instructions outdated/poor


yuri-rage:
In the [brushed motor section](https://ardupilot.org/rover/docs/common-brushed-motors.html) for Rover, we link to a [repository](https://github.com/jazzl0ver/ardupilot-rover-l298n) that discusses using a typically problematic 2.4.8 clone board and includes a script that looks like it probably works for skid steering but is written strangely for that purpose.

I'll dig around to see if I have any hardware that would support some testing of an updated script that we could include as an applet. If I achieve success, I'll provide an update.

I'm also happy to defer that work to another dev if someone beats me to it.

Hwurzburg:
Closing as resolved by #6961. The current Brushed Motors documentation now includes native L298N setup instructions, relay/output assignments, inversion settings, and a wiring diagram: https://ardupilot.org/rover/docs/common-brushed-motors.html#l298n-motor-driver

PR #6961, opened 2025-06-29
Brushed Motors: add L298N native support instructions

### LINKED ISSUE #1069: Rover: add wiring diagram for BrushedWithRelay motor type
Our Motors & Servo wiki page describes how to connect the flight controller to motor driver's using "Normal" pwm input (i.e. in the range 1ms ~ 2ms).  We should add wiring diagrams and instructions on how BrushedWithRelay and perhaps BrushedBiPolar motor types can be used.  Ideally we should include links to motor drivers that use these inputs.

These pages should be added to:
http://ardupilot.org/rover/docs/rover-motor-and-servo-connections.html
http://ardupilot.org/rover/docs/rover-motor-and-servo-configuration.html#rover-motor-and-servo-configuration

yuri-rage:
As of 4.6, AP supports inverted relay outputs, deprecating the previously linked script, which was an incomplete and slightly awkward implementation.

This should close #1069 and #6351. It also renders https://github.com/ArduPilot/ardupilot/pull/28488 unnecessary, though the PR script could be a means to more gracefully deprecate the previously linked script here.


rmackay9:
great to see this!

PR #6972, opened 2025-07-04
4.7:PiPilot

Hwurzburg:


rmackay9:
Ah, it's also called PilotPi strangely instead of PiPilot.

Hwurzburg:
@rmackay9 sorry, since this has docs in PX4, and I am now operating at 3X my AP designated hours, I cant author almost  almost everything any more....I am trying to get back to 2x max, so that is not on my list

PR #6974, opened 2025-07-04
4.7:EKF3 Source option

Hwurzburg:


Hwurzburg:
> Hi @Hwurzburg, thanks for this. Re the existing source set auxiliary switch, it actually has no effect I think if this option is set
> 
> FYI @rishabsingh3003

this option is used at ekf init...andI see nothing that prevents the aux switch from changing the source sets after,if used...

rishabsingh3003:
@Hwurzburg  I think Randy is correct, the source switch aux would basically be ignored if that bit is set..

PR #6984, opened 2025-07-05
Cruise alf floor units fix

landswellsong:
As per https://github.com/ArduPilot/ardupilot/commit/8151647e0465b5fe8b82b5f352f0f22b96025d33 the parameter is set in meters, I think that relates to 4.5.0 rescales of parameters away from centimeters. [The code](https://github.com/ArduPilot/ardupilot/blob/05dbacebd24e3354469a4c908fe8461d20ee1813/ArduPlane/altitude.cpp#L406) clearly expects meters in parameter too.


Hwurzburg:
thanks

PR #6991, opened 2025-07-06
Update common-power-module-configuration-in-mission-planner.rst

felixconsulting:
I added information about BATT_OPTIONS bit 0 and detailed why battery level can remain at 0% because of the droneCAN SOC's information being prioritized.

source: [https://discuss.ardupilot.org/t/battery-remain-percent-is-always-0/109230/4](https://discuss.ardupilot.org/t/battery-remain-percent-is-always-0/109230/4)


Hwurzburg:
lgtm..thanks

PR #6994, opened 2025-07-07
Update ros2.rst: Add ArduPilot Build Environment Setup to ROS2 Documentation

mueller-lukas:
## Summary
Fixes build errors when running `colcon build --packages-up-to ardupilot_dds_tests` by adding explicit instructions for setting up the ArduPilot build environment using the provided installation script referenced in [https://ardupilot.org/dev/docs/building-setup-linux.html#install-some-required-packages](https://ardupilot.org/dev/docs/building-setup-linux.html#install-some-required-packages)

## Problem
The current documentation only mentions "ensure you have set up your ArduPilot build environment" without providing specific instructions. This causes build failures later in the process because required dependencies are missing.

## Solution
- Added a new section with instructions for using the `install-prereqs-ubuntu.sh` script
- Positioned after the rosdep installation step, before MicroXRCEDDSGen installation
- Includes proper bash commands and a warning about Linux distribution updates

### DIFF
--- a/dev/source/docs/ros2.rst
+++ b/dev/source/docs/ros2.rst
@@ -55,6 +55,20 @@ Now update all dependencies:
     source /opt/ros/humble/setup.bash
     rosdep install --from-paths src --ignore-src -r -y
 
+Now we install some required packages:
+
+If you are on a debian based system (such as Ubuntu or Mint), we provide `a script <https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-ubuntu.sh>`__ that will do it for you. 
+This script does NOT support building on operating systems that have reached end of support such as Ubuntu Bionic (18.04).
+
+.. code-block:: bash
+
+    cd ~/ardu_ws/src/ardupilot/
+    Tools/environment_install/install-prereqs-ubuntu.sh -y
+    # Reload the path (log-out and log-in to make it permanent):
+    . ~/.profile
+
+.. warning:: if you update your Linux distribution to a later version, then you will need to repeat this (rerun the provided script) step to assure that all the packages needed will be re-installed for this Linux version
+
 Installing the `MicroXRCEDDSGen` build dependency:
 
 .. code-block:: bash


Hwurzburg:
@Ryanf55 review please

Ryanf55:
If you follow the link `https://ardupilot.org/dev/docs/building-the-code.html#setting-up-the-build-environment>` which is given as a prepreqquisite, then you run the script. I do not see value in duplicating more instructions, and this doesn't explain other things beyond running the setup script that are necessary. Perhaps there is a way to make it more clear that it's a prerequisite?

mueller-lukas:
Thanks for the feedback @Ryanf55! I still think this addition would be really helpful for users working through the ROS2 tutorial.

The core issue is that the current reference to _"ensure you have set up your ArduPilot build environment"_ is easy to miss, as it’s just a quick mention without clear action steps. For users following the ROS2 documentation linearly, I can imagine this prerequisite often slips through the cracks, and they only realize it’s missing when ``colcon build`` fails with error messages.

My goal is to make the setup process smoother, especially for newcomers who might not be familiar with the broader ArduPilot ecosystem. ROS2 integration is often their first touchpoint, and early build errors can be discouraging, especially since the instructions for getting ``ArduPilot + ROS 2 + SITL + Gazebo`` up and running are spread across multiple subpages, making it easy to miss steps.

I totally understand the concern about duplicating content. Maybe instead of repeating instructions, a prominent callout box or warning that explicitly directs users to run the installation script would be a good solution?
I’m open to any idea that makes this critical step harder to miss while keeping the documentation clean and maintainable.

Maybe something like
```
Install required packages:
If you have not already done so, run the install-prereqs-ubuntu.sh script `as instructed here <https://ardupilot.org/dev/docs/building-setup-linux.html#install-some-required-packages>`__ to install the required packages for your Debian-based system (e.g., Ubuntu).
.. warning::
The next steps, such as colcon build, will fail if the required packages are not installed. This script does NOT support end-of-life operating systems like Ubuntu Bionic (18.04). If you update your Linux distribution, you must rerun this script to ensure compatibility.
```

Hwurzburg:
@Ryanf55 ?? response

Ryanf55:
> Thanks for the feedback @Ryanf55! I still think this addition would be really helpful for users working through the ROS2 tutorial.
> 
> The core issue is that the current reference to _"ensure you have set up your ArduPilot build environment"_ is easy to miss, as it’s just a quick mention without clear action steps. For users following the ROS2 documentation linearly, I can imagine this prerequisite often slips through the cracks, and they only realize it’s missing when `colcon build` fails with error messages.
> 
> My goal is to make the setup process smoother, especially for newcomers who might not be familiar with the broader ArduPilot ecosystem. ROS2 integration is often their first touchpoint, and early build errors can be discouraging, especially since the instructions for getting `ArduPilot + ROS 2 + SITL + Gazebo` up and running are spread across multiple subpages, making it easy to miss steps.
> 
> I totally understand the concern about duplicating content. Maybe instead of repeating instructions, a prominent callout box or warning that explicitly directs users to run the installation script would be a good solution? I’m open to any idea that makes this critical step harder to miss while keeping the documentation clean and maintainable.
> 
> Maybe something like
> 
> ```
> Install required packages:
> If you have not already done so, run the install-prereqs-ubuntu.sh script `as instructed here <https://ardupilot.org/dev/docs/building-setup-linux.html#install-some-required-packages>`__ to install the required packages for your Debian-based system (e.g., Ubuntu).
> .. warning::
> The next steps, such as colcon build, will fail if the required packages are not installed. This script does NOT support end-of-life operating systems like Ubuntu Bionic (18.04). If you update your Linux distribution, you must rerun this script to ensure compatibility.
> ```

If you can rework this to have a callout box or other method without duplicating content, I would accept the contribution.

PR #6996, opened 2025-07-10
GitHub Actions: Lint Python code with ruff

cclauss:
https://github.com/astral-sh/ruff

Why ruff?
* https://github.com/mavlink/mavlink/pull/2303#issuecomment-3029944846

If ruff is a superset of flake8, why doesn't it complain about the rules in this repo's `.flake8`? https://github.com/ArduPilot/ardupilot_wiki/blob/703219ca36a9f8df9741f19810965aa7da2ea8f7/.flake8#L1-L16

Most of these are whitespace errors that are better handled by a code formatter like ruff format or psf/black, so ruff check ignores them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,11 @@ jobs:
                   exit 1
               fi
           done
-      - name: Check python3 flake8 formatting
-        run : |
+      - name: Check python formatting
+        run: |
           set -eux -o pipefail
           flake8 .
+          pipx run ruff check
       - name: Test build
         run: |
           export AP_SPHINXOPTS=-W


peterbarker:
Query: do we need to use random isolation tools like pipx to do this stuff?

If I'm thinking right this actually increases the changes of the job failing as it installs the tool from the internet as part of running `pipx` here - in a CI job that's evrey time we run the job.  Th jobs on the main repository fail not infrequently because of external dependencies (including failed connections to github.com resources themselves!) so I don't know if this is an appropriate way of running the tool.  I'm certainly not growing fonder of restarting CI jobs that have failed with "Connection refused" to some random thing on the internet.

Installing it on the runner image would seem to be a better option.

@khancyr hard to install `ruff` on the image we use for testing the Wiki?

Hwurzburg:
@cclauss no need for merge commits....it just cluters up....should be a single commit for this PR...squash please

peterbarker (changes requested):
We are non-compliant.

We've already broken vast amounts of stuff with a previous merge, where I had assumed that because this PR passes CI everything was OK.

peterbarker:
> We are non-compliant.
> 
> We've already broken vast amounts of stuff with a previous merge, where I had assumed that because this PR passes CI everything was OK.

... nope, I ran it on the wrong repo.  *sigh*

PR #7008, opened 2025-07-14
Add missing SID_AXIS param settings to table

bnsgeyer:
The listing of the available options for the system ID SID_AXIS parameter was updated to be a table but not all of the axis settings were included.  This PR updates the table to include the missing position controller axis settings.


Hwurzburg:
thanks

PR #7010, opened 2025-07-15
common_theme_override.css: fix anchor link positioning

ES-Alexander:
The existing "hack" wasn't working for me (anchor links were still ending up underneath the top menu when jumping to them), and there is [a dedicated feature for this](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top), which has been well-supported by browsers since 2021.

### DIFF
--- a/common/source/_static/common_theme_override.css
+++ b/common/source/_static/common_theme_override.css
@@ -54,11 +54,8 @@ table.partners-table td {
     }
 }
 
-/* wiki top menu makes anchor links disappear under it, this hack fixes it */
-.section {
-    padding-top: 40px;
-    margin-top: -40px;
-}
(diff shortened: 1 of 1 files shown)


Hwurzburg:
dont understand the issue

ES-Alexander:
@Hwurzburg currently, when I use a URL to jump to a heading, the heading gets hidden behind the top bar (i.e. the "hack" in the CSS which is intended to correct for that doesn't work on my machine):
Screenshot 2025-07-16 at 3 45 07 am

With this change, the heading shows up properly (and uses an intended CSS feature to achieve it, instead of a hack):
Screenshot 2025-07-16 at 3 50 06 am

PR #7019, opened 2025-07-16
Update of Plane PID diagram

grep265:
This is an update of the ArduPlane PID schematic to reflect the speed scaling and slew rate limiter.  It also includes a brief description of both features to guide the user during the tuning process.


plane/source/docs/new-roll-and-pitch-tuning.rst
```diff
@@ -223,6 +223,24 @@ The new PID-FF controller in Plane also has a number of lowpass filters whose fr
 - :ref:`RLL_RATE_FLTE<RLL_RATE_FLTE>`: filter on roll rate error
 - :ref:`RLL_RATE_FLTD<RLL_RATE_FLTD>`: filter on D term (tends to be noisy)
 
+Speed Scaling (SS)
+==================
+
+ArduPilot automatically adjusts the stabilization gains in the PID-FF loops based on the airspeed. This helps avoid oscillations when
```
IamPete1, line 229: I think its worth adding a note or warning that changing this parameter will mess up the tune.

images/PID-controller.jpg
IamPete1, line 1: Makes no difference to the result, but the FLTT and SS^2 should be the other way round.

plane/source/docs/new-roll-and-pitch-tuning.rst
Hwurzburg: should say changing requires re-running AUTOTUNE

images/PID-controller.jpg
grep265: I updated the diagram. Thanks

plane/source/docs/new-roll-and-pitch-tuning.rst
grep265: I added a warning message with your suggestions. Thanks

Hwurzburg (approved):

images/PID-controller.jpg
isso11: @grep265
Hey, thanks for sharing this diagram 🙌
I compared it with master and noticed a few small differences:

1- Target Rate Scaling
 → The diagram scales `q_cmd `before the PID using `SS `and `1 / (SS × EAS2TAS`),
 → But in master, the target is unscaled. Aairspeed scaling happens only at the final output.

2- I-Term Clamping
 → master clamps the integrator to `±IMAX`, not shown here.

Overall though.. really nice breakdown, captures the structure well 👏

isso11: This is just my attempt at updating the diagram after checking and aligning it with the current master code.
Happy to hear any thoughts or corrections!

Picture4

grep265: Thanks for the comments and schematic. 

I double checked the master branch and I updated the diagram. Here is a list of some of the changes:

1. Add Anti-windup
2. Fix the units
3. Add rad to deg conversion at the output
4. Update FF component
5. Add PD scale limit

The scaling speed is applied on the desired and measured rates. At the end the output gets converted from radians to degrees. Also, I was thinking about a more general and simplified schematic that applies for roll and pitch. That's why I didn't include features such as the coordination rate, ground mode suppression, etc. I can add them if needed.

isso11, line 1: @grep265 
I believe the _DFF term is applied to the derivative of the target rate, not the target rate itself.
Also, I couldn’t find where the SS scaling is applied to the target or actual rates.. would you mind pointing me to that part if I missed it?

grep265: Updated. Thanks again, I appreciate it.

Ofc. Check [this line](https://github.com/ArduPilot/ardupilot/blob/32c3bfa4b9086d2890e06811de4ac3c45dc39fb1/libraries/APM_Control/AP_FW_Controller.cpp#L55)

Hwurzburg:
@IamPete1 okay to merge?

grep265:
Updated. Thanks.

Hwurzburg:
still awaiting @IamPete1 approval before merging

IamPete1 (commented):
Looks like we don't do the ```1/ ss * eas2tas``` for DFF. Maybe that is a bug and we should do it per your diagram.

It would also be nice to have the file from the tool that was used to create this diagram in tree somewhere, so if we need to update it in the future its easy rather than having to start over.

IamPete1:
> Do you mind if I submit a PR suggesting a fix?

A PR would be great!

> Could anyone please suggest a location to upload the file of the diagram? I made it in PowerPoint.

I guess it could in a folder in the wiki somewhere, @Hwurzburg any thoughts? If nothing else then you could upload it here (if GitHub allows).

Hwurzburg:
pr it to the /data folder....I will rename it /misc...or you can in the pr and I will approve and merge

grep265:
Thanks to both of you. I added the file and submitted a PR.

Hwurzburg (approved):

PR #7020, opened 2025-07-19
Adding Beyond Robotix AUAV sensor

o-gent:
We've created hardware around the new AUAV sensor. the sensor has both dynamic and static pressure measurement meaning it can be used for both airspeed and barometer. We've combined the hardware with a can node so the sensor has a DroneCAN interface.


common/source/docs/common-baro-external.rst
```diff
@@ -15,7 +18,8 @@ Most autopilots include a barometer directly on the autopilot but attaching an e
 Where to Buy
 ------------
 
-- `MS5611 from Drotek <https://drotek.com/shop/en/imu/44-ms5611-pressure-barometric-board.html>`__ (recommended)
+- `Beyond Robotix Air Data Module (AUAV) <https://www.beyondrobotix.com/products/air-data-module>`_
```
Hwurzburg: this should in a DroneCan catergory, not with the i2c sensors...

plane/source/docs/airspeed.rst
```diff
@@ -173,6 +173,9 @@ I2C
 DroneCAN
 --------
 
+- AUAV (static and dynamic pressure)
```
Hwurzburg: does this use a std sensor, like from Amphenol? then a separate caterory naming the sensor should be used...unless AUAV markets a new chip of their own in which case we would need to also link the chip datasheet also

Hwurzburg (changes requested):
this needs to be added to tthe DroneCan peripherals page also

common/source/docs/common-baro-external.rst
o-gent: Ah ok, so this external barometer page is for i2c sensors only? it would be quite good to have a general page about external barometers ... but if not I can remove it

plane/source/docs/airspeed.rst
o-gent: I tried to follow the existing structure, for example the next item is DVLR sensors.. So I made a new category with AUAV. Where would you suggest a new header should be?

common/source/docs/common-baro-external.rst
o-gent: If you'd like, i can make some changes to the page to explain more about external barometers, why you'd use them etc

Hwurzburg:
@o-gent tell me which sensor IC this module uses, I will fix up the PR and merge (if yo give me permission in your github...you did not check the allow devs to make changes box it appears)

o-gent:
Thank you! I'll try and change that, sorry didn't realise it was an option. Might be when I'm back at my laptop this evening.. 


Here's the sensor details, we can fit any of the speed variants. They just called it the AUAV sensor.. 

https://allsensors.eu/wp-content/uploads/2024/07/PR-AUAV.pdf

https://www.digikey.lt/en/product-highlight/a/amphenol-all-sensors/auav-dual-air-speed-and-altitude-pressure-sensor

o-gent:
You have write access to our repository.. not sure what else I can do but I'll have a better look when I'm back home

o-gent:
All squashed and added Amphenol to each instance

PR #7026, opened 2025-07-24
common-upixels-tx-series.rst: fix the errors in the document

Ainnety:
common-upixels-tx-series.rst: fix the errors in the document


Hwurzburg:
thanks

PR #7030, opened 2025-07-26
Update documentation of the parameters that limit the steering

grep265:
This updates the description of key parameters for tuning ground steering by adding their full names, links, and units.


Hwurzburg:
thanks

PR #7036, opened 2025-08-01
Sub: cover default RC channel mappings

ES-Alexander:
Update RCMAP and aux func defaults for Sub vehicles.


Hwurzburg:
thanks...LGTM

PR #7050, opened 2025-08-11
Update ros2.rst

frieddeeu:
Use a pinned version of OpenJDK, 22.04 default-jre does not work.


Hwurzburg:
@Ryanf55 review requested

Hwurzburg:
@frieddeeu ???

PR #7043, opened 2025-08-08
document network capture

tridge:
https://github.com/ArduPilot/ardupilot/pull/30637

PR #7057, opened 2025-08-14
Document network capture""

Hwurzburg:
Rcovers ArduPilot/ardupilot_wiki#7043


rmackay9:
I think maybe we also need to mention that the file is delete and re-created once it reaches 1MB?  While using this I found that I sometimes failed to download the file because it had disappeared and it seemed to me this happened when it was close to 1MB in size

PR #7058, opened 2025-08-14
add readme.md templare and rename data dir to misc

Hwurzburg:
I created README.md template for a fictional XYZ board with everything I could think of for required info/sections  and formatted such that conversion to .rst is minimalized in terms of effort...


misc/readme_template.md
```diff
@@ -0,0 +1,184 @@
...
+|6     |  USART3 |  RCin          |  ✔    |   ✔   |
+|7     |   USB   |  SLCAN         |  ✘    |   ✘   |
+
+USART7 and USART 2 have flow control pins.
+
+RC input
+========
+<!--This is the most difficult section and varies widely.-->
+
+The SBUS pin, can be used for all ArduPilot supported receiver protocols, except CRSF/ELRS and SRXL2 which require a true UART connection. However, FPort, when connected in this manner, will only provide RC without telemetry.
```
andyp1per: This reads quite confusingly to me, I feel like it could be clearer

Hwurzburg: thanks will change

Hwurzburg: thanks....fixing

PR #7059, opened 2025-08-15
Add Rover speed controller diagram

grep265:
This adds the controller schematic of the Rover speed controller. It also includes some minor changes in the titles to match the diagram and fix the units.


rmackay9 (approved):
Thanks for this.  I think this looks correct

PR #7082, opened 2025-09-02
dev: Fix spelling and pinout number typo

Ryanf55:
FYI @srmainwaring  you found the pinout number issue, thanks for the spot!


Hwurzburg:
thanks

PR #7116, opened 2025-09-17
Sub: add actuator button functions

ES-Alexander:
ArduSub 4.5.5 renamed the old `servoN_*` (N=1/2/3) functions to `actuator_N_*`, and added options for N=4/5/6

I had to make the second column wider to fit the new options, so the whole table is flagged as changed.


Williangalvani (approved):

Hwurzburg:
thanks

PR #7119, opened 2025-09-19
common-ice: expand and update documentation

robertlong13:
When discussing https://github.com/ArduPilot/ardupilot/pull/30201 on the dev call, Tridge requested that I overhaul the whole ICE wiki page, and go into more detail on disarmed operation.

I've been through the source in detail and tried to make sure every nuance is captured here.

In the past I really struggled to get the ICE library to play nicely with handheld starters when you also have an RPM sensor. The approach I added in here has worked well for me for such systems.


Hwurzburg:
havent reviewed fully but all parameters that can be, should be properly linked, ie
instead of ``ICE_START_PCT``  it should be :ref:`ICE_START_PCT<ICE_START_PCT>`

robertlong13:
Some of the history there
https://github.com/ArduPilot/ardupilot/pull/11109
https://github.com/ArduPilot/ardupilot/pull/25053
https://github.com/ArduPilot/ardupilot/pull/25138

I don't think the idle percentage parameter should ever have been tied to the behavior of whether to run while disarmed, but it's been in there long enough that I doubt I can fix that without breaking current users' setups.

robertlong13:
I've corrected it now. Hopefully in a way that isn't terribly confusing.

PR #7133, opened 2025-09-27
Common: Caddx input channel requirement added

rmackay9:
This adds an important setting to the CADDX gimbal setup page which is required for the most recent version of the manufacturer's gimbal software.  The related user discussion is here https://discuss.ardupilot.org/t/caddx-gm3-does-not-work-with-firmware-3-4/137918

I've tested the settings change and built this page locally and it all looks OK to me.


common/source/docs/common-caddx-gimbal.rst
```diff
@@ -65,6 +65,8 @@ Firmware upgrade instructions are in the "Upgrade" section of the `user manual <
     :target: ../_images/caddx-firmware-update-PC.png
     :width: 450px
 
+- Ensure the channel down-downs highlighted in orange above are set to CH01
```
Hwurzburg:
```suggestion
- Ensure the channel drop-downs highlighted in orange above are set to CH01
```

Hwurzburg (dismissed):

common/source/docs/common-caddx-gimbal.rst
rmackay9: thanks for that, fixed!

PR #7134, opened 2025-09-27
Dev: BlueOS setup improvements

rmackay9:
This adds two small improvements to the BlueOS setup instructions:

1. Small section highlighting that the Ochin carrier board's ethernet solder bridge must be soldered
2. BlueOS initial setup gets info on using the BlueOS's wifi AP which is slightly easier than directly connect a PC to the I/O board

I've built this page locally and it looks OK to me


Hwurzburg:
LGTM...thanks

rmackay9:
Thanks!

PR #7135, opened 2025-09-27
Common: optical flow setup instruction improvements

rmackay9:
This includes a few changes to the optical flow setup pages based on difficulties that @andyp1per recently experienced.  I don't think these changes are sufficient but they're a start:

1. move the rangefinder check section up to be just below the raw sensor value check
2. improve the sensor check section to include setting the EKF origin and recommend moving the vehicle around to confirm the vehicle icon on the map also moves
3. remove the out-of-date suggestion to change the arming checks.  This is not required

As I said, these updates are not complete so any other comments from AndyP are very much welcome!


common/source/docs/common-optical-flow-sensor-setup.rst
```diff
@@ -14,6 +16,32 @@ With the sensor connected to the autopilot, connect to the autopilot with the Mi
...
+
+Check the rangefinder's reported distance is between the :ref:`RNGFND1_MIN <RNGFND1_MIN>` and :ref:`RNGFND1_MAX <RNGFND1_MAX>` parameter values when on the ground and when held up
+
+.. image:: ../../../images/mp_rangefinder_lidarlite_testing.jpg
+    :target: ../_images/mp_rangefinder_lidarlite_testing.jpg
+    :width: 250px
+
+You may also download an onboard log and check the ``EKF5.meaRng`` message to confirm
+
+#. There is continuous range measurement
```
andyp1per, line 38: Might need a comment here about the special case of 0.

andyp1per: So I think min->max should set the usable range and RNGFND1_GNDCLEAR should set the value on the ground (which is 0 in my case)

PR #7145, opened 2025-10-01
Dev: add simulated slung payload info

rmackay9:
This adds info to the dev wiki on how to setup a simulated slung payload.  I've also made a small correction to the user facing slung payload page.

I noticed these issues as I prepare for my annual search and rescue competition.


Hwurzburg (changes requested):
All params should be hotlinked

PR #7168, opened 2025-10-14
dev: BlueOS LTE setup gets minor improvement

rmackay9:
This improves the BlueOS LTE setup instructions a bit:

1. image shows pin assignments for the connection between the Ochin carrier board and LTE modem
2. new images shows that a "Dynamic IP" is used


rmackay9:
Thanks!

PR #7172, opened 2025-10-15
Copter: Blueos Precision Landing

rmackay9:
This adds a new page to Copter's Precision Landing section to cover the setup using BlueOS.  This is a 4.7 feature so we could delay for a few weeks before merging this

I also made a small drive-by fix to the IR-Lock title underlining


Hwurzburg:
the other one is one the 4.7 list and is covered by another PR from you already....just didn't mention BlueOS, so that confused me....and this PrecLand one I did not mark as wiki needed since it did not seem to have user visible impact, so this PR  for BlueOS is great...

PR #7183, opened 2025-10-16
Dev: improve ROS2-SITL adding  simulator build instructions and noting DDS configuration differences for ArduPilot 4.7 and earlier

ZhongmouLi:
This PR updates the `ROS 2 with SITL` and `Install ROS 2` setup pages to address several issues encountered by both me and @M-Wasif-Raza  during setup and testing..

For `ROS 2 with SITL` ,
1. detail the building instructions for SITL using the source code cloned instructed in the  `Install ROS2` and remove hyperlink of sitl-simulator-software-in-the-loop.
2. clarify the DDS enablement differences between ArduPilot 4.7 and earlier versions such as 4.6.2.
3. explain the functions of the ROS 2 packages involved in launching SITL under ROS 2
4. add methods of changing Ardupilot parameters in SITL wih mavproxy (with example), Mission Planner etc. 

For `Install ROS2` ,

1. include explicit git clone commands to illustrate the vcs import --input Tools/ros2/ros2.repos step and show the two repositories being cloned, which will be reused in `ROS 2 with SITL` to avoid redundancy 
2. add a tag to `Installation (Ubuntu)` to allow cross-referencing from ROS 2 with SITL
3. remove the redundant  `Set up your ArduPilot build environment` subsection (now fully detailed in ROS 2 with SITL)
4. keep `Micro-XRCE-DDS-Gen` usage for consistency with the existing document style

The steps have been tested in SITL with ROS2 humble.

### DIFF
--- a/dev/source/docs/ros2-sitl.rst
+++ b/dev/source/docs/ros2-sitl.rst
@@ -35,7 +35,7 @@ Then, build Ardupilot for SITL with DDS enabled. The example below shows how to 
 
         cd ardu_ws/src/ardupilot
         ./waf distclean
-        ./waf configure --board sitl --enable-dds # Note --enable-DDS for version after 4.7
+        ./waf configure --board sitl --enable-DDS # Note --enable-DDS for version after 4.7
         ./waf copter
 
 You can build the ROS2 packages `ardupilot_msgs`, `micro_ros_agent`, `ardupilot_sitl` and `ardupilot_dds_tests` as 


ZhongmouLi:
@Hwurzburg  could you please assign a reviewer for this PR? Thanks in advance.

dev/source/docs/ros2-sitl.rst
```diff
@@ -4,9 +4,57 @@
...
+
+
+   .. group-tab:: After Ardupilot 4.7
+
+      .. code-block:: bash
+
+        cd ardu_ws/src/ardupilot
+        ./waf distclean
+        ./waf configure --board sitl --enable-DDS # Note --enable-DDS for version after 4.7
+        ./waf copter
```
Ryanf55, line 39: Once you are in a ROS 2 workspace, there is no reason to build this way.
If you do this, then colcon build later, you end up building the same repo twice.

dev/source/docs/ros2-sitl.rst
```diff
@@ -18,7 +66,19 @@ You will need to run this command on every new shell you open to have access to
         cd ~/ardu_ws/
         colcon build --packages-up-to ardupilot_sitl
         source install/setup.bash
-        ros2 launch ardupilot_sitl sitl_dds_udp.launch.py transport:=udp4 refs:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/dds_xrce_profile.xml synthetic_clock:=True wipe:=False model:=quad speedup:=1 slave:=0 instance:=0 defaults:=$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/copter.parm,$(ros2 pkg prefix ardupilot_sitl)/share/ardupilot_sitl/config/default_params/dds_udp.parm sim_address:=127.0.0.1 master:=tcp:127.0.0.1:5760 sitl:=127.0.0.1:5501
+        ros2 launch ardupilot_sitl sitl_dds_udp.launch.py \
```
Ryanf55, line 69: I like these formatting changes, these could be merged on their own.

dev/source/docs/ros2-sitl.rst
```diff
@@ -51,18 +128,22 @@ Once everything is running, you can now interact with ArduPilot through the ROS
     # Echo a topic published from ArduPilot
     ros2 topic echo /ap/geopose/filtered
 
-If the ROS 2 topics aren't being published, ensure the ardupilot parameter ref:`DDS_ENABLE<DDS_ENABLE>` is set to 1 and reboot the launch.
+DDS is responsible for ROS2 communication. Therefore, DDS must be enabled by setting the Ardupilot parameter `DDS_ENABLE` and make the parameter `DDS_DOMAIN_ID` match `ROS_DOMAIN_ID` of your ROS2 settings.
+
+For instance, if the ROS 2 topics aren't being published, first check if `DDS_ENABLE` is set to ``1`` or not. If not, set as ``1``. It can be done through Mission Planner, QGroundControl or the `param <https://ardupilot.org/mavproxy/docs/getting_started/cheatsheet.html>`__ command in the MAVProxy console. The second aspect to check, ensure the ArduPilot parameter `DDS_DOMAIN_ID` matches your environment variable ``ROS_DOMAIN_ID``. The default is ``0`` for ArduPilot, which corresponds to the environment variable being unset. You may need to relaunch the SITL after changing the parameters.
```
Ryanf55, line 133: I don't think it's necessary to explain how to use MAVProxy here. The parameter DDS_ENABLE is enabled by default, documented that it needs reboot, so I don't think this is actually making anything clearer.

ROS_DOMAIN_ID defaults to "", unset, not 0. They result in the same thing, but the info you supllied is not quite correct.

dev/source/docs/ros2-sitl.rst
```diff
@@ -51,18 +128,22 @@ Once everything is running, you can now interact with ArduPilot through the ROS
...
+To change the parameters using `mavproxy`, you can launch `mavproxy` in another terminal and connect to SITL through UDP
 
 .. code-block:: bash
 
     export PATH=$PATH:~/ardu_ws/src/ardupilot/Tools/autotest
-    sim_vehicle.py -w -v ArduPlane --console -DG --enable-dds
+    mavproxy.py --master udp:127.0.0.1:14550
 
     param set DDS_ENABLE 1
+    param set DDS_DOMAIN_ID 1 # if you set export ROS_DOMAIN_ID=1
```
Ryanf55, line 143: I would not put this here. People will blindly copy that. 

Previously, this code block was for `If the ROS 2 topics aren't being published, ensure the ardupilot parameter ref:`DDS_ENABLE<DDS_ENABLE>` is set to 1 and reboot the launch.`

Now, you are using it to show how to change parameter that most people don't need to do.

I think we should just delete the block now that DDS_ENABLE is defaulted to 1.

dev/source/docs/ros2.rst
```diff
@@ -43,9 +44,20 @@ To make installation easy, we will clone the required repositories using `vcs` a
...
 
-Now update all dependencies:
+You can also do it manually by cloning the following repositories into the `src` folder of your workspace:
+
+.. code-block:: bash
+    
+    mkdir -p ~/ardu_ws/src
+    cd ~/ardu_ws
+    git clone  --recurse-submodules -b master https://github.com/ArduPilot/ardupilot.git src/ardupilot
+    git clone --recursive -b humble https://github.com/micro-ROS/micro-ROS-Agent.git src/micro_ros_agent
```
Ryanf55, line 56: Remove this. I don't want people cloning manually. We should use vcs when working with ROS 2. 
Otherwise there are now multiple places to maintain the repos. If we are going to support manually cloning, it might be worth showing how to symlink an existing ardupilot repo into a ROS 2 workspace layout, but that is definately an advanced use case that would confuse the heck out of most devs.

dev/source/docs/ros2-sitl.rst
ZhongmouLi: @Ryanf55 , thanks for your comment. A new PR is submitted just for this format at https://github.com/ArduPilot/ardupilot_wiki/pull/7220#issue-3576618982

PR #7197, opened 2025-10-22
MAV_CMD_DO_CHANGE_ALTITUDE in Plane

peterbarker:


Hwurzburg (changes requested):
add to mission command list

Hwurzburg:
I was going to add the info here for you, but I realized during research this is not a mission command but a MAVLink only command....so I added in the wiki notes as needed to clarify mission vs MAVLink...merging...thanks

PR #7203, opened 2025-10-24
4.7/29843/roll limits option

rubenp02:
Fixed minor punctuation and consistency issues, and updated the contents to include bits 15 and 16. The addition of bit 16 is the wiki change needed for ArduPilot/ardupilot#29843.


Hwurzburg:
> I'll update those other pages as well. Is this a 4.6 backport? I found it on the Discord list for the wiki requirements for the 4.7 beta. In terms of the conflict with #7016, note that this also does a cleans up the puctuation and spacing of the text as well. So ideally the second commit of this PR should be dropped and replaced with the one in #7016.

no its 4.7...too many lists on my notpad..I changed the title (titles are really just for my convenience when searching)..just copy the same bit 15 as in 7016 and other wont be a merge conflict....adding to places that reference ROLL_LIMIT_DEG and adding info on how enabling this option can reduce the limit (FBWA doc, Navigation) should be enough....I don't think you have to add everywhere stabilization is used(AUTO, etc.)

PR #7217, opened 2025-10-31
Fix LOITER_TO_ALT

robertlong13:
also driveby clarification/fix of the crosstrack parameter description for LOITER_TURNS


robertlong13:
This broke in https://github.com/ArduPilot/ardupilot_wiki/pull/6535. I'm guessing you were just trying to delete the old Heading Required in param1, which was removed in the code long ago, but the Radius should not have moved; Radius has always been param2 on this.

ArduPilot source: [ArduPilot/ardupilot@d2a361a/libraries/AP_Mission/AP_Mission.cpp#L1163-L1168](https://github.com/ArduPilot/ardupilot/blob/d2a361afa744223354faa2cf989740458fe5b38c/libraries/AP_Mission/AP_Mission.cpp#L1163-L1168)

MAVLink spec:
[mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT)

PR #7231, opened 2025-11-10
Common: watchdog reporting clarified

rmackay9:
This attempts to resolve issue https://github.com/ArduPilot/ardupilot/issues/31404 by making it more clear to users how to to report, erase or ignore the crash dump file

This has been tested locally and it looks OK to me.  I've also done some drive-by changes to wording on the page to fix grammatical errors and make it more succinct.

### DIFF
--- a/common/source/docs/common-watchdog.rst
+++ b/common/source/docs/common-watchdog.rst
@@ -4,18 +4,40 @@
 Independent Watchdog and Crash Dump
 ===================================
 
-In ArduPilot 4.0 and later firmware revisions, the autopilot cpu's internal independent watchdog has been enabled. It can be disabled (NOT recommended!) by setting :ref:`BRD_OPTIONS<BRD_OPTIONS>` = 0. The cpu will be reset if a peripheral or code takes too long to complete its scheduled software thread or "hangs" the cpu, and will restart the cpu. This may or may not save the vehicle if in flight or motion.
+ArduPilot enables the autopilot CPU's internal independent watchdog which causes the CPU to be reset and restarted if a peripheral or some piece of code takes too long to complete or "hangs" the CPU.  This may or may not save the vehicle if in flight or motion.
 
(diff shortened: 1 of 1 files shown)


rmackay9:
@Hwurzburg,

AP attempts to create a crash_dump file on every watchdog but it may not always succeed.  The most common reason that it won't succeed is if the autopilot has no flash space left because we've used it all up for code :-)

common/source/docs/common-watchdog.rst
```diff
@@ -4,19 +4,41 @@
...
 
-.. warning:: In the case of a "HARD FAULT" (illegal instruction, out of bounds memory access,etc.) instead of a delay in completing a thread in timely manner, before resetting, the code will attempt to write a file called crash_dump.bin in the flash capturing the cpu state and registers for later analysis (See section below for more details). A "HARD FAULT" is extremely serious and should be reported to ArduPilot. The vehicle is probably not safe to fly until the cause is resolved.
+While not recommended, this feature can be disabled by setting :ref:`BRD_OPTIONS<BRD_OPTIONS>` = 0.
 
+In the case of a "HARD FAULT" (e.g an illegal instruction, out of bounds memory access, etc.) before resetting, the watchdog handler code will attempt to write a crash_dump.bin file to the "@SYS" flash area.  This file contains the cpu state and registers for later analysis (See section below for more details). A "HARD FAULT" is extremely serious and should be reported to ArduPilot. The vehicle is probably not safe to fly until the cause is resolved.
```
Hwurzburg:
```suggestion
In the case of a "HARD FAULT" (e.g an illegal instruction, out of bounds memory access, etc.) a "watchdog" event will occur and log it and send a GCS "WDG" message, and reset the autopilot.  Before resetting, the watchdog handler code will attempt to write a crash_dump.bin file to the "@SYS" flash area.  This file contains the cpu state and registers for later analysis (See section below for more details). A "HARD FAULT" is extremely serious and should be reported to ArduPilot. The vehicle is probably not safe to fly until the cause is resolved.
```

rmackay9: ok, thanks for the review.  I've updated the wording a bit although it's still not exactly as you suggest.  I'm not sure it helps to tell users about "a watchdog event" because they don't really know what events are.  The new wording does more completely tell them what will happen (e.g. WDG log, WDG text message, crash_dump.bin file)

PR #7241, opened 2025-11-20
Update/Expand Mount page explanations

Hwurzburg:


common/source/docs/common-mount-targeting.rst
```diff
@@ -1,72 +1,84 @@
 .. _common-mount-targeting:
 
-
 =======================
 Gimbal / Mount Controls
 =======================
+ArduPilot provides mechanisms to control the pointing direction (aka targeting) automatically for camera mounts (know hereafter as ``mounts``), as well as by pilot commands/controls. Mounts can be commanded point in at least six different ways (``targeting modes``) by ArduPilot.  This page provides an overview of these controls and their setup.
```
rmackay9, line 6: small typo, "Mounts can be commanded **to** point in ...".  If you're editing that line, I'd probably also remove "by ArduPilot" because this is an AP wiki so it goes without saying

common/source/docs/common-mount-targeting.rst
```diff
@@ -1,72 +1,84 @@
...
+
+.. note:: ArduPilot 4.5 (and higher) automatically switches the mount to RC Targeting Mode if the pilot moves any configured Roll/Pitch/Yaw RC targeting inputs (see below) by the larger of 10uS or ``RCx_DZ``.  The only exception is if the mounts is in RETRACT mode in which case the mount mode will not be automatically changed.
 
-While the gimbal is in "RC Targeting" mode (see above for how to change modes), the pilot can control the gimbal's target roll, pitch and yaw angles using an RC transmitter channel for each axis. For example (RC channels shown are for example only, and can be any channel not otherwise used):
+Mount Internal Axis Locks
+=========================
+The above modes concern how pitch, roll, and yaw axis *TARGETs* are sent to the mount. But many mounts have internal controls for how those targets for pitch, roll, and yaw are interpreted. Those positioning commands can be interpreted by the mount as being with respect to Earth frame reference or Body frame. Body frame are angle targets on each axis measured with respect to the mount itself, while Earth frame is interpreted to be with respect the horizon and compass heading.
```
rmackay9, line 48: I'm not sure we should actually explain the internal controls.  It's just confusing and users don't really need to know

common/source/docs/common-mount-targeting.rst
```diff
@@ -77,55 +89,59 @@ Mission Planner uses MAVLink commands to control the gimbal.  The gimbal's angle
...
 - Check "Enable on Screen Camera Control" and set "Control type" to "Click to point"
 - Enter the camera's Horizontal FOV and Vertical FOV
-- Maximise the real-time video and click anywhere on the screen and the gimbal should move appropriately
+- Maximise the real-time video and click anywhere on the screen and the mount should move appropriately
 
 .. image:: ../../../images/mount-targeting-qgc-clicktopoint.png
     :target: ../_images/mount-targeting-qgc-clicktopoint.png
 
-To point the gimbal at a Location
+To point the mount at a Location
```
rmackay9, line 122: is replacing "gimbal" with "mount" a move forwards or backwards? :-).  we call it "mount" in ArduPilot but i think that was a mistake made long ago.  I think the more commonly used term is "gimbal" and we should move towards that

common/source/docs/common-mount-targeting.rst
Hwurzburg: fixed typo

I disagree, users have already demonstrated that they do not understand the difference in what AP is doing versus how they have configured the mount with its setup program, if it exists...I  think this is necessary clarification...and its not incorrect.

common/source/docs/common-mount-targeting.rst
Hwurzburg: I have added a note in the RC targeting section and removed this...

common/source/docs/common-mount-targeting.rst
Hwurzburg: at the very top, we state that "mount" is synonymous with "gimbal" throughout this document....

I agree Mount was a poor prior choice, but unless we are willing to change all the MNTx params to GIMx or something else, it will still be the same potential confusion....plus "mount" is still used throughout the wiki....and it does apply to non-gimbaled retractable mountings..so there is that

PR #7242, opened 2025-11-20
Common: rplidar recommends loop rate of 400 for rovers

rmackay9:
This adds a recommendation that the main loop rate be increased to 400 on rovers and boats in order to consume all the data from the sensor.

This is the result of [this support discussion](https://discuss.ardupilot.org/t/object-avoidance-bug/140929)

We don't need to recommend this for Plane because it doesn't support proximity sensors as far as I know


rmackay9:
thanks!

PR #7252, opened 2025-11-23
4.7:31563/ Compass X not healthy

muramura:
https://github.com/ArduPilot/ardupilot/pull/31563

In earlier versions, this message appeared as “Compass not healthy”, which made it difficult to identify which compass was reporting an issue on systems equipped with multiple compasses.

To address this, ArduPilot now includes the compass index in the message (e.g., “Compass 1 not healthy”, “Compass 2 not healthy”). This allows users to easily determine which compass is unhealthy and troubleshoot the correct hardware.


Hwurzburg:
Thanks for this

Issue #6065, opened 2024-06-24
Dev: list firmware-supported MAVLink messages and commands


ES-Alexander:
Near term:
- [x] Create firmware source parser for messages and commands that each vehicle type handles and/or sends
    - ArduPilot/ardupilot#27226
- [x] Store those outputs somewhere for existing firmwares
    - ArduPilot/ParameterRepository#7
- [ ] Add per-vehicle dropdown version selection support, [like the parameter lists](https://github.com/ArduPilot/ardupilot_wiki/blob/master/build_parameters.py#L338)
- [ ] Include in the wiki
    - Pull the files in during generation
    - Vaguely relevant to #1884, although that's focused more on higher level overviews
    - Also somewhat relevant to #74, although the current auto-parsing is only for listing message names and the source file they're handled in / sent from

Longer term:
- [ ] Set up automatic generation per vehicle, per version
    - May be done via autotest, [like for parameters](https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/autotest.py#L194)
        - Moving the script to autotest could be useful for notifying of changes in MAVLink support on PRs, but is probably too much effort for now
    - Will likely get done externally, in what is currently [the parameters repository](https://github.com/ArduPilot/ParameterRepository/blob/main/scripts/run_param_parse.py#L132)
        - Extra attractive because the script requires Python &geq; 3.11, which will be easier to specify there than forcing autotest to update
        - **WIP**: ArduPilot/ParameterRepository#8
- [ ] Account for / mention preprocessor directives, where relevant
    - e.g. add a "notes" column, and include things like "message only included if firmware compiled with `XXX` flag"
- [ ] Include supported fields, and any definition/range specifics for the selected vehicle
    - For automatic parsing would likely require an explicit comment syntax for within the firmware source

PR #7256, opened 2025-11-27
rover: remove MAV_CMD_NAV_SET_YAW_SPEED from mavlink interface descri…

peterbarker:
…ption

this hasn't been compiled in by default for a while, and we are planning on removing the code

See https://github.com/ArduPilot/ardupilot/pull/24776


rmackay9 (approved):
This looks fine to me.

As a side note, we should replace the link to "companion computers" with a link to the Mavlink Interface page for Rovers https://ardupilot.org/dev/docs/mavlink-rover-commands.html

peterbarker:
> @peterbarker what about the other vehicles? the auto generated wiki with cmds per vehicle shows it for them also...may need to be recreated, but I dont know how to refresh...its a manual task

Link?

Hwurzburg:
@peterbarker  https://ardupilot.org/copter/docs/ArduCopter_MAVLink_Messages.html for copter..its in every vehicle wiki

peterbarker:
> @peterbarker https://ardupilot.org/copter/docs/ArduCopter_MAVLink_Messages.html for copter..its in every vehicle wiki

Huh.

I have no idea what process updates that page.  Doesn't seem to be anything in the master code in either ardupilot or ardupilot_wiki which does it, so I'm guessing someone ran a script and pushed stuff in.

The auto-generator (wherever it is) wouldn't be able to easily tell that the code was compiled out by default.  Heck, it might even be fooled by comments for all I know!

Unfortunately, while the comments in the files say that they're automatically generated, many patches have been made by hand to that file.  Good changes, and those will be lost of the tool is run again.

Additionally, the command I'm removing *only* ever worked on Rover, so its inclusion in those other files was incorrect.

We could consider having in-the-code documentation for mavlink messages and commands that we understand, in the same way that we do for log messages and parameters, extracting those comments into a generated Wiki page.  We have to be a bit careful about phrasing anyway, as a lot of messages/commands won't be available on all boards (e.g. `GENERATOR` isn't produced except on CubeRedPrimary and one or two other boards) unless you compile your own firmware.

In any case, I've pushed up a patch to remove the lines from these "generated" files.

Hwurzburg:
Thanks...merging...fyi here is the @ES-Alexander original work...for 4.7, when it releases, it would be nice to have a new generated set of pages....or you and he wok together to make it automatic, perhaps?? https://github.com/ArduPilot/ardupilot/pull/27226

ES-Alexander:
They currently get auto-generated [here](https://github.com/ArduPilot/ParameterRepository), but they aren't yet auto-pulled or pushed into the wiki (and IIRC doing so would require modifying the filenames and adding the first line).

ES-Alexander:
Noting that the lack of prepocessor directive support (like compilation flag handling) is currently mentioned in #6065, along with other desirable features (like automatic wiki inclusion).

PR #7271, opened 2025-12-01
sub: leak-failsafe: clarify GPIO requirement

ES-Alexander:
Ran into some user confusion, because `AUX6` on Pixhawk used to be the default leak sensor port for ArduSub, but when they updated to 4.5 it wasn't working anymore because the output wasn't assigned as a GPIO (thanks to the removal of `BRD_PWM_COUNT`). We're going to be working on ArduSub detecting and dealing with this, but it also seems relevant to document how users are supposed to configure it in the first place.

This approach was the most straightforward way I could think of linking these things without going into detail about specific flight controller boards, or `SERVO_GPIO_MASK` shenanigans...

## Change Preview:
Screenshot 2025-12-02 at 5 01 14 pm

### DIFF
--- a/sub/source/docs/internal-leak-failsafe.rst
+++ b/sub/source/docs/internal-leak-failsafe.rst
@@ -10,11 +10,13 @@ For the remainder of this article, the first leak detector's parameters will use
 
 When the failsafe will trigger
 ==============================
-If :ref:`FS_LEAK_ENABLE<FS_LEAK_ENABLE>` is non-zero, and a detector is attached to either an analog input or GPIO pin on the autopilot, set by the :ref:`LEAK1_PIN<LEAK1_PIN>`. Then if the detector is analog , it's considered active ("1") if its voltage is > 2.0V. Digital sensors present a normal digital logic level.
+If :ref:`FS_LEAK_ENABLE<FS_LEAK_ENABLE>` is non-zero, and a detector is attached to either an analog input or GPIO pin on the autopilot, set by the :ref:`LEAK1_PIN<LEAK1_PIN>`. Then if the detector is analog, it's considered active ("1") if its voltage is > 2.0V. Digital sensors present a normal digital logic level.
 
 Whether the active level of an analog sensor or the logic level from a digital sensor is taken as "dry" is determined by the :ref:`LEAK1_LOGIC<LEAK1_LOGIC>` parameter.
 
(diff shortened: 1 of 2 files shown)


sub/source/docs/internal-leak-failsafe.rst
```diff
@@ -16,6 +16,8 @@ Whether the active level of an analog sensor or the logic level from a digital s
 
 .. note:: In ArduSub firmware prior or equal to 4.5, the input pin must be declared to be analog or digital with the ``LEAKx_TYPE`` parameter.
 
+.. note:: Digital leak detectors connected to an :ref:`AUX port <main-aux-out>` must be configured as :ref:`common-gpios`.
```
Hwurzburg: doesnt this also apply to main outputs as well?

ES-Alexander: I was under the (apparently mistaken) impression that flight controller boards that make a distinction between MAIN and AUX ports do so because the auxiliary ones can be used as digital inputs/outputs. I'm not sure why there are boards that do that differently, but docs should indeed reflect reality rather than what seems to make sense 🤷‍♂️

Adjusted - thanks :-)

Hwurzburg: okay, there is a difference between IOMCU outputs as GPIOs and FMU outputs as GPIOS...both can be GPIOs used as outputs, only FMU outputs can be used as inputs....so maybe it should be:
.. note:: Digital leak detectors must be connected to an output which can be used as an input (AUX outputs if the autopilot uses an IOMCU, any output if it does not). See :ref:`main-aux-out`. And  the output must be configured as a :ref:`GPIO <common-gpios>`.

ES-Alexander: I think this is the kind of technical complexity and explanation redundancy I was hoping to avoid here, by linking to existing sources of truth that explain the nuances well and comprehensively. I feel like the flight controller I/O and GPIOs pages are that source of truth, and if they're not already then I feel like it at least makes sense for them to be.

There are also alternative cases where the leak detectors can be connected to a pin that is only a GPIO (e.g. the Navigator has a dedicated GPIO pin that is intended to be used for a leak detector, which I wouldn't consider to be "an output which can be used as an input"). I'll try to come up with something that captures that distinction...

ES-Alexander: I think I've come up with a reasonable compromise. Let me know if I've misused any of the terminology :-)

PR #7272, opened 2025-12-02
sub: random fixes

ES-Alexander:
- Removed some text in the battery failsafe page (likely an accidental copy when the page was being made)
- Fixed a broken link in the companion computers page


Hwurzburg:
thanks

PR #7274, opened 2025-12-03
update.py: smarter common files handling

khancyr:
just remove what need remove and copy if necessary to preserve sphinx cache
Should give some speed up on rebuild

With the --fast parameter , it allow instant rebuild and true incremental build

### DIFF
--- a/update.py
+++ b/update.py
@@ -57,8 +57,6 @@ from typing import Optional, Dict, List
 
 from sphinx.application import Sphinx
 import rst_table
-
-from codecs import open
 from datetime import datetime
 # while flake8 says this is unused, distutils.dir_util.mkpath fails
 # without the following import on old versions of Python:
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -994,7 +1046,7 @@ def create_features_page(features, build_options_by_define, vehicletype):
     parser.add_argument(
         '--clean',
         action='store_true',
-        help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output).",
+        help="Does a very clean build - resets git to master head (and TBD cleans up any duplicates in the output), and force copy common files in wikis directories.",
```
Hwurzburg: still unclear..why the to be determined (TBD)...it means a future change..resets git to master head means that it only builds a master version, even if on a change branch, correct?

and the force clean is still not in help as far as I can see..I assume the clean will remove the common files and build on the local branch source

khancyr: I didn't touch this. So it  appears that we the git reset feature doesn't exist anymore. I have correct the comments et fix the clean parameter as it wasn't doing anything too.

PR #7290, opened 2025-12-10
reference-frames-hexsoon: fix URL typo

yuri-rage:
URL without proper capitalization leads to 404 error.


Hwurzburg:
thanks, but Randy already has this fixed in #7286  and is still reorganizing the frames stuff in master and will submit a new wiki PR once that is settled for the 4.7 wiki...so closing this

PR #7301, opened 2025-12-15
Sub: internal-leak-failsafe: improve warning description

ES-Alexander:
Cover LED and buzzer behaviours, and the specific warning message that is sent.


Hwurzburg:
thanks

PR #7312, opened 2025-12-26
Copter: update DO_SET_RELAY screenshot and docs

### LINKED ISSUE #7277: Copter: Do Set Relay mission command docs outdated
Happened upon this while looking for something else.

The screenshot shows an old interface (with mislabeled columns), and [the description](https://github.com/ArduPilot/ardupilot_wiki/blob/master/copter/source/docs/mission-command-list.rst?plain=1#L384) references an unlinked Issue that is presumably [this one](https://github.com/ArduPilot/MissionPlanner/issues/372), which was resolved in 2014.

I would submit a PR, but I don't have mission planner so can't take an updated screenshot :-\

deepak61296:
Updated screenshot to current Mission Planner interface and removed outdated note about mislabeled columns.

Fixes #7277


Hwurzburg (changes requested):
thanks for this...you should get rid of the alt of 10 shown...that field does nothing...
also, the DO_REPEAT_RELAY command image could be updated since Relay No is shown as the first column header now...if you want another easy one

PR #7319, opened 2025-12-27
dev: images: fix incorrect PID parameters

### LINKED ISSUE #7139: Copter: The parameter names in this diagram are all set to ROLL.
## Bug report

**Issue details**

Although it refers to the PID for ROLL, PITCH, and YAW, the parameter names are shown as ATC_RAT_RLL_XXX. In fact, they should be ATC_RAT_RLL_XXX, ATC_RAT_PIT_XXX, and ATC_RAT_YAW_XXX. Since this diagram is an image, it cannot be corrected.

Image

[_Please describe the problem_](https://ardupilot.org/dev/_images/copter-v4-small-attitude-PIDs.png)

**Version**

AC V4.X or later

**Platform**
[  ] All
[  ] AntennaTracker
[X] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**

ALL

**Hardware type**

ALL

**Logs**

NONE

ES-Alexander:
Fixes #7139

The required parameter snippets (`PIT`/`YAW`) were already elsewhere in the image, so it was possible to correct using just image manipulation (instead of needing to find a matching font, or actually regenerate the diagram).

For ease of review, the changed portion is just the parameter names in the "Body Frame Controllers" box in the top right. GitHub's "Swipe" comparison tool is the easiest to see the difference.

---

Noting for potential future reference that this kind of diagram would ideally be represented by an SVG, which would make the text selectable/copyable (and also possible to assign hyperlinks to), as well as indefinitely scalable (instead of somewhat pixelated).


Hwurzburg:
LGTM...thanks

PR #7321, opened 2025-12-28
Add note about AirMode to stabilize/acro behavior.

LupusTheCanine:
Clarify behavior of failsafe in Stabilize or Acro modes with AirMode active. See https://github.com/ArduPilot/ardupilot/issues/31798#issuecomment-3694961314

### DIFF
--- a/copter/source/docs/radio-failsafe.rst
+++ b/copter/source/docs/radio-failsafe.rst
@@ -37,7 +37,7 @@ When a radio failsafe is triggered, the copter can be configured via parameters 
 
 - If the copter is disarmed, no failsafe will take place.
 - If the copter is armed but has landed, the copter will immediately disarm.
-- If the copter is armed in Stabilize or Acro modes, and the throttle input is at minimum, the copter will immediately disarm.
+- If the copter is armed in Stabilize or Acro modes, and the throttle input is at minimum, and the vehicle is not in :ref:`AirMode<airmode>` then the copter will immediately disarm. 
 - Otherwise, the copter will take the actions as configured in the parameters described below.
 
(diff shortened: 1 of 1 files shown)


copter/source/docs/radio-failsafe.rst
```diff
@@ -37,7 +37,7 @@ When a radio failsafe is triggered, the copter can be configured via parameters
 
 - If the copter is disarmed, no failsafe will take place.
 - If the copter is armed but has landed, the copter will immediately disarm.
-- If the copter is armed in Stabilize or Acro modes, and the throttle input is at minimum, the copter will immediately disarm.
+- If the copter is armed in Stabilize or Acro modes, and the throttle input is at minimum, the copter will immediately disarm. If AirMode is active throttle is never considered to be at minimum.
```
rmackay9: thanks for this.  I wonder if the wording below might be easier for users to understand because they don't need to think through the logic?

`If the copter is armed in Stabilize or Acro modes, and the throttle input is at minimum, and the vehicle is not in AirMode then the copter will immediately disarm.
`

We could make the word "AirMode" a hyperlink to this page https://ardupilot.org/copter/docs/airmode.html

LupusTheCanine: Your wording is nicer; I hope the hyperlink is correct.

On one hand AirMode isn't the default so we shouldn't burden users that go through failsafe configuration with it, especially if they will use stabilize only during initial testing. On the other hand, more users should be made aware of it (I know mentioning it in the wiki isn't an effective way of informing users 😅).

andyp1per (approved):

Hwurzburg:
thanks

PR #7340, opened 2026-01-15
Fixed MAX_PITCH and MIN_PITCH parameters for Siyi gimbal

zlixOS:
This is my first pull request, I have not much idea what I am doing, only that when using my Siyi A8 mini I noticed the the max pitch was severely gimped!


Hwurzburg:
@rmackay9 my gimbal does not do this, does yours?

PR #7357, opened 2026-01-24
Dev: Add missing GStreamer plugins for Gazebo video streaming

litski0:
Added missing GStreamer plugins to installation instructions.


khancyr (changes requested):
doesn't match official instructions : https://github.com/ArduPilot/ardupilot_gazebo

Why do you think we need this ?

rmackay9:
Hi @Hwurzburg, I'm not too familiar with this area so I'll leave it to @khancyr to make the call on whether it is correct or not

@litski0, thanks very much for noticing the problem and creating this PR!

rmackay9:
I have no idea if this is correct or not.  I think that @snktshrma, @srmainwaring or @Ryanf55 might know

Ryanf55:
Can you share what your error was before you installed these? Or, do you not have that anymore?

PR #7361, opened 2026-01-26
fix some unlinked params

Hwurzburg:


copter/source/docs/drift-mode.rst
```diff
@@ -46,7 +46,7 @@ How Drift Mode works
 -  Pilot’s throttle input controls the average motor speed meaning that
    constant adjustment of the throttle is required to maintain
    altitude.  If the pilot puts the throttle completely down the motors
-   will go to their minimum rate (MOT_SPIN_ARMED) and if the vehicle
```
rmackay9, line 49: thanks for this.  I wonder if this is correct though?  I think MOT_SPIN_ARM would be correct

Hwurzburg: I thought while in the air MIN was the lowest possible level? I thought that long ago ARMED was the lowest in all cases but then ARM and MIN params were added, so MIN was the lowest value in the air...but if I am wrong, its an easy change...let me know

rmackay9: yes, you're right, sorry!!

PR #7374, opened 2026-01-29
addition of instructions for using CEE-LINE sonar

altmattr:
CEE LINE has been tested thoroughly with ardupilot and provides another sonar options in a small footprint.  It is small enough for quadcopters if anyone wants to try that.  I've documented it mainly for rover, but left instructions open for other types.


common/source/docs/common-ceeline.rst
```diff
@@ -0,0 +1,69 @@
+.. _common_ceeline:
+
+========================
```
rmackay9: we should extend these double lines to match the text

common/source/docs/common-underwater-sonars-landingpage.rst
```diff
@@ -18,6 +18,7 @@ Follow the links below (or in the sidebar) for configuration information based u
     Underwater Sonar (Analog) <common-underwater-sonar-analog>
     Triducer <common-underwater-triducer>
     Simple sonar <common-sonar-L04xMTW-GL04xMT.rst>
+    CEELine single-beam echosounder <common-ceeline>
```
rmackay9: This list is suppose to be in alphabetical order although it's not quite at the moment.  In any case I think it would be good to move this up in the list so that it is below the "BlueRobotics Ping"

altmattr:
txs all, much appreciated.

PR #7377, opened 2026-01-30
Planner: Update release requirements and enabling of Drone ID tab

Ryanf55:
The note is misleading, it's been part of mission planner for a while.
I checked the changelog to see when a bulk of the development was done.

It's also not enabled by deafult, so I add instructions how to make it appear.

image


Hwurzburg:
LGTM...thanks

PR #7390, opened 2026-02-07
Update DDFP tail thrust expo directions

thespova:
Added some more detailed guidance on thrust expo settings along with a starting point setting that will be much more affective that default for most new configurations.


thespova:
@bnsgeyer This is just those notes Matt and I talked about on the wiki for DDFP tails.

bnsgeyer (approved):
@Hwurzburg I am happy with this change.   Please review and merge.
Thanks,
Bill

PR #7395, opened 2026-02-10
alpha unmanned systems tracking system mavlink product

JavierGalanM:
Tracking systems have been added as a section within the peripheral hardware for both copter and plane platforms. This section includes the Alpha Unmanned Systems tracking antenna, which is fully compatible with MAVLink and ArduPilot devices when used as a ground station.
We would like to include this system in the documentation for users who believe our solutions may be adaptable to their MAVLink environment.
The system has been tested with ArduPilot simulators and Pixhawk-based hardware.


JavierGalanM:
Thank you for your feedback. As you mentioned, this tracking system does not include a built-in ardupilot antenna tracker like the other solutions you provide. In this case, our system receives telemetry information from the pixhawk located in the aircraft and passes this data through a radio link directly to the ground pc using the corresponding software (mission planner, qgroundcontrol, etc.).

Because of this architecture, we require a dedicated section for our system and other similar solutions. Regarding the radio link, our system is modular, and the supported radio band can be either 900 MHz or 2.4 GHz.

I will add some additional information in another branch so that you can review and potentially approve this pull request. this is the first time we have created a pull request for your team, so any feedback is greatly appreciated.

PR #7403, opened 2026-02-11
dev: improve recommendations for new PRs

ES-Alexander:
Due to some recent cases of people submitting PRs to solve issues without realising there's one already being worked on, it seems worthwhile to recommend checking that before starting to work on an issue. It's also common for @peterbarker to suggest a change needs an autotest, which we may as well mention in the submission recommendations.

Complements ArduPilot/ardupilot#32173.

### DIFF
--- a/dev/source/docs/submitting-patches-back-to-master.rst
+++ b/dev/source/docs/submitting-patches-back-to-master.rst
@@ -117,7 +117,7 @@ You can track the state of your PR from the `Pull Requests list <https://github.
 PRs are more likely to be merged quickly if:
 
 - The PR clearly states what changes in behaviour are expected
-- Good testing evidence is provided.  This could be graphs of logs recorded before and after the change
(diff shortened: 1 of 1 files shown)


dev/source/docs/submitting-patches-back-to-master.rst
```diff
@@ -115,7 +117,7 @@ You can track the state of your PR from the `Pull Requests list <https://github.
 PRs are more likely to be merged quickly if:
 
 - The PR clearly states what changes in behaviour are expected
-- Good testing evidence is provided.  This could be graphs of logs recorded before and after the change
+- Good testing evidence is provided.  This could be graphs of logs recorded before and after the change, possibly from a :ref:`simulation <simulation-2>`. Check for existing :ref:`autotests <the-ardupilot-autotest-framework>`, and update them or add new ones to cover new functionality you're introducing.
```
Hwurzburg:
```suggestion
- Good testing evidence is provided.  This could be graphs of logs recorded before and after the change, possibly from a :ref:`simulation <simulation-2>`. Check for existing :ref:`autotests <the-ardupilot-autotest-framework>`, and update them if they get broken by the change. Add new ones to cover new functionality you're introducing, or bug fixes presently not tested, if you are capable.
```

PR #7408, opened 2026-02-15
Update GPS2_TYPE reference to GPS_TYPE2 

aritrobasu2003:
I have edited the common/source/docs/common-gps-blending.rst where i have the updated the parameter from GPS2_TYPE to GPS_TYPE2.


Hwurzburg:
this is incorrect...there is no param named GPS_TYPE2

PR #7424, opened 2026-02-20
Docs: add Zorin OS mono-complete troubleshooting fix

BeastAyyG:


peterbarker:
> Instead of a wiki note, it'd be best to identify where the env install script misses this package and include it properly.

Given this is a MissionPlanner-on-Linux page, I don't think there *is* a relevant env install script?

PR #7426, opened 2026-02-21
Add codespell to GitHub Actions

cclauss:
Automatically search for typos in future code changes using https://pypi.org/project/codespell

% `git mv ./copter/source/images/telemtry.jpg ./copter/source/images/telemetry.jpg`
To match the changes to the file `copter/source/docs/xbee-radios.rst` in:
* #7422


.codespellrc
```diff
@@ -0,0 +1,4 @@
+[codespell]
+check-filenames = true
+ignore-words-list = bu,diamon,dout,empy,extint,fram,inh,minite,ned,parm,produkt,ser,sitl,te,theses,thi,tje,ue,xwindows
```
ES-Alexander: Having a list like this makes me a bit uncomfortable, because it seems to imply every non-standard-english term/name needs to be manually whitelisted. I'd personally be more inclined for codespell to be able to flag _potential_ typos, and allow the developer or a maintainer to decide whether they're genuine issues.

That said, if we need such a list, here are a couple more terms I found as part of #7428:
```suggestion
ignore-words-list = bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm,produkt,retuned,ser,sitl,te,theses,thi,tje,ue,xwindows
```

cclauss:
> [codespell] would only run on the actual changes, not on full files every time

Checking all files takes a second or two of runtime.

khancyr:
I agree we should restrict the check at least to only the source files ... my wiki repo is huge so recursion on all dir and build files will be a waste !

cclauss:
How long does `codespell` take to run on your wiki repo?

ES-Alexander:
> Checking all files takes a second or two of runtime. It is fast enough that many projects run it in pre-commit.

A pre-commit hook should (at least in principle) be able to filter by which files were changed, and where/how, rather than needing to process every file every time.

That said, my primary concern here is contributors needing to update the whitelist whenever a new word or term is introduced, rather than a maintainer just allowing its usage in one pull request and that part of the codebase not needing to be checked again. A whitelist is probably ok for commonly used technical terms, but adding things to it that are only used once seems onerous and excessive (to me at least).

Runtime is secondary - first a feature has to work well for our use cases, then if/once that’s established we can consider it for a regular Action, and if it’s fast enough we can make that a required PR check and/or a recommended pre-commit step.

> we should restrict the check at least to only the source files

I’m less sure on this one. A spell-checker should definitely avoid default auto-_fixing_ of anything that may be part of machine interpreted syntax somewhere, but it could still be useful to at least have warnings that something _might_ be spelled incorrectly, even if it’s a variable name or library import in a code/build file.

PR #7436, opened 2026-02-25
Create README: link to relevant information

ES-Alexander:
In case someone gets to the repo without knowing their way around the wiki itself, it seems worth having some relevant links and info.

I've currently got two visual approaches for the same info - not sure whether the sections or the numbered list is preferred. [Preview here](https://github.com/ES-Alexander/ardupilot_wiki/tree/add-README?tab=readme-ov-file#ardupilot-documentation-source-files).

Not sure if there's a public link I should add for where/how the wiki site actually gets built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,17 @@
+# ArduPilot Documentation Source Files
+
+1. See [the wiki site](https://ardupilot.org/ardupilot/index.html) for the human-friendly view
+2. See [the Issues](https://github.com/ArduPilot/ardupilot_wiki/issues) for known missing, incorrect, or outdated content
+    - Submit a new issue if you've found a problem that isn't already listed
+    - DO NOT use the wiki issues for support questions - they will be immediately closed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,33 @@
...
+
+> **NOTE:** Include `Resolves #ISSUE_NUMBER` in the description to automatically close a listed Issue if/when your pull request gets merged.
+
+---
+
+# ArduPilot Documentation Source Files
+
+1. See [the wiki site](https://ardupilot.org/ardupilot/index.html) for the human-friendly view
+2. See [the Issues](https://github.com/ArduPilot/ardupilot_wiki/issues) for known missing, incorrect, or outdated content
+    - Submit a new issue if you've found a problem that isn't already listed
```
Hwurzburg, line 5:
```suggestion
    - Submit a new issue if you've found a problem that isn't already listed, but do not use the wiki issues for support question. Instead use [Discord] (https://ardupilot.org/dev/docs/ardupilot-discord-server.html) or [Discuss](https://discuss.ardupilot.org/). Support questions entered as wiki issues will be immediately closed.
```

PR #7437, opened 2026-02-25
Remove deprecated dependency python-future

cclauss:
[`python-future`](https://pypi.org/project/future) was deprecated two years ago.
* https://python-future.org/whatsnew.html#what-s-new-in-version-1-0-0-2024-02-21

Remove mention of an end-of-life version of Raspbian.

Use https://docs.python.org/3/library/venv.html as recommended in https://peps.python.org/pep-0668

Upgrade to a non-EOL version of Python as discussed at https://devguide.python.org/versions

Apply [`ruff rule UP010`](https://docs.astral.sh/ruff/rules/unnecessary-future-import)

### DIFF
--- a/dev/source/docs/building-for-navio2-on-rpi3.rst
+++ b/dev/source/docs/building-for-navio2-on-rpi3.rst
@@ -23,14 +23,6 @@ Setup
 Use an ssh terminal program such as `Putty <http://www.putty.org/>`__ to
 log into the Navio2 board's RPI3.
 
-.. note::
-    
-    On Raspbian Stretch, one of the Python requirements might be missing, so please install future by
-
-::
-
-    pip install future
-
 .. include:: building-setup-linux.rst
     :start-after: Setup on Ubuntu
     :end-before: Setup for other Distributions
--- a/dev/source/docs/building-setup-mac.rst
+++ b/dev/source/docs/building-setup-mac.rst
@@ -39,8 +39,7 @@ Setup steps
    ::
 
        brew update
-       brew install genromfs
-       brew install gcc-arm-none-eabi
+       brew install gcc-arm-none-eabi genromfs python3
 
 #. Install the latest version of awk using brew (make sure
    **/usr/local/bin** takes precedence in your path):
@@ -49,12 +48,11 @@ Setup steps
 
        brew install gawk
 
-#. Install *pip* and *pyserial* using the following commands:
+#. Install *empy* and *pyserial* using the following commands:
 
    ::
 
-       sudo easy_install pip
-       sudo pip install pyserial future empy
+       python3 -m pip install empy pyserial
        
    ::
    


dev/source/docs/building-for-navio2-on-rpi3.rst
```diff
@@ -23,14 +23,6 @@ Setup
 Use an ssh terminal program such as `Putty <http://www.putty.org/>`__ to
 log into the Navio2 board's RPI3.
 
-.. note::
```
peterbarker, line 26: Patch is unrelated to PR title.  Or the commit title it is included in

dev/source/docs/building-setup-mac.rst
```diff
@@ -53,8 +52,9 @@ Setup steps
 
    ::
 
-       sudo easy_install pip
-       sudo pip install pyserial future empy
+       python3 -m venv .venv
+       source .venv/bin/activate
+       sudo pip install pyserial empy
```
peterbarker: Using the venv is the right way to move people, but these instructions aren't going to do the same thing for people as the originals.

Problem is that the initial instructions persist across people's sessions.  The new instructions people will need to know to source the venv activation script to work in that environment, and they probably won't know (or forget) that they need to do that.

Same comment applies to the other places in here.

Not saying we can't go this way, just that it's not just a matter of changing the environment setup instructions and calling it a day.

update.py
```diff
@@ -31,8 +31,6 @@
 Parameters files are fetched from autotest using requests
 
 """
-from __future__ import print_function, unicode_literals
```
peterbarker, line 34: Is this change doable independent of all the other changes?

cclauss: Yes.  Moved to.
* #7438

dev/source/docs/building-for-navio2-on-rpi3.rst
cclauss: What does that mean?!?  This block is about installing `python-future`, which is no longer needed because we no longer maintain compatibility with legacy Python 2, which died more than 6 years ago on 1/1/2020.

dev/source/docs/building-setup-mac.rst
cclauss: `easy_install` was deprecated even before legacy Python 2.

`brew install python3` will install `pip`.

Removed the use of `venv`.

`sudo pip` should not be used as explained in https://peps.python.org/pep-0668

cclauss:
> Has cygwin been tested?

I reverted all Cygwin changes because that platform only supports EoL versions of Python.

dev/source/docs/building-for-navio2-on-rpi3.rst
peterbarker: Bleh, sorry.  I missed the removal of `future`.

One way or the other the lede was thoroughly buried here.

Had you just removed `future` in these lines and the actual library import this would have been a trivial merge.  If you want to move everyone to venv's, great - but do that as a separate PR!

peterbarker:
> cygwin is mostly used only to make binarie for MP sitl, we shouldn't recommand it as build environment anymore

... and esp. not the Wiki build env ;-)

PR #7449, opened 2026-02-27
common_conf.py: simplier mapping

khancyr:


common_conf.py
```diff
@@ -31,30 +31,16 @@
...
-                                  None),  # noqa: E128
-                       'blimp': (intersphinx_base_url % 'blimp',
-                                  None),  # noqa: E127
-                      }  # noqa: E124
+# Known wiki keys (single source of truth)
+WIKI_KEYS = [
+    'copter', 'plane', 'rover', 'sub', 'planner',
+    'planner2', 'dev', 'antennatracker', 'ardupilot',
+    'mavproxy', 'blimp'
+]
```
peterbarker: Could we have one-per-line and sorted, please?

peterbarker (changes requested):
Very happy with this apart from the formatting request.

cclauss:
Was this modification triggered by looking at the changes in, but not reviewing:
* #7456
* #7458
* #7467

PR #7450, opened 2026-02-27
Copy simplify

khancyr:
simplify some code
prevent debug logging the common file copy (useless)
Permit sphinx to use more CPU core to parse files
Don't generate sphinx link for versionned parameters: this simplify the the parsing/generation and we don't use it anyway. default parameter file is untouched
fix image.nonlocal_uri as the old code is deprecated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 # Build outputs
 *.html
 *.pyc
+.cache/
 
 # static html pages
 !/frontend/*.html
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,3 +21,4 @@ sub
 
 
 !Sphinxsetup.sh
+.cache/


common_conf.py
```diff
@@ -31,6 +31,11 @@
 html_context = {'target': '/'}
 # This needs to change to the actual URL root once the theme updated.
 
+# Don't generate search index for versioned parameter pages
+html_search_options = {
+    'dict_max_word_length': 40,  # Skip very long parameter names
+}
```
cclauss, line 37: It slows down the reader when we spread code across three lines just to add a comment.
```suggestion
# Skip very long parameter names
html_search_options = {'dict_max_word_length': 40}
```

khancyr: I prefer to left it as it as the file is config file and quite simple, so if won't bother much make next change easier

peterbarker: Agree with @khancyr 

Also means shorter deltas when adding and removing things from the dict.

common_conf.py
```diff
@@ -56,22 +61,11 @@
...
-    def _warn_node(self, msg, node, **kwargs):
-        if not msg.startswith('nonlocal image URI found:'):
-            self._warnfunc(msg, '%s:%s' % get_source_line(node), **kwargs)
-
-    sphinx.environment.BuildEnvironment.warn_node = _warn_node
-# ENDPATH
+# Suppress warnings that slow down builds (already have nitpicky = False)
+suppress_warnings = [
+    'epub.unknown_project_files',  # Suppress epub warnings
+    'image.nonlocal_uri',  # Suppress external image warnings
```
peterbarker: You appear to be making this unconditional?

.gitignore
```diff
@@ -15,6 +15,7 @@
 # Build outputs
 *.html
 *.pyc
+.cache/
```
hunt0r, line 18: Thank you for this!

khancyr: nothing use this yet ...

common_conf.py
khancyr: added back, I am not sure anybody use it but the option is still there !

.gitignore
hunt0r: I'm interpreting your comment as an implicit question about why I appreciate this.

I thanked you because my LSP uses it locally. (I've had a similar local-only modification that I keep rebasing to wherever I'm working, then rebasing away before I PR the result. Having this in main means I don't need it local-only anymore.)

(I assumed you were adding it because you and others also have the same minor friction-point in your workflow.)

hunt0r: OH! I'm a dummy 🤦 
This is the wiki repo, not the main repo. (My local-only gitignore .cache is in the main one.)

I'm sorry for the distraction.

PR #7456, opened 2026-03-02
Format Python code with ruff format

cclauss:
Blocked by: #7458 or #7467
* #7458
* #7467 

https://docs.astral.sh/ruff/formatter

% `ruff format --config="format.quote-style='preserve'" ` # as discussed in
* #6996 

The goal of this pull request is explained in the first paragraph of the [`psf/black` docs](https://black.readthedocs.io):
> By using `Black`, you agree to cede control over minutiae of hand-formatting. In return, `Black` gives you speed, determinism, and freedom from `pycodestyle` nagging about formatting. You will save time and mental energy for more important matters.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,8 +8,8 @@ ignore =
     E226,
     # E261 at least two spaces before inline comment
     E261,
-    # W504 line break after binary operator
-    W504,
+    # W503 line break before binary operator
+    W503,
     # E203 whitespace before ':'
     E203,
     # E221 multiple spaces before operator
--- a/antennatracker/source/conf.py
+++ b/antennatracker/source/conf.py
@@ -148,10 +148,12 @@ html_favicon = '../../images/favicon_default.ico'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
+# fmt: off
 html_js_files = [
     './useralerts.js',
     ('https://plausible.ardupilot.org/js/script.outbound-links.js', {"data-domain": "ardupilot.org", "defer": "defer"}),
 ]
+# fmt: on
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
@@ -230,13 +232,13 @@ html_context = common_conf.html_context
 latex_elements = {
     # The paper size ('letterpaper' or 'a4paper').
     # 'papersize': 'letterpaper',
-
+    #
     # The font size ('10pt', '11pt' or '12pt').
     # 'pointsize': '10pt',
-
+    #
     # Additional stuff for the LaTeX preamble.
     # 'preamble': '',
-
+    #
     # Latex figure (float) alignment
     # 'figure_align': 'htbp',
 }
@@ -244,10 +246,12 @@ latex_elements = {
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
+# fmt: off
 latex_documents = [
     (master_doc, 'ArduPilot.tex', 'ArduPilot Documentation',
      'ArduPilot Dev Team', 'manual'),
 ]
+# fmt: on
 
 # The name of an image file (relative to this directory) to place at the top of
 # the title page.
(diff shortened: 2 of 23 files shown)


antennatracker/source/conf.py
```diff
@@ -245,8 +245,13 @@
...
 latex_documents = [
-    (master_doc, 'ArduPilot.tex', 'ArduPilot Documentation',
-     'ArduPilot Dev Team', 'manual'),
+    (
+        master_doc,
+        'ArduPilot.tex',
+        'ArduPilot Documentation',
+        'ArduPilot Dev Team',
+        'manual',
+    ),
```
cclauss: `# fmt: off` and `# fmt: on` directives could be used here to keep the current formatting.
* https://docs.astral.sh/ruff/formatter/#format-suppression

.flake8
```diff
@@ -8,8 +8,8 @@ ignore =
     E226,
     # E261 at least two spaces before inline comment
     E261,
-    # W504 line break after binary operator
-    W504,
+    # W503 line break before binary operator
+    W503,
```
cclauss, line 12: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#w503

antennatracker/source/conf.py
```diff
@@ -230,13 +233,10 @@
 latex_elements = {
     # The paper size ('letterpaper' or 'a4paper').
     # 'papersize': 'letterpaper',
-
```
peterbarker, line 233: This change is bogus

antennatracker/source/conf.py
```diff
@@ -150,7 +150,10 @@
 
 html_js_files = [
     './useralerts.js',
-    ('https://plausible.ardupilot.org/js/script.outbound-links.js', {"data-domain": "ardupilot.org", "defer": "defer"}),
+    (
+        'https://plausible.ardupilot.org/js/script.outbound-links.js',
+        {"data-domain": "ardupilot.org", "defer": "defer"},
```
peterbarker: Not a fan of having hash entries packed so that name/value pairs take up a neat 80 characters.  Should be one line or one-line-per-name/value pair

antennatracker/source/conf.py
```diff
@@ -245,8 +245,13 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-    (master_doc, 'ArduPilot.tex', 'ArduPilot Documentation',
```
peterbarker, line 248: This format is currently better - it matches the documentation at the top of the definitions.

antennatracker/source/conf.py
```diff
@@ -274,10 +279,7 @@
 
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
-man_pages = [
-    (master_doc, 'ardupilot', 'ArduPilot Documentation',
-     [author], 1)
-]
+man_pages = [(master_doc, 'ardupilot', 'ArduPilot Documentation', [author], 1)]
```
peterbarker: This makes no sense.  Taking the CR out of the middle of the declaration would

build_parameters.py
```diff
@@ -1,27 +1,27 @@
...
-        named new_params_mvesion/ and old_params_mversion/
+* Before start:
+    * Folder and file management tested only in linux;
+    * It is supposed to have the wiki repo in a same level of an ArduPilot repo and two other folders
+    named new_params_mvesion/ and old_params_mversion/
 
-    * First step is go to each vehicle on firmware.ardupilot.org and get all available versions namely as stable/beta/latest.
-        * For each version it gets a board and get git_version.txt file;
-        * It parsers that to get the version and commit hash;
-        * It creates a dictionary with vehicles, versions and commit hashes.
```
peterbarker, line 18: This seems a bit random.

peterbarker (commented):
(very) partial review only.

This is a very large PR to review.  I think we need to know what we would get at the end of all this.

Certainly I disagree with a chunk of the changes being made by the tool here, and that's always been our gripe with formatting tools.  Exactly the same language construct might require different formatting depending on content of the construct.

build_parameters.py
cclauss: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings says:
> `Black` also processes docstrings. Firstly, the indentation of docstrings is corrected for both quotations and the text within, although relative indentation in the text is preserved.

cclauss:
Closing in favor of:
* #7458

PR #7463, opened 2026-03-04
Fix capitalization in analog pins documentation

demetalgimp:
Correct capitalization in note about input voltage range.

### DIFF
--- a/common/source/docs/common-analog-pins.rst
+++ b/common/source/docs/common-analog-pins.rst
@@ -12,7 +12,7 @@ In addition to digital input and output pins, autopilots have analog input pins 
 - Analog Sonar/Rangefinder Input
 - Analog Temperature Sensor Input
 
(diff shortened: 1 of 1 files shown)


common/source/docs/common-analog-pins.rst
```diff
@@ -12,7 +12,7 @@ In addition to digital input and output pins, autopilots have analog input pins
 - Analog Sonar/Rangefinder Input
 - Analog Temperature Sensor Input
 
-.. note:: all pins can accept 0-3.3v inputs. Unless the documentation (OEM or Wiki) specifically specifies 6.6V input range, only apply 3.3 maximum!
+.. Note:: All pins can accept 0-3.3v inputs. Unless the documentation (OEM or Wiki) specifically specifies 6.6V input range, only apply 3.3 maximum!
```
Hwurzburg, line 15: fyi lower case note is the proper directive, but luckily either can be used and still result in a capitalized appearance so I wont request a change

PR #7479, opened 2026-03-06
pre-commit: run in GitHub Actions

cclauss:
In a GitHub Action, run `pre-commit run --all-files` on all pull requests.  While some developers will `pre-commit install` locally, this will ensure that `pre-commit` passes even for those who have not done so.

Like:
* https://github.com/ArduPilot/ardupilot/pull/30270

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,15 @@ env:
   TEST: "1"
 jobs:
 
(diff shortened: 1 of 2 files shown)


peterbarker (approved):
LGTM.

But shouldn't we also remove `- name: Check python3 flake8 formatting` as redundant?

PR #7482, opened 2026-03-07
wiki: update AltHold tuning parameters for Copter 4.7 SI unit changes

lthall:
Update the AltHold tuning instructions to reflect the Copter 4.7 parameter renames and scaling changes introduced as part of the SI unit migration.

PSC_ACCZ_P/I have been replaced by PSC_D_ACC_P/I with values scaled 10x smaller. The documentation now presents the 4.7 parameters as the primary instructions and retains the ≤4.6 equivalents as a compatibility note. Also update PSC_VELZ_P to PSC_D_VEL_P in the oscillation guidance to match the 4.7 parameter rename.

This keeps the existing guidance intact while ensuring users following the wiki with Copter 4.7+ see the correct parameters.


rmackay9 (dismissed):
we could change the parameters to be links to the wiki but other wise it's fine, thanks!

Hwurzburg:
tweaked and merged...thanks

PR #7467, opened 2026-03-05
ruff format --exclude=*/source/conf.py --line-length=127

cclauss:
Like #7458 but with `--line-length=127` as requested at:
* https://github.com/ArduPilot/ardupilot_wiki/pull/7458#issuecomment-4003497512

The goal of this pull request is explained in the first paragraph of the [`psf/black` docs](https://black.readthedocs.io):
> By using `Black`, you agree to cede control over minutiae of hand-formatting. In return, `Black` gives you speed, determinism, and freedom from `pycodestyle` nagging about formatting. You will save time and mental energy for more important matters.

`pyproject.toml`:
```toml
[tool.ruff]
format.exclude = ["*/source/conf.py"]
format.quote-style="preserve"
line-length = 127
```
@khancyr

PR #7494, opened 2026-03-09
Replace Python printf strings with f-strings

cclauss:
Python f-strings are shorter, faster, and easier to read than printf-style strings.

Avoids some code formatting issues discussed at:
* https://github.com/ArduPilot/ardupilot_wiki/pull/7467#discussion_r2901508713

% `ruff check --select=UP031,UP032 --statistics`
```
67	UP031	printf-string-formatting
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,3 +2,11 @@
 check-filenames = true
 ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm,produkt,retuned,ser,sitl,som,te,theses,thi,tje,ue,xwindows"
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
+
+[tool.ruff]
+target-version = "py310"
+line-length = 127
+lint.extend-select = [
+  "UP031",  # printf-string-formatting
+  "UP032",  # f-string
+]
--- a/update.py
+++ b/update.py
@@ -324,7 +324,7 @@ def check_build(site):
             continue
         index_html = os.path.join(wiki, "build", "html", "index.html")
         if not os.path.exists(index_html):
-            fatal("%s site not built - missing %s" % (wiki, index_html))
+            fatal(f"{wiki} site not built - missing {index_html}")
 
 
 def copy_build(site, destdir):
@@ -338,21 +338,21 @@ def copy_build(site, destdir):
             continue
         if wiki == 'frontend':
             continue
-        debug('Copy: %s' % wiki)
+        debug(f'Copy: {wiki}')
         targetdir = os.path.join(destdir, wiki)
         debug("Creating backup")
         olddir = os.path.join(destdir, 'old')
-        debug('Recreating %s' % olddir)
+        debug(f'Recreating {olddir}')
         if os.path.exists(olddir):
             shutil.rmtree(olddir)
         os.makedirs(olddir)
         if os.path.exists(targetdir):
-            debug('Moving %s into %s' % (targetdir, olddir))
+            debug(f'Moving {targetdir} into {olddir}')
             shutil.move(targetdir, olddir)
         # copy new dir to targetdir
         # progress("DEBUG: targetdir: %s" % targetdir)
         # sourcedir='./%s/build/html/*' % wiki
-        sourcedir = './%s/build/html/' % wiki
+        sourcedir = f'./{wiki}/build/html/'
         # progress("DEBUG: sourcedir: %s" % sourcedir)
         # progress('DEBUG: mv %s %s' % (sourcedir, destdir) )
 
@@ -369,7 +369,7 @@ def copy_build(site, destdir):
         os.makedirs(os.path.join(targetdir, '_static'), exist_ok=True)
 
         # delete the old directory
-        debug('Removing %s' % olddir)
+        debug(f'Removing {olddir}')
         shutil.rmtree(olddir)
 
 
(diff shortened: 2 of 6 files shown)


frontend/scripts/get_discourse_posts.py
```diff
@@ -98,7 +98,8 @@ def execute_http_request_json(self, url: str) -> Any:
             if cache_path.exists():
                 with open(cache_path, "r", encoding="utf-8") as f:
                     return json.load(f)
-            raise RequestExecutionError(f"Request failed with {err}. URL: {url}")
+            msg = f"Request failed with {err}. URL: {url}"
+            raise RequestExecutionError(msg)
```
peterbarker: Please eliminate the vast majority of these intermediate variables that have been created.

```suggestion
            raise RequestExecutionError(f"Request failed with {err}. URL: {url}")
```

.... why would you do that?!

cclauss: Using the intermediate variable improves the readability of the stack traces as discussed in:
% `ruff rule EM102 ` # f-string-in-exception

We can ignore the EM rules if you would prefer.

peterbarker: OK, thanks for the explanation.  That `ruff rule EM102` is *very* good documentation.

However, I'm unconvinced, I'd prefer to *not* create the `msg` variables. 

Can we except just EM102 rather than all EM rules?  Don't want to go overboard ignoring stuff.

cclauss:
Dropped `EM102`.  Let's ignore all three [`EM`](https://docs.astral.sh/ruff/rules/#flake8-errmsg-em) ruff rules because they do the same thing but with:
1. `r"string"`
2. `f"string"`
3. `"string".format()`

update.py
```diff
@@ -384,42 +384,44 @@ def make_backup(site, destdir, backupdestdir):
...
-                    debug('Deleting folder %s' % str(backup_folders[i]))
+                    debug(f'Deleting folder {str(backup_folders[i])}')
                     shutil.rmtree(str(backup_folders[i]))
                 else:
-                    debug('Ignoring folder %s because it does not look like a auto generated wiki backup folder' %
-                          str(backup_folders[i]))
+                    debug(
+                        f'Ignoring folder {str(backup_folders[i])} because '
+                        'it does not look like a auto generated wiki backup folder'
+                    )
```
peterbarker: I'd really rather a # noqa: xxx on this sort of thing.  Magic string appendy stuff is annoying - esp. with f-strings

pyproject.toml
```diff
@@ -2,3 +2,10 @@
 check-filenames = true
 ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm,produkt,retuned,ser,sitl,som,te,theses,thi,tje,ue,xwindows"
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
+
+[tool.ruff]
+line-length = 127
+lint.extend-select = [
+  "UP031",  # printf-string-formatting
+  "UP032",  # f-string
```
peterbarker, line 11: Not going to guarantee we'll always keep this.  Sometimes format-strings seem to work better.

cclauss: `str.format()` is useful for tuple unpacking like `"Python {}.{}.{}".format(*sys.version_info)` (which the ruff UP rules will not flag), but I have not seen other reasons to continue using it.

hunt0r: Just in case anyone didn't know, one other niche to avoid f-strings is in logging:
https://google.github.io/styleguide/pyguide.html#3101-logging

update.py
```diff
@@ -384,42 +384,41 @@ def make_backup(site, destdir, backupdestdir):
...
         if len(backup_folders) > n_to_keep:
             for i in range(0, len(backup_folders) - n_to_keep):
                 if '-wiki-bkp' in str(backup_folders[i]):
-                    debug('Deleting folder %s' % str(backup_folders[i]))
+                    debug(f'Deleting folder {str(backup_folders[i])}')
                     shutil.rmtree(str(backup_folders[i]))
                 else:
-                    debug('Ignoring folder %s because it does not look like a auto generated wiki backup folder' %
-                          str(backup_folders[i]))
+                    debug(f'Ignoring folder {str(backup_folders[i])} because it does not look like a auto generated wiki backup folder')  # noqa: E501
```
hunt0r, line 417: To avoid the noqa, wouldn't implicit line continuation work well here?
```suggestion
                    debug(f'Ignoring folder {str(backup_folders[i])} because it '
                          f'does not look like a auto generated wiki backup folder')
```

update.py
```diff
@@ -844,12 +834,12 @@ def check_imports():
...
     required_packages = ["sphinx_rtd_theme>=1.3.0", "sphinxcontrib.youtube>=1.2.0", "sphinx>=7.1.2", "docutils<0.19"]
     for package in required_packages:
-        debug("Checking for %s" % package)
+        debug(f"Checking for {package}")
         try:
             importlib.metadata.version(package.split("<")[0].split(">=")[0])
         except importlib.metadata.PackageNotFoundError as ex:
             progress(ex)
-            fatal("Require %s\nPlease run the wiki build setup script \"Sphinxsetup\"" % package)
+            fatal(f"Require {package}\nPlease run the wiki build setup script \"Sphinxsetup\"")
```
hunt0r, line 842: optional, just in case you like the improved readability of swapping the quote choice
```suggestion
            fatal(f'Require {package}\nPlease run the wiki build setup script "Sphinxsetup"')
```

pyproject.toml
cclauss: As discussed above the ruff `EM` solution was rejected.
* https://github.com/ArduPilot/ardupilot_wiki/pull/7494#discussion_r2943991826

update.py
cclauss: I am going to pass on making manual edits given the multi-day GitHub Action runs.

update.py
cclauss: I had already made this change, but was asked to revert to using ___# noqa: 501___ to avoid Python's "magic string concatenation".

PR #7522, opened 2026-03-13
Common copy fix2

khancyr:
fixup common files comparison to prevent copies (cannot use shallow cmp as file got wiki directives)
allow to only copy builded wiki files

master (copter only):
Capture d’écran du 2026-03-13 16-03-33

PR (copter only):
Capture d’écran du 2026-03-13 15-57-30

Master (all):
Capture d’écran du 2026-03-13 16-05-48

PR (all):
Capture d’écran du 2026-03-13 16-05-07

### DIFF
--- a/update.py
+++ b/update.py
@@ -714,7 +714,7 @@ def fetch_versioned_parameters(site=None):
                     for filename in old_parameters_files:
                         debug("Erasing rst " + filename)
                         os.remove(filename)
-                except Exception as e:
+                except OSError as e:
                     error(e)
                     pass
 
@@ -741,7 +741,7 @@ def fetch_versioned_parameters(site=None):
                     debug("Moving " + vehicle_json_file)
                     # os.rename(vehicle_json_file, new_file)
                     shutil.copy2(vehicle_json_file, new_file)
-                except Exception as e:
+                except OSError as e:
                     error(e)
                     pass
 
@@ -752,7 +752,7 @@ def fetch_versioned_parameters(site=None):
                     new_parameters_files = [
                         f for f in glob.glob(new_parameters_folder + "*.rst")
                     ]
-                except Exception as e:
+                except OSError as e:
                     error(e)
                     pass
                 for filename in new_parameters_files:
@@ -780,7 +780,7 @@ def fetch_versioned_parameters(site=None):
                             debug("Copying %s to %s" % (filename, new_file))
                             shutil.copy2(filename, new_file)
 
-                    except Exception as e:
+                    except OSError as e:
                         error(e)
                         pass
 
@@ -847,7 +847,7 @@ def cache_parameters_files(site=None):
                           (built, old_parameters_folder))
                     shutil.copy2(built, old_parameters_folder)
 
-            except Exception as e:
+            except OSError as e:
                 error(e)
                 pass
 
@@ -873,7 +873,7 @@ def put_cached_parameters_files_in_sites(site=None):
                         debug("Reusing built %s in %s " %
                               (built, vehicle_folder))
                         shutil.copy(built, vehicle_folder)
-            except Exception as e:
+            except OSError as e:
                 error(e)
                 pass
 
@@ -903,7 +903,7 @@ def copy_static_html_sites(site, destdir):
             targetdir = os.path.join(destdir, folder)
             shutil.rmtree(targetdir, ignore_errors=True)
             shutil.copytree(site_folder, targetdir)
-        except Exception as e:
+        except OSError as e:
             error(e)
             pass
 


update.py
```diff
@@ -453,92 +508,120 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
-                        destination_file.write(content)
+                    try:
+                        write_if_different(targetfile, content, encoding='utf-8')
+                    except Exception as e:
+                        error(f"Failed to write {targetfile}: {e}")
 
-    progress(f"Common files: {files_copied} copied, {files_skipped} unchanged, {files_removed} removed")
+    site_label = 'all' if site is None else site
+    progress("Common files (%s): %d copied, %d unchanged, %d removed" %
+             (site_label, files_copied, files_skipped, files_removed))
```
peterbarker: This looks like a regression to me - why move away from the f-string?

update.py
```diff
@@ -453,92 +508,120 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
-                    # debug(targetfile)
-                    with open(targetfile, 'w', encoding='utf-8') as destination_file:
-                        destination_file.write(content)
+                    try:
+                        write_if_different(targetfile, content, encoding='utf-8')
+                    except Exception as e:
+                        error(f"Failed to write {targetfile}: {e}")
 
-    progress(f"Common files: {files_copied} copied, {files_skipped} unchanged, {files_removed} removed")
+    site_label = 'all' if site is None else site
```
peterbarker: I'm really not a fan of using the ternary-operator style of assignment to default a value.

update.py
```diff
@@ -426,24 +426,79 @@ def delete_old_wiki_backups(folder, n_to_keep):
         error('Error on deleting some previous wiki backup folders: %s' % e)
 
 
-def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
+def write_if_different(path: str, content: Union[str, bytes], encoding: str = 'utf-8') -> bool:
+    """Write ``content`` to ``path`` only when it's different from existing file.
+
+    - Accepts ``str`` or ``bytes`` for ``content``.
```
peterbarker: Explain why

update.py
```diff
@@ -426,24 +426,79 @@ def delete_old_wiki_backups(folder, n_to_keep):
...
+    - Normalizes CRLF for text comparisons.
```
peterbarker: Why would we need to do this?  The source files should be identical to the source files at the end of this process.  Are we programatically changing the files in the destination in some later step?!

If we don't care about the content then we can just find some "checksum this file" thing.

In fact, I'm really not sure why we don't (a) delete everything that could possibly be stale and (b) invoke `rsync -a` to get the common files into place.  This is likely to be faster than looking at the content of each file.   You could also unconditionally copy the files using `shutil.copy2` to preserve the mtime and thus avoid unnecessarily sphinx-work.

update.py
```diff
@@ -453,92 +508,120 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
+        # Clean existing common topics only for allowed_wikis
+        for wiki in allowed_wikis:
             files = glob.glob('%s/source/docs/common-*.rst' % wiki)
             for f in files:
                 debug('Remove existing common: %s' % f)
-                os.remove(f)
+                try:
+                    os.remove(f)
+                except Exception:
+                    debug(f"Could not remove {f}")
```
peterbarker: It hasn't previously been a problem to leave the exception to run.  Why are you catching and ignoring now?

peterbarker (commented):
partial review only

update.py
khancyr: I tested again and I don't see where bytes or CRLF pop up again (this comes from an old PR I keep rebasing)
So I will discard this check.

For the copy strategy: 

Non copy version for all vehicles : 
```
[update.py]: [INFO]: Time elapsed so far: 4.97 seconds
[update.py]: [INFO]: Common files (all): 0 copied, 3056 unchanged, 0 removed
[update.py]: [INFO]: === Step 5: Building documentation with Sphinx ===
[update.py]: [INFO]: Time elapsed so far: 5.10 seconds
```
so 0.13s

Master with copy version :
```
[update.py]: Copying common source files to each Wiki
[update.py]: Common files: 2064 copied, 992 unchanged, 0 removed
[update.py]: Copying common source files took 0.211005 seconds
```
So that about the same. We gain to no have to do IO on filesystem and sphinx faster rebuild.

shutil.copy2 cannot be used for files in common as those have the copy directive in them. So if you follow the process, we read the file to know where to copy them and then strip the directive. So that is why the current version I push doesn't works with the shallow cmp unfortunately.
Rsync would mean to call subproccess which would be slower (guessing) and would need python manipulation anyway to get the copy directive ... so that cannot be used.

update.py
khancyr: yep it is spammy. 
This comes from docker usage where permission could be messup so this was showing the remove failure instead of beeing silent

update.py
khancyr: see next response

khancyr:
@peterbarker comment address and corrected were needed

update.py
```diff
@@ -426,24 +426,59 @@ def delete_old_wiki_backups(folder, n_to_keep):
...
+        os.makedirs(dirpath, exist_ok=True)
+
+    # Text content
+    if os.path.exists(path):
+        try:
+            with open(path, 'r', encoding=encoding) as f:
+                existing = f.read()
+            if existing == content:
+                return False
+        except Exception:
```
cclauss: ruff rule BLE001

update.py
```diff
@@ -453,92 +488,118 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
+                for wiki in allowed_wikis:
                     dst = '%s/source/_static/%s' % (wiki, file)
-                    # Only copy if different
-                    if not clean_common and os.path.exists(dst) and filecmp.cmp(src, dst, shallow=False):
-                        continue
+                    if not clean_common and os.path.exists(dst):
+                        try:
+                            if filecmp.cmp(src, dst, shallow=True):
+                                continue
+                        except Exception as e:
```
cclauss: ruff rule BLE001

update.py
```diff
@@ -453,92 +488,118 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
+                            debug(f"filecmp shallow failed for {src} vs {dst}: {e}")
+                        try:
+                            if filecmp.cmp(src, dst, shallow=False):
+                                continue
+                        except Exception as e:
```
cclauss: ruff rule BLE001

update.py
```diff
@@ -453,92 +488,118 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
-                            debug(f"filecmp failed for {source_file_path} vs {targetfile}: {e}")
-                            # treat as different and fall through to write
+                            debug(f"filecmp shallow check failed for {source_file_path} vs {targetfile}: {e}")
 
-                    # debug(targetfile)
-                    with open(targetfile, 'w', encoding='utf-8') as destination_file:
-                        destination_file.write(content)
+                    try:
+                        write_if_different(targetfile, content, encoding='utf-8')
+                    except Exception as e:
```
cclauss: ruff rule BLE001

update.py
```diff
@@ -737,8 +798,14 @@ def create_latest_parameter_redirect(default_param_file, vehicle):
...
     filename = vehicle + "/source/docs/parameters.rst"
-    with open(filename, "w") as text_file:
-        text_file.write(out_line)
+
+    # Write only-if-different to avoid touching mtime unnecessarily
+    try:
+        changed = write_if_different(filename, out_line)
+        if not changed:
+            debug(f"No change for redirect {filename}")
+    except Exception as e:
```
cclauss: ruff rule BLE001

update.py
```diff
@@ -920,8 +987,15 @@ def create_features_pages(site):
...
-        with open(destination_filepath, "w") as f:
-            f.write(content)
+
+        # Write only if content changed to avoid touching mtime and forcing
+        # unnecessary Sphinx rebuilds.
+        try:
+            changed = write_if_different(destination_filepath, content)
+            if not changed:
+                debug(f"No change for {destination_filepath}")
+        except Exception as e:
```
cclauss: ruff rule BLE001

update.py
peterbarker: Wait.  As my old UPS manual said, "silencing the alarm does not solve the problem that caused it".

Is there some reason the Docker image can't be fixed?

khancyr: yes. The technic we currently use for docker is to mount the wiki directory inside the docker container. The container have a non root user that should match the local user u/g id by passing -u "$(id -u):$(id -g)" . 
So that is quite closed in terms of permission, any deviation make permissions issues

The only way would be to run docker with root but that would nuke all the permission locally ..., so we can live without notice like before or I can create a notify once function so that won't be spammy and report at least once

update.py
```diff
@@ -453,92 +487,118 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
...
-            if filename not in expected_common_files[wiki]:
+            if filename not in expected_common_files.get(wiki, set()):
                 debug(f'Removing stale common file: {filepath}')
-                os.remove(filepath)
-                files_removed += 1
+                try:
+                    os.remove(filepath)
+                    files_removed += 1
+                except OSError as e:
+                    debug(f"Failed to remove stale common file {filepath}: {e}")
```
peterbarker, line 507: I'm really rather loathe to do this because Docker can't get its stuff together.

There's no reason this shouldn't work on a well-behaved machine image.

update.py
```diff
@@ -514,24 +536,39 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: Opt
 
                     # Only write if content has changed (preserves timestamps for unchanged files)
                     if not clean_common and os.path.exists(targetfile):
+                        # Fast path: raw source == target (very common when there are no shortcodes)
                         try:
-                            if filecmp.cmp(source_file_path, targetfile, shallow=False):
+                            if filecmp.cmp(source_file_path, targetfile, shallow=True):
```
peterbarker, line 541: This seems to be doing the opposite of what your comments in the PR description says is happening.

i.e. after this we are trusting the metadata to tell us what's identical.

PR #7523, opened 2026-03-13
common: add new extension to not use parameters files labels

khancyr:
add a new extension that prevent parsing and creating label for versionned parameters, only keep for latest that the default use across the wikis

Capture d’écran du 2026-03-13 16-31-49


khancyr:
@Hwurzburg no it doesn't. It doesn't touch the parameter links either

PR #7526, opened 2026-03-13
Should we accelerate build tests by upgrading to Python 3.14?

cclauss:
### Our slowest GitHub Action workflow is sped up by 90 seconds...

Will it pass?
* Yes, with setuptools on Py3.12 and later.

Will it be faster?
* Yes, save 1m24s on the Test build step and 1m30s on the overall workflow.

1. [ModuleNotFoundError: No module named 'distutils'](https://docs.python.org/3/library/distutils.html)
2. pip install setuptools
3. Test build step: 8m51s on Py3.14 vs. 10m15s on Py3.10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel if a job from the matrix fails
       matrix:
-        python-version: ["3.10", "3.14"]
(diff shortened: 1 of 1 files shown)


cclauss:
Ubuntu 26.04 LTS Resolute Raccoon is scheduled to be released in ~10 days on April 23rd.
* https://documentation.ubuntu.com/release-notes/26.04/schedule/#resolute-raccoon-schedule

PR #7533, opened 2026-03-16
4.7: Document Acro Rate Loop Only I-term scaling

andyp1per:
## Summary
- Document the I-term scaling behavior when Rate Loop Only mode is enabled in Acro mode (ACRO_OPTIONS bit 1)
- When enabled, the rate controller I-term is automatically scaled by the angle P gain for each axis
- This mimics Betaflight tuning approach, making tune transitions easier for pilots switching from Betaflight

Relates to ArduPilot PR #31171.

### DIFF
--- a/copter/source/docs/acro-mode.rst
+++ b/copter/source/docs/acro-mode.rst
@@ -13,6 +13,17 @@ See :ref:`common-imu-notch-filtering`  for more information.
 
 Pure rate mode stabilization, utilizing only the gyros with no attitude feedback from the accelerometers, can be forced by setting bit 1 of :ref:`ACRO_OPTIONS<ACRO_OPTIONS>` to one. Attitude will still have an open loop correction applied in this mode, similar to "heading hold" tail gyros, but attitude can drift over time.
 
+Rate Loop Only with I-term Scaling
+===================================
+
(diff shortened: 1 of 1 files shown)


copter/source/docs/acro-mode.rst
```diff
@@ -13,6 +13,17 @@ See :ref:`common-imu-notch-filtering`  for more information.
 
 Pure rate mode stabilization, utilizing only the gyros with no attitude feedback from the accelerometers, can be forced by setting bit 1 of :ref:`ACRO_OPTIONS<ACRO_OPTIONS>` to one. Attitude will still have an open loop correction applied in this mode, similar to "heading hold" tail gyros, but attitude can drift over time.
 
+Rate Loop Only with I-term Scaling
+===================================
+
+When the "Rate Loop Only" option is enabled (bit 1 of :ref:`ACRO_OPTIONS<ACRO_OPTIONS>`), the rate controller's I-term (integral term) is automatically scaled by the angle P gain for each axis. This makes the I-term contribution proportional to the attitude correction strength, mimicking the tuning approach used by Betaflight. Pilots transitioning from Betaflight should find that their existing tune translates more naturally with this option enabled.
+
+In this mode:
```
rmackay9: Let's use another word instead of "mode" because it is easily confused with "flight mode"

Instead lets say something like, "with I-term scaling enabled:"

Is the advantage only that it helps users transitioning from Betaflight?  I think tuning experts will understand the description involving I-terms but I wonder if regular users will.  If at all possible I think a simpler explanation for those who are not control experts would be good.

andyp1per: fixed

Hwurzburg:
@andyp1per I have tweaked the wording a bit...if its okay with you, I will merge...thanks for documenting

Hwurzburg (approved):

Issue #4274, opened 2022-04-02
Wiki needed 4.4 and earlier


Hwurzburg:
4.2 Wiki:
- [x] 16166 External AHRS (@tridge )
- [ ] 16255 ModalFC-V1 board (@auturgy )
- [ ] 16436 Print ASSERTS (@andyp1per )
- [ ] 17539 Guide mode tools (@lthall )
- [x] 17613 Ride Along JSON (@IamPete1)
- [x] 17651 ONIF camera in SITL/Linux
- [ ] 17930 Precision Landing (@rishabsingh)
- [ ] 17992 USB passthru with baudrate support (@tridge )
- [x] 18232 SDP3X baro check (@tridge)
- [ ] 18307 BirdCANdy device support (@MaarrrK)
- [x] 18337 Webots2012b compatibility (@HefnySco)
- [x] 18464 PMW node in AP_Periph(@tridge )
- [x] 18465 FlyWooF745 limitations (@tridge)
- [x] 18837 Lord EAHRS (@davisschenik)
- [x] 18954 ESP32 docs (@davidbuzz )
- [ ] 19514 PixC4-Jetson board (@climr)
- [ ] 19814 JHEMCU GSF045 board (@ot0tot
- [ ] 20937 Mamba F405 MK4 (@andyp1per )
- [ ] 20993 Filter tool(@tridge)
 
4.3 Wiki :
- [ ] 19220 Rover S-Curves and WP_OVERSHOOT (@rmackay9 )
- [x] 19882 High Latency Links (@stephendade )
- [x] 20069 Alarm PWM (@andyp1per  )
- [x] 20070 Button PWM range check (@IamPete1 )
- [ ] 20250 Sierra F9P (@MallikarjunSE 
- [x] 20851 NAV_ATT_TIME (@rmackay9 )
- [x] 20979 GPS_OPTIONS for ellipsoid hght(@tridge )
- [x] 21037 Foxeer Reaper F745 board(@andyp1per )
- [x] 21038 Concurrent pre/post filter logging(@andyp1per )
- [x] SkyStars H743 board (@magicrub @andyp1per )
- [x] 21130 Gripper Autoclose (@lthall )
- [x] 21142 Set ESC RPM via LUA(@tridge )
- [ ] 21158 Throttle slew time(@lthall )
- [ ] 21168 Benewake H30 Lidar(@tridge )
- [ ] 21215 PH4-mini-bdshot(@pallidin82 )
[...]

PR #7538, opened 2026-03-17
common-positioning: add new GPS driver options

yuri-rage:
Add new GPS driver/configuration options.

With @bugobliterator 's approval, closes ` 31217 GPS Config options` in https://github.com/ArduPilot/ardupilot_wiki/issues/4274

### DIFF
--- a/common/source/docs/common-positioning-landing-page.rst
+++ b/common/source/docs/common-positioning-landing-page.rst
@@ -132,6 +132,25 @@ Moving Baseline (GPS for Yaw) Capable
     Holybro RTK mosaic-H GPS (Dual Antenna Heading) <https://holybro.com/collections/h-rtk-gps/products/h-rtk-mosaic>
     Qiotek DroneCAN RTK-F9P GPS <https://www.qio-tek.com/index.php/product/qiotek-zed-f9p-rtk-and-compass-dronecan-module>
     Synerx MDU-2000 RTK + LTE GPS <common-synerex-mdu-2000>
+    
+GPS Auto Configuration
+======================
+
+The :ref:`GPS_AUTO_CONFIG<GPS_AUTO_CONFIG>` parameter provides several automatic configuration options.
+
+- 0: Disable automatic configuration
+    Not recommended when using supported hardware.
+    
+- 1: Enable automatic configuration for Serial GPSes only
+    Default value; supports many common modules.
+    
+- 2: Enable automatic configuration for DroneCAN as well
+    Intended for use when configuring DroneCAN GPS modules for :ref:`common-gps-for-yaw` operation. 
+    
+.. note:: After selecting option 2 for :ref:`common-gps-for-yaw` configuration, reverting to default (non-moving-baseline) operation requires connecting to the DroneCAN peripheral(s) and manually restoring the internal GPSx_TYPE parameters to their default values.
+
+- 3: Clear all configurations not set by ArduPilot (u-Blox only)
+    On next boot, resets u-Blox GPS modules to default before applying ArduPilot configuration values (u-Blox Generation 9 and newer).
 
 GPS Driver Options
 ==================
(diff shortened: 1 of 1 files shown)


common/source/docs/common-positioning-landing-page.rst
```diff
@@ -132,6 +132,23 @@ Moving Baseline (GPS for Yaw) Capable
...
+    Not recommended when using supported hardware.
+    
+- 1: Enable automatic configuration for Serial GPSes only
+    This is the default value to support most common modules.
+    
+- 2: Enable automatic configuration for DroneCAN as well
+    Use this when at least one suppported GPS module is connected via DroneCAN.
+    
+- 3: Clear all configurations not set by ArduPilot
+    On next boot, resets u-Blox GPS modules to default before applying ArduPilot configuration values (u-Blox Generation 9 and newer).
```
khancyr, line 153: does that work only on ublox ?

yuri-rage: @khancyr, correct. It's a feature of the v2 u-Blox driver that's new in 4.7.

khancyr: Probably better to write it more explicitly on the wiki page. The parameter mark up specify it. Could be my english, but I don't this restricted to ublox only , sorry

Hwurzburg: I think its clear enough...but option 2 needs different explanation

common/source/docs/common-positioning-landing-page.rst
```diff
@@ -132,6 +132,23 @@ Moving Baseline (GPS for Yaw) Capable
...
+The :ref:`GPS_AUTO_CONFIG<GPS_AUTO_CONFIG>` parameter provides several automatic configuration options.
+
+- 0: Disable automatic configuration
+    Not recommended when using supported hardware.
+    
+- 1: Enable automatic configuration for Serial GPSes only
+    This is the default value to support most common modules.
+    
+- 2: Enable automatic configuration for DroneCAN as well
+    Use this when at least one suppported GPS module is connected via DroneCAN.
```
Hwurzburg: fixes CI failure for miss-spelling also
```suggestion
    This is only used on DroneCAN GPS in its parameter set (i.e., internal to the peripheral) to enable auto configuration.
```

yuri-rage: @Hwurzburg I don't think your suggested fix is the correct interpretation. This is an autopilot-intended parameter to signal DroneCAN connected modules (often running periph) to auto-configure, as far as I can see via code review, though I think I found an issue with it (see Discord). But by intent, I think it'd be more correct to simply fix my typo, no?

Hwurzburg: that param is internal to any DroneCan GPS and the option is used only there...you never need to set it for an attached DroneCAN GPS in the autopilot params...the AP_Periph code uses it when setting up the periph's GPS...then it send GPS info in DroneCAN format to the autopilot....at least that is how Tridge explained, IRC....

but looking at the code, it seems that it makes all serial AND all DroneCAN GPSes auto configure...soemthing that a DroneCAN GPS would not normally want to happen, since it knows what GPS is attached and should configure it locally....

so maybe this is more correct:
"This is used when the user wants to autoconfigure ALL attached GPSes, including any DroneCAN GPSes (not normally required)"

yuri-rage: It appears the intent is to force nodes to set their GPSx_TYPE params (as in, change from 1 to 17/18 for moving baseline). But the firmware code as-is appears non-functional (or VERY cryptic). I'd like to resolve this with @peterbarker and/or @tridge before proceeding.

yuri-rage: Walked the code more completely and better understand its intent. Force pushed new verbiage based on Discord chat with Peter.

PR #7548, opened 2026-03-18
update readme template

Hwurzburg:
get rid of last "Sphinxisms"

### DIFF
--- a/misc/readme_template.md
+++ b/misc/readme_template.md
@@ -81,11 +81,11 @@ The SBUS pin, can be used for all ArduPilot supported receiver protocols, except
 
 To allow CRSF/ELRS and embedded telemetry available in Fport, CRSF, and SRXL2 receivers, a full UART, such as SERIAL6 (UART4) would need to be used for receiver connections. Below are setups using SERIAL6.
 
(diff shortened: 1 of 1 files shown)


misc/readme_template.md
```diff
@@ -129,11 +129,11 @@ An on-board IIS2MDC compass is provided. However, often will disabled this compa
 
 ## RSSI
 
-If the RSSI pin is used for analog RSSI input. Set :ref:`RSSI_ANA_PIN<RSSI_ANA_PIN>` to 91. Set :ref:`RSSI_TYPE<RSSI_TYPE>` to "3" if the RC protocol provides rssi data.
+If the RSSI pin is used for analog RSSI input. Set RSSI_ANA_PIN to 91. Set :ref:`RSSI_TYPE<RSSI_TYPE>` to "3" if the RC protocol provides rssi data.
```
rmackay9: this is not really my area but maybe the two other ":ref:" should be removed as well?  the RSSI_TYPE and ARSPD_TYPE?  I guess they're being removed because they don't work when inside an .md file

Hwurzburg: missed em

PR #7549, opened 2026-03-19
Common: add parameter lockdown to advanced configuration

rmackay9:
This adds a new parameter-lockdown page under Advanced Configuration.

### DIFF
--- a/common/source/docs/common-parameter-lockdown.rst
+++ b/common/source/docs/common-parameter-lockdown.rst
@@ -1,0 +1,39 @@
+.. _common-parameter-lockdown:
+
+==================
+Parameter Lockdown
+==================
+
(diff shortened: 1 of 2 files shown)


common/source/docs/common-parameter-lockdown.rst
```diff
@@ -0,0 +1,39 @@
...
+
+    -- handle PARAM_SET message
+    local parameters_which_can_be_set = {}
+    parameters_which_can_be_set["MAV_OPTIONS"] = true
+    parameters_which_can_be_set["PARAM_SET_ENABLE"] = true
+    parameters_which_can_be_set["BATT_ARM_MAH"] = true
+
+This list can be customised by adding or removing lines from the list and then re-uploading the script to the autopilot
+
+[copywiki destination="copter,plane,rover,dev"]
```
Hwurzburg: build error here....its not currently in DEV toctree...and IS in SUB....so modify the sites in this line
we could also keep it in dev and add a toctree and link in the secure-firmware page of DEV...if you wish

rmackay9: ah, right, I'll remove it from dev and add to sub

PR #7550, opened 2026-03-19
Ruff rules UP for pyupgrade

cclauss:
Make all of [ruff's pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up) mandatory in pre-commit.

This pull request makes the ___minimum Python v3.10___, which matches the default Python of the oldest Standard Support Ubuntu LTS as recommended at:
* https://github.com/ArduPilot/ardupilot/pull/32453#discussion_r2951643145

Related to:
* #7551

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,13 +6,15 @@ skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 [tool.ruff]
 target-version = "py310"
 line-length = 127
-lint.extend-select = [
+lint.select = [
   # "C4",   # flake8-comprehensions
+  "E",      # pycodestyle
+  "F",      # pyflakes
   "FURB",   # refurb
   "I",      # isort
   # "PERF", # Perflint
   "Q003",   # avoidable-escaped-quote
-  "UP031",  # printf-string-formatting
-  "UP032",  # f-string
-  # "UP",   # pyupgradde
+  "UP",     # pyupgrade
+  "W",      # pycodestyle
 ]
+lint.future-annotations = true
--- a/update.py
+++ b/update.py
@@ -31,6 +31,7 @@ Build notes:
 Parameters files are fetched from autotest using requests
 
 """
+from __future__ import annotations
 
 import argparse
 import errno
@@ -49,7 +50,6 @@ import sys
 import time
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
-from typing import Dict, List, Optional
 from urllib.parse import urlparse
 
 import requests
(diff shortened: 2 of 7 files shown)


khancyr (changes requested):
conflict with https://github.com/ArduPilot/ardupilot_wiki/pull/7554, https://github.com/ArduPilot/ardupilot_wiki/pull/7555, https://github.com/ArduPilot/ardupilot_wiki/pull/7552 need wait

pyproject.toml
```diff
@@ -4,11 +4,13 @@ ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm
...
 line-length = 127
 lint.extend-select = [
+  # "C4",   # flake8-comprehensions
   "FURB",   # refurb
+  # "I",    # isort
+  # "PERF", # Perflint
   "Q003",   # avoidable-escaped-quote
-  "UP031",  # printf-string-formatting
-  "UP032",  # f-string
+  "UP",     # pyupgradde
```
peterbarker: spelling?

update.py
```diff
@@ -50,17 +50,18 @@
...
 from frontend.scripts import get_discourse_posts
 
-if sys.version_info < (3, 8):
-    print("Minimum python version is 3.8")
+
+if sys.version_info < (3, 10):  # noqa: UP036
+    # Default Python of the oldest Standard Support Ubuntu LTS.
+    # https://releases.ubuntu.com
+    print("Minimum Python version is 3.10")
     sys.exit(1)
```
peterbarker: This doesn't belong in this PR AFAICS.

peterbarker (changes requested):
I'm OK with this PR except for the enforced upgrade.

Not saying we can't do it, but not part of a general other PR and not without general agreement.

I'd like to have minimum Python version the same for both the Wiki and the normal firmware, just for the simpler story.

pyproject.toml
cclauss: Good eyes!

update.py
cclauss: See the commit description above.  The typing syntax `str | None` requires Py3.10.

pyproject.toml
```diff
@@ -4,11 +4,14 @@ ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 
 [tool.ruff]
-target-version = "py310"
+target-version = "py38"
```
khancyr: we don't want that

cclauss: [Ruff rules UP](https://docs.astral.sh/ruff/rules/#pyupgrade-up) are sensitive to `tool.ruff.target-version`, and we do not break users on valid platforms.
* #7599

Examples, but perhaps not exhaustive:
* https://docs.astral.sh/ruff/rules/lru-cache-with-maxsize-none could break users on Vagrant.
* https://docs.astral.sh/ruff/rules/non-pep585-annotation could break users on Vagrant.
* https://docs.astral.sh/ruff/rules/non-pep604-annotation-union could break users on Debian Bullseye and Vagrant.
* https://docs.astral.sh/ruff/rules/timeout-error-alias could break users on Debian Bullseye and Vagrant.

PR #7552, opened 2026-03-19
update.py: only copy target wiki files

khancyr:
only copy file for the wiki we are building. this prevent uncessary remove and copy.

tested using 
docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 update.py --site copter --verbose
Output is the same as master

only the wiki selected files are copied

using 
docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 update.py  --verbose

Output is the same as master
files for all wiki are copied

### DIFF
--- a/update.py
+++ b/update.py
@@ -550,9 +550,10 @@ def delete_old_wiki_backups(folder, n_to_keep):
         error(f'Error on deleting some previous wiki backup folders: {e}')
 
 
-def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
+def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: Optional[str] = None):
     """
-    copies files common to all Wikis to the source directories for each Wiki
+    copies files common to all Wikis (or a single wiki when `site` is provided) to the
+    source directories for each Wiki.
 
     Args:
         start_dir: Directory containing common source files
@@ -560,14 +561,21 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
                      If False, only copy files that have changed (faster incremental builds).
     """
 
+    # Determine which wikis we will operate on
+    if site:
+        allowed_wikis = {site}
+        debug(f"copy_common_source_files: restricted to site '{site}'")
+    else:
+        allowed_wikis = set(ALL_WIKIS)
+
     # Create destination folders that might be needed (if don't exist)
-    for wiki in ALL_WIKIS:
+    for wiki in allowed_wikis:
         os.makedirs(f'{wiki}/source/docs', exist_ok=True)
         os.makedirs(f'{wiki}/source/_static', exist_ok=True)
 
     # Build a set of expected common files per wiki (to detect stale files)
     # Format: {wiki: set of filenames that should exist}
-    expected_common_files = {wiki: set() for wiki in ALL_WIKIS}
+    expected_common_files = {wiki: set() for wiki in allowed_wikis}
 
     # First pass: determine which files should exist in each wiki
     for root, dirs, files in os.walk(start_dir):
@@ -577,12 +585,14 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
                 with open(source_file_path, 'r', encoding='utf-8') as f:
                     source_content = f.read()
                 targets = get_copy_targets(source_content)
+                # Only record expected files for wikis we're operating on
                 for wiki in targets:
-                    expected_common_files[wiki].add(file)
+                    if wiki in allowed_wikis:
+                        expected_common_files[wiki].add(file)
 
     # Remove stale common files (files that exist but shouldn't)
     files_removed = 0
-    for wiki in ALL_WIKIS:
+    for wiki in allowed_wikis:
         existing_common_files = glob.glob(f'{wiki}/source/docs/common-*.rst')
         for filepath in existing_common_files:
             filename = os.path.basename(filepath)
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -422,24 +422,32 @@ def delete_old_wiki_backups(folder, n_to_keep):
         error('Error on deleting some previous wiki backup folders: %s' % e)
 
 
-def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
+def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: Optional[str] = None):
```
hunt0r, line 553: optional detail: did you consider using empty-string as the signal to do all wikis?
The "advantage" I see is a simpler typehint.
(No problem if we prefer the style like "always using None as the default value to indicate no-value-provided".)
```suggestion
def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: str = ""):
```

By "optional" I mean to imply no response is necessary, and you're welcome to resolve this as-is without reply.

update.py
```diff
@@ -422,24 +422,32 @@ def delete_old_wiki_backups(folder, n_to_keep):
...
         clean_common: If True, delete and recopy all common files (old behavior).
                      If False, only copy files that have changed (faster incremental builds).
     """
 
+    # Determine which wikis we will operate on
+    if site:
+        allowed_wikis = {site}
+        debug(f"copy_common_source_files: restricted to site '{site}'")
+    else:
+        allowed_wikis = set(ALL_WIKIS)
```
hunt0r, line 569: optional: I don't know if this different implementation any easier to read. In case you like it, I thought I'd share
```suggestion
   allowed_wikis = set(ALL_WIKIS)
   if site:
        allowed_wikis = {site}
        debug(f"copy_common_source_files: restricted to site '{site}'")
```

By "optional" I mean totally fine to resolve this without change nor response.

update.py
```diff
@@ -534,7 +550,10 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
                     with open(targetfile, 'w', encoding='utf-8') as destination_file:
                         destination_file.write(content)
 
-    progress(f"Common files: {files_copied} copied, {files_skipped} unchanged, {files_removed} removed")
+    site_label = site
+    if site_label is None:
+        site_label = "all"
```
hunt0r, line 664: optional: Above, I think `if site` is used, but here the check is slightly different `if site is None`. Shouldn't the two match, for consistency's sake, no matter which is chosen?

hunt0r: optional: Also, I think a one-liner looks nice here, in case you like it
```suggestion
    site_label = site or "all"
```
But (as you probably know) it only works if empty-string and None are distinct representations.

hunt0r (approved):
Looks good. I left some optional suggestions, but they are only "if you like them". Feel free to resolve/ignore them all.

update.py
Hwurzburg: I am not going thru all this...but the command update.py alone should build ALL sites... I have many  bash shortcuts that rely on this

update.py
khancyr: out of the scope of this PR

update.py
khancyr: @Hwurzburg if you don't pass the arguement --site copter/rover/etc , it always default to all sites !

peterbarker (dismissed):
Please fix the conflicts

update.py
```diff
@@ -421,24 +421,32 @@ def delete_old_wiki_backups(folder, n_to_keep):
         error(f'Error on deleting some previous wiki backup folders: {e}')
 
 
-def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
+def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: Optional[str] = None):
```
cclauss, line 553:
```suggestion
def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False, site: str = ""):
```

update.py
```diff
@@ -421,24 +421,32 @@ def delete_old_wiki_backups(folder, n_to_keep):
...
+    copies files common to all Wikis (or a single wiki when `site` is provided) to the
+    source directories for each Wiki.
 
     Args:
         start_dir: Directory containing common source files
         clean_common: If True, delete and recopy all common files (old behavior).
                      If False, only copy files that have changed (faster incremental builds).
     """
 
+    # Determine which wikis we will operate on
```
cclauss, line 564:
```suggestion
    # Determine which wiki we will operate on
```
Line 437 is going to create a set that only contains a single string.
If we want to allow multiple sites then we would need a `sites` collection (tuple, list, set, etc.).

update.py
```diff
@@ -421,24 +421,32 @@ def delete_old_wiki_backups(folder, n_to_keep):
...
+    if site:
+        allowed_wikis = {site}
+        debug(f"copy_common_source_files: restricted to site '{site}'")
+    else:
+        allowed_wikis = set(ALL_WIKIS)
```
cclauss, line 569:
```suggestion
    if site:
        debug(f"copy_common_source_files: restricted to site '{site}'")
    allowed_wikis = set(site or ALL_WIKIS)
```

peterbarker:
How was this tested?

khancyr:
PR : 
image

Master : 
image

(look at the Common fileq line)

Full build still working as master
image

peterbarker (approved):
LGTM

PR #7554, opened 2026-03-19
update.py: update code style

khancyr:
make update.py more python3 compliant
add typing for easier checking
use f-string
use Logging with some color
rename shadowing variable

Output is the same as master, feature page is the same 
Capture d’écran du 2026-03-19 13-01-06

### DIFF
--- a/update.py
+++ b/update.py
@@ -50,7 +50,7 @@ import time
 import requests
 from urllib.parse import urlparse
 from concurrent.futures import ThreadPoolExecutor
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, Union
 
 
 from sphinx.application import Sphinx
@@ -58,6 +58,54 @@ import rst_table
 from datetime import datetime
 
 from frontend.scripts import get_discourse_posts
+import logging
+
+
+# Configure logging
+class ColoredFormatter(logging.Formatter):
+    """Simple ANSI-coloured formatter for terminal output."""
+    COLORS = {
+        logging.DEBUG: '\033[36m',      # cyan
+        logging.INFO: '\033[32m',       # green
+        logging.WARNING: '\033[33m',    # yellow
+        logging.ERROR: '\033[31m',      # red
+        logging.CRITICAL: '\033[1;31m', # bold red
+    }
+    RESET = '\033[0m'
+
+    def format(self, record):
+        # Apply colour only when output is a tty
+        if hasattr(sys.stdout, 'isatty') and sys.stdout.isatty() \
+                and not os.environ.get('CI') and not os.environ.get('GITHUB_ACTIONS'):
+            color = self.COLORS.get(record.levelno, '')
+            record.levelname = f"{color}{record.levelname}{self.RESET}"
+        return super().format(record)
+
+
+class ErrorStoreHandler(logging.Handler):
+    """Allow to store errors for later usage."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.error_messages = []
+
+    def emit(self, record):
+        self.error_messages.append(record.getMessage())
+
+
+# The StreamHandler logs to the console
+stream_handler = logging.StreamHandler(sys.stdout)
+stream_handler.setFormatter(ColoredFormatter('[update.py]: [%(levelname)s]: %(message)s'))
+stream_handler.setLevel(logging.INFO)
+logging.basicConfig(level=logging.INFO, handlers=[stream_handler])
+logger = logging.getLogger(__name__)
+
+# The ErrorStoreHandler stores the messages
+error_store_handler = ErrorStoreHandler()
+error_store_handler.setLevel(logging.ERROR)
+logger.addHandler(error_store_handler)
+
+# Keep noisy third-party network logs quiet by default
+logging.getLogger('urllib3').setLevel(logging.WARNING)
 
 if sys.version_info < (3, 8):
     print("Minimum python version is 3.8")
@@ -107,36 +155,32 @@ LOGMESSAGE_SITE = {
     'antennatracker': 'Tracker',
     'blimp': 'Blimp',
 }
-error_log = list()
+
 N_BACKUPS_RETAIN = 10
 
-VERBOSE = False
+
+def info(str_to_print: str) -> None:
+    """Info output."""
+    logger.info(str_to_print)
 
 
-def debug(str_to_print):
+def debug(str_to_print: str) -> None:
     """Debug output if verbose is set."""
-    if VERBOSE:
-        print(f"[update.py]: {str_to_print}")
+    logger.debug(str_to_print)
 
 
-def progress(message, file=sys.stdout, end="\n"):
-    print(f"[update.py]: {message}", file=file, end=end)
-
-
-def error(str_to_print):
+def error(str_to_print: Union[str, Exception]) -> None:
     """Show and count the errors."""
-    global error_log  # noqa: F824
-    error_log.append(str_to_print)
-    print(f"[update.py][error]: {str_to_print}", file=sys.stderr)
+    logger.error(f"{str_to_print}")
 
 
-def fatal(str_to_print):
-    """Show and count the errors."""
-    error(str_to_print)
+def fatal(str_to_print: Union[str, Exception]) -> None:
+    """Show and exit on errors."""
+    logger.critical(f"{str_to_print}")
     sys.exit(1)
 
 
-def remove_if_exists(filepath):
+def remove_if_exists(filepath: str) -> None:
     try:
         os.remove(filepath)
     except OSError as e:
@@ -146,13 +190,13 @@ def remove_if_exists(filepath):
 
 def fetch_and_rename(fetchurl: str, target_file: str, new_name: str) -> None:
     fetch_url(fetchurl, fpath=new_name, verbose=False)
-    progress(f"Renaming {new_name} to {target_file}")
+    info(f"Renaming {new_name} to {target_file}")
     os.replace(new_name, target_file)
 
 
 def fetch_url(fetchurl: str, fpath: Optional[str] = None, verbose: bool = True) -> None:
     """Fetches content at url and puts it in a file corresponding to the filename in the URL"""
-    progress(f"Fetching {fetchurl}")
+    info(f"Fetching {fetchurl}")
 
     if verbose:
         total_size = get_request_file_size(fetchurl)
@@ -167,7 +211,7 @@ def fetch_url(fetchurl: str, fpath: Optional[str] = None, verbose: bool = True) 
 
     with open(filename, 'wb') as out_file:
         if verbose:
-            progress("Completed : 0%", end='')
+            print("[update.py]: Completed : 0%", end='', file=sys.stdout)  # intentionally use of print for formatting
         completed_last = 0
         for chunk in response.iter_content(chunk_size=chunk_size):
             out_file.write(chunk)
@@ -177,10 +221,10 @@ def fetch_url(fetchurl: str, fpath: Optional[str] = None, verbose: bool = True) 
             if verbose:
                 completed = downloaded_size * 100 // total_size
                 if completed - completed_last > 10 or completed == 100:
-                    print(f"..{completed}%", end='')
+                    print(f"..{completed}%", end='', file=sys.stdout)
                     completed_last = completed
         if verbose:
-            print()  # Newline to correct the console cursor position
+            print(file=sys.stdout)  # Newline to correct the console cursor position
 
 
 def get_request_file_size(url: str) -> int:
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -107,36 +156,32 @@
...
+    logger.debug(str_to_print)
 
 
-def error(str_to_print):
+def error(str_to_print: Union[str, Exception]) -> None:
     """Show and count the errors."""
-    global error_log  # noqa: F824
-    error_log.append(str_to_print)
-    print(f"[update.py][error]: {str_to_print}", file=sys.stderr)
+    logger.error(f"{str_to_print}")
```
hunt0r, line 174: Optional: Do you know about advice not to use f-strings in logging? Here's one example about it https://klichx.dev/2019/11/09/f-strings-and-python-logging/

update.py
```diff
@@ -177,10 +222,10 @@ def fetch_url(fetchurl: str, fpath: Optional[str] = None, verbose: bool = True)
             if verbose:
                 completed = downloaded_size * 100 // total_size
                 if completed - completed_last > 10 or completed == 100:
-                    print(f"..{completed}%", end='')
+                    print(f"..{completed}%", end='', file=sys.stdout)
```
hunt0r, line 224: Optional, for my own learning: Why specify the default value for `file` explicitly?
(Of course it's fine as-is! No problem to ignore, and resolve with no response.)

update.py
```diff
@@ -272,13 +317,13 @@ def build_one(wiki, fast):
...
-def sphinx_make(site, parallel, fast):
+def sphinx_make(site: Optional[str], parallel: int, fast: bool) -> None:
     """
     Calls 'make html' to build each site
     """
-    done = set()
-    wikis = set(ALL_WIKIS[:])
-    procs = []
+    done: set[str] = set()
+    wikis: set[str] = set(ALL_WIKIS[:])
```
hunt0r: optional: I know you didn't touch it, but since the line is changing, consider removing the unnecessary bit?
```suggestion
    wikis: set[str] = set(ALL_WIKIS)
```

(I'm hoping that my understanding is wrong, and that removing it is not correct for some reason that I'll learn.)

update.py
```diff
@@ -272,13 +317,13 @@ def build_one(wiki, fast):
     app.build()
 
 
-def sphinx_make(site, parallel, fast):
+def sphinx_make(site: Optional[str], parallel: int, fast: bool) -> None:
     """
     Calls 'make html' to build each site
     """
```
hunt0r, line 322: optional: Do `None` and empty-string mean the same thing as values for `site`? If yes, I'd suggest 'restricting' to avoid None via the simpler typehint `str` and using empty-string as the falsy signal.
(I prefer the style of keeping types "as simple as possible" over the style of "always use None to indicate that no argument value was provided". This becomes relevant for null-ish alternatives.)
(Also, I put 'restricting' in quotes because I understand Python's typehint enforcement. Or rather, lack thereof.)
```suggestion
def sphinx_make(site: str, parallel: int, fast: bool) -> None:
    """
    Calls 'make html' to build each site. Pass in site=="" to 'make all sites in the wiki'.
    """
```

I believe accepting this suggestion would mean also changing a line in the body which has `site is not None`. (Fortunately, it might be a simple delete?! I didn't parse carefully tho.)

hunt0r: And I see other fuctions use the same Optional[str] pattern. I won't comment on them, but the same suggestion probably applies.

update.py
```diff
@@ -390,39 +435,39 @@ def make_backup(site, destdir, backupdestdir):
...
             for i in range(0, len(backup_folders) - n_to_keep):
                 if '-wiki-bkp' in str(backup_folders[i]):
-                    debug('Deleting folder %s' % str(backup_folders[i]))
+                    debug(f'Deleting folder {backup_folders[i]}')
                     shutil.rmtree(str(backup_folders[i]))
                 else:
-                    debug('Ignoring folder %s because it does not look like a auto generated wiki backup folder' %
-                          str(backup_folders[i]))
+                    debug(f'Ignoring folder {backup_folders[i]} because it does not '
+                          f'look like a auto generated wiki backup folder')
```
hunt0r, line 462: Love it!

But also, it looked very familiar... and then I realized where I'd seen it. On the last PR I reviewed. #7494 .
Mentioning just in case you weren't aware, looks like merge conflicts will hit whoever lands second.

hunt0r: Or worse, they don't, and that PR accidentally un-does good work done by yours!

update.py
```diff
@@ -632,49 +677,45 @@ def fetch_versioned_parameters(site=None):
...
                 if 'antennatracker' in key.lower():  # To main the original script approach instead of the build_parameters.py approach.  # noqa: E501
                     old_parameters_mask = (os.getcwd() +
-                                           '/%s/source/docs/parameters-%s-' %
-                                           ("AntennaTracker", "AntennaTracker"))
+                                           f"/{key}/source/docs/parameters-{key.title()}-")
                 else:
                     old_parameters_mask = (os.getcwd() +
-                                           '/%s/source/docs/parameters-%s-' %
-                                           (key, key.title()))
+                                           f"/{key}/source/docs/parameters-{key.title()}-")
```
hunt0r: optional: I don't know the details/history here. But if your changes are right, wouldn't it be easier to eliminate the if-case as no longer necessary? Don't both branches have the same code-line?

(Of course if I'm wrong, a simple "no" is fine! I'm not expecting you to explain what error I have in my thinking.)

hunt0r: This same suggestion applies several spots below. But I won't duplicate it to them, in case it's wrong everywhere.

update.py
```diff
@@ -829,32 +866,32 @@ def copy_static_html_sites(site, destdir):
...
     for package in required_packages:
-        debug("Checking for %s" % package)
+        debug(f"Checking for {package}")
         try:
             importlib.metadata.version(package.split("<")[0].split(">=")[0])
         except importlib.metadata.PackageNotFoundError as ex:
-            progress(ex)
-            fatal("Require %s\nPlease run the wiki build setup script \"Sphinxsetup\"" % package)
+            error(ex)
+            fatal(f"Require {package}\nPlease run the wiki build setup script \"Sphinxsetup\"")
```
hunt0r: optional: https://github.com/ArduPilot/ardupilot_wiki/pull/7494/changes#r2960078893

update.py
khancyr: likely irelevant in our context

update.py
khancyr: I won't change this

update.py
khancyr: out of the scope of this PR

update.py
khancyr: out of the scope of this PR

update.py
cclauss: Type hints are not needed because all the type checkers and humans can intuit that the data type is set.

You can verify this by trying to assign a str to wikis and see the type checker complain.

hunt0r: @cclauss I understand your point, but your comment seems to be about a different aspect of this line than mine. Did you perhaps mean to place it somewhere else, or in a new thread?

(My comment was about removing the unnecessary `[:]`. Yours seems about the typehint `: set[str]`.)

cclauss: I think you are correct... `[:]` makes a separte copy of `ALL_WIKIS` but `set(ALL_WIKIS)` also makes a separate copy.  Proof:
```
>>> abc = list("abc")
>>> abc_again = abc
>>> abc_copy = abc[:]
>>> abc_set = set(abc)
>>> abc, abc_again, abc_copy, abc_set
(['a', 'b', 'c'], ['a', 'b', 'c'], ['a', 'b', 'c'], {'c', 'a', 'b'})
>>> abc.pop()
'c'
>>> abc, abc_again, abc_copy, abc_set
(['a', 'b'], ['a', 'b'], ['a', 'b', 'c'], {'c', 'a', 'b'})
```
Both `abc` and `abc_again` are modified by `list.pop()` but `abc_copy` and `abc_set` are not.

khancyr: solved

update.py
```diff
@@ -50,14 +50,63 @@
...
+        logging.INFO: '\033[32m',       # green
+        logging.WARNING: '\033[33m',    # yellow
+        logging.ERROR: '\033[31m',      # red
+        logging.CRITICAL: '\033[1;31m', # bold red
+    }
+    RESET = '\033[0m'
+
+    def format(self, record):
+        # Apply colour only when output is a tty
+        if hasattr(sys.stdout, 'isatty') and sys.stdout.isatty() \
```
cclauss, line 78: PEP8: Avoid backslash line continuation because whitespace to the right of the backslash can break the script on a change that is invisible to the reader.

update.py
```diff
@@ -838,13 +884,13 @@ def check_imports():
         try:
             importlib.metadata.version(package.split("<")[0].split(">=")[0])
         except importlib.metadata.PackageNotFoundError as ex:
-            progress(ex)
+            error(ex)
             fatal(f"Require {package}\nPlease run the wiki build setup script \"Sphinxsetup\"")
```
cclauss:
```suggestion
            fatal(f'Require {package}\nPlease run the wiki build setup script "Sphinxsetup"')
```
% `ruff check --select=Q003 --output-format=concise`
```
build_parameters.py:437:19: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:462:27: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:466:35: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:466:122: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:466:150: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:469:35: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:469:126: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:469:139: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:472:35: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:472:126: Q003 [*] Change outer quotes to avoid escaping inner quotes
build_parameters.py:472:154: Q003 [*] Change outer quotes to avoid escaping inner quotes
update.py:726:17: Q003 [*] Change outer quotes to avoid escaping inner quotes
update.py:726:87: Q003 [*] Change outer quotes to avoid escaping inner quotes
update.py:841:19: Q003 [*] Change outer quotes to avoid escaping inner quotes
update.py:861:31: Q003 [*] Change outer quotes to avoid escaping inner quotes
update.py:863:31: Q003 [*] Change outer quotes to avoid escaping inner quotes
Found 16 errors.
[*] 16 fixable with the `--fix` option.
```

hunt0r: https://github.com/ArduPilot/ardupilot_wiki/pull/7554#discussion_r2960907806

cclauss: @hunt0r Would you be willing to create a ___separate___ pull request that focuses on fixing [`ruff rule Q003`](https://docs.astral.sh/ruff/rules/#flake8-quotes-q)?

Add a line to `pyproject.toml`...
```diff
[tool.ruff]
lint.extend-select = [
+  "Q003",  # avoidable-escaped-quote
]
```
And then run:
% `ruff check --select=Q003 --fix`

I would avoid fixing other `Q` rules in this pull request to make it easier to review and merge.

Like:
* #7550
* #7560
* #7562
* #7563

hunt0r: Sure! I'm not working this weekend, but in a few days I'll post back here with a link to it.

cclauss: Great. If you @mention me on your pull request, then I will be happy to review it.

hunt0r: https://github.com/ArduPilot/ardupilot_wiki/pull/7572

That was _trivially_ easy... did I misunderstand something?

update.py
cclauss: I also dislike `Optional[str]` all over the place because type checkers (mypy, ty, pyrefly, etc.) are going to be raising `not-iterable` and `not-subscriptable` errors.
* https://docs.astral.sh/ty/reference/rules/#not-iterable

tpwrules: They are not smart enough to see it as type syntax??

cclauss: They are able to see that None is neither iterable nor subscriptable.
```python
def func(s: str | None = "hello") -> None:
    for c in s:  # type: ignore[not-iterable]
        print(c)
    print(s[0])  # type: ignore[not-subscriptable]
    try:
        print(s[100])  # type: ignore[not-subscriptable]
    except IndexError:
        print("short string")
    except TypeError:
        print("'NoneType' object is not iterable")

if __name__ == "__main__":
    func()
    func(None)
```

peterbarker:
@khancyr conflicting again...

cclauss:
The pull request that caused these conflicts was a nice readability improvement, which is not useless.

update.py
```diff
@@ -50,14 +50,63 @@
...
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, Union
 
 
 from sphinx.application import Sphinx
 import rst_table
 from datetime import datetime
 
 from frontend.scripts import get_discourse_posts
+import logging
```
cclauss, line 61: Belongs above with the other Standard Library imports.

update.py
```diff
@@ -50,14 +50,63 @@
...
+from build_options import BUILD_OPTIONS, Feature
```
cclauss: Please do not unconditionally import a file that may not exist.  The CI failures of this unforced error are blocking review of other pull requests that are [`DelayMerge`](https://github.com/ArduPilot/ardupilot_wiki/issues?q=state%3Aopen%20label%3ADelayMerge) behind this one.

PR #7555, opened 2026-03-19
build_parameters.py: use http cache, session, and parallele resquests

khancyr:
use http session for better request
use parallele request for faster grabbing
use cache with 6h timeout to not request to much on the server as parameters docs change slowly.

This reduce the build_parameters.py call by 4min.

No change on output
tested with 
docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 build_parameters.py --destination /ardupilot_wiki/new_params_mversion

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -32,16 +32,24 @@ import time  # noqa: F401
 import logging
 import json
 from html.parser import HTMLParser
-import urllib.request
 import re
 import glob
 import argparse
+import requests
+from concurrent.futures import ThreadPoolExecutor
 
 parser = argparse.ArgumentParser(description="python3 build_parameters.py [options]")
 parser.add_argument("--verbose", dest='verbose', action='store_false', default=True, help="show debugging output")
 parser.add_argument("--ardupilotRepoFolder", dest='gitFolder', default="../ardupilot", help="Ardupilot git folder. ")
 parser.add_argument("--destination", dest='destFolder', default="../../../../new_params_mversion", help="Parameters*.rst destination folder.")  # noqa: E501
 parser.add_argument('--vehicle', dest='single_vehicle', help="If you just want to copy to one vehicle, you can do this. Otherwise it will work for all vehicles (Copter, Plane, Rover, AntennaTracker, Sub, Blimp)")  # noqa: E501
+DEFAULT_CACHE_TIME = 6 * 3600  # 6 hours in seconds
+
+# Get the directory where this script is located
+script_dir = os.path.dirname(os.path.abspath(__file__))
+default_cache_dir = os.path.join(script_dir, '.cache')
+
+parser.add_argument("--cache-dir", dest='cache_dir', default=default_cache_dir, help="Directory to cache HTTP responses")
 args = parser.parse_args()
 
 error_count = 0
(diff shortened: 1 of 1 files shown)


build_parameters.py
```diff
@@ -510,13 +584,13 @@ def move_results(vehicles):
...
                 os.remove(old_file)
 
-            # Moving files.
+            # Moving files (use shutil.move for cross-device compatibility)
             files_to_move = [f for f in glob.glob("parameters-" + vehicle + "*")]
             for file in files_to_move:
-                if ("latest" not in file):  # Trying to re-enable toc list on the left bar on the wiki by forcing latest file name.  # noqa: E501
-                    os.rename(file, folder + file)
+                if "latest" not in file:  # Trying to re-enable toc list on the left bar on the wiki by forcing latest file name.  # noqa: E501
+                    shutil.move(file, folder + file)
```
khancyr, line 591: need for using on docker as we mount the wiki directory into docker otherwise it os.rename trigger a filesytem change error

peterbarker (changes requested):
Please fix the conflicts

build_parameters.py
```diff
@@ -327,17 +397,22 @@ def fetch_commit_hash(version_link, board, file):
...
-    if len(commite_and_codes_cleanned) == 0:
+    commits_and_codes_cleanned = {}
+    for i, cc in enumerate(commits_and_codes):
+        if cc[0] != 'error':
+            commits_and_codes_cleanned[i] = cc
+
+    if len(commits_and_codes_cleanned) == 0:
         error("Expected at least one commit")
-    return commite_and_codes_cleanned
+    return commits_and_codes_cleanned
```
hunt0r, line 419: Optional: "cleanned" is harmless typo.
```suggestion
    commits_and_codes_cleaned = {}
    for i, cc in enumerate(commits_and_codes):
        if cc[0] != 'error':
            commits_and_codes_cleaned[i] = cc

    if len(commits_and_codes_cleaned) == 0:
        error("Expected at least one commit")
    return commits_and_codes_cleaned
```

build_parameters.py
```diff
@@ -236,27 +303,33 @@ def get_last_board_folder(url):
...
             lParser.close()
+
+        if links:  # Only proceed if we have links
             last_item = links.pop()
             last_folder = last_item['href']
             debug("Returning link of the last board folder (" + last_folder[last_folder.rindex('/')+1:] + ")")
             return last_folder[last_folder.rindex('/')+1:]  # clean the partial link
+        else:
+            error(f"No board folders found for {url}")
+            return None
```
hunt0r, line 336: Optional code-style idea: Do you like [early return](https://dev.to/malaquiasdev/why-i-like-to-use-early-returns-pattern-52b2)?

(I put in the f-string convert and a tiny DRY suggestion too.)
```suggestion
        if not links:
            error(f"No board folders found for {url}")
            return None
        
        last_item = links.pop()
        last_folder = last_item['href']
        cleaned_link = last_folder[last_folder.rindex('/')+1:]  # clean the partial link
        debug(f"Returning link of the last board folder ({cleaned_link})")
        return cleaned_link
```

build_parameters.py
```diff
@@ -174,53 +233,61 @@ def setup():
...
+        firmware_base_url = firmware_url[:-1]
+
+        if "stable" in version_folder and not version_folder.endswith("stable"):
+            return firmware_base_url + version_folder
+        elif "latest" in version_folder:
+            return firmware_base_url + version_folder
+        elif "beta" in version_folder:
+            return firmware_base_url + version_folder
+        else:
+            return None
```
hunt0r, line 278: Optional style suggestion:
```suggestion
    def get_vehicle_firmware_links(vehicle_url, version_folder):
        if (("stable" in version_folder and not version_folder.endswith("stable"))
            or ("latest" in version_folder)
            or ("beta" in version_folder):
            return firmware_url[:-1] + version_folder
        return None
```

build_parameters.py
```diff
@@ -72,10 +81,60 @@ def format(self, record):
...
+
+        return content
+    except Exception as e:
+        error(f"Failed to fetch {url}: {e}")
+        # Try to use old cached version if available
+        if os.path.exists(cache_path):
+            debug(f"Using stale cached content for {url}")
+            with open(cache_path, 'r', encoding='utf-8') as f:
+                return f.read()
+        raise
```
hunt0r, line 126: Imagine the case where the fetch works, but caching it fails. Won't the code discard the good fresh-fetch and either fail or use a stale version?

I expected to use the fresh, and simply skip the cache that failed. Here's that in code-form:

```suggestion
    # Fetch fresh content
    try:
        debug(f"Fetching fresh content from {url}")
        response = session.get(url, timeout=30)
        response.raise_for_status()
        content = response.text
    except Exception as e:
        error(f"Failed to fetch {url}: {e}")
        # Try to use old cached version if available
        if os.path.exists(cache_path):
            debug(f"Using stale cached content for {url}")
            with open(cache_path, 'r', encoding='utf-8') as f:
                return f.read()
        raise
    
    # Cache the content
    try:
        with open(cache_path, 'w', encoding='utf-8') as f:
            f.write(content)
     except:
         pass  # a failure-to-cache is not worth flagging

    return content
```

build_parameters.py
```diff
@@ -174,53 +233,61 @@ def setup():
...
+                if tag == 'a':
+                    attr = dict(attrs)
+                    if 'href' in attr:
+                        self.links.append(attr['href'])
+
+        html_parser = ParseText()
         try:
-            debug("Fetching " + firmware_url)
-            lParser.feed(urllib.request.urlopen(firmware_url).read().decode('utf8'))
+            debug("Fetching " + vehicle_url)
```
hunt0r: optional, aligning with the f-string style:
```suggestion
            debug(f"Fetching {vehicle_url}")
```

build_parameters.py
```diff
@@ -73,10 +81,64 @@ def format(self, record):
...
+        cache_age = time.time() - os.path.getmtime(cache_path)
+        if cache_age < DEFAULT_CACHE_TIME:
+            debug(f"Using cached content for {url}")
+            with open(cache_path, 'r', encoding='utf-8') as f:
+                return f.read()
+
+    # Fetch fresh content
+    try:
+        debug(f"Fetching fresh content from {url}")
+        response = session.get(url, timeout=30)
```
peterbarker, line 115: This should be using `If-Modified-Since`, and the cache file's mtime deliberately modified to have the timestamp of the fetched data.

Alternatively, apparently `CacheControl` is an add-on which makes this easy...

build_parameters.py
```diff
@@ -148,9 +210,9 @@ def setup():
     if args.single_vehicle in ALLVEHICLES:
         global VEHICLES
         VEHICLES = [args.single_vehicle]
-        progress("Running only for " + str(args.single_vehicle))
+        progress(f"Running only for {args.single_vehicle}")
     else:
-        progress(f"Vehicle {str(args.single_vehicle)} not recognized, running for all vehicles.")
+        progress(f"Vehicle {args.single_vehicle} not recognized, running for all vehicles.")
```
peterbarker, line 215: This is *all* unrelated.

build_parameters.py
```diff
@@ -175,53 +237,61 @@ def setup():
...
         """
         Fetch firmware.ardupilot.org/baseURL all first level folders for a given base URL.
-
         """
-        links = []
-        # Define HTML Parser
+        class ParseText(HTMLParser):
+            def __init__(self):
+                super().__init__()
+                self.links = []
```
peterbarker, line 249: Is this related?!

build_parameters.py
khancyr: so you mean to ask the server to serve the response only if it think it has been modified on the server side, right ?
That is another layer, but can add it

PR #7556, opened 2026-03-19
build_parameters.py: don't update submodule on parameter update

khancyr:
this is slow and not use to generate the parameters


peterbarker (approved):
LGTM

PR #7557, opened 2026-03-19
build_parameters.py: use true json formating

khancyr:
simplify the syntax and validate the json file

tested comparing output from master. it match


peterbarker:
This looks really nice.

> tested comparing output from master. it match

Can you explain more about that, please?  How did you test master vs this branch?

khancyr:
@peterbarker 
image

image

PR #7558, opened 2026-03-19
build_parameters.py: use subprocess to call param_parse.py

khancyr:
subprocess is safer and faster than os.system 
we don't need os.chdir() (one is left since git command are still using os.system)
we need the cgi_escape patch otherwise we failed to build since subprocess is forwarding error and are handling them correctly now.
So it bring new parameter to build (one one's)

Tested by comparing parameters directories generated with master build version
image

image
Screenshot_2026-04-20_22-16-34

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -32,6 +32,7 @@ import logging
 import os
 import re
 import shutil  # noqa: F401
+import subprocess
 import sys
 import time  # noqa: F401
 from html.parser import HTMLParser
@@ -250,6 +251,7 @@ def setup():
         global BASEPATH
         BASEPATH = os.getcwd()
         check_temp_folders()
+        os.chdir(BASEPATH) # Need to call git command correctly
     except Exception as e:
         error(f"ArduPilot Repo folder not found (cd {args.gitFolder} failed)")
         error(e)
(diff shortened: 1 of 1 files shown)


build_parameters.py
```diff
@@ -406,11 +407,28 @@ def replace_anchors(source_file, dest_file, version_tag):
...
-                os.system("python3 ./param_parse.py --vehicle " + vehicle_new_to_old_name[vehicle])  # option "param_parse.py --format rst" is not available in all commits where param_parse.py is found  # noqa: E501
+            param_metadata_dir = os.path.join(BASEPATH, "Tools", "autotest", "param_metadata")
+            # Build the command
+            if ('rover' in vehicle.lower()) and ('v3.' not in version.lower()) and ('v4.0' not in version.lower()):
+                vehicle_name = 'Rover'
+            else:
+                vehicle_name = vehicle_new_to_old_name[vehicle]
+
+            cmd = ["python3", "./param_parse.py", "--vehicle", vehicle_name]
+
```
cclauss, line 511: My recommendation would be to move the `try:` from line 408/409 down to here and dedent the intervening lines.  If an exception is raised in those intervening lines, it is a programming error that should be raised instead of being written to a log.

If this change is made, then I would recommend that we modify line 412 to call `vehicle.lower())` once instead of three times in rapid succession.

https://peps.python.org/pep-0008/#programming-recommendations
> Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs:

khancyr: out of scope of this PR and will delay it

cclauss (changes requested):
https://github.com/ArduPilot/ardupilot_wiki/pull/7558#discussion_r2986547974

hunt0r (approved):

cclauss (approved):

peterbarker (approved):
LGTM!

PR #7560, opened 2026-03-21
Ruff rules for comprehensions and performance

cclauss:
Add ruff rules for [comprehensions](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) and [performance](https://docs.astral.sh/ruff/rules/#perflint-perf). 
```diff
[tool.ruff]
lint.extend-select = [
+  "C4",     # flake8-comprehensions
+  "PERF",   # Perflint
]
```
% `ruff check --select=C4,PERF --statistics`
```
9	C416   	unnecessary-comprehension
2	PERF203	try-except-in-loop
1	C408   	unnecessary-collection-call
1	PERF401	manual-list-comprehension
Found 13 errors.
```
% `ruff rule C416`
# unnecessary-comprehension (C416)

Derived from the **flake8-comprehensions** linter.

Fix is always available.

## What it does
Checks for unnecessary dict, list, and set comprehension.

## Why is this bad?
It's unnecessary to use a dict/list/set comprehension to build a data structure if the
elements are unchanged. Wrap the iterable with `dict()`, `list()`, or `set()` instead.

## Example
```python
{a: b for a, b in iterable}
[x for x in iterable]
{x for x in iterable}
```

Use instead:
```python
dict(iterable)
list(iterable)
set(iterable)
```

## Known problems

This rule may produce false positives for dictionary comprehensions that iterate over a mapping.
The dict constructor behaves differently depending on if it receives a sequence (e.g., a
list) or a mapping (e.g., a dict). When a comprehension iterates over the keys of a mapping,
replacing it with a `dict()` constructor call will give a different result.

For example:

```pycon
>>> d1 = {(1, 2): 3, (4, 5): 6}
>>> {x: y for x, y in d1}  # Iterates over the keys of a mapping
{1: 2, 4: 5}
>>> dict(d1)               # Ruff's incorrect suggested fix
{(1, 2): 3, (4, 5): 6}
>>> dict(d1.keys())        # Correct fix
{1: 2, 4: 5}
```

When the comprehension iterates over a sequence, Ruff's suggested fix is correct. However, Ruff
cannot consistently infer if the iterable type is a sequence or a mapping and cannot suggest
the correct fix for mappings.

## Fix safety
Due to the known problem with dictionary comprehensions, this fix is marked as unsafe.

Additionally, this fix may drop comments when rewriting the comprehension.

### DIFF
--- a/update.py
+++ b/update.py
@@ -106,7 +106,7 @@ LOGMESSAGE_SITE = {
     'antennatracker': 'Tracker',
     'blimp': 'Blimp',
 }
-error_log = list()
+error_log = []
 N_BACKUPS_RETAIN = 10
 
 VERBOSE = False
@@ -593,7 +593,7 @@ def logmatch_code(matchobj, prefix):
     for i in range(9):
         try:
             progress(f"{prefix} m{i}: {matchobj.group(i)}")
-        except IndexError:  # The object has less groups than expected
+        except IndexError:  # noqa: PERF203 The object has less groups than expected
             progress(f"{prefix}: except m{i}")
 
 
@@ -639,8 +639,7 @@ def fetch_versioned_parameters(site=None):
                 else:
                     old_parameters_mask = f"{os.getcwd()}/{key}/source/docs/parameters-{key.title()}-"
                 try:
-                    old_parameters_files = [
-                        f for f in glob.glob(f"{old_parameters_mask}*.rst")]
+                    old_parameters_files = list(glob.glob(f"{old_parameters_mask}*.rst"))
                     for filename in old_parameters_files:
                         debug(f"Erasing rst {filename}")
                         os.remove(filename)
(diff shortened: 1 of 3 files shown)


update.py
```diff
@@ -983,9 +972,11 @@ def create_features_page(features, build_options_by_define, vehicletype):
...
         index += f'- :ref:`{board}<{reference_for_board(board)}>`\n\n'
 
-    all_features_rows = []
-    for feature in sorted(build_options_by_define.values(), key=lambda x : (x.category + x.label).lower()):
-        all_features_rows.append([feature.category, feature.label, feature.description])
+    all_features_rows = [
+        [feature.category, feature.label, feature.description]
+        for feature
+        in sorted(build_options_by_define.values(), key=lambda x : (x.category + x.label).lower())
+    ]
```
peterbarker: Please revert this patch.  The original code is simply *better*.

peterbarker (changes requested):
Apart from that one patch this is a great improvement.

update.py
cclauss: Will do, but the [`ruff rule PERF401`](https://docs.astral.sh/ruff/rules/manual-list-comprehension) says:
> the list comprehension is ~10% faster on Python 3.11, and ~25% faster on Python 3.10.

But we probably have few build_options.

khancyr (changes requested):
conflict with https://github.com/ArduPilot/ardupilot_wiki/pull/7554, https://github.com/ArduPilot/ardupilot_wiki/pull/7555, https://github.com/ArduPilot/ardupilot_wiki/pull/7552 need wait

PR #7563, opened 2026-03-21
Ruff rules FURB for refurb

cclauss:
% `ruff check --select=FURB --statistics ` # https://docs.astral.sh/ruff/rules/#refurb-furb
```
3	FURB167	[*] regex-flag-alias
1	FURB122	[*] for-loop-writes
Found 4 errors.
[*] 4 fixable with the `--fix` option.
```
% `ruff check --select=FURB --fix`

% `ruff rule FURB167`
# regex-flag-alias (FURB167)

Derived from the **refurb** linter.

Fix is always available.

## What it does
Checks for the use of shorthand aliases for regular expression flags
(e.g., `re.I` instead of `re.IGNORECASE`).

## Why is this bad?
The regular expression module provides descriptive names for each flag,
along with single-letter aliases. Prefer the descriptive names, as they
are more readable and self-documenting.

## Example
```python
import re

if re.search("^hello", "hello world", re.I):
    ...
```

Use instead:
```python
import re

if re.search("^hello", "hello world", re.IGNORECASE):
    ...
```

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -477,7 +477,7 @@ def generate_json(vehicles):
         json_filename = "parameters-" + vehicle + ".json"
         try:
             with open(json_filename, 'w') as f:
-                for lines in json_lines:
+                for lines in json_lines:  # noqa: FURB122
                     f.write(f"{lines}\n")
         except Exception as e:
             error("Error while creating the JSON file " + vehicle + " in folder " + str(os.getcwd()))  # noqa: E501
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
       - id: flake8
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.6
+    rev: v0.15.7
     hooks:
       - id: ruff-check
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 target-version = "py310"
 line-length = 127
 lint.extend-select = [
+  "FURB",   # refurb
   "UP031",  # printf-string-formatting
   "UP032",  # f-string
 ]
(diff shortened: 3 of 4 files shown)


build_parameters.py
```diff
@@ -477,8 +477,7 @@ def generate_json(vehicles):
         json_filename = "parameters-" + vehicle + ".json"
         try:
             with open(json_filename, 'w') as f:
-                for lines in json_lines:
```
peterbarker, line 480: This patch is unrelated to the topic of the PR, *and* the original code is better.

cclauss: This is `FURB122` which has a ___why this is bad___ section like all ruff rules.

peterbarker: `ruff fule FURB122` is hilariously bad at advocating for this change.  "it's more idiomatic to use a single method".

I don't buy that for a second.  I've written Perl.  I've written *bad* Perl.  I can't help but wonder whether the same thing is happening in Python as it it typically did in Perl - it's possible to be *just so clever* with the language that you can't help but to do this sort of thing, making code that is difficult for people new to the language to understand and for the original author to understand 6 months later.

"Every program can be reduced by one line, and every program has at least one bug."

... and if I *were* to buy the "single method" thing it I'd phrase it:

```
f.write("".join([line + "\n"] for line in json_lines))
```

(and we do actually use that somewhere in our Python scripts)

cclauss: Reverted.

Function call overhead is a steep price to pay in Python.  Avoiding function call overhead is how FURB113 delivers faster performance with a single function call than the repeated function calls that it replaces.
```python
>>> from timeit import timeit
>>> timeit(stmt="for i in range(100): lst.append(i)", setup="lst=[]")
1.535065249998297
>>> timeit(stmt="lst.extend(range(100))", setup="lst=[]")  # FURB113
0.5653657080001722
```
Please stop comparing Python to Perl.  It is like comparing apples to hand grenades.

peterbarker (approved):
LGTM

PR #7571, opened 2026-03-23
Common: param lockdown script rename fix (flight code PR:32489)

rmackay9:
Flight code PR https://github.com/ArduPilot/ardupilot/pull/32489 renamed param-set.lua to param-lockdown.lua so this updates the param lockdown wiki page to reflect this.

I've also added a new section pointing users at the OEM Customisation page which has an alternative method of accomplishing something similar.

### DIFF
--- a/common/source/docs/common-parameter-lockdown.rst
+++ b/common/source/docs/common-parameter-lockdown.rst
@@ -4,7 +4,7 @@
 Parameter Lockdown
 ==================
 
-To avoid accidental changes to critical parameters, ArduPilot provides a parameter lockdown feature using the `param_set.lua <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/param-set.lua>`__ Lua applet.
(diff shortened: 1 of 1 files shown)


common/source/docs/common-parameter-lockdown.rst
```diff
@@ -4,7 +4,7 @@
 Parameter Lockdown
 ==================
 
-To avoid accidental changes to critical parameters, ArduPilot provides a parameter lockdown feature using the `param_set.lua <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/param-set.lua>`__ Lua applet.
+To avoid accidental changes to critical parameters, ArduPilot provides a parameter lockdown feature using the `param_lockdown.lua <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/param-lockdown.lua>`__ Lua applet.
```
ES-Alexander:
```suggestion
To avoid accidental changes to critical parameters, ArduPilot provides a parameter lockdown feature using the `param-lockdown.lua <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/applets/param-lockdown.lua>`__ Lua applet.
```
The underscore there doesn't match the actual filename. Is there a reason for it?

rmackay9: thanks very much for the review, fixed!

PR #7572, opened 2026-03-23
Fix Ruff Q003

hunt0r:
This PR is the result of running `ruff check --select=Q003 --fix` after adding Q003 in `pyproject.toml`.

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -434,7 +434,7 @@ def generate_rst_files(commits_to_checkout_and_parse):
 
             os.chdir(BASEPATH)
         except Exception as e:
-            error("Error while parsing \"Parameters.rst\" | details:\t" + vehicle + "\t" + version  + "\t" + commit_id)
+            error('Error while parsing "Parameters.rst" | details:\t' + vehicle + "\t" + version  + "\t" + commit_id)
             error(e)
             # sys.exit(1)
         debug("")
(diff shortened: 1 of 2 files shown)


build_parameters.py
```diff
@@ -434,7 +434,7 @@ def replace_anchors(source_file, dest_file, version_tag):
 
             os.chdir(BASEPATH)
         except Exception as e:
-            error("Error while parsing \"Parameters.rst\" | details:\t" + vehicle + "\t" + version  + "\t" + commit_id)
+            error('Error while parsing "Parameters.rst" | details:\t' + vehicle + "\t" + version  + "\t" + commit_id)
```
hunt0r, line 437: Want this replaced with an f-string? ([Google style guide 3.10](https://google.github.io/styleguide/pyguide.html#310-strings) prefers that.)
```suggestion
            error(f'Error while parsing "Parameters.rst" | details:\t{vehicle}\t{version}\t{commit_id}')
```

hunt0r: Or a slightly DRY-er alternative, if you like this style
```suggestion
            details = "\t".join(["details", vehicle, version, commit_id])
            error('Error while parsing "Parameters.rst" | ' + details)
```

cclauss: Please save these modifications for a ___separate___ pull request.

hunt0r: per https://github.com/ArduPilot/ardupilot_wiki/pull/7572#pullrequestreview-3999036842: "Do not prefer to make this change here."

PR #7575, opened 2026-03-25
Use `with open() as:` to close file handles

cclauss:
Like:
* ArduPilot/ardupilot#32533

% [`ruff rule SIM115`](https://docs.astral.sh/ruff/rules/open-file-with-context-handler)
# open-file-with-context-handler (SIM115)

Derived from the **flake8-simplify** linter.

## What it does
Checks for cases where files are opened (e.g., using the builtin `open()` function)
without using a context manager.

## Why is this bad?
If a file is opened without a context manager, it is not guaranteed that
the file will be closed (e.g., if an exception is raised), which can cause
resource leaks. The rule detects a wide array of IO calls where context managers
could be used, such as `open`, `pathlib.Path(...).open()`, `tempfile.TemporaryFile()`
or`tarfile.TarFile(...).gzopen()`.

## Example
```python
file = open("foo.txt")
...
file.close()
```

Use instead:
```python
with open("foo.txt") as file:
    ...
```

## References
- [Python documentation: `open`](https://docs.python.org/3/library/functions.html#open)

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -352,8 +352,8 @@ def generate_rst_files(commits_to_checkout_and_parse):
         For each parameter file generate by param_parse.py, it inserts a version tag in anchors
         to do not make confusing in sphinx toctrees.
         """
-        file_in = open(source_file, "r")
-        file_out = open(dest_file, "w")
+        file_in = open(source_file, "r")  # noqa: SIM115
+        file_out = open(dest_file, "w")  # noqa: SIM115
         found_original_title = False
         if "latest" not in version_tag:
             file_out.write(':orphan:\n\n')
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 
 [tool.ruff]
-target-version = "py310"
+target-version = "py38"
 line-length = 127
 lint.extend-select = [
   # "C4",   # flake8-comprehensions
@@ -12,6 +12,7 @@ lint.extend-select = [
   "I",      # isort
   # "PERF", # Perflint
   "Q003",   # avoidable-escaped-quote
+  "SIM115", # open-file-with-context-handler
   "UP031",  # printf-string-formatting
   "UP032",  # f-string
   # "UP",   # pyupgradde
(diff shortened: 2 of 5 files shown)


build_parameters.py
```diff
@@ -213,12 +213,10 @@ def handle_starttag(self, tag, attrs):
...
             version_folder = str(folder)
-            if version_folder.find("stable") > 0 and not version_folder.endswith("stable"): # If finish with
-                stableFirmwares.append(firmware_url[:-1] + version_folder[10:-2])
-            elif version_folder.find("latest") > 0 :
-                stableFirmwares.append(firmware_url[:-1] + version_folder[10:-2])
-            elif version_folder.find("beta") > 0:
+            if "stable" in version_folder and not version_folder.endswith("stable"): # If finish with
                 stableFirmwares.append(firmware_url[:-1] + version_folder[10:-2])
+            elif "beta" in version_folder or "latest" in version_folder:
+                stableFirmwares.append(firmware_url[:-1] + version_folder[10:-2])  # noqa: SIM114
```
peterbarker: What does this have to do with filehandles?

build_parameters.py
```diff
@@ -414,10 +412,8 @@ def replace_anchors(source_file, dest_file, version_tag):
 
             # create a filename for new parameters file
             filename = "parameters-" + vehicle
-            if ("beta" in version or "rc" in version): # Plane uses BETA, Copter and Rover uses RCn
+            if "latest" in version or "beta" in version or "rc" in version: # Plane uses BETA, Copter and Rover uses RCn
                 filename += "-" + version  + ".rst"
-            elif ("latest" in version):
-                filename += "-" + version + ".rst"
             else:
```
peterbarker: What does this have to do with filehandles?

Similarly elsewhere

peterbarker (changes requested):
I don't have a problem with adding SIM in.  But the title on this PR is clear and the patches don't match it.

Please have your PRs do *one* thing and do it well.

cclauss:
Can you please recommend a test plan?

In all cases, we read or write whole files in a single `.read()` or `.write()`, so there is only one command under the context manager and no intervening code to test between the open()/.read()/.close() or open/.write()/.close().

peterbarker:
> Can you please recommend a test plan?

Work out how to run each script so that the changes lines are crossed.  Run the script.  Make sure the script still does what it is supposed to do.

Incidentally, why are we not changing all of these to use `pathlib.Path(...).read_text(encoding=utf8')`?!

PR #7586, opened 2026-03-28
Add warning that QuadPlane VTOL mission legs do not support Terrain waypoints

NDevDrone:
As pointed out by a Partner who came across this not so obvious gap in current function of qplane, this warns users that QuadPlane mission legs flown in VTOL mode do not currently support waypoints with altitude set to Terrain.

Adds the same warning in:
- QuadPlane AUTO missions
- Terrain Following

### DIFF
--- a/common/source/docs/common-terrain-following.rst
+++ b/common/source/docs/common-terrain-following.rst
@@ -60,6 +60,10 @@ In Plane terrain following is available in the following flight modes:
 -  GUIDED - "fly to" waypoints
 -  AUTO - fully autonomous missions
 
+.. warning:: QuadPlane mission legs flown in VTOL mode do not currently
+   support waypoints with altitude set to Terrain. Use Relative or Absolute
+   altitudes for VTOL waypoints.
+
 Use of terrain following in RTL, LOITER, CRUISE, FBWB and GUIDED modes
 is controlled by the :ref:`TERRAIN_FOLLOW<TERRAIN_FOLLOW>` parameter. That parameter defaults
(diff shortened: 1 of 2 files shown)


IamPete1:
Did you test test in SITL? Its possible that there was some other issue for the partner.

In general I think its a good idea to put the warning on VTOL auto. I would call it "experimental".

Hwurzburg:
@IamPete1 ..what is experimental?
> In general I think its a good idea to put the warning on VTOL auto. I would call it "experimental".

IamPete1:
> @IamPete1 ..what is experimental?
> 

In my opinion any VTOL navigation and "do transition" way-points in auto are "experimental", that is to say they might cause you issues if you just blindly run them. For example "do transition" does not slow down or do any approach stuff it just bangs you into VTOL at any speed.

Hwurzburg:
@NDevDrone are you up to adding something like that in the auto mode page along with the warning?

IamPete1 (approved):
Looks good, I have confirmed in the sim.

IamPete1:
If you wanted to add a more general warning about VTOL flight in auto it might be something like this:
"Care must be taken when planning missions using `DO_VTOL_TRANSITION` waypoints. Transition will begin immediately when the vehicle gets to the waypoint. This should be on a relatively straight mission leg and the vehicle should be slowed down before it approaches the transition."

Hwurzburg:
@NDevDrone  please expand notees to include Pete's language

PR #7597, opened 2026-03-31
build_parameters.py: fix htmlparser

khancyr:
fix class (python use capitalize name, links was missing from class but use in exception handling)
fix error handling
fix parsing element
simplify elements parsed and return: get the href directly as that is what interest us and simplify the parsing, pre-filter links to keep only the necessary, do better return to avoid broad exception on missing directory

image

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -400,7 +400,9 @@ def fetch_releases(firmware_url, vehicles):
                 if tag == 'a':
                     attr = dict(attrs)
                     href = attr.get('href')
-                    self.links.append(href)
+                    if href and vehicle.lower() in href.lower():
+                        # Only select folders with the vehicle name in it
+                        self.links.append(href)
 
         html_parser = ParseText()
         try:
@@ -454,7 +456,8 @@ def get_commit_dict(releases_parsed):
                 if tag == 'a':
                     attr = dict(attrs)
                     href = attr.get('href')
-                    self.links.append(href)
+                    if href:
+                        self.links.append(href)
 
         html_parser = ParseText()
         try:
@@ -467,6 +470,9 @@ def get_commit_dict(releases_parsed):
         except Exception as e:
             error(f"Board folders list download error: {e}")
         finally:
+            if len(html_parser.links) == 0:
+                error(f"No board folders found in {url}, skipping this release.")
+                return ""
             last_folder = html_parser.links.pop()
             board_name = os.path.basename(last_folder)
             debug(f"Returning link of the last board folder ({board_name})")
@@ -481,6 +487,9 @@ def get_commit_dict(releases_parsed):
         fetch_link = f"{version_link}/{board}/{file}"
 
         progress(f"Processing link...\t{fetch_link}")
+        if board == "":
+            error(f"Board folder not found in {version_link}, skipping this release.")
+            return "error", "error", "error"
 
         try:
             response = session.get(fetch_link, timeout=30)


khancyr:
rebased on https://github.com/ArduPilot/ardupilot_wiki/pull/7650

khancyr:
rebased after https://github.com/ArduPilot/ardupilot_wiki/pull/7659

build_parameters.py
```diff
@@ -268,14 +268,17 @@ def __init__(self):
             def handle_starttag(self, tag, attrs):
                 if tag == 'a':
                     attr = dict(attrs)
-                    self.links.append(attr)
+                    href = attr.get('href')
```
khancyr, line 402: we just need the link so no need to capture the whole <a> section

build_parameters.py
```diff
@@ -287,13 +290,13 @@ def handle_starttag(self, tag, attrs):
 
         for folder in page_links:  # Non clever way to filter the strings insert by makehtml.py, unwanted folders, and so.
             version_folder = str(folder)
-            firmware_link = f"{firmware_url[:-1]}{version_folder[10:-2]}"
+            firmware_version_url = f"{firmware_url[:-1]}{version_folder}"
```
khancyr, line 428: as we got on the href, we don't need to discard the <a> marking

build_parameters.py
```diff
@@ -317,19 +320,24 @@ def __init__(self):
...
             html_parser.feed(urllib.request.urlopen(url).read().decode('utf8'))
         except Exception as e:
-            error(f"Folders list download error:{e}")
+            error(f"Board folders list download error: {e}")
         finally:
-            last_item = html_parser.links.pop()
-            last_folder = last_item['href']
-            debug(f"Returning link of the last board folder ({last_folder[last_folder.rindex('/')+1:]})")
-            return last_folder[last_folder.rindex('/')+1:]  # clean the partial link
+            if len(html_parser.links) == 0:
```
khancyr, line 473: fix the case of empty list (that shouldn't happens but ...)

build_parameters.py
```diff
@@ -317,19 +320,24 @@ def __init__(self):
...
+                error(f"No board folders found in {url}, skipping this release.")
+                return ""
+            last_folder = html_parser.links.pop()
+            board_name = os.path.basename(last_folder)
```
khancyr, line 477: this is simpler to get the last part of the url

build_parameters.py
```diff
@@ -340,6 +348,9 @@ def fetch_commit_hash(version_link, board, file):
         fetch_link = f"{version_link}/{board}/{file}"
 
         progress(f"Processing link...\t{fetch_link}")
+        if board == "":
```
khancyr, line 490: handle the empty list error

build_parameters.py
peterbarker: Won't this happen any time the server returns `HTTP 500 Still dividing by zero`?

khancyr: yes but that would need to be handled on the Exception level. This just handle the fact the sever respond by with empty or not well formatted page and prevent raising an exception.

Unless you prefer to raise on such case ? But if that happens how do we get notification as this happens on the server ? (no CI for parameter versionning yet)

build_parameters.py
khancyr: master:
version folder: {'href': '/Copter/stable-4.6.2'}
firmware version url: https://firmware.ardupilot.org/Copter/stable-4.6.2

PR:
version folder: /Copter/stable-4.6.3
firmware version url: https://firmware.ardupilot.org/Copter/stable-4.6.3

build_parameters.py
khancyr: attr : {'href': '/Copter/stable-4.4.2'}
href : /Copter/stable-4.4.2

Hwurzburg:
@peterbarker  ping

PR #7551, opened 2026-03-19
Vagrantfile: Add Ubuntu 24.04 LTS noble and 26.04 LTS resolute

cclauss:
All Ubuntu releases in `Vagrantfile` are no longer under Ubuntu Standard Support, so upgrade them to align with:
* https://ubuntu.com/about/release-cycle

Why?  Ubuntu 20.04 default Python 3.8.10 is also [`end-of-life`](https://devguide.python.org/versions) and can no longer be proven to be compatible with this project's code.

% `docker run -it ubuntu:20.04`
\# `apt update ; apt install --yes lsb-release python-is-python3 ; lsb_release -ds ; python -V && exit`
```
Ubuntu 20.04.6 LTS
Python 3.8.10
```
Should we ___wait a month___ and also add Ubuntu 26.04?

> [!CAUTION]
> Ubuntu 20.04 LTS Resolute Raccoon is scheduled to be released on Thu 23rd Apr 2026 (~10 days).
https://documentation.ubuntu.com/release-notes/26.04/schedule/#resolute-raccoon-schedule

Related to:
* #6316

I have never used Vagrant so ___this pull request has not been tested in any way!!!___

The `vb.customize` memory settings should be carefully reviewed by someone who knows Vagrant.

PR #7599, opened 2026-03-31
pyproject.toml: target-version = "py38" to match Vagrantfile

cclauss:
Related to:
* https://docs.astral.sh/ruff/settings/#target-version
* #7551
* #7635

Vagrant still uses Ubuntu 20.04 focal which defaults to Python 3.8.
https://github.com/ArduPilot/ardupilot_wiki/blob/3464385e1521f3c76bead96fd4af8fafa2c1a54c/Vagrantfile#L10-L11
To upgrade to `py39`, we would need to merge #7551 (or similar).

Debian Bullseye LTS end-of-life is [2026-08-31](https://wiki.debian.org/DebianReleases#Versions_of_Debian_stable), which is why the main repo is pinned to `py39`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 
 [tool.ruff]
-target-version = "py310"
+target-version = "py38"
 line-length = 127
 lint.extend-select = [
   "FURB",   # refurb


pyproject.toml
```diff
@@ -4,7 +4,7 @@ ignore-words-list = "bu,collets,diamon,dout,empy,extint,fram,inh,minite,ned,parm
 skip = "ARCHIVED/*,LICENSE,*.ai,*.pdf,*.svg"
 
 [tool.ruff]
-target-version = "py310"
```
khancyr, line 7: nah, we should keep 3.10 by default as that what is on ubuntu22.04. Just need to unsure what is on the server and what Henry use the most if we want to raise up to 3.12

khancyr (changes requested):

pyproject.toml
cclauss: Vagrant still uses Ubuntu 20.04 focal which defaults to Python 3.8.

To upgrade to `py39`, we would need to merge #7551 (or similar) as stated above.

Also, Debian Bullseye LTS end-of-life is [2026-08-31](https://wiki.debian.org/DebianReleases#Versions_of_Debian_stable), which is why the main repo is pinned to `py39`.

peterbarker (approved):

PR #7600, opened 2026-04-01
build_parameters.py: use subprocess for git commands

khancyr:
use subprocess instead of os.system for git operation. 
fix the exception scope

Capture d’écran du 2026-04-01 16-29-27

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -46,6 +46,16 @@ parser.add_argument("--destination", dest='destFolder', default="../../../../new
 parser.add_argument('--vehicle', dest='single_vehicle', help="If you just want to copy to one vehicle, you can do this. Otherwise it will work for all vehicles (Copter, Plane, Rover, AntennaTracker, Sub, Blimp)")  # noqa: E501
 args = parser.parse_args()
 
+
+# Parameters
+COMMITFILE = "git-version.txt"
+BASEURL = "https://firmware.ardupilot.org/"
+ALLVEHICLES = ["AntennaTracker", "Copter", "Plane", "Rover", "Sub", "Blimp"]
+VEHICLES = ALLVEHICLES
+# Filter out versions below this semantic version threshold.
+PARAM_PARSE_MINIMUM_VERSION = (3, 9, 0)
+
+BASEPATH = ""
 error_count = 0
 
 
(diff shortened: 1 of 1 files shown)


build_parameters.py
```diff
@@ -73,6 +74,80 @@ def format(self, record):
...
+            )
+            if result.stderr:
+                debug(f"Git stderr: {result.stderr}")
+            return result.stdout
+
+        except subprocess.CalledProcessError as e:
+            # Check if it's a lock file issue
+            if 'index.lock' in str(e.stderr) or 'Unable to create' in str(e.stderr):
+                debug(f"Git lock detected on attempt {attempt + 1}, cleaning up...")
+                cleanup_git_locks(cwd)
```
peterbarker: Doesn't this entirely defeat the purposes of said locks?!

khancyr: yes and no.
the lock prevent concurent access that shouldn't happen on manipulating the sub repo used for parameters generation

But if for any reason an issue happens on the subprocess call, and more rarely on the git operation, the lock file will prevent any other git operations until it is manually removed. This just enforce this. Normally, the lock should be cleaned up by git even on failure.

khancyr:
@peterbarker I have updated to only clean git lock when it is a staled/old one. So that should cope with the issues I have seen and retain the lock mecanism

peterbarker (changes requested):
Needs rebase

khancyr:
image

build_parameters.py
peterbarker: Sorry, but could you remove this stale-lock-cleanup from this PR, please?  Replace with a message to the user saying they've got stale locks and providing instructions on how to fix.

I am deeply dubious of this stale lock removal.  If we get this wrong then we end up corrupting people's git repositories, potentially losing them large amounts of work.

We can discuss on a fresh PR if you like.  I actually don't think it is unreasonable to ask a user to remove these locks by hand - it should only be exceptional cases where these locks exist.

build_parameters.py
```diff
@@ -83,16 +93,78 @@ def format(self, record):
...
+                cleaned += 1
+            except OSError as e:
+                debug(f"Could not remove git lock {lock_file}: {e}")
+
+    if cleaned > 0:
+        debug(f"Cleaned up {cleaned} git lock file(s)")
+    return cleaned
+
+
+def run_git_command_with_retry(cmd, cwd=None, check=True, max_retries=3):
```
```suggestion
def run_git(cmd, cwd=None, check=True, max_retries=3):
```

peterbarker (changes requested):

build_parameters.py
khancyr: @peterbarker sure, I can remove the git lock things. Don't forget this is used on the ardupilot git repo that is used to generate the parameters, not the wiki git repo.

 I don't think it should be used for anything else as we are running command like clean and reset so running the build_parametesr.py script will nuke the repo in some way.
 
 Let's see how it goes without the lock management, we can still add it back if it is too much trouble.

khancyr:
@peterbarker rebase and updated without the git lock removal and function name changed

working as expected 
image

peterbarker (approved):
Great, thanks.

PR #7602, opened 2026-04-01
build_parameters.py: fix parameters duplication issues for old and no…

khancyr:
…tify for new

This fix the parameters duplicate we got on some old parameters (4.0.X serie). Resolving the duplication fix also the RAM issue on building.
The ram issue comes from duplicated label and sphinx looping on duplicated label.

A new function report if duplicate a found

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -81,6 +81,82 @@ VEHICLES = ALLVEHICLES
 
 BASEPATH = ""
 
+
+def rst_has_duplicate_labels(filepath: str) -> bool:
+    """
+    Check an RST file for duplicate label definitions.
+    Returns True if duplicates are found, False otherwise.
+
+    RST labels look like: .. _label_name:
+    """
+    try:
+        with open(filepath, 'r', encoding='utf-8') as f:
+            content = f.read()
+    except (UnicodeDecodeError, OSError) as e:
+        error(f"Error checking RST file {filepath}: {e}")
+        return False
+
+    # Find all RST label definitions
+    labels = re.findall(r'^\.\. _([^:]+):', content, re.MULTILINE)
+
+    # Check for duplicates
+    seen = set()
+    duplicates = []
+    for label in labels:
+        label_lower = label.lower()  # RST labels are case-insensitive
+        if label_lower in seen:
+            duplicates.append(label)
+        seen.add(label_lower)
+
+    if duplicates:
+        logger.warning(f"Found {len(duplicates)} duplicate RST labels in {filepath}: {duplicates[:5]}")
+        return True
+    return False
+
+
+def dedupe_rngfnd_parameters_sections(filepath: str) -> None:
+    """Keep only the first RNGFNDx_* Parameters section and drop subsequent duplicate sections."""
+    label_re = re.compile(r'^\.\. _parameters_RNGFND([0-9A-Za-z]+)_.*:$', re.IGNORECASE)
+    section_label_re = re.compile(r'^\.\. _RNGFND([0-9A-Za-z]+).*:$', re.IGNORECASE)
+
+    try:
+        with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
+            lines = f.readlines()
+    except (UnicodeDecodeError, OSError) as e:
+        debug(f"Failed to dedupe RNGFND parameters in {filepath}: {e}")
+        return
+
+    seen_labels = set()
+    saved_lines = []  # Lines to write back to the file after deduplication
+    i = 0
+    while i < len(lines):
+        current_line = lines[i]
+        current_line_stripped = current_line.strip()
+
+        rangefinder_label = label_re.match(current_line_stripped)
+        if rangefinder_label:
+            rngfnd_id = rangefinder_label.group(1)
+            if rngfnd_id in seen_labels:
+                # Don't save the duplicated line on new file, skip to next section
+                i += 1
+                while i < len(lines) and not section_label_re.match(lines[i].strip()):
+                    i += 1
+                continue
+
+            seen_labels.add(rngfnd_id)
+
+        saved_lines.append(current_line)
+        i += 1
+
+    if len(saved_lines) != len(lines):
+        try:
+            with open(filepath, 'w', encoding='utf-8') as f:
+                f.writelines(saved_lines)
+            debug(f"Dedupe applied to RNGFND sections in {filepath}")
+        except (UnicodeDecodeError, OSError) as e:
+            debug(f"Failed to dedupe RNGFND parameters in {filepath}: {e}")
+
+
 # Dicts for name replacing
 vehicle_new_to_old_name = { # Used because "param_parse.py" args expect old names
     "Rover": "APMrover2",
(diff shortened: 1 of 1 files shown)


build_parameters.py
```diff
@@ -81,6 +81,84 @@ def format(self, record):
 
 BASEPATH = ""
 
+
+def check_rst_for_duplicate_labels(filepath):
```
cclauss:
```suggestion
def check_rst_for_duplicate_labels(filepath: str) -> bool:
```

build_parameters.py
```diff
@@ -81,6 +81,84 @@ def format(self, record):
...
+        if duplicates:
+            logger.warning(f"Found {len(duplicates)} duplicate RST labels in {filepath}: {duplicates[:5]}")
+            return True
+        return False
+    except Exception as e:
+        error(f"Error checking RST file {filepath}: {e}")
+        return False
+
+
+def dedupe_rngfnd_parameters_sections(filepath):
```
cclauss:
```suggestion
def dedupe_rngfnd_parameters_sections(filepath: str) -> None:
```

build_parameters.py
```diff
@@ -81,6 +81,84 @@ def format(self, record):
...
+            label_lower = label.lower()  # RST labels are case-insensitive
+            if label_lower in seen:
+                duplicates.append(label)
+            seen.add(label_lower)
+
+        if duplicates:
+            logger.warning(f"Found {len(duplicates)} duplicate RST labels in {filepath}: {duplicates[:5]}")
+            return True
+        return False
+    except Exception as e:
```
cclauss: This over broad exception covers too many lines.  This can easily make programming errors hard to discover.

build_parameters.py
```diff
@@ -81,6 +81,84 @@ def format(self, record):
...
+
+            out_lines.append(line)
+            i += 1
+
+        if len(out_lines) != len(lines):
+            with open(filepath, 'w', encoding='utf-8') as f:
+                f.writelines(out_lines)
+            debug(f"Dedupe applied to RNGFND sections in {filepath}")
+
+    except Exception as e:
```
cclauss: Massive try except block plus overly broad exception.

cclauss (commented):
Type hints please on new functions.

cclauss (approved):
Nice improvement!  Thanks.

build_parameters.py
```diff
@@ -428,6 +509,12 @@ def replace_anchors(source_file, dest_file, version_tag):
                 replace_anchors("Parameters.rst", filename, filename[10:-4])
                 os.remove("Parameters.rst")
                 debug(f"File {filename} generated. ")
+                # Remove duplicate RNGFNDx_ Parameters sections before checking labels.
+                dedupe_rngfnd_parameters_sections(filename)
+                # Check for duplicate RST labels in the generated file
+                if check_rst_for_duplicate_labels(filename):
```
```suggestion
                if rst_has_duplicate_labels(filename):
```

(you're not interested in whether you successfully checked for duplicate labels (that's what exceptions are for), you're checking whether there *are* duplicate labels.

build_parameters.py
```diff
@@ -81,6 +81,87 @@ def format(self, record):
...
+        seen.add(label_lower)
+
+    if duplicates:
+        logger.warning(f"Found {len(duplicates)} duplicate RST labels in {filepath}: {duplicates[:5]}")
+        return True
+    return False
+
+
+def dedupe_rngfnd_parameters_sections(filepath: str) -> None:
+    """Keep only the first RNGFNDx_* Parameters section and drop subsequent duplicate sections."""
```
peterbarker, line 118: Why are the rangefinder parameters specifically a problem?  Should we fix the root cause here rather than adding an exception here?

If I'm thinking right this code is what would make this script useful where `rstcheck` would not be.

peterbarker (commented):
I'm keen to get this in to fix the immediate problem, but I am curious to know why not `rstcheck`; the equivalent for markdown seems to have worked quite well to clean things up on the main ardupilot firmware repo....

I'm happy with this going in whenever.

build_parameters.py
```diff
@@ -81,6 +81,87 @@ def format(self, record):
...
+                        break
+                    i += 1
+                continue
+            seen.add(idx)
+            out_lines.append(line)
+            i += 1
+            continue
+
+        out_lines.append(line)
+        i += 1
```
hunt0r: optional: I think a subfunction + 'early continue' version is slightly easier to understand. And it avoids duplicated calls to `line.strip()`. Here it is, if you like it too:
```suggestion
    def _find_block_end(lines, start):
        i = start + 1
        while i < len(lines):
            stripped = lines[i].strip()
            if label_re.match(stripped):
                break
            if include_re.match(stripped):
                break
            i += 1
        return i

    seen = set()
    out_lines = []
    i = 0
    while i < len(lines):
        line = lines[i]
        stripped = line.strip()
        m = label_re.match(stripped)

        if not m:
            out_lines.append(line)
            i += 1
            continue

        idx = m.group(1)
        block_end = _find_block_end(lines, i)

        if idx not in seen:
            seen.add(idx)
            out_lines.extend(lines[i:block_end])

        i = block_end
```

(By "optional" I explicitly mean: OK to resolve this comment without response.)

build_parameters.py
hunt0r: optional: I agree the rename is better, but also I'm ok with this as-is.

hunt0r (approved):

build_parameters.py
hunt0r: I should flag that I used (abused?) the PEP8 "leading underscore = private" convention for the `_find_block_end( )` function ([link](https://peps.python.org/pep-0008/#method-names-and-instance-variables)).
I like that leading-underscore to imply "this is a special-purpose function, not generally useful".
But the fact that it's nested basically implies the same. So if the leading-underscore isn't common in our codebase, probably better to omit it.

cclauss: https://dbader.org/blog/meaning-of-underscores-in-python

hunt0r: @cclauss What's your opinion? Should this nested function have a leading underscore, or no?
(Also fine if you don't care either way 😄)

peterbarker:
@khancyr I've tested this locally and it doesn't fix the build for me on a 16GB virtual machine - the oomkiller still rears its ugly head.

Raising to 20GB does fix it.

Things still work, and @cclauss thinks this is better, so I'm inclined to merge anyway.  But while I have a reproducible failure here is there some instrumentation or whatnot you'd like to add?

Also failed with
```
ESC[1;31m[update.py][error]: __init__() got an unexpected keyword argument 'allowed_methods'ESC[0m
[update.py][error]: 1 errors during Wiki build
```

with 20GB of memory, so that's fun

peterbarker:
.... more testing here.... the build completes successfully on jammy with and without your patches here.

khancyr:
@peterbarker I cannot find reference to `allowed_methods`, how are you testing this ?
are you sure you generate the parameters with build_parameters.py and copy them properly to use them on the update.py, and not use some old one's on update.py ?

I test using the docker way and the leak was done ... I will do some more tests tonigth ?

peterbarker:
> @peterbarker I cannot find reference to `allowed_methods`, how are you testing this ? are you sure you generate the parameters with build_parameters.py and copy them properly to use them on the update.py, and not use some old one's on update.py ?

I think that failure can probably be ignored for now.

> I test using the docker way and the leak was done ... I will do some more tests tonigth ?

I was using the testing instructions in the top of `Vagrantfile`.

I'm happy to just hit "merge" here.  Do you want me to do that or keep investigating?

Were we expecting to build in 16GB of RAM?! :-)

khancyr:
@peterbarker we can merge, it was clearly fixing stuff on my side and the infinite recursion that was cause the ram issue. 
After merge, I will look again but with Vagrant workflow to see where the diff is and if there is another issue.

peterbarker (approved):
LGTM

Issue #7604, opened 2026-04-03
ESCs and Motors: link to HobbyWing Telemetry Hub broken


rmackay9:
On our [ESCs and Motors page](https://ardupilot.org/copter/docs/common-escs-and-motors.html) under the telemetry section we have a broken link to Hobbywing Telemetry Hub (DatalinkV2)(uses LUA driver)

The current broken link points to: https://www.hobbywingdirect.com/products/data-link-v2 which should be replaced by a link to our lua script's md file:  https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/drivers/Hobbywing_DataLink.md

The above linked .md file also has a broken link to https://www.hobbywing.com/en/products?id=59 which should probably be updated to point to https://www.hobbywing.com/en/products/datalinkv2273

PR #7605, opened 2026-04-03
Common: Hobbywing DataLink v2 link fix

rmackay9:
This resolves issue https://github.com/ArduPilot/ardupilot_wiki/issues/7604 by fixing the link to the Hobbywing DataLink v2 to point to our Lua driver.  I've also slightly renamed the wording attached to the link to make it more closely match the product itself.

Related flight code PR: https://github.com/ArduPilot/ardupilot/pull/32668

I've built this locally and it looks good to me.

### DIFF
--- a/common/source/docs/common-escs-and-motors.rst
+++ b/common/source/docs/common-escs-and-motors.rst
@@ -75,7 +75,7 @@ Telemetry
     :maxdepth: 1
 
     ESC Telemetry <common-esc-telemetry>
-    Hobbywing Telemetry Hub (DatalinkV2)(uses LUA driver) <https://www.hobbywingdirect.com/products/data-link-v2>
+    Hobbywing DataLink V2 (Lua Driver) <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/drivers/Hobbywing_DataLink.md>
 
 .. note:: Currently ArduPilot only supports telemetry on BLHeli or DroneCAN/CAN ESCs, not throttle signal wire reported telemetry that some single unit ESCs report.
 


common/source/docs/common-escs-and-motors.rst
```diff
@@ -75,7 +75,7 @@ Telemetry
     :maxdepth: 1
 
     ESC Telemetry <common-esc-telemetry>
-    Hobbywing Telemetry Hub (DatalinkV2)(uses LUA driver) <https://www.hobbywingdirect.com/products/data-link-v2>
+    Hobbywing DataLink V2 (Lua Driver) <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/drivers/Hobbywing_DataLink.md>
```
Hwurzburg, line 78:
```suggestion
    `Hobbywing DataLink V2 (Lua Driver) <https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/drivers/Hobbywing_DataLink.md>`__
```

Hwurzburg (changes requested):

common/source/docs/common-escs-and-motors.rst
rmackay9: Txs for the review, I don't think this is a correct change though.  The original works but the suggested change looks like this

image

Hwurzburg: must be because its in a toctree...I will find to way make it a hotlink....stdby

rmackay9: OK thanks.  .. but what I have there actually works and is consistent with how it was done previously.  I was also a little surprised that we could have links to external sites in the toctree but it seems to work.
image

Hwurzburg: yes....the url is hotlinked...my mistake

PR #7607, opened 2026-04-04
Common: DroneBridge setup clarified

rmackay9:
This clarifies the DroneBridge setup based on my recent experiences using it.

The changes include:

1. Title changed to "DroneBridge for ESP32" to make consistent with what the manufacturer calls it
2. Images copied locally instead of being pulled from the manufacturer's site.  This allowed some customisation (e.g. removing the names of other flight code stacks)
3. Images added to some steps (like flashing the firmware)
4. Emphasize the official hardware which is easier to setup and also supports the DroneBridge project
5. Lead the user towards the ESP-NOW configuration which should be the most popular considering its 1km range
6. Simplified the wording and setup order

@seeul8er could you confirm that you're OK with these changes?


seeul8er (approved):
I like it and it's all correct. Thanks for putting in the work. I will link your page today.

PR #7608, opened 2026-04-04
Copter: Add doc on newly supported auto_trim in position hold modes and fix punctuation

EvansCesilThomas:
Auto Trim is supported in position hold modes in Copter 4.7. Add documentation to cover this.

### DIFF
--- a/copter/source/docs/autotrim.rst
+++ b/copter/source/docs/autotrim.rst
@@ -20,19 +20,19 @@ With AHRS AutoTrim the roll and pitch corrections are captured as you fly in a s
 .. image:: ../images/MP_SaveTrim_Ch7PWMCheck.png
     :target: ../_images/MP_SaveTrim_Ch7PWMCheck.png
 
-2. Set the CH7 Option to "AHRS AutoTrim" on the Software > Copter Pids screen and press the "Write Params" button
+2. Set the CH7 Option to "AHRS AutoTrim" on the Software > Copter Pids screen and press the "Write Params" button.
 
-3. Find a wind free environment with sufficient space to fly your copter without crashing into something
+3. Find a wind free environment with sufficient space to fly your copter without crashing into something.
 
-4. Put the vehicle in STABILIZE or ALT_HOLD mode.
+4. Put the vehicle in manual Stabilize/AltHold mode or a position hold Loiter/PosHold mode (recommended). 
 
-5. Arm the vehicle and lift into a hover. The vehicle must be flying in order to enable autotrim. Flip the switch that you setup(CH7) to its High position (>1800us). "Autotrim running" message will be sent to the GCS.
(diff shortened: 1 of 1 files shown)


copter/source/docs/autotrim.rst
```diff
@@ -20,19 +20,19 @@ With AHRS AutoTrim the roll and pitch corrections are captured as you fly in a s
...
 
 5. Arm the vehicle and lift into a hover. The vehicle must be flying in order to enable autotrim. Flip the switch that you setup(CH7) to its High position (>1800us). "Autotrim running" message will be sent to the GCS.
 
-6. Fly your copter for about 25 seconds in a stable hover in position, keeping it from drifting with the pitch and roll sticks. As the trim is adjusted, less and less corrections with the sticks will be required.
+6. If in a manual mode, fly your copter for about 25 seconds in a stable hover in position, keeping it from drifting with the pitch and roll sticks. As the trim is adjusted, less and less corrections with the sticks will be required. If using a position hold mode, take off and let the copter hover in place while not touching the sticks. 
 
 7. Once satisfied, flip the switch low, saving the trims, and land. "Trim save" message will be sent to the GCS.
 
-8. Take off again in stabilize mode and check if your copter is flying level now. If not repeat steps 4 - 7 again.
+8. Take off in stabilize mode and check if your copter is flying level now. If not repeat steps 4 - 7 again.
```
rmackay9: maybe we could remove the mention of stabilize mode here completely

rmackay9 (approved):
txs for this.  It would be good if we updated the screen shots and wording to reflect how MP look now but it's still an improvement

copter/source/docs/autotrim.rst
EvansCesilThomas: I think the best way of gauging the new trims is by flying in stabilize/loiter. What do you think?

EvansCesilThomas:
> txs for this. It would be good if we updated the screen shots and wording to reflect how MP look now but it's still an improvement

Will look at it and edit accordingly

copter/source/docs/autotrim.rst
rmackay9: I think a lot of users struggle with stabilize mode and AltHold mode should work just as well

txs again for all this

EvansCesilThomas: Right, will edit to AltHold. By the way I meant "AltHold" instead of "loiter" in the above comment.

EvansCesilThomas (commented):
@rmackay9 I edited the text now. I also added to the last note under Auto_Trim that it may be better to yaw the copter during the trimming process as suggested by @lthall in https://discuss.ardupilot.org/t/calibrating-level-from-loiter/142290/16?u=strikeeagle. Maybe I should edit that to say to do a continuous 360 deg yaw?

As for the images, I will try to update those I can. But since I don't have an RC with me now, I cannot update the images showing the RC calibration screen. Hopefully someone else can do it?

Hwurzburg:
update the images in a new PR since you are using web editor...

PR #7610, opened 2026-04-05
the TF02-Pro can be used in I2C

IPv777:


Hwurzburg:
This not really useful, since it is added only for the discontinued version, not the pro. and more importantly, the I2C setup is not detailed on this page....if you can add that, then noting that the PRO (not the dicontinued version) has I2C would be worthwhile...

Hwurzburg:
you have not modified this PR as requested.....currently not approved....

PR #7615, opened 2026-04-07
Copter: AutoTrim - Add an image and correct a description to reflect new MP version

EvansCesilThomas:
Due to not having an RC with me now, I couldn't update the first two images. Hopefully someone can do that.


rmackay9 (approved):
LGTM, thanks!

Hwurzburg:
that image is not the one being called from the wiki page   its calling from copter/source/images not images directory...look at the url  ../images vs ../../../images which is where you made the change

PR #7616, opened 2026-04-07
Dev: update release procedures page

rmackay9:
This updates the release procedures with our latest procedure which includes all the vehicles.

### DIFF
--- a/dev/source/docs/release-procedures.rst
+++ b/dev/source/docs/release-procedures.rst
@@ -1,12 +1,12 @@
 .. _release-procedures:
 
-=========================
-Copter Release Procedures
-=========================
+==================
+Release Procedures
+==================
 
 ArduPilot is very reliable because bug fixes are prioritised over new features and releases are performed carefully and only after sufficient testing has been completed.
 
-This page outlines the steps that are normally followed for a Copter release.  Listed mostly as a reference so that we don't forget the steps.
+This page outlines the steps that are normally followed for a new release.  Listed mostly as a reference so that we don't forget the steps.
 
 Alpha Testing
 =============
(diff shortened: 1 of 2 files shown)


dev/source/docs/release-procedures.rst
```diff
@@ -22,76 +22,139 @@ Releasing Beta Versions / Release Candidates
...
+Create a Release Pull Request
+-----------------------------
 
-- ``git checkout Copter-4.3`` to switch to the existing release branch ("Copter-4.3" should be replaced with the correct major and minor version numbers)
+1. In your local repo, create a new branch based on the upstream release branch:
 
-Pull in changes from master
----------------------------
+  - ``git checkout upstream/ArduPilot-4.7`` ("ArduPilot-4.7" should be replaced with the correct major version number)
+  - ``git checkout -b ap-470-xxx`` to create a new branch for the PR
```
ES-Alexander: This naming format seems likely to get confusing if/when we get to 4.10

dev/source/docs/release-procedures.rst
```diff
@@ -22,76 +22,139 @@ Releasing Beta Versions / Release Candidates
...
 
-OR
+4. As PRs are included in the new branch, move the PR from the "No Status" or "Pending" column of the project to the column for this release
 
-- use ``gitk master`` to cherry-pick in just some changes from master
+5. Update version, release notes and tags
 
-Update version, release notes and tags
---------------------------------------
+  - update the THISFIRMWARE and FIRMWARE_VERSION definitions in `version.h <https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/version.h>`__
```
ES-Alexander, line 62: Presumably this should apply for each vehicle too, like the point after?

dev/source/docs/release-procedures.rst
```diff
@@ -22,76 +22,139 @@ Releasing Beta Versions / Release Candidates
...
+  - update `ReleaseNotes.txt <https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/ReleaseNotes.txt>`__ for each vehicle
+
+6. Create the PR and gather signatures from other developers
```
ES-Alexander, line 65: This could be clearer on who is expected to "sign off". Is it all maintainers? At least one maintainer per vehicle subsystem? All maintainers whose feature sets have had a code change in the PR?

dev/source/docs/release-procedures.rst
```diff
@@ -22,76 +22,139 @@ Releasing Beta Versions / Release Candidates
...
 Announce the release to the beta testers
 ----------------------------------------
 
-Let Beta testers know the new version is available by posting on these sites:
+Let Beta testers know the new version is available by posting on discuss and facebook
 
+- `Discuss's Plane category <https://discuss.ardupilot.org/c/arduplane>`__ (create a new category if required)
 - `Discuss's Copter category <https://discuss.ardupilot.org/c/arducopter>`__ (create a new category if required)
+- `Discuss's Rover category <https://discuss.ardupilot.org/c/ardurover>`__ (create a new category if required)
+- `Discuss's Sub category <https://discuss.ardupilot.org/c/sub>`__ (create a new category if required)
```
ES-Alexander, line 148: Possibly out of scope for these instructions, but it is appreciated for [the BR forum](https://discuss.bluerobotics.com/c/bluerobotics-software/ardusub/16) to get a post for new releases too, at least insofar as they concern Sub.

That said, I don't think that needs to be the role of ArduPilot - it's ok to leave that to BR if it's deemed too much hassle to do 🤷‍♂️ :-)

dev/source/docs/release-procedures.rst
rmackay9: yes, in a year or two we will need to decide if we move to 4.10 or 5.0.  In the past we've always upgraded the major version before the minor reached 10.

dev/source/docs/release-procedures.rst
rmackay9: yes it should.  We should probably mention that the release notes are copied across to each vehicle directory

dev/source/docs/release-procedures.rst
rmackay9: yes txs.  It should be all vehicle maintainers + Systems lead at least.

dev/source/docs/release-procedures.rst
rmackay9: Good point, I did post in the BlueRobotics forum at the start of 4.7.0 beta testing so at least the major announcements should go there.

PR #7619, opened 2026-04-07
4.8:30298/Currawong ESC info on using custom build server

rmackay9:
Flight code PR https://github.com/ArduPilot/ardupilot/pull/30298 means that PiccoloCAN must be enabled using the custom build server.  This wiki PR makes that clear for the Currawong ESCs which is the major user of this protocol.

As a drive-by change I've also updated the custom builder server info for the XFRobot camera gimbal.  Originally it was only available via the custom build server but now it's enabled with all the other camera gimbal


SchrodingersGat:
@rmackay9 thanks, those changes look great

rmackay9:
I've updated this now to move the DroneCAN instructions above the PiccoloCAN.  I've built it locally and it looks OK to me.

There is some duplication on the page in that we specify the autopilot parameters twice but they're slightly different and I think users will appreciate not having to jump around on the page.

I think we should merge this before the 4.8 release because DroneCAN is supported now and it's best if users use that.  The warning about version (4.7 and 4.8) is valid even now and won't change once 4.8 is released.

Hwurzburg:
> I think we should merge this before the 4.8 release because DroneCAN is supported now and it's best if users use that. The warning about version (4.7 and 4.8) is valid even now and won't change once 4.8 is released.

I will work on that...needs upcoming update also to let people know something changed...will do as in the past when merging..

Hwurzburg (approved):

rmackay9:
Thanks very much!

PR #7620, opened 2026-04-07
pre-commit: Detect errors in scripts with shellcheck

cclauss:
Like:
* ArduPilot/ardupilot#32621

### Shellcheck
* https://www.shellcheck.net -- 'shellcheck --help'
* https://github.com/koalaman/shellcheck

Localhost testing:
% `find . -type f -name "*.sh" -exec shellcheck {} +`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ env:
 jobs:
 
   pre-commit:
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
     - uses: actions/setup-python@v6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,11 @@ repos:
     hooks:
       - id: ruff-check
 
+  - repo: https://github.com/koalaman/shellcheck-precommit
+    rev: v0.11.0
+    hooks:
+      - id: shellcheck
+
   - repo: https://github.com/sphinx-contrib/sphinx-lint
     rev: v1.0.2
     hooks:
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,10 @@
+# https://www.shellcheck.net -- 'shellcheck --help'
+# https://github.com/koalaman/shellcheck
+# 'find . -type f -name "*.sh" -exec shellcheck {} +'
+
+# To improve our code quality, comment out one for the following and create a pull request:
+disable=SC2006  # 1 instance
+disable=SC2046  # 3 instances
+disable=SC2086  # Many instances
+disable=SC2162  # 3 instances
+disable=SC2185  # 1 instance


.pre-commit-config.yaml
```diff
@@ -23,6 +23,11 @@ repos:
     hooks:
       - id: ruff-check
 
+  - repo: https://github.com/koalaman/shellcheck-precommit
```
peterbarker, line 26: Again, supply-chain attack issues.

Personally I'd like to reduce our exposure to these.  Cloning this into our own org might be an option.

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ env:
 jobs:
 
   pre-commit:
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-latest
```
peterbarker, line 12: Why?

.pre-commit-config.yaml
cclauss: Both the pre-commit hook and shellcheck are released by the same author, as described in the documentation above.  They were released 8+ months ago and are the same shellcheck version that can be apt-installed in Ubuntu 26.04.  Their versions will remain fixed until:
1. The original author publishes a new release `___AND___`
2. We create a new pull request and merge it into this repo that does:
    * % `pre-commit autoupdate && pre-commit run --all-files` 

As we discussed on last week's DevCall, all of our pre-commit hooks are finding errors but not fixing them.

All our hooks either run in a separate Python venv or (like this one) in a Docker container.  This is safer than pip installing dependencies.  None of our hooks has the ability to modify the pull requests they are checking or existing Python dependencies.

We also discussed at last week's DevCall a `repo: local` alternative, but I am not a fan.
* ArduPilot/ardupilot#32644

.github/workflows/test.yml
cclauss: * https://github.com/ArduPilot/ardupilot/pull/32621#pullrequestreview-4056146663

.pre-commit-config.yaml
peterbarker: > Both the pre-commit hook and shellcheck are released by the same author, as described in the documentation above. They were released 8+ months ago and are the same shellcheck version that can be apt-installed in Ubuntu 26.04. Their versions will remain fixed until:

To be clear here - I have absolutely no reason to doubt the author's good intentions, these are fantastic tools.  But I'd also think the same of the trivy maintainer who got duped - by an attack which probably also would have duped me!
 
>     1. The original author publishes a new release `___AND___`
> 
>     2. We create a new pull request and merge it into this repo that does:
>        
>        * % `pre-commit autoupdate && pre-commit run --all-files`

Did you catch the fact that it wasn't the source code that was modified in the trivy attack - *the releases were replaced with malware*!

This places anybody that does `pre-commit install` or `autoupdate` or whatever - especially those that are doing this in CI - at risk.

> As we discussed on last week's DevCall, all of our pre-commit hooks are finding errors but not fixing them.

What the hooks are *supposed* to do isn't relevant.

> All our hooks either run in a separate Python venv or (like this one) in a Docker container. This is safer than pip installing dependencies. None of our hooks has the ability to modify the pull requests they are checking or existing Python dependencies.

They may run locally on people's machines.  A venv is not a great sandbox for the sorts of things that you can do in Python.  Docker rather better, but impractical?   Having pre-commit `hooks:` be able to enumerate the permissions the thing requires (e.g. read-dir, read-file which is all these things should require) would be a good step forward in terms of isolation; plenty of ways you could enforce that sort of thing in Linux, at least.

PR #7625, opened 2026-04-08
common: use-cases-and-applications: expand Sub section

ES-Alexander:
@Williangalvani I've transferred across the Sub use-cases [from the original docs](https://www.ardusub.com/introduction/applications.html), with some video examples where I could find them.

I've also moved the use-cases and applications to the Introduction section of the sidebar, because I don't expect many users will find it in the "Additional Information" down the bottom. Unfortunately there is [a Sphinx limitation](https://github.com/sphinx-doc/sphinx/issues/8287) that means it needs to be at the top of the sidebar section it's in, but doesn't feel _too_ problematic 🤷‍♂️ 

Screenshot 2026-04-08 at 8 33 52 pm

---

~~@Hwurzburg While I was at it I decided to add site filtering to minimise the irrelevant use-cases in each wiki, and in doing so found that the non-gps content was being filtered out of vehicles that support it, so fixed that too.~~
Moved to #7648

### DIFF
--- a/common/source/docs/common-use-cases-and-applications.rst
+++ b/common/source/docs/common-use-cases-and-applications.rst
@@ -16,11 +16,49 @@ Use Cases
 
 .. image:: ../../../images/case-sub.jpg
 
+- Observation and exploration
+
+.. youtube:: T_FMx7C5P_8
+   :width: 100%
+
+- Wreck discovery and documenting
+
+.. youtube:: BV91zgzEFHs
+   :width: 100%
+
+- Photography and videography
+
+.. youtube:: 8IOn9yDXBLM
+   :width: 100%
+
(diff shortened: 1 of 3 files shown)


Hwurzburg:
traveling...will review before next week...lets squash this to a single commit...thanks

Hwurzburg (changes requested):
the use cases look mashed up....I dont think they are segregated correctly/completely...lets split this into two PRs...one sub specific changes....and one that reorgs the use case pages

ES-Alexander:
> lets split this into two PRs...one sub specific changes....and one that reorgs the use case pages

Done - the reorg is now in #7648

Hwurzburg (approved):

PR #7627, opened 2026-04-08
scripts/initvagrant.sh: Drop Ubuntu 18.04 and Python 3.6

cclauss:
Ubuntu 18.04 LTS Standard Support EOL was in May 2023. -- https://ubuntu.com/about/release-cycle 

Python 3.6 EOL was 2021-12-23. -- https://devguide.python.org/versions


peterbarker (approved):
LGTM

I might have left in the skeleton of the check for future use.  But all good.

PR #7629, opened 2026-04-08
parameters: version-selector: select the current file

### LINKED ISSUE #7609: Bug in version dropdown on Parameter List page(s)
The dropdown box always has the top item selected, regardless of which actual version's page is being viewed.
(For me, now, that's V4.6.3.)
This makes it impossible to actually select that page (since clicking the 'already selected' one does nothing).

To workaround, I can edit the URL directly to get to the desired page. So please consider this a "low priority" issue.

Example screenshot: Notice that V4.5.7 is the actual page, but dropdown shows V4.6.3 as selected.
Image

Browser: Safari.

ES-Alexander:
Defaults the param version selector to match the current page (pulled from the URL). Without a default specified, a HTML `<select>` element defaults the first option in the list (which then makes that un-selectable, in this case).

I've tested this in a browser console (on the live wiki site), and it works as far as I can tell. I did confirm that the pathname doesn't include fragment IDs (`#...`), so it's not affected by the page having optionally been reached with a link to a specific parameter.

As I understand it, [the cache gets invalidated for changed files](https://github.com/ArduPilot/ardupilot_wiki/blob/master/update.py#L696), so this should fix all the versioned parameter files at the next site build after it's merged.

Fixes #7609.

### DIFF
--- a/antennatracker/source/_static/parameters_versioning_script.inc
+++ b/antennatracker/source/_static/parameters_versioning_script.inc
@@ -20,6 +20,9 @@ document.addEventListener("DOMContentLoaded", function() {
       var opt = document.createElement('option');
       opt.value = json[key];
       opt.innerHTML = key;
+      if (json[key] === window.location.pathname.split('/').pop()) {
+        opt.selected = true;
+      }
       selectPicker.appendChild(opt);
     }
   }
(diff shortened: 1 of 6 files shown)


ES-Alexander:
Noting this won’t currently set a default for the bare `parameters.html` file. If we want to do that I’m unsure whether it’s best to make a new “master” item, or just treat it as the latest beta or something.

khancyr (approved):
We could have use   var currentPage = window.location.pathname.split('/').pop();
 to check it once, but the page is already slow as hell, so that won't matter much
Thanks for this

ES-Alexander:
> We could have use `var currentPage = window.location.pathname.split('/').pop();` to check it once

Yeah, in retrospect I really should have thought of that.

Thanks for the review and merge :-)

PR #7636, opened 2026-04-10
Simplify command-line testing of get_discourse_posts.py

cclauss:
Add PEP 723 Inline script metadata to get_discourse_posts.py.  This simplifies command-line testing via:
% `pipx run frontend/scripts/get_discourse_posts.py ` # or
% `uv run frontend/scripts/get_discourse_posts.py`

https://peps.python.org/pep-0723

### How was this tested?
Run both commands above to verify correct execution.  Also,

% `uv run frontend/scripts/get_discourse_posts.py --help`
```
usage: get_discourse_posts.py [-h] [--n_posts N_POSTS] [--verbose]

python3 get_discourse_posts.py

options:
  -h, --help         show this help message and exit
  --n_posts N_POSTS  Number of posts to retrieve
  --verbose          show debugging output
```

### DIFF
--- a/frontend/scripts/get_discourse_posts.py
+++ b/frontend/scripts/get_discourse_posts.py
@@ -2,6 +2,14 @@
 """
 Script to get last blog entries on Discourse (https://discuss.ardupilot.org/)
 """
+# PEP 723 Inline script metadata enables testing via:
+# pipx run frontend/scripts/get_discourse_posts.py  # or
+# uv run frontend/scripts/get_discourse_posts.py
+# /// script
+# dependencies = [
+#   "requests",
+# ]
+# ///
 import argparse
 import hashlib
 import json


frontend/scripts/get_discourse_posts.py
```diff
@@ -2,6 +2,14 @@
 """
 Script to get last blog entries on Discourse (https://discuss.ardupilot.org/)
 """
+# PEP 723 Inline script metadata enables testing via:
+# pipx run frontend/scripts/get_discourse_posts.py  # or
+# uv run frontend/scripts/get_discourse_posts.py
+# /// script
+# dependencies = [
+#   "requests",
```
peterbarker, line 10: Why only "requests" here?

I'm guessing everything else ships with Python and requests typically does not?

How does one know what to put in `dependencies` here - trial and error?!

cclauss: Isort puts a blank line between Python Standard Library imports and third-party imports (like `requests`).
So we need to only focus on the imports below that line.  We only need `requests` because https://requests.readthedocs.io states that it is built on top of `urllib3`.
> Keep-alive and HTTP connection pooling are 100% automatic, thanks to [urllib3](https://github.com/urllib3/urllib3).

peterbarker: I think you missed my point.

Why `requests` and not `hashlib`?

How does someone adding another `import` here know not to break the pep723 block here?

cclauss: The PEP723 header automates away the need for a separate `python3 -m pip install ...` command.

`hashlib` is in https://docs.python.org/3/library, so it never needs to be pip-installed.
`requests` is in https://pypi.org/project/requests, so it must be pip-installed.

This difference is also why isort puts them above and below that blank line.

> How does someone adding another import here know not to break the pep723 block here?

https://packaging.python.org/en/latest/specifications/inline-script-metadata/#specification
> Any Python script may have ___top-level___ comment blocks that MUST...

It is kinda like `from __future__ import xyz` MUST come before any other import.
https://docs.python.org/3/reference/simple_stmts.html#future
> A `future statement` must appear near the top of the module.

khancyr:
We got a requirements file, why bother with inline comments ? It just looks bloatware for me

cclauss:
% `uv run --with-requirements=requirements.txt frontend/scripts/get_discourse_posts.py --help`
Is more to remember, more to type, and it loads up a bunch of unnecessary Sphinx stuff.

khancyr:
https://github.com/ArduPilot/ardupilot_wiki/blob/master/frontend/scripts/requirements.txt ... it doesn't need any sphinx stuff.
Instead of those inline things that won't be maintained, I rather prefer to have a better pyproject.toml that define things properly

cclauss:
Closing...  I like your approach of putting requirements in `pyproject.toml` with [dependency groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups) much better than multiple `requirements.txt` files in different directories.

frontend/scripts/get_discourse_posts.py
peterbarker: > The PEP723 header automates away the need for a separate `python3 -m pip install ...` command.
> 
> `hashlib` is in https://docs.python.org/3/library, so it never needs to be pip-installed. `requests` is in https://pypi.org/project/requests, so it must be pip-installed.
> 
> This difference is also why isort puts them above and below that blank line.

Where is `isort` getting that information?

> > How does someone adding another import here know not to break the pep723 block here?
> 
> https://packaging.python.org/en/latest/specifications/inline-script-metadata/#specification

Interesting, but doesn't answer my question.

> > A `future statement` must appear near the top of the module.

Great.  Simple rule, easy to follow.

Where's the same rule for these blocks?

I'd be quite happy with a rule of "always add it in here if present.  Or "just let it break, someone who uses pipx or uv will fix it if they care".  But "go and do a few minutes of research" is an absolute non-starter.  I'm OK to have a little bit of overhead here, but not a lot.

Also... why can't `uv` or `pipx` work this out for themselves?!  i.e. run the things in "download the internet" mode where anything missing gets magically installed into that `uv` env....

cclauss: Isort's sense of Python Standard Library modules can be found in:
https://github.com/PyCQA/isort/tree/main/isort/stdlibs

This PR is closed because we can do without it.

PR #7638, opened 2026-04-11
Run shellcheck in GitHub Actions CI

cclauss:
### https://github.com/koalaman/shellcheck
`shellcheck` is [preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md#installed-apt-packages) on GitHub Actions runners.

Closes: #7620
* #7620

Like:
* ArduPilot/ardupilot#32715

### How was this tested?
Compare the results of:
% `find . -type f -name "*.sh" -exec shellcheck {} +`
with this GitHub Action job to ensure they match when modifying the config file `.shellcheckrc`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,12 @@ jobs:
         python-version: "3.10"
     - uses: pre-commit/action@v3.0.1
 
+  shellcheck:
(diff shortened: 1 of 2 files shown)


peterbarker (approved):
LGTM

Issue #7639, opened 2026-04-11
Plane Failsafe Function docs: BATT_FS params incorrect mapping


gdavud:
Docs at [Plane Failsafe Function - Two-Stage Battery Failsafe](https://ardupilot.org/plane/docs/apms-failsafe-function.html#two-stage-battery-failsafe) say:

"A reasonable setup would be to have [BATT_FS_LOW_ACT](https://ardupilot.org/plane/docs/parameters.html#batt-fs-low-act) = 2 (RTL) and [BATT_FS_CRT_ACT](https://ardupilot.org/plane/docs/parameters.html#batt-fs-crt-act) = 1 (Land)"

Those parameter values do NOT align with the labels (in parenthesis) according to [Complete Parameter List](https://ardupilot.org/plane/docs/parameters.html#complete-parameter-list).

PR #7641, opened 2026-04-11
Isse #7639: Fix BATT_FS param values in plane failsafe docs

gdavud:
Documentation update for `BATT_FS_LOW_ACT` and `BATT_FS_CRT_ACT` values which were incorrect (based on complete params list). Labels (in parenthesis) remain the same but now matching the action values.


Hwurzburg:
already generated PR...thanks

PR #7642, opened 2026-04-12
Python shebangs: #!/usr/bin/env python3

cclauss:
Some were `python` and others were `python3` so let's pick just one.

As discussed at https://github.com/ArduPilot/ardupilot_wiki/pull/7551#issuecomment-4213536493

% `git grep "#\!" **/*.py`
```
build_parameters.py:#!/usr/bin/env python  # <--
frontend/scripts/get_discourse_posts.py:#!/usr/bin/env python3
frontend/scripts/serve.py:#!/usr/bin/env python3
scripts/build_motor_diagrams.py:#!/usr/bin/env python  # <--
scripts/cap_params.py:#!/usr/bin/env python  # <--
scripts/generate_motor_json.py:#!/usr/bin/env python  # <--
scripts/rename_params.py:#!/usr/bin/env python3
update.py:#!/usr/bin/env python3
```
`python-is-python3` is [preinstalled](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md#installed-apt-packages) on GitHub Actions runners, so this does not matter in workflow jobs.

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
     This script aims to provide multiple parameters source files for each vehicle based on the versions available at
     https://firmware.ardupilot.org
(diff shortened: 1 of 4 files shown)


khancyr (approved):
Yep, that is the right way.
Thanks

PR #7651, opened 2026-04-14
Revise RTL altitude parameters and descriptions

### LINKED ISSUE #32741: Potential mismatch in RTL_ALT handling for zero and very large values
# Bug Report
### Issue details

The RTL documentation states that RTL_ALT=0 means return at the current altitude. In SITL, with RTL_CLIMB_MIN=0, I triggered RTL at about 0.2 m above home. When RTL_ALT=0, the parameter was accepted, but the vehicle did not stay at the current altitude during RTL. Is this functionality not implemented?

I also found that RTL_ALT=500000 (5000 m) was accepted. This suggests there may be a code-side problem in how RTL_ALT special values and extreme values are handled.

Furthermore, the documented parameter range for RTL_ALT also seems likely to be incorrect or outdated.
https://ardupilot.org/copter/docs/rtl-mode.html

**Version**
Copter4.7.0 & 4.6.3

**Platform**
[  ] All
[  ] AntennaTracker
[×] Copter
[  ] Plane
[  ] Rover
[  ] Submarine

**Airframe type**
Quadcopter

**Hardware type**
SITL

Non1187:
Fix https://github.com/ArduPilot/ardupilot/issues/32741
Clarifies the RTL_ALT_M=0 wording, and corrects a few nearby range/unit inconsistencies.

### DIFF
--- a/copter/source/docs/rtl-mode.rst
+++ b/copter/source/docs/rtl-mode.rst
@@ -50,16 +50,16 @@ Options (User Adjustable Parameters)
    minimum altitude the copter will move to before returning to launch.
 
    -  Set to zero to return at the current altitude.
-   -  The return altitude can be set from 1 to 8000 centimeters.
-   -  The default return altitude Default is 15 meters (1500)
+   -  The return altitude can be set from 0.3 to 3000 meters.
+   -  The default return altitude Default is 15 meters.
 
 -  :ref:`RTL_ALT_FINAL_M<RTL_ALT_FINAL_M>`: The
    altitude the copter will move to at the final stage of "Returning to
    Launch" or after completing a Mission.
 
    -  Set to zero to automatically land the copter. See :ref:`land-mode`.
-   -  The final return altitude may be adjusted from 0 to 1000
-      centimeters.
+   -  The final return altitude may be adjusted from 0 to 10
+      meters.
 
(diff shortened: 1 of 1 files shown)


copter/source/docs/rtl-mode.rst
```diff
@@ -49,17 +49,17 @@ Options (User Adjustable Parameters)
 -  :ref:`RTL_ALT_M<RTL_ALT_M>`: The
    minimum altitude the copter will move to before returning to launch.
 
-   -  Set to zero to return at the current altitude.
```
rmackay9, line 52: I know it isn't perfectly correct but I'd prefer to leave the line as it is.  I think the new wording is unfortunately harder to understand than the simple line that is there now.

rmackay9:
most of this looks good, thanks very much!

rmackay9 (changes requested):
After the, "Set to zero to return at the current altitude." line is reverted to the original I'm happy for this to be merged, thanks!

Non1187:
> After the, "Set to zero to return at the current altitude." line is reverted to the original I'm happy for this to be merged, thanks!

Thanks very much for the review — I’ve reverted that line to the original wording.

rmackay9 (approved):
LGTM, thanks!

PR #7656, opened 2026-04-15
update.py: preserve latest downloaded file if unchanged

khancyr:
Improve a bit build time by note force replacing latest parameters.rst and logmessage.rst

### DIFF
--- a/update.py
+++ b/update.py
@@ -144,7 +144,20 @@ def remove_if_exists(filepath):
 
 
 def fetch_and_rename(fetchurl: str, target_file: str, new_name: str) -> None:
+    # Fetch into a temporary filename (new_name) and only replace the
+    # real target if content actually changed. This avoids touching
+    # mtimes when the fetched content is identical and prevents
+    # unnecessary Sphinx rebuilds.
     fetch_url(fetchurl, fpath=new_name, verbose=False)
+
+    try:
+        # If target exists and is identical, remove fetched temp and skip replace
+        if os.path.exists(target_file) and filecmp.cmp(new_name, target_file, shallow=False):
+            debug(f"No change for {target_file} (fetched content identical)")
+            os.remove(new_name)
+            return
+    except OSError as e:
+        debug(f"Failed to compare fetched file and target: {e}")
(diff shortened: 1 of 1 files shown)


Hwurzburg:
is --fast still needed? I thought it did this

khancyr:
that another thing. using --fast or --cached-parametes-files will skip dowloading.
If you don't use --fast, it will download and erase the file each time even if that is the same file. This cause sphinx to have to read and parse it again.
This patch allow to prevent this. If the file is the same, we don't have to bother erasing the old one and let sphinx use it cache 

This progress to the goal to have sphinx only update the changed files ! (got a branch working with this) and give us close to instant wiki rebuild

Hwurzburg:
> that another thing. using --fast or --cached-parametes-files will skip dowloading. If you don't use --fast, it will download and erase the file each time even if that is the same file. This cause sphinx to have to read and parse it again. This patch allow to prevent this. If the file is the same, we don't have to bother erasing the old one and let sphinx use it cache
> 
> This progress to the goal to have sphinx only update the changed files ! (got a branch working with this) and give us close to instant wiki rebuild

would --fast then be obsolete when that goal is reached?

khancyr:
yes

update.py
```diff
@@ -144,7 +144,23 @@ def remove_if_exists(filepath):
...
     fetch_url(fetchurl, fpath=new_name, verbose=False)
+
+    try:
+        # If target exists and is identical, remove fetched temp and skip replace
+        if os.path.exists(target_file) and filecmp.cmp(new_name, target_file, shallow=False):
+            debug(f"No change for {target_file} (fetched content identical)")
+            try:
+                os.remove(new_name)
+            except OSError:
+                pass
```
peterbarker: Please remove this `try` block - unless you think there's a legitimate reason we can't remove it!

You already have the outer `try` block to catch it, so it's relatively safe.

peterbarker (changes requested):

update.py
khancyr: ha yes, I missed this !

peterbarker (approved):
LGTM

PR #7658, opened 2026-04-15
update.py: use class for main

khancyr:
use a small class for the updater. 
This will allow to prevent using global variable and permit parallelism on some functions

### DIFF
--- a/update.py
+++ b/update.py
@@ -372,7 +372,7 @@ def copy_build(site, destdir):
         shutil.rmtree(olddir)
 
 
-def make_backup(site, destdir, backupdestdir):
+def make_backup(building_time, site, destdir, backupdestdir):
     """
     backup current site
     """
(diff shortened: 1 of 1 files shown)


peterbarker:
OK, this is fanstastic - thanks!

How did you do this?

How did you test this?

khancyr:
@peterbarker extracted from https://github.com/ArduPilot/ardupilot_wiki/pull/7554
Tested with `docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 update.py --site copter --verbose
`
And make a diff against master branch output

Noted that this one is also covered by CI . 
The only functional change is 

make_backup() that need the building_time now. I just create the class but didn't modify anything, so global variable are still there and all we don't use it fully for now

peterbarker (approved):
LGTM

PR #7659, opened 2026-04-16
build_parameters.py: avoid use of poorly-scoped variables in html par…

peterbarker:
…sing

also remove unnecessary cleanup and comments

we have an instance, use it for storing state rather than more-widely scoped variables

More extracting from https://github.com/ArduPilot/ardupilot_wiki/pull/7597

.... I can't say I'm happy to be extracting that last link in the `finally` block, but it's pre-existing

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -260,28 +260,24 @@ def fetch_releases(firmware_url, vehicles):
         Fetch firmware.ardupilot.org/baseURL all first level folders for a given base URL.
 
         """
-        links = []
-        # Define HTML Parser
-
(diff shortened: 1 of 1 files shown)


khancyr:
Shouldn't forget than in the first instance the finally is useless since we are sys.exit(1) on exception
and html_parser.links = [] does nothing since this class didn't have a links attribut!

khancyr (approved):
It is ok and progressing
Still got important issue that would be resolved in the linked PR
(for example: what appends when the links list is empty ?

PR #7661, opened 2026-04-16
update.py: use logging for logging

khancyr:
Same as for build_parameters.py , use logging lib and add some color.

Error handler allow to still have error resume at the end of the build 
(example with intensional errors)
image

image

tested with : 
docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki python3 update.py --site copter --verbose

### DIFF
--- a/update.py
+++ b/update.py
@@ -39,6 +39,7 @@ import glob
 import gzip
 import hashlib
 import json
+import logging
 import multiprocessing
 import os
 import platform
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -58,6 +59,53 @@
...
+        logging.WARNING: '\033[33m',    # yellow
+        logging.ERROR: '\033[31m',      # red
+        logging.CRITICAL: '\033[1;31m', # bold red
+    }
+    RESET = '\033[0m'
+
+    def format(self, record):
+        # Apply colour only when output is a tty
+        if hasattr(sys.stdout, 'isatty') and sys.stdout.isatty() \
+                and not os.environ.get('CI') and not os.environ.get('GITHUB_ACTIONS'):
```
cclauss, line 82: GitHub Actions defines both, so it is sufficient to only test for "CI".
https://docs.github.com/en/actions/reference/workflows-and-actions/variables

update.py
```diff
@@ -58,6 +59,53 @@
...
+
+# The ErrorStoreHandler stores the messages
+error_store_handler = ErrorStoreHandler()
+error_store_handler.setLevel(logging.ERROR)
+logger.addHandler(error_store_handler)
+
+# Keep noisy third-party network logs quiet by default
+logging.getLogger('urllib3').setLevel(logging.WARNING)
+
 if sys.version_info < (3, 8):
```
cclauss: Please keep this nearthe top of the file.  There is no need to set up logging if we are going to sys.exit().

update.py
```diff
@@ -58,6 +59,53 @@
...
+logging.basicConfig(level=logging.INFO, handlers=[stream_handler])
+logger = logging.getLogger(__name__)
+
+# The ErrorStoreHandler stores the messages
+error_store_handler = ErrorStoreHandler()
+error_store_handler.setLevel(logging.ERROR)
+logger.addHandler(error_store_handler)
+
+# Keep noisy third-party network logs quiet by default
+logging.getLogger('urllib3').setLevel(logging.WARNING)
```
cclauss, line 111: Can we put all of this in a function that returns the logger so we can reuse that function in other files?

cclauss (changes requested):
Super cool.

update.py
khancyr: nice one, corrected!

update.py
khancyr: I will keep it like this as it match what is in the other file. We can consolidate after

update.py
khancyr: nice idea, but I won't bring more complexity in this PR to not delay it. we can consolidate after (still lot to do on wiki)

peterbarker (approved):
LGTM

PR #7666, opened 2026-04-17
build_parameters.py: tidy parsing of list html

peterbarker:
the existing code is parsing the stringification of a structure, which is just bad form.

Extract the href directly from the HTML attributes and use it for further parsing

More stuff extracted from https://github.com/ArduPilot/ardupilot_wiki/pull/7597/changes

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -268,7 +268,8 @@ def fetch_releases(firmware_url, vehicles):
             def handle_starttag(self, tag, attrs):
                 if tag == 'a':
                     attr = dict(attrs)
-                    self.links.append(attr)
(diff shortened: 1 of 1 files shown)


khancyr:
@peterbarker that is ok. Following PR will do the error management as here we don't catch the errors (as before)

khancyr (approved):
Tested and working

peterbarker:
> @peterbarker that is ok. Following PR will do the error management as here we don't catch the errors (as before)

Yeah, I have reservations about catching and throwing away those errors - I wanted to study them a lot more!

peterbarker:
> Tested and working

My testing also went well :-)

PR #7667, opened 2026-04-17
Dev: clarify commit length and allow multiple commits per subsystem

Ryanf55:
* Github breaks display after 72 lines. see https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

### DIFF
--- a/dev/source/docs/submitting-patches-back-to-master.rst
+++ b/dev/source/docs/submitting-patches-back-to-master.rst
@@ -29,7 +29,8 @@ Preparing commits
   means that intermediate commits break the build. This also implies that
   moving code from one library to another happens in two commits: one that
   "deletes" it from the old spot, and a separate one that "creates" it at
-  the new spot.
+  the new spot. It's better for unrelated changes in a single library
+  to be separate commits in the pull request.
 
 - Do not create 'only fix formatting' commits. The disruption they cause
   to git history is more significant than their improvement to the code.
@@ -66,6 +67,8 @@ Preparing commits
         APM_Control: reduce the number of parameter saves in autotune
 
         do not save a parameter unless it has changed by 0.1%
+
+- The brief description (summary) line should be less than 72 characters so it renders correctly in Github.
 
 - **clean up your local commit history** :ref:`using interactive rebase <git-interactive-rebase>`
   (i.e. ``git rebase -i "HEAD~10"``) to re-arrange patches and fold things together. The idea is to present


dev/source/docs/submitting-patches-back-to-master.rst
```diff
@@ -137,7 +141,7 @@ You can track the state of your PR from the `Pull Requests list <https://github.
 
 PRs are more likely to be merged quickly if:
 
-- The PR clearly states what changes in behaviour are expected
+- The PR clearly states what changes in behavior are expected
```
robertlong13: I don't think we should fight back and forth on the wiki between American and Commonwealth spelling. I personally use American spelling when I write stuff, but I'd leave either alone if existing.

peterbarker: Yeah.  The rule is "if it's valid British English, American English or Australian English or Australian slang then leave it alone".

I may be stretching things a *little*.

dev/source/docs/submitting-patches-back-to-master.rst
```diff
@@ -47,6 +48,7 @@ Preparing commits
   organization. Defer to a reviewer's judgement to resolve any
   disagreements regarding this.
 
+
```
hunt0r: optional: Was this newline intentional?

dev/source/docs/submitting-patches-back-to-master.rst
```diff
@@ -29,7 +29,8 @@ Preparing commits
   means that intermediate commits break the build. This also implies that
   moving code from one library to another happens in two commits: one that
   "deletes" it from the old spot, and a separate one that "creates" it at
-  the new spot.
+  the new spot. It is ok to have unrelated changes in a single library
+  as separate commits in a single pull request.
```
hunt0r: Thank you for this!

optional: I would even favor strengthening the language, if others would.
```suggestion
  the new spot. It's better for unrelated changes in a single library
  to be separate commits in the pull request.
```
But also, the word "unrelated" invites criticism: Why is a PR doing two "unrelated" things at all? I wonder if we could think of a word like "easily separable" or "independent" or "orthogonal" or something to avoid that.


(By "optional", I mean this is totally fine as-is! Ok to resolve this comment without response/changes.)

hunt0r (approved):

dev/source/docs/submitting-patches-back-to-master.rst
Hwurzburg: I am in favor of this rigorous approach!

dev/source/docs/submitting-patches-back-to-master.rst
Ryanf55: Nope, will remove that unintentional change. tx.

dev/source/docs/submitting-patches-back-to-master.rst
Ryanf55: Dropped that, oops. Night brain forgot that spellcheck is not right all the time.

dev/source/docs/submitting-patches-back-to-master.rst
hunt0r: optional: On my side, the newline still seems to be here? Can't tell if that's a git bug, or an oversight.

hunt0r (approved):

dev/source/docs/submitting-patches-back-to-master.rst
Ryanf55: It just needed a refresh I guess, looks fine now.

Hwurzburg (approved):

tpwrules (approved):
Great, thank you! I am a big proponent of fine-grained changes.

PR #7669, opened 2026-04-17
update.py: raise file chunk size for better download performance

khancyr:
forward declare total_size to prevent checker to complain

tested with docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki22 python3 update.py --site blimp

### DIFF
--- a/update.py
+++ b/update.py
@@ -166,6 +166,8 @@ def fetch_url(fetchurl: str, fpath: Optional[str] = None, verbose: bool = True) 
     """Fetches content at url and puts it in a file corresponding to the filename in the URL"""
     progress(f"Fetching {fetchurl}")
 
+    total_size = 0
+
     if verbose:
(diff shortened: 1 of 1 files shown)


peterbarker:
> forward declare total_size to prevent checker to complain
> 
> tested with docker run --rm -it -v "$(pwd):/ardupilot_wiki" -u "$(id -u):$(id -g)" ardupilot_wiki22 python3 update.py --site blimp

.... and what were the results?

How much faster were things after this?

peterbarker (approved):
LGTM

khancyr:
@peterbarker around 1-2s gain . not that much, but with today connexion speed we can allow it, mostly only usefull for the parameters and logmessage rst as those are the biggest (> 1mb)

PR #7670, opened 2026-04-17
Update parallele

khancyr:
download the files in parallele instead of waiting (works best with session)

image


peterbarker (approved):
LGTM

PR #7671, opened 2026-04-17
build_parameters.py: use http session

khancyr:
use http session to limit connexion on server and make faster requests
for just copter we down from 77s to 30s on the version fetching !
image


khancyr:
master : 
image

peterbarker (approved):
LGTM

PR #7672, opened 2026-04-17
Copter: clarify FS_OPTIONS bit 5 in radio, GCS, and battery failsafe docs

jinmussj:
The current failsafe pages list FS_OPTIONS bits 0..4 only, but the parameter definitions already include bit 5 ("Release Gripper").


Hwurzburg (commented):
LGTM...thats an oldie...thanks

Hwurzburg (approved):

PR #7681, opened 2026-04-21
build_parameters.py: use local cache and server cache to get files

khancyr:
add local cache as parameters files don't change often
use HEAD request before getting the file to know if it changed or not.

Cache work in 3 steps:
- Uses cached content if the file is still fresh (ie < 6h)
- ask the server about the file Last-Modified, if it match cache, use cache
- if cache cannot be used, get the file from the server.

6h time is totally arbitrary. As we don't update often parameter that look sane to test on 4 time per day and reduce lot the request on the server !
We could work on clever way later for old parameters and perhaps reporting if a modification is noted on old files

image

From testing our server don't return http 304

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -35,8 +35,10 @@ import shutil  # noqa: F401
 import subprocess
 import sys
 import time  # noqa: F401
+import urllib.parse
 from concurrent.futures import ThreadPoolExecutor
 from html.parser import HTMLParser
+from pathlib import Path
 
 import requests
 from requests.adapters import HTTPAdapter
@@ -48,6 +50,15 @@ parser.add_argument("--verbose", dest='verbose', action='store_false', default=T
 parser.add_argument("--ardupilotRepoFolder", dest='gitFolder', default="../ardupilot", help="Ardupilot git folder. ")
 parser.add_argument("--destination", dest='destFolder', default="../../../../new_params_mversion", help="Parameters*.rst destination folder.")  # noqa: E501
 parser.add_argument('--vehicle', dest='single_vehicle', help="If you just want to copy to one vehicle, you can do this. Otherwise it will work for all vehicles (Copter, Plane, Rover, AntennaTracker, Sub, Blimp)")  # noqa: E501
+
+DEFAULT_CACHE_TIME = 6 * 3600
+
+# Get the directory where this script is located
+script_dir = os.path.dirname(os.path.abspath(__file__))
+default_http_request_cache_dir = os.path.join(script_dir, '.cache')
+
+parser.add_argument("--cache-dir", dest='cache_dir', default=default_http_request_cache_dir,
+                    help="Directory to cache HTTP responses")
 args = parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


build_parameters.py
```diff
@@ -43,6 +43,14 @@
 parser.add_argument("--ardupilotRepoFolder", dest='gitFolder', default="../ardupilot", help="Ardupilot git folder. ")
 parser.add_argument("--destination", dest='destFolder', default="../../../../new_params_mversion", help="Parameters*.rst destination folder.")  # noqa: E501
 parser.add_argument('--vehicle', dest='single_vehicle', help="If you just want to copy to one vehicle, you can do this. Otherwise it will work for all vehicles (Copter, Plane, Rover, AntennaTracker, Sub, Blimp)")  # noqa: E501
+
+DEFAULT_CACHE_TIME = 6 * 3600
+
+# Get the directory where this script is located
+script_dir = os.path.dirname(os.path.abspath(__file__))
+default_cache_dir = os.path.join(script_dir, '.cache')
```
peterbarker: Maybe a bit generic a name

build_parameters.py
```diff
@@ -82,6 +90,95 @@ def format(self, record):
...
+
+def get_cached_url(url, cache_dir=None):
+    """Get URL content with caching to avoid repeated downloads"""
+    if cache_dir is None:
+        cache_dir = args.cache_dir
+
+    os.makedirs(cache_dir, exist_ok=True)
+
+    # Create cache filename from URL
+    cache_filename = re.sub(r'[^\w\-_.]', '_', url) + '.cache'
```
peterbarker: This is the wrong sort of regex to use here.  You should be choosing which characters to keep, not which ones to throw away.

`urllib` has a `quote` method available.

build_parameters.py
```diff
@@ -284,9 +381,7 @@ def handle_starttag(self, tag, attrs):
         try:
             debug(f"Fetching {firmware_url}{vehicle}")
 
-            response = session.get(firmware_url + vehicle, timeout=30)
-            response.raise_for_status()
-            content = response.text
+            content = get_cached_url(firmware_url + vehicle)
```
peterbarker: That's not what this really does.

peterbarker (changes requested):
It would be nice if this was all tied up in a little object rather than being a bunch of methods, but I'm not going to insist.

build_parameters.py
khancyr: default_http_request_cache_dir ok ?

build_parameters.py
khancyr: I didn't know about urllib.quote, using it instead of the regex

build_parameters.py
khancyr: yup, corrected

khancyr:
@peterbarker comments addressed !

khancyr:
@peterbarker , when cache is done in build_parameters.py and update.py, I will consolidate both script with a common lib (logging, http and git things) that will make things easier at a central place

build_parameters.py
```diff
@@ -48,6 +49,15 @@
...
 parser.add_argument('--vehicle', dest='single_vehicle', help="If you just want to copy to one vehicle, you can do this. Otherwise it will work for all vehicles (Copter, Plane, Rover, AntennaTracker, Sub, Blimp)")  # noqa: E501
+
+DEFAULT_CACHE_TIME = 6 * 3600
+
+# Get the directory where this script is located
+script_dir = os.path.dirname(os.path.abspath(__file__))
+default_http_request_cache_dir = os.path.join(script_dir, '.cache')
+
+parser.add_argument("--cache-dir", dest='cache_dir', default=default_http_request_cache_dir,
+                    help="Directory to cache HTTP responses")
```
peterbarker, line 61: I think we really should be better at making this one-keyword-per-line, it's better all round.

Since I'm recently guilty of using the same format I definitely can't ask for it to be changed here :-)

build_parameters.py
```diff
@@ -103,6 +113,102 @@ def format(self, record):
...
 
+def fetch_url_with_cache(url, cache_dir=None):
+    """Fetch URL content with caching to avoid repeated downloads."""
+    if cache_dir is None:
+        cache_dir = args.cache_dir
+
+    os.makedirs(cache_dir, exist_ok=True)
+
+    # Create cache filename from URL
+    cache_filename = quote(url, safe='') + '.cache'
```
```suggestion
    cache_filename = urllib.parse.quote(url, safe='') + '.cache'
```

`quote` really just too generic a name to leave lying around as an import

build_parameters.py
```diff
@@ -103,6 +113,102 @@ def format(self, record):
...
+    cache_path = os.path.join(cache_dir, cache_filename)
+    cache_meta_path = cache_path + '.meta'
+
+    def load_cached_content():
+        with open(cache_path, 'r', encoding='utf-8') as f:
+            return f.read()
+
```
peterbarker, line 131: I'd prefer a one-liner using pathlib

build_parameters.py
```diff
@@ -103,6 +113,102 @@ def format(self, record):
...
+    def load_cache_metadata():
+        if not os.path.exists(cache_meta_path):
+            return {}
+        try:
+            with open(cache_meta_path, 'r', encoding='utf-8') as f:
+                return json.load(f)
+        except OSError:
```
peterbarker: You're ignoring all sorts of things here.

When do you think it would be OK for the file to exist but us not to be able to read it?

I'd actually rather not catch any exception here.

build_parameters.py
```diff
@@ -103,6 +113,102 @@ def format(self, record):
...
+            return {}
+
+    def save_cache(content, response):
+        with open(cache_path, 'w', encoding='utf-8') as f:
+            f.write(content)
```
peterbarker: pathlib?

peterbarker (dismissed):
LGTM

build_parameters.py
khancyr: updated

build_parameters.py
khancyr: right, removed.

build_parameters.py
khancyr: updated all open occurent to use pathlib instead

khancyr:
Tested locally, and parameters match to master generated one's
last push was fixing ruff check with an import that wasn't ordered correctly

peterbarker (approved):

PR #7682, opened 2026-04-21
Clarify RTL overview wording for pre-arm GPS and position checks

jinmussj:
The previous text could be read as an unconditional guarantee that default pre-arm checks always ensure a 3D GPS lock with sufficient HDOP and a working compass before arming. In practice, the exact arming and pre-arm gating depends on the selected mode and configuration. 

For example, arming in Stabilize can still be allowed under GPS/HDOP conditions that would not be accepted for a position-dependent mode.


copter/source/docs/rtl-mode.rst
```diff
@@ -23,7 +23,7 @@ The altitude reference frame is set by either the rally point, if proceeding to
...
 
-RTL mode requires a reliable position estimate to work properly, most commonly provided by GPS and compass. Default prearm checks will ensure a 3D GPS lock with sufficient HDOP is acquired and your mag is working as expected prior to arming. When using non-default arming checks, make sure you do have a sufficient GPS lock and / or a reliable position estimate for RTL to perform as expected.
+RTL mode requires a reliable position estimate to work properly, most commonly provided by GPS and compass. Default prearm checks will ensure a 3D GPS lock with sufficient HDOP is acquired and your mag is working as expected prior to arming, when required by the selected mode and configuration. When using non-default arming checks, make sure you do have a sufficient GPS lock and / or a reliable position estimate for RTL to perform as expected.
```
Hwurzburg: lets expand this a bit while we are at it

```suggestion
RTL mode requires a reliable position estimate to work properly, most commonly provided by GPS and compass. Default prearm checks will ensure a 3D GPS lock with sufficient HDOP is acquired and your mag is working as expected prior to arming, if required by the selected mode and configuration during arming (ie STABILIZE could be armed without a reliable position and a switch into RTL would be refused without it). When using non-default arming checks, make sure you do have a sufficient GPS lock and / or a reliable position estimate for RTL to perform as expected.
```

Hwurzburg (changes requested):

Hwurzburg (approved):

PR #7685, opened 2026-04-21
Build parameters filter

khancyr:
We don't necessary have firmware for all vehicles < 3.9 so no need to serve parameters for those and they weren't build do to parsing issue anyway

Fix parameters generation by live fixing the parse_parm.py that we checkout on version revision that still have python2 support.

image


peterbarker (approved):
LGTM

PR #7701, opened 2026-04-24
Common: joystick mode channel warning

### LINKED ISSUE #32862: RC_CHANNEL_OVERRIDE fallback to normal RC may cause invalid channel input to be consumed
# Bug Report

We believe we have a safety vulnerability in setups where both RC_CHANNEL_OVERRIDES (aka [joystick](https://ardupilot.org/copter/docs/common-joystick.html)) are used along with a real RC but where the RC_CHANNEL_OVERRIDES are only applied to some channels (not all).

The issue occurs when the RC_CHANNEL_OVERRIDES values timeout (e.g. loss of contact with the GCS), the fallback to the real RC may lead to the RC values even thought they are stale (e.g. the RC connection was lost and timed out)

The issue can be re-created by doing the following:

1. Connect MAVLink to an autopilot which has a real RC and is running Copter (any version)
2. message graph RC_CHANNELS.chan1_raw RC_CHANNELS.chan2_raw RC_CHANNELS.chan3_raw RC_CHANNELS.chan4_raw RC_CHANNELS.chan5_raw
3. set each flight mode switch position to different modes
    - param set FLTMODE1 1 (Acro)
    - param set FLTMODE2 2 (AltHold)
    - parma set FLTMODE3 3 (Auto)
    - param set FLTMODE4 4 (Guided)
    - param set FLTMODE5 5 (Loiter)
    - param set FLTMODE6 6 (RTL)
4. turn on the RC receiver and move the flight mode channel to the lowest position (should switch to Acro)
5. turn off the RC receiver (the flight mode should not change)
6. message RC_CHANNELS_OVERRIDE 0 0 65535 65535 65535 65535 **1900** 65535 65535 65535 (vehicle should switch to RTL)
7. message RC_CHANNELS_OVERRIDE 0 0 65535 65535 65535 65535 **0** 65535 65535 65535 (vehicle will switch to Acro)

This last step is the issue.  The GCS has handed back control of the flight mode switch and it has resulted in the mode change even though the real RC input is stale.  The correct behaviour would be for the mode to remain unchanged.

[Explanation of the RC_CHANNEL_OVERRIDE feature can be found here on the dev wiki](https://ardupilot.org/dev/docs/mavlink-rcinput.html#rc-channels-override).
 
If the RC was never turned on (e.g. skip step 4 above) the vehicle's mode does not change upon loss of joystick.

This issue was first reported in https://github.com/ArduPilot/ardupilot/issues/32653.  This was discussed by @rmackay9 and @peterbarker on the 21-Apr-2026 dev call

**Version**
What version was the issue encountered with

**Platform**
[x] All
[ ] AntennaTracker
[ ] Copter
[ ] Plane
[ ] Rover
[ ] Submarine

**Hardware type**

Real RC and RC_CHANNEL_OVERRIDES (aka joystick)

rmackay9:
This makes two sets of changes to the joystick page:

1. minor updates and corrections to wording
2. strengthens the advice to NOT use the joystick to control the flight mode channel nor auxiliary function channels

This should help users avoid this issue https://github.com/ArduPilot/ardupilot/issues/32862 which involves stale input being consumed from the regular RC after a fallback from the joystick to the regular RC

fixes ArduPilot/ardupilot/issues/32862

### DIFF
--- a/common/source/docs/common-joystick.rst
+++ b/common/source/docs/common-joystick.rst
@@ -63,10 +63,9 @@ Setup with the Mission Planner
 
 -  Repeat for *Pitch*, *Throttle* and *Rudder* and push the **Save** button
 -  You will likely want to set-up other buttons for *Arm* and *Disarm*
--  Instead of setting up a stick or button for ch5 (the flight mode
-   channel), it is better to set individual buttons to initiate a flight
-   mode by selecting **Change Mode** and then push the **Settings**
-   button and select the flight mode from the drop-down (see pic above)
+-  The flight mode channel (e.g. RC5 or RC8) and :ref:`auxiliary function channels <common-auxiliary-functions>`
+   should not be controlled from the joystick especially if a regular RC receiver is also connected.  Instead it is better to configure buttons to set the flight
+   mode by selecting **Change Mode** and then push the **Settings** button and select the flight mode from the drop-down (see pic above).  Details on why this is important are documented in `issue #32862 <https://github.com/ArduPilot/ardupilot/issues/32862>`__.
 -  When done, push the **Save** button. This records the min and max calibrations and records the neutral positions of the joystick for the trim value. Mission Planner will then scale the joystick signal to the RC channel's ``RCx_MIN/MAX/TRIM`` parameters in the autopilot when sending RC override commands by the Joystick.
 
 Autopilot Setup
--- a/common/source/docs/common-joystick.rst
+++ b/common/source/docs/common-joystick.rst
@@ -4,10 +4,9 @@
 Joystick/Gamepad
 ================
 
-This article explains how you can control your vehicle with a Joystick or Gamepad using the Mission Planner which sends "RC Override" messages to the vehicle.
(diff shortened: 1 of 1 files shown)


common/source/docs/common-joystick.rst
```diff
@@ -41,14 +39,14 @@ First you must calibrate the joystick using the Windows joystick/game controller
 Setup with the Mission Planner
 ==============================
 
--  Connect your USB joystick/gamepad to the laptop computer
+-  Connect your USB joystick/gamepad to the PC
```
Hwurzburg, line 42: I think you also need to get the PC to recognize the joystick via its joystick setup panel, dont you?

rmackay9: I've never had to do anything special.. the joystick has always just appeared in the MP drop-down.

Hwurzburg: your are right I was remembering the serial joystick days....just too old....HID class USB joysticks dont need setup

Hwurzburg (approved):

rmackay9:
Great, thanks very much!

PR #7706, opened 2026-04-24
versionned parameters cleanup

khancyr:
If you build versionned parameters, they stay. building without should remove the copied .rst as we dont keep the cache so they end up builded again, and as we update main parameters.rst, they aren't even served.

parameters-*.json are useless on normal build and are generated  each time we build with versionned parameters, so no need to track

os.getcwd() usage was removed as not needed and the target_json_file wasn't using it anyway.

Capture d’écran du 2026-04-24 15-51-33
tested working with and without the versionning 

Capture d’écran du 2026-04-24 17-58-30


peterbarker (approved):
LGTM

PR #7708, opened 2026-04-24
Update.js

khancyr:
Some slight cleanup on the common copy file :
remove useless exception : if filecmp.cmp it would be from a filesystem issue so the following write on will also raise an error, therefore that is useless so we can simplify

unifiy the variables names as they are all the same
For the useralert.js, we don't need clever directive, just copying to the vehicle _static directory is enough
Correct the default vehicle copy list by adding antennatracker and blimp

after git clean
Capture d’écran du 2026-04-27 09-57-28

after run
Capture d’écran du 2026-04-27 09-57-35

Capture d’écran du 2026-04-27 09-57-49

image

### DIFF
--- a/update.py
+++ b/update.py
@@ -585,9 +585,9 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
 
     for root, dirs, files in os.walk(start_dir):
         for file in files:
+            source_file_path = Path(f"{root}/{file}")
             if file.endswith(".rst"):
                 # debug("  FILE: %s" % file)
-                source_file_path = Path(f"{root}/{file}")
                 source_content = source_file_path.read_text(encoding='utf-8')
                 targets = get_copy_targets(source_content)
                 for wiki in targets:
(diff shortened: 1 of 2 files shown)


update.py
```diff
@@ -584,10 +584,9 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
 
     for root, dirs, files in os.walk(start_dir):
         for file in files:
+            file_path = os.path.join(root, file)
             if file.endswith(".rst"):
-                # debug("  FILE: %s" % file)
-                source_file_path = os.path.join(root, file)
-                source_file = open(source_file_path, 'r', encoding='utf-8')
+                source_file = open(file_path, 'r', encoding='utf-8')
```
peterbarker: `pathlib` would be nice here

peterbarker (commented):
Testing evidence?

update.py

khancyr:
@peterbarker updated with pathlib: base on https://github.com/ArduPilot/ardupilot_wiki/pull/7710 to make the review and rebase easier

khancyr:
more testing show that blimp doesn't support useralert yet and antenna tracker is outputing some build issue if common are updated. So we need to revert those

PR #7710, opened 2026-04-24
update.py: correct common rst copy check

khancyr:
filecmp.cmp is useless for most files that have a copy directives, so we need to do bytes to bytes comparison.
That is still cheap to do and faster than having sphinx to reread everything.

image

### DIFF
--- a/update.py
+++ b/update.py
@@ -596,15 +596,12 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
                     targetfile = f'{wiki}/source/docs/{file}'
 
                     # Only write if content has changed (preserves timestamps for unchanged files)
-                    # Use byte-accurate file comparison against the source file.
+                    # Use byte-accurate file comparison against the source file stripped of copywiki shortcodes.
                     if not clean_common and os.path.exists(targetfile):
-                        try:
-                            if filecmp.cmp(source_file_path, targetfile, shallow=False):
+                        with open(targetfile, 'r', encoding='utf-8') as f:
+                            if f.read() == content:
                                 files_skipped += 1
                                 continue
-                        except Exception as e:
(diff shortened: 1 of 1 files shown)


update.py
```diff
@@ -596,15 +596,12 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
                     targetfile = f'{wiki}/source/docs/{file}'
 
                     # Only write if content has changed (preserves timestamps for unchanged files)
-                    # Use byte-accurate file comparison against the source file.
+                    # Use byte-accurate file comparison against the source file stripped of copywiki shortcodes.
                     if not clean_common and os.path.exists(targetfile):
-                        try:
-                            if filecmp.cmp(source_file_path, targetfile, shallow=False):
+                        with open(targetfile, 'r', encoding='utf-8') as f:
+                            if f.read() == content:
```
peterbarker: Please use `pathlib` rather than open+read

peterbarker (changes requested):
I'd really like us to remove the open+read pattern in favour of `pathlib` in our new code.

Primarily to avoid the calls to use contexts to avoid dangling filehandles...

update.py

khancyr:
after $ git clean -f -f -x -d -d 
Capture d’écran du 2026-04-27 09-39-26

after rebuild
Capture d’écran du 2026-04-27 09-39-37

peterbarker (approved):
Thanks for the pathlib change.  Much nicer there.  I'll have to bend my brain to like the "/" operator those things, 'though :-)

LGTM

khancyr:
@peterbarker ha crap, I let one passing ... I hate them too, I much prefer a single string contruction (even if it could be slower, it is easier)

PR #7711, opened 2026-04-24
Images: Update RC calib images

EvansCesilThomas:
Update the images showing the RC page to reflect new MP version.


Hwurzburg:
does not seem to replace anything...just adds an image that is not referenced

EvansCesilThomas:
@Hwurzburg I replaced the original image with the new one. ie: the filename of the image is the same as before

Hwurzburg (approved):

rmackay9:
very nice!

EvansCesilThomas:
@Hwurzburg Seems like this new image is not being used in the website. It still shows the old images. Anything wrong with this PR? Or is it that the system only builds what changed and since this PR doesn't rename the filename, it didn't get built?

Hwurzburg:
yes, even when I build locally it does not update the image...and its correct in the source...muat be due to a changee that @khancyr has made trying to optimize by not rebuilding unchanged files...will make a note in infrastructure channel

EvansCesilThomas:
Okay, tell me if you want me to change the filename then so it gets built

Hwurzburg:
okay the issue is that the image you updated is NOT the image used in the Copter wiki page...it uses an image from copter/source/images directory...this image is not used in the wiki at all

EvansCesilThomas:
Great. I see my previous PR also had this mistake. I'll correct that. But this means we have duplicate images in the outer images folder also

PR #7718, opened 2026-04-28
Update pathlib

khancyr:
simplify the copy_common_source_files syntax with pathlib @peterbarker 
fix the js copy
I chose to not consolidate more and keep js and rst handling separated (their target isn't hte same but we could have merge them)

### DIFF
--- a/update.py
+++ b/update.py
@@ -585,9 +585,9 @@ def copy_common_source_files(start_dir=COMMON_DIR, clean_common=False):
 
     for root, dirs, files in os.walk(start_dir):
         for file in files:
+            source_file_path = Path(root) / file
             if file.endswith(".rst"):
                 # debug("  FILE: %s" % file)
-                source_file_path = Path(root) / file
                 source_content = source_file_path.read_text(encoding='utf-8')
                 targets = get_copy_targets(source_content)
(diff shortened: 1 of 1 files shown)


peterbarker (commented):
I'm not sure the constructor syntax is better, myself.  But other people must.

How was this tested?

khancyr:
I don't like it much either, but what decide me to use it is : 
Avoids double slashes — if root already ends with /, the f-string version could produce dir//file; Path.__truediv__ handles this correctly

PR :
rebuild
Capture d’écran du 2026-04-29 10-44-51

after git clean
Capture d’écran du 2026-04-29 10-45-38

check useralerts are in right place
Capture d’écran du 2026-04-29 10-45-46

check content
Capture d’écran du 2026-04-29 10-46-36


Master: 
after git clean
Capture d’écran du 2026-04-29 10-47-13

after rebuild
Capture d’écran du 2026-04-29 10-47-33

peterbarker (changes requested):
Unrelated patches snuck in!

khancyr:
@peterbarker corrected, I forget to remove it

peterbarker (approved):
Nice ,thanks!

PR #7719, opened 2026-04-28
Precommit conveninent checkers

khancyr:
add some more conveninent checker on pre-commit :
check xml, json and yaml files are valid (parsing check)
check python files are valid (ast check)

Check permission are settled correctly on files : git handling permission flag correctly so we should mark executable file as such and remove permissions on other.

Add a merge conflict tag checker. This one prevent some mistake on rebase, and merge tag are valid rst syntax so we won't necessarly failed the build


Ryanf55 (approved):
I'm assuming these are enforced in CI?

peterbarker (approved):
LGTM.

PR #7723, opened 2026-04-29
update.py: print we got sphinx warning on update end

khancyr:
make sphinx warning appears as error on update.py
This is non blocking and not using sphinx Warning is error system that would be blocking the build.

image

There we just report we had warning during build. That would prevent missing some warning line.

On a second step, I can make CI to catch an reply on the PR status if we got such errors


Hwurzburg (approved):
adding summary line if warning happened is fine with me

peterbarker (approved):
LGTM

PR #7724, opened 2026-04-29
scripts: add new dedupe_params script and use it for rangefinder and …

khancyr:
…periph net parameters

Result for versionned parameters against master :
Capture d’écran du 2026-04-29 17-28-23
(show dedup is still working as expected)

For periph parameters : 
Capture d’écran du 2026-04-29 17-30-57

image


peterbarker (changes requested):
CI must pass

peterbarker (approved):
LGTM

PR #7727, opened 2026-04-30
Checkers rst

khancyr:
move check-ref-directives from update.py to pre-commit (gain 6s on build)
add another check on :ref: directives.
Add unittitest for both rst checks
make both rst check use pre-commit venv to run and not system one's.

tested locally with pre-commit run --all-files (all passed) and CI validate this also

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,3 +59,11 @@ repos:
         entry: python scripts/check_youtube_timestamps.py
         language: python
         types: [rst]
+
+  - repo: local
+    hooks:
+      - id: check-ref-directives
+        name: Check ref directive formatting
+        entry: python scripts/check_ref_directives.py
+        language: python
+        types: [rst]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,8 +49,8 @@ repos:
       - id: rst-literal-block-spacing
         name: Check ReST literal block spacing
         entry: python scripts/check_rst_literal_blocks.py
-        language: system
-        files: '\.rst$'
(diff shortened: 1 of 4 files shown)


Hwurzburg:
does this mean those checks do not occur on local builds? I think that this means one would HAVE to run not only update.py but pre-commit to check local builds thoroughly before pushing as a PR to master, correct? would impact my workflow, but probably mainly only me

khancyr:
pre-commit should be part of the workflow anyway now, as we require it to pass CI. That is why I moved the check from update.py to pre-commit. 
Another advantage is that is speed up a bit update.py so we need only one run with pre-commit to catch all the syntax mistakes

Hwurzburg:
@peterbarker ping

peterbarker (commented):
Does this mean you can make lots of changes and only discover there's a problem when you go to commit, rather than when you run update.py?

khancyr:
That is already the case, we have only one syntax check on update.py and it is not blocking, all other checks are on pre-commit call.
Notice that those check aren't failing sphinx build that why we added them.

If you prefer the other way, we could call pre-commit checks from update.py, that would and make it blocking. That would force the check to run without being too slow

Hwurzburg:
and there is a difference between server builds and local developer builds....AFAIAC, having update.py warnings for errors on local builds (no blocking to the site build locally) is fine....

khancyr:
@Hwurzburg @peterbarker I have rebased this and make the new script always run as part of update.py like it was before. 
I would still want to decouple building and checking the syntax, so we can keep the building fast. But that will be in another PRs.

PR #7730, opened 2026-05-01
dev: add Python style guide

peterbarker:


dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,39 @@
...
+New functions should have type annotations.
+
+If you are modifying a function's signature please consider adding type annotations.
+
+Automated Linting
+=================
+
+Generally ``ruff`` is used to lint the codebase.
+
+It is often triggered via ``pre-commit``
```
khancyr: We are migrating to `pre-commit` to handle the linting and python correctness. You can use  `pre-commit run --all-files` to validate your changes are correct. This check is enforced on CI

dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,39 @@
...
+================
+
+New functions should have type annotations.
+
+If you are modifying a function's signature please consider adding type annotations.
+
+Automated Linting
+=================
+
+Generally ``ruff`` is used to lint the codebase.
```
khancyr, line 47: we should remove flake8 test on ardupilot as for now we got both. Ruff covered most flake8 check so that would be reduncdant

khancyr (dismissed):
Good start. That is simple enough !

dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,39 @@
...
+ArduPilot's Python has not traditionally used ``_`` as a hint that a member is intended for internal use.
+
+They should be omitted until we adopt the convention.
+
+Type Annotations
+================
+
+New functions should have type annotations.
+
+If you are modifying a function's signature please consider adding type annotations.
```
Hwurzburg, line 32: example??

dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,39 @@
...
+============================
+
+In order to improve the readability of the ArduPilot code, and help new
+users pick up the code more quickly, please use the following styles. Code submitted may be rejected if these styles are not followed.
+
+.. note::
+
+   Some parts of the code may not conform due to historic reasons but
+   new additions should!
+
```
Hwurzburg, line 14:
```suggestion
- For Python changes follow the :ref:`style guide <python-style-guide>` so your code fits in with the existing code.
```

Hwurzburg (changes requested):

dev/source/docs/python-style-guide.rst
peterbarker: Errr... this is *in* that very document you are referencing?!

dev/source/docs/python-style-guide.rst
peterbarker: I've added a "Right" but omitted the "Wrong" as I'm not being strict about enforcing the style at the moment.  Slow-start to it, basically.

dev/source/docs/python-style-guide.rst
peterbarker: "most" is not "all".  I do agree that if we can show flake8 is entirely redundant we could remove it.

dev/source/docs/python-style-guide.rst
peterbarker: I've added the second part of this.  The intent of where we are going doesn't need to be conveyed.

dev/source/docs/python-style-guide.rst
Hwurzburg: duh...nm

dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,49 @@
...
+ArduPilot supports Python versions back to 3.9 - scripts should be written with an eye to working back that far.
+
+Method Names
+============
+
+ArduPilot's Python has not traditionally used ``_`` as a hint that a member is intended for internal use.
+
+They should be omitted until we adopt the convention.
+
```
hunt0r, line 26: optional: Is this up for discussion? (Totally fine if "no", please state such and resolve this comment.)

[In case it is up for discussion]
The original source (to my knowledge) is [here in PEP8](https://peps.python.org/pep-0008/#descriptive-naming-styles), which is... 25 years old?
With the convention already being "weak" / "internal", forbidding it seems unnecessary.

Of course I would also never "mandate" folks use it either, both because its a 'weak' convention and also because consistency-with-legacy seems a good reason to not force.

So my proposal is that here we should either explicitly "permit but not require" it, or just not mention it which seems to achieve the same.
```suggestion
```

dev/source/docs/python-style-guide.rst
```diff
@@ -0,0 +1,49 @@
...
+**Right:**
+
+::
+
+    def set_parameter_bit(self, name: str, bit_offset: int) -> None:
+        '''set bit in parameter to true, preserving values of other bits'''
+        value = int(self.get_parameter(name))
+        value |= 1 << bit_offset
+        self.set_parameter(name, value)
```
hunt0r, line 42: optional: Would you like more examples contributed? I will voluntarily author some, either for addition to this PR or as a follow-up (your preference).

[Ideas]
I think having at least one non-None return value in addition to this one which does not `return` (and thus `-> None` is correct).
I also think showing an example of `typing.Optional` and `typing.Union` will be value-added.

I'll also write us a note about a future preference for `A | B` over `typing.Union[A, B]` when we advance our version. (I think 3.10 fixes that, but I'd check first.)

I think a note about `from __future__ import annotations` being preferred over string-literal ones is appropriate.

A harder typehint topic is 'be as general as possible on inputs, and as specific as possible on outputs.' I worry that going there right now will be 'biting off too much' for folks, but I have written that opinion before and will contribute it if you'd like.

hunt0r (approved):

tridge (approved):

Hwurzburg (approved):

PR #7739, opened 2026-05-05
update.py: old urllib3 don't have retry

khancyr:
need urllib3 > 2.x so we cannot have this on hold ubuntu with python 3.8 (tested with Henry)


Hwurzburg (approved):
tested...merging...thanks

PR #7744, opened 2026-05-12
Update Auto mode GPS HDOP and landing-cancel wording

jinmussj:
1. Update the default GPS_HDOP_GOOD value The Auto mode page still refers to an HDOP of under 2.0. The current default GPS_HDOP_GOOD value is 1.4. 
2. Clarify high-throttle landing cancel behavior in Auto In current Copter behavior, high throttle cancels an AUTO landing by switching to Loiter, or AltHold if Loiter is unavailable, without advancing the mission to the next item. This was also verified in SITL with `PILOT_THR_BHV=2`.

### DIFF
--- a/copter/source/docs/auto-mode.rst
+++ b/copter/source/docs/auto-mode.rst
@@ -19,7 +19,7 @@ control from :ref:`Loiter mode <loiter-mode>` and should not
 be attempted before these modes are flying well.  All the same
 requirements apply including ensuring that :ref:`vibration levels <common-measuring-vibration>` and compass
 interference levels are acceptable and that the GPS is functioning well
-including returning an HDOP of under 2.0.
+including returning an HDOP of under 1.4.
 
 Overview
 ========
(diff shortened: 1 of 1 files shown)


copter/source/docs/auto-mode.rst
```diff
@@ -80,7 +80,7 @@ position which is the location where the copter was armed.
...
+.. note:: for a NAV_LAND, the :ref:`LAND_SPD_MS<LAND_SPD_MS>` and the :ref:`LAND_SPD_HIGH_MS<LAND_SPD_HIGH_MS>` parameters affect descent speed just like in the :ref:`land-mode`. Unlike the LAND mode, the switch point altitude is set by the mission altitude parameter, will be 10m above that altitude in whatever reference frame being used for the altitude parameter. The :ref:`PILOT_THR_BHV<PILOT_THR_BHV>` option bit 1, when set, allows high throttle stick position to cancel the landing, and switch to a position-hold mode until the mode is changed or mission restarted. During landing the pilot can re-position the vehicle using the pitch and roll sticks to avoid obstacles unless the :ref:`LAND_REPOSITION <LAND_REPOSITION>` parameter is changed to "0".
```
rmackay9: thanks for this.  we can be a bit more specific though and say loiter or alt hold mode.

rmackay9: I'd say we could remove this part, "until the mode is changed or mission restarted".  I think it's obvious that if the vehicle is switched out of Auto mode that the mission will not continue

jinmussj: Thanks, that makes sense. Updated to mention Loiter/AltHold directly and removed the last clause.

rmackay9 (approved):
looks good, thanks!

PR #7749, opened 2026-05-13
Common: remove out-of-date commercial support entries

rmackay9:
This removes some of the out-of-date entries from the commercial support page.

I've contacted every individual or company on the list by email, direct message (on AP's discord server) or linked-in to confirm that they wished to stay on the list and were still intending to provide commercial support.

The removals are all the result of the individual/company saying they were not providing support OR I was unable to contact them despite at least 3 attempts.

We can of course reinstate entries in the future if I've made a mistake here.


Hwurzburg (approved):

Hwurzburg:
thanks

PR #7751, opened 2026-05-15
Revise circle mode documentation for clarity

jinmussj:
- The rate-limit wording is updated because Circle mode uses about half of the configured horizontal acceleration value when limiting the achievable circle rate. https://github.com/ArduPilot/ardupilot/blob/29a29550b1347d607c5e22347f3fd21b83444300/libraries/AC_WPNav/AC_Circle.cpp#L354

- The in-flight tuning text is clarified because RC Channel 6 tuning with TUNE=39 uses the TUNE_MIN/TUNE_MAX range rather than an implied fixed scaling of CIRCLE_RATE.

- The stick-control reference is also updated from CIRCLE_CONTROL  to CIRCLE_OPTIONS bit 0, and the later roll/pitch note is made conditional on that bit not being enabled.


rmackay9 (approved):
looks good, thanks!

PR #7752, opened 2026-05-15
remove land alt line from copter auto mode

### LINKED ISSUE #7747: Copter: auto mode page should lose reference to LAND alt
As mentioned in the comments of https://github.com/ArduPilot/ardupilot_wiki/pull/6945 we should remove this line from the [Copter auto mode page](https://ardupilot.org/copter/docs/auto-mode.html).  It's potentially a good idea but that's not how it works in 4.7 at least and from inspection I don't think it worked like that in older versions either

`Unlike the LAND mode, the switch point altitude is set by the mission altitude parameter, will be 10m above that altitude in whatever reference frame being used for the altitude parameter`

Hwurzburg:
resolves #7747


rmackay9 (dismissed):

rmackay9:
Thanks very much!

rmackay9 (approved):

PR #7753, opened 2026-05-15
Sub: document existing MOT_FV_CPLNG_K behaviour

ES-Alexander:
The [specificity of this feature](https://github.com/ArduPilot/ardupilot/pull/33076) upsets me a bit, but docs should reflect the way things are, so here we find ourselves 🤷‍♂️ 

Screenshot 2026-05-16 at 6 51 18 am


Williangalvani (approved):

ES-Alexander:
@Hwurzburg that ArduPilot PR just clarifies the parameter docs - this feature has been like this since it was implemented near the start of ArduSub existing, it's just that the frame limitation (and feature more generally) weren't documented.

PR #7754, opened 2026-05-16
Common: clarify RTL rally-point selection wording when Home is included

### LINKED ISSUE #7632: Clarify RTL rally-point selection wording when Home is included
In testing, I observed that even with rally points configured, RTL can still return to Home when Home is closer.

That seems different from the current documentation, which suggests:
- RTL returns to the closest rally point when rally points are configured
- `RALLY_LIMIT_KM = 0` means the closest rally point is always used

From `libraries/AP_Rally/AP_Rally.cpp`, `RALLY_LIMIT_KM = 0` appears to only disable the rally distance limit, while the final choice is still made by `calc_best_rally_or_home_location()`. When `RALLY_INCL_HOME == 1`, Home can still be selected if it is closer.

It may be worth updating the wording in both places to match the current implementation more precisely.

evanofficial:
Resolves #7632

The current Rally Points page suggests that setting RALLY_LIMIT_KM=0 means the closest rally point is always used, but the actual selection in libraries/AP_Rally/AP_Rally.cpp::calc_best_rally_or_home_location() only treats LIMIT_KM as a distance gate. When RALLY_INCL_HOME=1 (the default on Copter and Rover), Home is also considered as a candidate and is chosen whenever it is closer than the nearest rally point, regardless of LIMIT_KM.

Rewrite the RALLY_LIMIT_KM and RALLY_INCL_HOME bullets in common-rally-points.rst so the two parameters' roles (distance gate vs. Home-as-candidate) are clearly separated, and add a short note in the overview pointing readers at RALLY_INCL_HOME for the Home-vs-rally tie-break.

Only common/source/docs/common-rally-points.rst is changed; that page has copywiki destination="copter,plane,rover,sub,planner" so a single edit fans out to every sub-project that documents rally points. The RTL mode pages (copter/plane) already link to this page via :ref: and have been left untouched to keep the diff minimal.

### DIFF
--- a/common/source/docs/common-rally-points.rst
+++ b/common/source/docs/common-rally-points.rst
@@ -21,8 +21,10 @@ to the point of takeoff.
 For this reason we now support the creation of multiple Rally Points.
 Should an aircraft enter RTL and Rally Points have been defined then it
 will proceed to the closest Rally Point, rather than proceeding to the
-Home position. Plane will then loiter at that location, and Copter will
-perform an automated landing there.
+Home position (although Home itself can still win when it is closer than
+any Rally Point -- see :ref:`RALLY_INCL_HOME <RALLY_INCL_HOME>` below).
+Plane will then loiter at that location, and Copter will perform an
+automated landing there.
 
 .. figure:: ../../../images/mp_flight_plan_with_three_rally_points.jpg
    :target: ../_images/mp_flight_plan_with_three_rally_points.jpg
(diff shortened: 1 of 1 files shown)


IamPete1:
I think the only thing I would add is that it will always go home is there is no other option. For example if you have set some rally limit distance such that all your rally points are too far away, home will be used independent of `RALLY_INCL_HOME`.

common/source/docs/common-rally-points.rst
```diff
@@ -21,8 +21,10 @@ to the point of takeoff.
 For this reason we now support the creation of multiple Rally Points.
 Should an aircraft enter RTL and Rally Points have been defined then it
 will proceed to the closest Rally Point, rather than proceeding to the
-Home position. Plane will then loiter at that location, and Copter will
-perform an automated landing there.
+Home position (although Home itself can still win when it is closer than
```
rmackay9, line 24: TBH, I'd prefer if we don't include the exceptions up here in the high level description.

common/source/docs/common-rally-points.rst
```diff
@@ -81,10 +83,20 @@ The following MAVLink parameters control Rally Point behavior:
    is the maximum distance a Rally Point may be from the aircraft to be
    considered for an RTL event.  If all Rally Points are greater than
    this distance from the aircraft, then the Home location is used for
-   RTL events (at altitude :ref:`RTL_ALTITUDE <RTL_ALTITUDE>`)
+   RTL events (at altitude :ref:`RTL_ALTITUDE <RTL_ALTITUDE>`).
```
rmackay9, line 86: it's an existing issue but I think we should replace "events" with "location".  In this case we could just end the sentence at "used".

rmackay9:
Txs for this.

I actually don't think there's any issue with the current wording though.

Hwurzburg:
> Txs for this.
> 
> I actually don't think there's any issue with the current wording though.

will you be the approver then?

rmackay9:
I'm sorry, I'm going to close this I think because I think the wording is ok.  Sorry but thanks for your efforts anyway!

PR #7756, opened 2026-05-18
Common: expand neopixel signal voltage workaround with diode details …

chobitsfan:
…and orientation

The current wiki page shows a diode workaround in an image, but does not provide a text description. This PR adds explicit instructions to use a 1N4001 diode and clarifies the required diode orientation.

The diode shown in the image appears to be a 1N4148 diode (~300mA max), which is typically used for signal applications. This PR also updates the recommendation to use a 1N4001 diode (~1A max), which is more appropriate for power supply use.

Tested on SDMODELH7V2


Hwurzburg (changes requested):
I dont agree with change in image to 1N4001...instead add comment that any silicon diode which can handle the current of the LED array can be used

chobitsfan:
Hi @Hwurzburg Thank you for reviewing. I reverted image and updated text.

Hwurzburg (approved):
tnx...lgtm

PR #7760, opened 2026-05-19
Clarify landing gear startup and altitude-trigger behavior

jinmussj:
1.Clarify LGR_STARTUP wait-for-input behavior

- The previous wording said that PWM signals are prevented from being active during startup. This can be read as meaning the landing gear output is disabled or has no valid PWM output.
- The wait-for-pilot-input behavior is that the landing gear output is left unchanged during startup until a pilot or automatic operation commands a deploy or retract.

2.Clarify altitude-triggered landing gear behavior

- The previous wording could be read as saying that the gear deploys or retracts whenever the vehicle is below or above the configured altitude.
- The implementation triggers these actions when the vehicle crosses the configured altitude: deploy while descending through LGR_DEPLOY_ALT, and retract while climbing through LGR_RETRACT_ALT.
- This better reflects that the altitude trigger occurs when the vehicle crosses the configured altitude.


common/source/docs/common-landing-gear.rst
```diff
@@ -9,11 +9,11 @@ The gear/ mount can be manually retracted or deployed with an auxiliary switch.
 The gear will also automatically deploy during landing and/or retract during takeoff.
 The output which controls the servo for landing gear/ camera mount is set by the SERVOx_FUNCTION =29.
 
-During startup, the status of the landing gear/ camera mount is determined by the :ref:`LGR_STARTUP<LGR_STARTUP>`  parameter. Starting either in deployed or retracted state can be selected, or to wait for pilot input, which prevents PWM signals from being active on the outputs whose SERVOx_FUNCTION = 29 (Landing Gear). This prevents the current position of those outputs from changing until the pilot or automatic operation commands a change.
+During startup, the status of the landing gear/ camera mount is determined by the :ref:`LGR_STARTUP<LGR_STARTUP>`  parameter. Starting either in deployed or retracted state can be selected, or to wait for pilot input, which leaves the outputs whose SERVOx_FUNCTION = 29 (Landing Gear) unchanged during startup. This prevents the current position of those outputs from changing until the pilot or automatic operation commands a change.
```
IamPete1: It does not output PWM on the channel. We can't do "unchanged" since we don't know what the value was before booting. The assumption is that the landing gear will not move if there is no PWM output.

jinmussj: > It does not output PWM on the channel. We can't do "unchanged" since we don't know what the value was before booting. The assumption is that the landing gear will not move if there is no PWM output.

One point I wanted to check: in SITL, with LGR_STARTUP=WaitForPilotInput and SERVO9_FUNCTION=29, I still see SERVO_OUTPUT_RAW reporting a value on that channel before any landing gear command. Is that just the reported/latched servo output value, while the real hardware channel is not actually outputting PWM?

rmackay9 (approved):

PR #7767, opened 2026-05-24
Modify the outdated descriptions related to parachutes

jinmussj:
1. Clarify relay parachute release defaults
- The previous wording used `RELAY_DEFAULT`, but relay default parameters are numbered as `RELAYx_DEFAULT`.
- It also tied the relay default setting to the bootloader behavior, which can be confusing. The high-level trigger requirement is for the bootloader period, when GPIOs are forced low. The `RELAYx_DEFAULT` setting applies after boot, during autopilot initialization.

2. Clarify parachute relay function setup
    The current wording can be read as saying that setting RELAYx_PIN and CHUTE_TYPE alone is sufficient to select the relay used for parachute release.
    In current Copter, the relay used for parachute release is selected through the RELAYx_FUNCTION parameter set to Parachute. 
    
3. Clarify parachute RC switch behavior
Clarifies that RCx_OPTION=21 enables or disables parachute release generally, rather than only automatic release.

4. Related to ArduPilot/ardupilot#4702
This separates the wording for the regular Crash Check and the automatic parachute release logic, so the page does not imply that Crash Check directly triggers the parachute release.

### DIFF
--- a/common/source/docs/common-parachute.rst
+++ b/common/source/docs/common-parachute.rst
@@ -40,7 +40,7 @@ parachute vendors:
 Connecting to the Autopilot
 ===========================
 
-The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, use a release mechanism that requires a high output level to trigger, since GPIOs are forced low during the bootloader period. Set the corresponding ``RELAYx_DEFAULT`` to OFF or No Change to avoid triggering the release during autopilot initialization.
+The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, use a release mechanism that requires a high output level to trigger, since GPIOs are forced low during the bootloader period, and set the corresponding ``RELAYx_DEFAULT`` to OFF or No Change to avoid triggering the release during autopilot initialization.
 
 
 .. image:: ../../../images/Parachute_Pixhawk.jpg
--- a/common/source/docs/common-parachute.rst
+++ b/common/source/docs/common-parachute.rst
@@ -121,7 +121,7 @@ Setting bit 1 high of :ref:`CHUTE_OPTIONS<CHUTE_OPTIONS>` will not disarm the ve
 RC Disable/Enable of Parachute
 ------------------------------
 
-You can disable or enable parachute release using an RC channel/switch: set an ``RCx_OPTION`` to 21. A high position enables parachute release; other positions disable it. If set to 23 (Parachute 3Pos), the low position disables, middle position enables operation, and a high will attempt to force release, as explained above, assuming the enabling conditions described below.
+You can disable or enable parachute release using an RC channel/switch: set an ``RCx_OPTION`` to 21. A high position enables parachute release; other positions disable it. If an ``RCx_OPTION`` is set to 23 (Parachute 3Pos), the low position disables parachute release, middle position enables operation, and a high will attempt to force release, as explained above, assuming the enabling conditions described below.
 
 When will the parachute deploy?
 ===============================
--- a/common/source/docs/common-parachute.rst
+++ b/common/source/docs/common-parachute.rst
@@ -40,7 +40,7 @@ parachute vendors:
 Connecting to the Autopilot
 ===========================
 
-The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). Be sure that if a Relay pin is used, that the parachute release is active on a high level and that the Relay pin default, ``RELAY_DEFAULT`` is OFF or No Change, to avoid triggering the release during initialization, since all GPIOs are forced low during the bootloader period.
+The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, use a release mechanism that requires a high output level to trigger, since GPIOs are forced low during the bootloader period. Set the corresponding ``RELAYx_DEFAULT`` to OFF or No Change to avoid triggering the release during autopilot initialization.
 
 
 .. image:: ../../../images/Parachute_Pixhawk.jpg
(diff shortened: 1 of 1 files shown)


common/source/docs/common-parachute.rst
```diff
@@ -40,7 +40,7 @@ parachute vendors:
 Connecting to the Autopilot
 ===========================
 
-The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). Be sure that if a Relay pin is used, that the parachute release is active on a high level and that the Relay pin default, ``RELAY_DEFAULT`` is OFF or No Change, to avoid triggering the release during initialization, since all GPIOs are forced low during the bootloader period.
+The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, use a release mechanism that requires a high output level to trigger, since GPIOs are forced low during the bootloader period. Set the corresponding ``RELAYx_DEFAULT`` to OFF or No Change to avoid triggering the release during autopilot initialization.
```
Hwurzburg:
```suggestion
The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, use a release mechanism that requires a high output level to trigger, since GPIOs are forced low during the bootloader period, and set the corresponding ``RELAYx_DEFAULT`` to OFF or No Change to avoid triggering the release during autopilot initialization.
```

common/source/docs/common-parachute.rst
```diff
@@ -72,10 +72,9 @@ Relay Release
...
-- Since GPIOs are always set low initially during the bootloader period, to avoid accidental release, always use a release mechanism that needs a high output level to trigger and set:
-- :ref:`RELAY1_DEFAULT<RELAY1_DEFAULT>` = 1 or 2 (low or no change), which determines how RELAY1 pins is set during the post-bootloader, autopilot initialization period.
-- :ref:`CHUTE_TYPE <CHUTE_TYPE>` = 0,1,2,or 3 to release with RELAY, RELAY2, RELAY3, or RELAY4 functions.
+- Set that pin number as one of the RELAY outputs and assign that relay to the parachute function. For example, for GPIO pin 51 using the first relay, set :ref:`RELAY1_PIN<RELAY1_PIN>` = 51 and :ref:`RELAY1_FUNCTION<RELAY1_FUNCTION>` = 3 (Parachute).
+- Since GPIOs are always set low initially during the bootloader period, to avoid accidental release, always use a release mechanism that needs a high output level to trigger and set.
+- :ref:`CHUTE_TYPE<CHUTE_TYPE>` = 0 to release with a relay. Set the corresponding relay function parameter, such as :ref:`RELAY1_FUNCTION<RELAY1_FUNCTION>`, to 3 (Parachute) to choose which relay output is used.
```
Hwurzburg:
```suggestion
- :ref:`CHUTE_TYPE<CHUTE_TYPE>` = 0 to release with a relay. Set the corresponding relay function parameter, such as :ref:`RELAY1_FUNCTION<RELAY1_FUNCTION>` (for RELAY1), to 3 (Parachute) to choose which relay output is used.
```

common/source/docs/common-parachute.rst
```diff
@@ -122,16 +121,12 @@ Setting bit 1 high of :ref:`CHUTE_OPTIONS<CHUTE_OPTIONS>` will not disarm the ve
 RC Disable/Enable of Parachute
 ------------------------------
 
-You can disable or enable the parachute automatic release using an RC channel/switch: set an ``RCx_OPTION`` to 21. A high enables the automatic release function, low disables it. Manual release is unaffected. If set to 23 (Parachute 3Pos), the low position disables, middle position enables auto operation, a high will attempt to force release, as explained above, assuming the enabling conditions described below.
+You can disable or enable parachute release using an RC channel/switch: set an ``RCx_OPTION`` to 21. A high position enables parachute release; other positions disable it. If set to 23 (Parachute 3Pos), the low position disables, middle position enables operation, and a high will attempt to force release, as explained above, assuming the enabling conditions described below.
```
Hwurzburg:
```suggestion
You can disable or enable parachute release using an RC channel/switch: set an ``RCx_OPTION`` to 21. A high position enables parachute release; other positions disable it. If an ``RCx_OPTION`` is set to 23 (Parachute 3Pos), the low position disables parachute release, middle position enables operation, and a high will attempt to force release, as explained above, assuming the enabling conditions described below.
```

Hwurzburg (changes requested):
clarified wordding a bit

common/source/docs/common-parachute.rst
hunt0r: Am I reading correctly that alerting that HIGH must be used to prevent trigger-on-init is duplicated here and down below in the GPIO subsection?

If yes, I suggest a "Don't Repeat Yourself" pattern, so that all information is written in one spot (with links/references as needed). So one option is to simply remove this here (because it is clearly written below). Another option is to "reference" it, perhaps like this.
```suggestion
The parachute release mechanism can be triggered from either a Relay or a PWM (i.e. Servo). If a Relay pin is used, pay careful attention to the information in the GPIO subsection below to prevent an unsafe & unexpected release.
```

But this is "just my opinion". I happily defer to the author's as well as all reviewers' judgements if anyone else prefers keeping the info in here twice. No problem!

hunt0r (approved):
I read this with the eyes of an experienced ArduPilot user who has never used a parachute. I believe it is well-written and clear.

(I'm not qualified to approve its correctness, but other reviewers can speak to that.)

common/source/docs/common-parachute.rst
Hwurzburg: In general, I agree...but when it comes to safety (you do not want to be around an deploying (explosively) parachute,  I accept duplications

hunt0r: We are probably already on the same page, but just in case:
- I understand and support bending all ideals in the name of improved safety. That's top priority.
- My concern with duplications is the future risk that it changes in one spot, but not the other... then the duplication suddenly (and unexpectedly) switches from a safety protection to a safety problem.

IMO a suitably strongly worded "reference" gains most of the safety benefit with minimal risk of future de-sync.

jinmussj: Thanks, I understand the DRY concern.

My preference is to keep the explicit warning here for now, since this is safety-related and it is useful for readers to see the relay/GPIO risk at this point in the page. I agree the detailed setup guidance should mainly live in the GPIO subsection below, and I’m happy to adjust if maintainers prefer the shorter reference-only wording.

hunt0r: (Thanks! I'm satisfied. Please resolve this comment whenever you like, I'm sadly unable to resolve my own comments.)

Hwurzburg (approved):

PR #7770, opened 2026-05-27
Frontend scaling

khancyr:
fixup frontend img scaling and first loading

Carrousel is now slightly bigger but scale correctly

Capture d’écran du 2026-05-27 17-04-51

image

### DIFF
--- a/frontend/css/main.css
+++ b/frontend/css/main.css
@@ -22,11 +22,19 @@ html {
 
 .carousel-inner {
     width: 100%;
-    height: auto;
+    aspect-ratio: 16 / 9;
+}
+
+.carousel-item,
+.carousel-item picture {
+    height: 100%;
 }
 
 .carousel-inner img {
-    max-height: 480px;
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    object-position: center;
 }
 
 .carousel-control-next,
--- a/frontend/css/main.css
+++ b/frontend/css/main.css
@@ -51,10 +51,24 @@ html {
 
 }
 
-.col_min {
-    aspect-ratio: 4/3;
+.blog-card-img-wrap {
+    display: block;
+    position: relative;
+    overflow: hidden;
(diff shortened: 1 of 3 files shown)


Hwurzburg:
doesn't this impact users with other display aspect ratios and widths? will it appear correct for someone with a VGA display?

khancyr:
no, I just adjust the image scaling inside the container, so it will still adapt to the screen size
Here is what we got on Pixel9 that is quite small 
image
image

and galaxy tab size : 
image
image

Georacer:
I don't see the banner being bigger in your screenshots of the standard computer screen aspect ratio.

But other than that this looks clean.
I absolutely cannot read .hmtl, btw.

khancyr:
@Georacer my bad ... that isn't the banner that is bigger but the carrousel !
otherwise, just trying on your laptop is also a way (I am not specialist on html/css either), using the docker container or just something like python -m http.server 8080 from the frontend directory is working

Georacer:
> @Georacer my bad ... that isn't the banner that is bigger but the carrousel ! otherwise, just trying on your laptop is also a way (I am not specialist on html/css either), using the docker container or just something like python -m http.server 8080 from the frontend directory is working

Do you have any idea how long the wiki takes to build?
image

Hwurzburg (approved):

khancyr:
@Georacer I did know, I am probably the only one that optimize the build time ! and that isn't that bad now ! 
But more importantly, you don't need to build anything to test the ardupilot.org front page ! that is static hmtl ! using the docker container will just add 2-3 min to simplify getting the blog and recent post !

Georacer:
> @Georacer I did know, I am probably the only one that optimize the build time ! and that isn't that bad now ! But more importantly, you don't need to build anything to test the ardupilot.org front page ! that is static hmtl ! using the docker container will just add 2-3 min to simplify getting the blog and recent post !

I know... I'm only using a joke to cover up the fact that I'm lazy.

PR #7776, opened 2026-05-31
dev: add a basic document on some autotest test utilities

peterbarker:
Horribly incomplete, but it would be several weeks of work to write a complete document.

Something here allows it to be fleshed out over time is the theory.

### DIFF
--- a/dev/source/docs/autotest-tests.rst
+++ b/dev/source/docs/autotest-tests.rst
@@ -1,0 +1,59 @@
+.. _autotest-tests:
+
+==============
+Autotest Tests
+==============
+
+The autotest suite is constantly evolving.  There are a lot of predefined convenience functions available (see `Vehicle tests <https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/vehicle_test_suite.py>`__ .  There are also many convenience functions within individual vehicle test suites like the `Plane tests <https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/arduplane.py>`__).  These functions can make writing tests fast, and there are a lot of things your test can assume when it starts and stops
+
+This document outlines some of the assumptions you can make and some of the utilities you can take advantage of.
+
+Assumptions Your test Can Make
+==============================
+
+Location
+--------
+For the Copter test suite you can assume the vehicle starts and the same location each time.
+
+Plane starts roughly in the CMAC location.
+
+QuadPlane always starts in the same location in Dalby.
+
+All other tests do not guarantee a starting location or rotation.
+
+Disarmed
+--------
+
+You can assume your vehicle starts disarmed (except for AntennaTracker, which is always armed).
+
+RC Inputs
+---------
+
+RC inputs are zeroed before your test starts
+
+Contexts
+========
+
+Contexts allow test suite state changes to be rolled back conveniently.
+
+Each test gets an implicit context, and you can push any number of contexts onto the stack.  You do not need to wrap your test in a context - you get one for free from the harness.
+
+When you pop a context, various state change are reverted to the way they were before the context was pushed.  Notably:
+
+ - parameters are reverted to their values before the context change
+ - message rates set via ``context_set_message_rate_hz`` are reverted
+ - filesystem files backed up with ``context_backup_file`` get their original contents back
+ - Lua scripts installed with ``install_script_content_context`` are removed
+
+Popping a context may involve rebooting the simulation.  We reboot SITL:
+ - if your test reboots the simulation (after resetting parameters to their pre-test values, in case there are reboot-required parameters)
+ - if your test set the vehicle home via ``self.set_home(loc)``
+
+Methods
+=======
+
+self.reboot_sitl()
+------------------
+
+Reboots not just the vehicle but the entire simulation.  Useful for reboot-required parameters.
+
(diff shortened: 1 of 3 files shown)


hunt0r (approved):
YES YES YES!
I am 100% in favor of helping folks understand and use the autotest suite.

Perfection (in my eyes) would be having autotest code which is so clearly written that it needs no further documentation. This document risks becoming a 'crutch'; counter-productively preventing the need for improving the code itself. If you agree, we could add lines here like "This document is a stop-gap solution. Information is documented here until the day it is made clear in the code itself. Then it should be removed from here."

Having said that, I overwhelmingly agree that having this information here (however imperfect) is better than not having it, in a pros-vs-cons sense.

Thank you for this!

dev/source/docs/autotest-tests.rst
```diff
@@ -0,0 +1,49 @@
+.. _autotest-tests.rst:
+
+==============
+Autotest Tests
+==============
+
+The autotest suite is constantly evolving.  There are a lot of convenience functions available to make writing tests fast, and there is a lot of things your test can assume when it starts and stops
```
Hwurzburg:
```suggestion
The autotest suite is constantly evolving.  There are a lot of predefined convenience functions available (see `Vehicle tests <https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/vehicle_test_suite.py>`__ , as well is individual vehicle test, like `Plane tests <https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/arduplane.py>`__) to make writing tests fast, and there are a lot of things your test can assume when it starts and stops
```

dev/source/docs/autotest-tests.rst
```diff
@@ -0,0 +1,49 @@
...
+Plane starts roughly in the CMAC location.
+
+QuadPlane always starts in the same location in Dalby.
+
+All other tests do not guarantee a starting location or rotation.
+
+Contexts
+========
+
+Contexts allow test suite state to be rolled back conveniently.
```
Hwurzburg:
```suggestion
Contexts allow test suite state changes to be rolled back conveniently.
```

dev/source/docs/autotest-tests.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Each test gets an implicit context, and you can push any number of contexts on top of the stack.  You do not need to wrap your test in a context - you get one for free from the harness.
```
Hwurzburg:
```suggestion
Each test gets an implicit context, and you can push any number of contexts on top of the stack.  You do not need to wrap your test in a context - you get one for free at the beginning each test module.
```

dev/source/docs/autotest-tests.rst
```diff
@@ -0,0 +1,49 @@
...
+
+When you pop a context various state is reverted to the way it was before the context was pushed.  Notably:
```
Hwurzburg:
```suggestion
When you pop a context, various state change are reverted to the way it was before the context was pushed.  Notably:
```

dev/source/docs/autotest-tests.rst
```diff
@@ -0,0 +1,49 @@
...
+
+ - parameters are unset
```
Hwurzburg:
```suggestion
 - parameters are reverted to their values before the context change
```

Hwurzburg (changes requested):
I need to improve the base framework doc also a bit...on my list

dev/source/docs/autotest-tests.rst
peterbarker: This is impinging on Python syntax with the word "module".  I'm not sure what you mean by it, actually - tests are functions, not modules (that's actually a lie, but now you're the second person on the planet to know that).

peterbarker:
> Perfection (in my eyes) would be having autotest code which is so clearly written that it needs no further documentation. This document risks becoming a 'crutch'; counter-productively preventing the need for improving the 

People have asked for a separate document before.  This document is only ever going to be a "introductory" document - not quite a quick-start guide but something akin to it.  Just somewhere the most critical functions are called out as things to pay attention to.  And also somewhere to document anti-patterns we are trying to get rid of.

> code itself. If you agree, we could add lines here like "This document is a stop-gap solution. Information is documented here until the day it is made clear in the code itself. Then it should be removed from here."

I *think* this will have utility past that point as a "here's what to know before you start".  The lengthy explanation of what you can assume from a context push/pop for example - just to get that concept into people's heads.  The mechanism is extremely useful but definitely the "most magic" thing we do in the suite.

peterbarker:
> I need to improve the base framework doc also a bit...on my list

I've applied most of your changes and cleaned up a little bit more.

dev/source/docs/autotest-tests.rst
Hwurzburg: Just looking for a term to quantify a test sequence..(subtest)..function is at bit confusing...maybe just test sequence?

peterbarker: It's the method which people create (and add to the list of tests) - which is invariable called a "test".

Anything else is a "subtest" (marked with a method call or otherwise).

... the other thing is that it *encloses* the test, it's not "at the beginning" of each test.

Maybe "you get one for free for your test"?

khancyr (commented):
fine by me. Better than nothing ... I fear this will bitroot and not be well maintained, but at least we got some a bit of docs so worth it !
I let the final validation to Henry for the contents !

Hwurzburg (approved):

PR #7778, opened 2026-06-02
Dev: companion computer link fixes

rmackay9:
This fixes a few links on the Dev wiki's companion computer page.

In addition I've removed the idrobotics void-ray link but their SkyHub is still linked a few lines lower down in the software section.

FYI @Karthikeyan-Idrobots, @yalexx,

### DIFF
--- a/common/source/docs/common-modalai-voxl2.rst
+++ b/common/source/docs/common-modalai-voxl2.rst
@@ -4,7 +4,7 @@
 ModalAI VOXL 2
 ==============
 
-[copywiki destination="copter,plane,rover,blimp"]
+[copywiki destination="copter,plane,rover,blimp,dev"]
 
 .. image:: ../../../images/modalai-voxl2.png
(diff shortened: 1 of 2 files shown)


Hwurzburg (commented):
While we are at it, shouldn't the voxl2 link be changed to our more complete wiki page? https://ardupilot.org/plane/docs/common-modalai-voxl2.html

rmackay9:
Hi @Hwurzburg,

Cool, txs.  I've updated the VOXL2 link to the copter wiki's page.  I initially tried to instead make the VOXL2 page part of the dev wiki but then it started complaining about not having it in a toc tree so I just used the direct link.

Hwurzburg:
I tweaked to use the common page directly instead of indirectly....just prefer to reduce inter site dependencies if I can...thanks

Hwurzburg (approved):

PR #7780, opened 2026-06-02
Dev: Rpanion setup pages

rmackay9:
This adds a few Rpanion setup pages to the AP wiki.  Rpanion already has its own pages but these are based on my personal experience and also rely on fairly specific hardware recommendations.

The tops of each new page include links back to the original Rpanion setup instructions so I think these pages add value.

I followed the style used for the BlueOS wiki pages and there's even one image which is shared between them.

I've built these locally and they all look good to me


Hwurzburg (dismissed):
A lot of work! thanks!

One spelling error is preventing merging

rmackay9:
@Hwurzburg,

Txs for the review, I've fixed the spelling mistake,txs!

stephendade:
Thanks for doing this @rmackay9! I've gone through the changes and they all look good for Rpanion-server.

Hwurzburg (approved):

PR #7784, opened 2026-06-04
Dev: Seperate out ROS1 and ROS2

stephendade:
This PR separates out the ROS1 and ROS2 instructions into separate sections, to prevent confusion about which sub-pages are for which ROS version.

Also removed duplicated TOC entry for ROS from the "Companion Computer" section.

Note the ``dev/source/docs/ros2.rst`` was renamed and edited into ``dev/source/docs/ros2-install.rst``, which is not quite clear in the Github GUI.

FYI @srmainwaring and @snktshrma


Ryanf55 (approved):
This is a good idea, ROS 1 is EOL and should be easily ignored for our users.

Hwurzburg (approved):

PR #7797, opened 2026-06-08
Common: dronebridge configuration gets button info

rmackay9:
For ESP32 running DroneBridge it is sometimes necessary to push the "Settings" button to make the WifiAP appear so I've added this to the wiki.


Hwurzburg (approved):
thanks, LGTM

PR #7803, opened 2026-06-12
Clarify Copter Land auto-disarm and landing detection conditions

jinmussj:
### Solved Problem

The Land Mode documentation describes the pilot's throttle being at minimum as a condition for automatic disarming and landing detection.

However, Land mode disarms after the landing detector reports `land_complete` and the motors reach `GROUND_IDLE`:

* [ModeLand::gps_run()](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/mode_land.cpp#L880-L883)
* [ModeLand::nogps_run()](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/mode_land.cpp#L952-L956)

The landing detector does not use pilot input throttle. Instead, it separately checks the motor lower limit and the attitude controller's throttle mix:

* [Input throttle limitation](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/land_detector.cpp#L954-L965)
* [Throttle-related landing checks](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/land_detector.cpp#L1043-L1047)

### Solution

Clarify that the copter automatically shuts down and disarms after landing is detected, and replace the ambiguous “Throttle is at minimum” condition with “The attitude controller's throttle mix is at minimum.”

### Test coverage

Documentation-only change, verified against the current Copter Land mode and landing detector implementations.

### DIFF
--- a/copter/source/docs/land-mode.rst
+++ b/copter/source/docs/land-mode.rst
@@ -20,8 +20,7 @@ features:
    .. image:: ../images/Land_DescentSpeed2.png
        :target: ../_images/Land_DescentSpeed2.png
 
--  Upon reaching the ground the copter will automatically shut-down the
-   motors and disarm the copter if the pilot's throttle is at minimum.
+-  After landing is detected, the copter will automatically shut down the motors and disarm.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Hwurzburg:
while technically code correct, the current wording is clearer to a user...if he has throttle above min, it certainly wont disarm

PR #7805, opened 2026-06-14
add JFB flight controllers

Hwurzburg:
while updarting for 4.7-b7 changes, noticed that all the JFB autopilots were missing from the wiki...


rmackay9 (approved):
Thanks so much for this!

PR #7812, opened 2026-06-17
Drotek removal

khancyr:
Remove drotek store : not partner and not working anymore
move pixawk3 pro to deprecated as not produced anymore.
remove old documentation for drotek external baro as no findable anymore


Hwurzburg:
Drotek still seems to advertise the Pixhawk3 Pro on its website

khancyr:
It is present as portfolio. But they don't produce it anymore, and they just do drone show now. They got another brand that sell their compass and gnss and we got no relationship with them, so no need to keep the brand

common/source/docs/common-baro-external.rst
```diff
@@ -28,17 +27,11 @@ Connecting and Configuring
 
 Connect the barometer to the autopilot's I2C port as shown below
 
-.. image:: ../../../images/baro-ms5611-pixhawk.jpg
-    :target: ../_images/baro-ms5611-pixhawk.jpg
```
rmackay9, line 32: do we need to delete this image as well?  maybe its used somewhere else but if not we should delete it I think

rmackay9:
In general this seems OK to me, txs!

common/source/docs/common-baro-external.rst
Hwurzburg: yes should delete it

Hwurzburg (changes requested):

common/source/docs/common-baro-external.rst
khancyr: removed !

khancyr:
@Hwurzburg , image removed

Hwurzburg (approved):

PR #7825, opened 2026-06-30
common: Revert magnetometer change for X_MAV_H743v2 in docs

TompsonTan:
This reverts the magnetometer reference from QMC5883P back to IST8310.

The previous PR mistakenly assumed a hardware change, but the X_MAV_H743v2 flight controller continues to use the IST8310 magnetometer as confirmed by the hardware schematic and current production specs.


Hwurzburg (approved):
V1 used the QMC compass....it was reported as an error in docs apparently before....V2 confirmed to use IST

PR #7826, opened 2026-06-30
Fix PILOT_Y_RATE scaling advice in input-shaping.rst

mbuzdalov:
Parameters for maximum acceleration were not only renamed but also scaled 100x down.

The previous advice to set PILOT_Y_RATE to 0.5% of the maximum rate (expressed in centidegrees per second squared) should have been adjusted to be just 0.5, but it has not been done. This PR fixes it.


rmackay9 (approved):

rmackay9:
Thanks very much for finding this!

PR #7831, opened 2026-07-08
Copter: archive gps glitch protection

rmackay9:
This moves the out-of-date GPS Glitch protection page to the Archive section.

I've moved the small amount of valuable information re the EKF's glitch protection to the Copter EKF failsafe page.


Hwurzburg (approved):
when we archive a page, we usually dont modify it..copying still relevant info to another page is fine....
I have reverted that portion
merging...thanks

rmackay9:
Great, thanks!

PR #7841, opened 2026-07-16
Move Advanced user tools to archived

khancyr:
Those tools are old and deprecated, we shouldn't link them in normal wiki but in archived


Hwurzburg (changes requested):
rare case of where you must update an already archived doc: common-ppm-encoder in order to pass CI

Hwurzburg (approved):

PR #7843, opened 2026-07-16
Dev cleanup

khancyr:
improve the toctree on dev wiki 
image

remove old and broken links
update some text to be more actual

add small tip as guideline for new developper as we do for other wiki


Hwurzburg:
@khancyr filing CI in dev....dont know why...fails locally also when building the dev site

khancyr:
@Hwurzburg I know why. I will split the PR

I have change the name of the pocketbeagle sphinx tag and we rely on it for dev wiki to build. 
Basically, dev wiki don't have common autopilot files, so it rely on intersphinx files to provide the correct link, but as I change the link name, it cannot find it anymore. 
The copter: force to use copter wiki links, otherwise it open pseudo-random wiki link, generally Sub, but could be others that don't have most board ...

khancyr:
@Hwurzburg fixed ! I had to rename the pocketbeagle file, but the links are preserved that way.

Hwurzburg (approved):
thaks

PR #7849, opened 2026-07-18
common: scripts: document extended range of scripting aux functions

IamPete1:
One of my outstanding wiki contributions, this is in 4.7.


Hwurzburg (approved):
thanks

PR #7853, opened 2026-07-18
Replace outdated macOS build steps with the install script

### LINKED ISSUE #7186: Mac Environment Build Instructions do not work for Sonoma
## Bug report

**Issue details**
1. The existing instructions to setup the build environment for MaxOX 14 do not work. Issues are faced with incorrect Python Version, incorrect G++ Version. 
Link & Video Referred to: https://ardupilot.org/dev/docs/building-setup-mac.html

2. The Automatic environment script for mac does not work either. I dont have the exact errors just now. I will have to re-run it on another mac to get the errors again.

Please note that in both cases I was attempting to setup the environment from scratch. 

**Version**
N/A - ./waf configure fails

**Platform**
N/A - ./waf configure fails

**Airframe type**
Copter however the issue would arise for any frame

**Hardware type**
Pixhawk6X however this is irrelevant becase - ./waf configure fails

**Logs**
I dont have the logs at the moment as the issue is now fixed but the problem is related to brew installing an older version 5.4.1 of G++ and the multiple conflicting versions of python. 
a. .pyenv did not work
b. after removing .pyenv version 3.9.6 did not work
c. finally pointed to 3.11.6 under homebrew and that worked.

ismetvahidzade986-sketch:
The macOS setup page still had the old manual steps, and they don't work on current macOS. The Homebrew install used the `/usr/bin/ruby` one-liner that Homebrew removed, and the Mojave SDK-header steps are long obsolete.

Per the discussion in #7186, this drops the manual list and points people at `Tools/environment_install/install-prereqs-mac.sh`, which the maintainers keep current. I kept a short note about using a virtualenv and not running things under sudo, since that trips people up.

Fixes #7186

### DIFF
--- a/dev/source/docs/building-setup-mac.rst
+++ b/dev/source/docs/building-setup-mac.rst
@@ -4,80 +4,34 @@
 Setting up the Build Environment (MacOSX)
 =========================================
 
-This article shows how to manually setup a minimal build environment on MacOS (ver 10.6 onwards).
+This article shows how to set up the build environment on macOS.
 
 ..  youtube:: wLK2wLwEXm4
     :width: 100%
 
 
-There is a `pre-built script  <https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-mac.sh>`__ that will install these pre-requisites.
-
-How to use it, cd to your ardupilot directory, and execute:
-
-::
-
-    sh ./Tools/environment_install/install-prereqs-mac.sh
-
-
-Setup steps
------------
-
-#. MacOS will alert you when you enter a command in the terminal that requires Xcode Command Line Tools. You can also install Xcode Command Line Tools manually
-
-   ::
-   
-       xcode-select --install
-
-#. Install `Homebrew <http://brew.sh>`__ for MacOS (Homebrew is a respected package manager for MacOS)
-
-   ::
-   
-      /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
- 
-#. Install the following packages using brew
-
-   ::
-
-       brew update
-       brew install gcc-arm-none-eabi genromfs python3
-
-#. Install the latest version of awk using brew (make sure
-   **/usr/local/bin** takes precedence in your path):
-
-   ::
-
-       brew install gawk
-
-#. Install *empy* and *pyserial* using the following commands:
-
-   ::
-
-       python3 -m pip install empy pyserial
-       
-   ::
-   
-       ** Starting with MacOS Mojave (10.14.x) you might want to install the SDK headers
-       
-       open /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg
-
-
-#. Follow the :ref:`MAVProxy documentation <mavproxy:mavproxy-downloadinstallmac>` if you plan to use the simulator.
-
-Now you should be able to build with waf as described in `BUILD.md <https://github.com/ArduPilot/ardupilot/blob/master/BUILD.md>`__.
-
-Additional Steps for macOS mojave
----------------------------------
-Due to some changes binutils installed via brew have stopped working for macOS mojave leading to crashing builds. So if installed, remove via following command:
+The recommended way to install the prerequisites is the environment install
+script. It is kept up to date with the current toolchain and Python
+requirements, so it is less likely to drift than a manual list of steps. From
+your ``ardupilot`` directory, run:
 
 ::
 
-    brew uninstall binutils
+    ./Tools/environment_install/install-prereqs-mac.sh -y
 
-Also you will need to install the c++ include headers to /usr/include to do that. Run the following in commandline and follow the installation routine:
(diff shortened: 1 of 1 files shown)


dev/source/docs/building-setup-mac.rst
```diff
@@ -4,80 +4,34 @@
...
-::
-
-    brew uninstall binutils
-
-Also you will need to install the c++ include headers to /usr/include to do that. Run the following in commandline and follow the installation routine:
-
-::
+    Do not run the script with ``sudo``, and avoid ``sudo`` for Python or pip
+    packages in general. Install Python packages into a virtual environment
+    (``python3 -m venv``) rather than system-wide. Homebrew will prompt for your
```
ES-Alexander: I find it unclear whether this is either
1. suggesting the prereqs script will automatically install things into a virtual environment
    - This does not seem to currently be the case, from a quick read of [the script](https://github.com/ArduPilot/ardupilot/blob/master/Tools/environment_install/install-prereqs-mac.sh)
2. recommending users should create and activate a virtual environment before running the prereqs script
    - If so, there should be clearer instructions on how to do so (e.g. `python3 -m venv .venv` followed by `source .venv/bin/activate`), and how to get out of it (i.e. `deactivate`)

dev/source/docs/building-setup-mac.rst
```diff
@@ -4,80 +4,34 @@
...
-   ::
-
-       brew update
-       brew install gcc-arm-none-eabi genromfs python3
-
-#. Install the latest version of awk using brew (make sure
-   **/usr/local/bin** takes precedence in your path):
-
-   ::
+    ./Tools/environment_install/install-prereqs-mac.sh -y
```
ES-Alexander, line 20: Perhaps worth mentioning that leaving off the `-y` argument allows more intentionally approving or declining each step of the process? `-y` is a nice convenience, but not using it provides a trust level between blind trust and manually reading and copying lines from the script.

Personally I'd prefer to let `uv` manage my Python versions and library installs, so I imagine others with similar inclinations may wish to avoid installing `pyenv` and a separate Python version without necessarily being aware of it. The script implementation details seem out of scope here - just mentioning as an example of why `-y` may not always be desirable.

ES-Alexander (commented):
I haven't tried running the script, but its functionality seems out of scope for a wiki PR (so I think that's unimportant to this review). I agree with the general idea that the prereqs script is at least more likely to be up to date than a list of manual instructions in the wiki, and this PR does seem to adequately reflect that change.

A minor gripe/confusion around the virtual environments comment. While I'm personally a fan of them to avoid messing up other parts of your system, I'm unsure how the rest of the ArduPilot dev team leans, and the comment is currently ambiguous (at least to me). 

---

As something of a side note, it _may_ be worth mentioning that it's also possible to [set up using Docker](https://ardupilot.org/dev/docs/building-setup-linux.html#setup-using-docker) for those who want to, but there are a couple of extra steps to doing so:

```bash
# Initial setup, assuming HomeBrew has already been installed
# Install Docker (to run isolated containers), and Colima (because macOS doesn't work with docker engine)
brew install docker colima
colima start  # (or `brew services start colima`, to auto-restart at login)
docker build . -t ardupilot --build-arg USER_UID=$(id -u) --build-arg USER_GID=1111

# Every time you want to do development, make sure Docker can run, and enter the ArduPilot docker container
colima start && docker run --rm -it -v "$(pwd):/ardupilot" -u "$(id -u):$(id -g)" ardupilot:latest bash

# When done, return to the normal Terminal
exit
```

ismetvahidzade986-sketch:
Sorry for the slow reply. You are right that the venv line was confusing, and it also did not match what the script does. install-prereqs-mac.sh runs python3 -m pip into whatever python3 is first on PATH, and only the pyenv prompt keeps that out of the system Python. I reworded the note to say exactly that.

On Docker, I would rather keep this PR to the script change. Happy to add a short Docker section in a follow-up if you think the extra macOS steps are worth documenting. Which steps did you have in mind beyond the linux container setup?

dev/source/docs/building-setup-mac.rst
```diff
@@ -4,80 +4,35 @@
...
 
-       brew install gawk
+The script installs the Xcode command line tools, Homebrew, the ARM toolchain,
+and the Python packages ArduPilot needs. If you would rather set things up by
+hand, read the script and run the steps it performs.
```
ES-Alexander, line 24:
```suggestion
and the Python version and packages ArduPilot needs. If you would rather set things up manually you can run without the ``-y`` flag (to be prompted for confirmation throughout), or read the script and run the steps by hand.
```

ES-Alexander (changes requested):
It's preferable to have text from a single paragraph in a single line of the file (instead of hard-wrapped to some fixed line length), to keep content changes in the git history semantic.

A suggestion to resolve my remaining previous comment.

ES-Alexander:
> Happy to add a short Docker section in a follow-up if you think the extra macOS steps are worth documenting. Which steps did you have in mind beyond the linux container setup?

The `colima` stuff and [manual GID](https://github.com/newsboat/newsboat/issues/2932#issuecomment-2525075678) specified in the code block in my previous comment.

PR #7855, opened 2026-07-19
4.8:32771/document VTOL landing descent throttle mix change

tridge:
Documents the behaviour change from ArduPilot/ardupilot#32771 (Plane 4.8 and later): attitude control is no longer prioritised over vertical speed control during the descent phase of a VTOL landing. A rapid descent on a quadplane often leads to poor attitude control, and the correct recovery is to slow the descent rather than put more power into attitude (see ArduPilot/ardupilot#32759 for the analysis).

Changes:
- `quadplane-auto-mode.rst`: note in the NAV_VTOL_LAND section describing the new 4.8 descent behaviour and pointing at the hover throttle recommendation.
- `quadplane-vtol-tuning-process.rst`: recommendation to set Q_M_HOVER_LEARN = 0 once a good hover throttle value has been learned, since typical quadplane flight profiles rarely include a steady hover and wing lift can corrupt the learned value, and an accurate Q_M_THST_HOVER is important for VTOL landing descent control.


Hwurzburg:
thanks!

PR #7857, opened 2026-07-19
Document LUA support absence for DAKEFPVH743/Pro

aelfakih:
Added section about absence of LUA support in this flight controller

### DIFF
--- a/common/source/docs/common-dakefpvh743pro.rst
+++ b/common/source/docs/common-dakefpvh743pro.rst
@@ -125,7 +125,7 @@ The HD VTX connector can have RX4 replaced by the analog VTX signal if that conn
 
 LUA Support
 ===========
-The DAKEFPVH743/Pro does not have an onboard SD Card resulting in the absence of LUA support.
+The DAKEFPVH743/Pro does not include an onboard microSD card, so Lua scripts cannot be loaded from removable storage in the normal way. Advanced users may still use Lua by building a custom ArduPilot firmware with the required scripts embedded in the firmware image.
 
 Loading Firmware
 ================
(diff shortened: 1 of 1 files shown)


common/source/docs/common-dakefpvh743pro.rst
```diff
@@ -123,6 +123,10 @@ The DAKEFPVH743/Pro has an onboard OSD using a MAX7456 chip and is enabled by de
 
 The HD VTX connector can have RX4 replaced by the analog VTX signal if that connector is used for analog VTX connection by using the DJI/VTX jumper pads.
 
+LUA Support
+===========
+The DAKEFPVH743/Pro does not have an onboard SD Card resulting in the absence of LUA support.
```
Hwurzburg: not entirely true....you could create local firmware which embeds LUA scripts into the firmware

there are many boards capable of LUA scripting (>2MB flash) which do not have SD cards...we dont bother adding this caveat anywhere else....anyone interested in LUA, the wiki page up front states that they are normally stored on an SD card...and a little further down states that they can run from embedded firmware if  custom built...

I dont have an issue adding this on this page (with explanation that scripts can be embedded in custom built firmware without the need for an sd card)

Hwurzburg (changes requested):

common/source/docs/common-dakefpvh743pro.rst
aelfakih: makes sense to be explicit about the availability with custom firmware. will edit and resubmit

Hwurzburg (approved):
LGTM...thanks

PR #7862, opened 2026-07-20
build_parameters.py: simplify nosearch marking for versionned files

khancyr:
image

Sphinx is doing weird stuff as including all versionned parameters into search, this PR remove the versionned parameter from the search, we should use only the latest parameter file.

### DIFF
--- a/build_parameters.py
+++ b/build_parameters.py
@@ -636,7 +636,7 @@ def generate_rst_files(commits_to_checkout_and_parse):
         file_out = open(dest_file, "w")
         found_original_title = False
         if "latest" not in version_tag:
-            file_out.write(':orphan:\n\n')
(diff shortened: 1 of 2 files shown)


Hwurzburg (commented):
Not sure I agree with this....otherwise a user of an older version searching for a param will lose the ability to find it...is it spammy for some users, yes...but necessary for others...

khancyr:
I would agree if the search was showing some consitency. But right now, the page is just spammed with all parameter revision and no sane way to filter the revision we want and it also hid other page since we got a lot of parameter revision, so that make the search close to useless.

Hwurzburg:
can we make the search fall back to the param revs IF it does not find anything using master param list? that way an old param will be found, but existing ones wont generate previous revs list

PR #7863, opened 2026-07-20
common: add missing temperature sensors

ES-Alexander:
Adds datasheet links and param references to the SHT3x (ArduPilot/ardupilot#29429) and TMP119 temperature sensors (ArduPilot/ardupilot#33322).


Hwurzburg (approved):
thanks...

PR #7869, opened 2026-07-22
docs: add electro-permanent magnet setup instructions

maksimdrachov:
Hi,

I've added a new sections to Grippers which details the instructions for setting up [FluxGrip FG40](https://fluxgrip.zubax.com/).

I've also wanted to suggest to remove the [Electro Permanent Magnet Gripper (OpenGrab EPM v3)](https://ardupilot.org/copter/docs/common-electro-permanent-magnet-V3.html) and [Electro Permanent Magnet Gripper (EPM688)](https://ardupilot.org/copter/docs/common-electro-permanent-magnet-gripper.html) sections since they're both outdated, as well as the product lines have been discontinued for quite some time now.


Hwurzburg (commented):
just put (dicountinued) after those entries...we wont archive their pages for a while in case someone still has them in use

Hwurzburg (dismissed):
lgtm

common/source/docs/common-electro-permanent-magnet.rst
```diff
@@ -0,0 +1,150 @@
+.. _common-electro-permanent-magnet:
+
+================================
+Electro-Permanent Magnet (EPM)
```
rmackay9: Txs for this.  I think if this is Zubax FluxGrip specific instructions then both this page's title and the entry on the landing page should make that clear.

BTW, it looks like the lines above and below the title are slightly longer than the text which makes one of the CI tests fail (I think)

Txs again

maksimdrachov: I think it should be good now

rmackay9 (approved):

PR #7879, opened 2026-07-26
4.8:33169/Yaw-during-landing

Hwurzburg:
## Summary

Documents the behaviour change from ArduPilot/ardupilot#33169, which fixed pilot yaw input being ignored/consumed inconsistently during landing:

- `LAND_REPOSITION` now also gates the pilot's yaw stick, not just position, while landing (LAND mode, RTL's final descent, and mission LAND commands).
- `RTL_OPTIONS`' "Ignore Pilot Yaw" bit is now documented, including that it's overridden by `LAND_REPOSITION` during RTL's final descent/land stage.
- `AUTO_OPTIONS`' yaw-ignore bit is now documented as being overridden by the sub-mode's own setting (`RTL_OPTIONS`/`LAND_REPOSITION` or `GUID_OPTIONS`) while a mission runs an RTL, LAND, or guided command.


rmackay9 (approved):
very nice!

PR #7882, opened 2026-07-28
4.8:/322804/add circuitStatus information

Jono453:
Add in information for circuitStatus parameter usage to align with existing approved PR. Explaining the assumption of matching the servo ID with the circuit ID for correct telemetry logging. https://github.com/ArduPilot/ardupilot/pull/32804


Hwurzburg:
thanks for this!

PR #7883, opened 2026-07-28
Common: XFRobot C-20T setup instruction fixes

rmackay9:
This improves the CADDX and XFRobot wiki pages to make it clear that the XFRobot C-20T should be setup using the CADDX camera gimbal driver.

I've done two things:

1. Fixed the XFRobot page so that it specifies that the MNTx_TYPE parameter is the CADDX value for the C-20T
2. changed the title of the CADDX page to make it clear that it also covers the XFRobot C-20T.
3. added notes to both pages referring to the other

This might be overkill as the 1st page listed above is probably sufficient to help users get the XFRobot C-20T setup correctly but I also think it may be helpful for users to realise they are the same product.  While this leads to some duplication and potential confusion from users as to which set of instructions they should follow, I think in practice either set of instructions will work and perhaps its unavoidable because both the CADDX and XFRobot each have their own stores, setup instructions, user manuals, firmwares and configuration apps.

This is in response to [this 4.7 discussion](https://discuss.ardupilot.org/t/xfrobot-gimbal-support/132467/12) where a user was misled by the existing instuctions.

### DIFF
--- a/common/source/docs/common-caddx-gimbal.rst
+++ b/common/source/docs/common-caddx-gimbal.rst
@@ -2,14 +2,16 @@
 
 [copywiki destination="plane,copter,rover,sub"]
 
-=============
-CADDX Gimbals
-=============
+=============================
+CADDX / XFRobot C-20T Gimbals
+=============================
 
-`CADDX <https://caddxfpv.com/products/caddxfpv-gm1-gm2-gm3.html>`__ GM1, GM2, GM3 gimbals are small 1, 2 and 3-axis gimbals weighing between 16g and 46g and can be paired with 19mm x 19mm FPV cameras. The gimbal mounted camera is always stabilized against short term movements in all axes.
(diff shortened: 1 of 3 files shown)


newcomer9747:
Well actually the XF C-20T/40T and Caddx GM3 share the exact same firmware, even the file name is the same. I flashed my XF C-40T with the GM3 firmware from Caddx site and it works exactly the same. 

Also MNT1_TYPE should be Caddx and SERIALx_PROTOCOL = Gimbal if one intents to use mission planner to control the gimbal via mavlink commands (Point here, ROI, Payload control etc), and map all channels to CH01 in the gimbalconfig.exe software). This also gives you headtracker control using trainer cable. RCx_OPTION should also be set according to ardu wiki for mount roll pitch yaw, with X corresponding to the aux channels on the radio(mavlink mode).

MNT1_TYPE if set to Mavlink1/2 will only give mavlink rc controls(Sticks or headtracker for roll pitch yaw), not mavlink messages commands like ROI.

Serial1_Protocol should be set to mavlink2 if radio link mode is on mavlink, assuming receiver is on UART1.

Hwurzburg (approved):
LGTM...thanks

PR #7888, opened 2026-07-29
common: make the EKF failsafe page common and cover Plane and Rover

### LINKED ISSUE #7370: Plane/Rover: EKF Failsafe isn't documented
The failsafe is referenced in [at least one common file](https://ardupilot.org/rover/docs/common-ac2_simple_geofence.html#warnings) (which should maybe be removed from Plane, since it explicitly mentions just Copter/Rover?), but the link goes out to Sub because [the EKF Failsafe page](https://ardupilot.org/sub/docs/ekf-inav-failsafe.html) isn't included in Rover or Plane.

Currently [both](https://ardupilot.org/plane/docs/apms-failsafe-function.html#apms-failsafe-function) firmware [have](https://ardupilot.org/rover/docs/rover-failsafes.html#rover-failsafes) single-page failsafe docs, neither of which mentions the EKF failsafe. Plane [mentions it in the quadplane docs](https://ardupilot.org/plane/docs/quadplane-flying.html#i-have-an-ekf-failsafe), but doesn't specify how to configure it.

I'm unsure whether the preferred solution here is to split the Plane and Rover failsafe docs into separate pages with a landing page (and making the EKF Failsafe page "common"), to match the Copter and Sub approaches, or have some kind of shorter form docs for the EKF failsafe that get included in their existing single-page docs.

Hwurzburg:
Fixes #7370

Copter and Sub each had a near-identical `ekf-inav-failsafe.rst`, while Plane and Rover had no EKF failsafe documentation at all despite both having the failsafe. As a result the EKF failsafe links in `common-ac2_simple_geofence` and `common-polygon_fence` resolved out to the **Sub** wiki when read from Plane or Rover.

This follows @Hwurzburg's suggestion in the issue: one common page tailored per vehicle site, referenced from each vehicle's failsafe list.

**What changed**

- merged the Copter and Sub pages into `common/source/docs/common-ekf-inav-failsafe.rst`, destination `copter,plane,rover,sub`, with `[site]` blocks for the parts that genuinely differ
- deleted `copter/source/docs/ekf-inav-failsafe.rst` and `sub/source/docs/ekf-inav-failsafe.rst`; moved the two images to the root `/images` directory
- added an EKF Failsafe section to `rover-failsafes.rst` and to `apms-failsafe-function.rst`, each linking to the common page, and added toctree entries for Plane and Rover
- updated the two toctree entries and the one hard-coded URL in `dev/security-landing-page.rst`

**New content, taken from the source**

- **Rover** — `FS_EKF_ACTION` 0 = disabled, 1 = Hold (default), 2 = report only; only acts in modes that require position (`Rover/ekf_check.cpp`)
- **Plane** — there is no `FS_EKF_ACTION`. The check only runs while armed **and** in a VTOL posvel mode; action is QLAND in VTOL AUTO, QHOVER otherwise; fixed wing falls back to DCM and takes no action (`ArduPlane/ekf_check.cpp`). This was previously only mentioned in passing in quadplane-flying.
- corrected the Sub page's `:ref:`AltHold <altholdmode>`` link, which pointed at an anchor that does not exist on the Sub wiki

**Two things worth your call**

1. **URL change.** `/copter/docs/ekf-inav-failsafe.html` becomes `/copter/docs/common-ekf-inav-failsafe.html`. The `.. _ekf-inav-failsafe:` anchor is kept in the common page so every existing `:ref:` link still resolves, but external links from the forum will break. Tell me if you would rather keep the Copter page at its current filename and handle Plane/Rover differently.
2. I put the **EKF Glitch Protection** section under `copter,plane,rover` and left it off Sub, and kept the dataflash log example and video Copter-only. Easy to change.

Built cleanly with `update.py --fast` for copter, plane, rover and sub — no Sphinx warnings, and no stray `[site]` shortcodes in the output.


rmackay9 (approved):
very nice!

PR #7889, opened 2026-07-29
4.9:/33840/ros2-interfaces: document DDS status text feedback and throttling

mgwedd:
Companion wiki update for ArduPilot/ardupilot#33840 (AP_DDS: throttled status
texts, marked WikiNeeded).

Adds a "Status Text Feedback" section to the ROS 2 interfaces page:

- commands and control inputs received over DDS now report their outcome as
  GCS status texts (including the previously-silent external control
  rejections);
- these messages are rate-limited per call site — first occurrence immediate,
  repeats suppressed for 5 s by default — because their rate is set by the
  remote ROS 2 node, not by ArduPilot;
- the interval is a compile-time knob, `AP_DDS_STATUSTEXT_MIN_INTERVAL_MS` in
  `AP_DDS_config.h`, linked in the existing compile-time-configuration style
  of the page.

Should merge only after ArduPilot/ardupilot#33840 merges. I will adjust the
wiki if the interval or message set changes during that review.


Hwurzburg:
thanks for this

Hwurzburg:
@mgwedd should be one commit....no merge commits allowed.....no need to rebase on master unless a merge conflict arises...

PR #7894, opened 2026-07-30
common: clarify EKF source sets per core

### LINKED ISSUE #7815: Clarify behaviour if bit 3 of EK3_SRC_OPTIONS is set
The [parameter description](https://ardupilot.org/rover/docs/parameters.html#ek3-src-options-ekf-source-options) and the last paragraph [here](https://ardupilot.org/copter/docs/common-ekf-sources.html#source-switching) leave it unclear, how lanes and source sets are assigned to each other in this case:
Consider a vehicle with 6 IMUS, 5 lanes (EK3_IMU_MASK = 0b00111101, i.e. IMUs 1, 3, 4, 5 and 6 are enabled) and bit 3 of EK3_OPTIONS set.
What Source Set is assigned to is the missing 
| Lane/Core | IMU | Source Set |
| -----------------| -------|----------|
| 1                 |  1      |    1|
| 2                 | 3      |   2 |
| 3                 | 4      |   3 =C ?|
| 4                 | 5     |   A  |
| 5                 | 6     |   B  |

What values for A,B and C describe the correct behaviour?

---
The second reference above contains

>  Alternatively, setting [EK3_OPTIONS](https://ardupilot.org/rover/docs/parameters.html#ek3-options) bit 3 will force each SRC set to be used by each IMU lane, ie EKF3 primary IMU lane will use the _SRC1 set, _SRC2 for the next, etc.) at initialization.

I believe "primary IMU lane" is a wrong description or at least misleading. I think it should say "first IMU lane".

Hwurzburg:
resolves #7815

### DIFF
--- a/common/source/docs/common-ekf-sources.rst
+++ b/common/source/docs/common-ekf-sources.rst
@@ -41,7 +41,18 @@ Ground stations or companion computers may set the source by sending a `MAV_CMD_
 
 This feature is especially helpful when using :ref:`common-non-gps-to-gps`.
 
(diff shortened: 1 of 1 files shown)


rishabsingh3003 (approved):

rmackay9:
thanks very much for this!

PR #7896, opened 2026-07-31
dev: correct MAVLink 1 frame byte indexes

### LINKED ISSUE #7741: MAVLink Basics diagram - Potential off by one error
In the article on https://ardupilot.org/dev/docs/mavlink-basics.html, the message format diagram appears to be incorrect. It says the data portion of the frame goes from byte 6 to n + 6, where n is presumably the length of the data depending on the message type. Then, it goes on to say the checksum should be in bytes n+7 and n+8.
I believe that the checksum should be n+6 and n+7, as otherwise there is one undefined byte at the end of the data portion. Additionally, frames are supposed to be between 8 and 263 bytes, but by the diagram, if the payload length was 0, there would be 9 bytes, and if it were 255, there would be 264.

Hwurzburg:
resolves #7741


dev/source/docs/mavlink-basics.rst
```diff
@@ -13,9 +13,40 @@ MAVLink Basics
 Message Format
 --------------
 
-.. image:: ../images/mavlink-frame.png
```
rmackay9, line 16: I wonder if we can delete the image as well?

Hwurzburg: good catch

rmackay9 (approved):

PR #7897, opened 2026-07-31
copter: clarify RTL cone altitude limits

### LINKED ISSUE #7683: RTL cone-slope wording conflicts with observed close-to-home RTL behavior
The [RTL Mode page](https://ardupilot.org/copter/docs/rtl-mode.html) says the climb altitude is limited by distance from return point * RTL_CONE_SLOPE, and that it “may be less depending on the other parameters, but not higher.” The same section also says that 2m is the minimum return altitude.

In a SITL test, I switched to RTL at about 1.5m altitude while very close to the return point, with RTL_CONE_SLOPE = 3 and RTL_CLIMB_MIN = 0. At RTL trigger, the horizontal distance was about 0.135 m, so distance * RTL_CONE_SLOPE was about 0.404 m, but the vehicle remained around 1.499 m during RTL.

This seems inconsistent with a strict reading of “not higher than distance * RTL_CONE_SLOPE”, and it also does not reflect a fixed 2m minimum return altitude in this scenario.

Whether the relevant description needs to be revised?

Hwurzburg:
resolves #7683

### DIFF
--- a/copter/source/docs/rtl-mode.rst
+++ b/copter/source/docs/rtl-mode.rst
@@ -21,7 +21,23 @@ The altitude reference frame is set by either the rally point, if proceeding to 
 .. image:: ../images/RTL.jpg
     :target: ../_images/RTL.jpg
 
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):
Looks good, thanks!

PR #7898, opened 2026-07-31
copter: clarify Traditional Heli land detection

### LINKED ISSUE #7587: Traditional Heli LAND mode note should clarify “below mid-position” wording and remaining lower-limit logic
### The Traditional Heli note on the LAND mode wiki page appears to need clarification.

**issue**
I already opened a corresponding issue in the main ArduPilot repository and received confirmation that this is a wiki wording issue: `[ArduPilot/ardupilot#32571](https://github.com/ArduPilot/ardupilot/issues/32571)` (“Question about ‘below mid-position’ in Traditional Heli LAND detection”).

Current wiki wording:
“For Traditional Heli, the low motor check in the above landing detection algorithm is replaced with a check that Collective output is below mid-position ...”（）

From the current implementation, there seem to be two separate wording issues:

1. The heli landing path uses `get_below_land_min_coll()`, whose code comment describes it as the land-min-collective flag, not simply “below mid-position”.

2. In `Copter::update_land_detector()`, there is also still an explicit lower-limit related branch:
   `|| ((!get_force_flying() || landing) && motors->limit.throttle_lower && pos_control->get_vel_desired_U_ms() < 0.0f);`

So the current wording “the low motor check is replaced” appears stronger than what the code currently does. It may be more accurate to clarify that heli landing detection involves land-min-collective / lower-limit related logic, rather than saying the low-motor check is fully replaced by a mid-position check.

**Relevant links:**
- LAND mode wiki page:
  https://ardupilot.org/copter/docs/land-mode.html

- `Copter::update_land_detector()`:
  https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/land_detector.cpp

- `get_below_land_min_coll()` and `get_coll_mid()`:
  https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Motors/AP_MotorsHeli.h#L922-L957

- heli motor code updating `below_land_min_coll`:
  https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Motors/AP_MotorsHeli_Single.cpp#L362

Hwurzburg:
resolves #7587

### DIFF
--- a/copter/source/docs/land-mode.rst
+++ b/copter/source/docs/land-mode.rst
@@ -42,7 +42,18 @@ features:
    **descent speed transition** from :ref:`LAND_SPD_HIGH_MS<LAND_SPD_HIGH_MS>`
    to :ref:`LAND_SPD_MS<LAND_SPD_MS>` — it has no role in landing detection or disarming.
 
-.. note:: For Traditional Heli, the low motor check in the above landing detection algorithm is replaced with a check that Collective output is below mid-position (controlled by the vertical position controller, ie in descent). The rotor still may be at governor speed up until Motor Interlock is removed and  disarming occurs.
+.. note::
+
+   Traditional Heli uses heli-specific collective logic in place of
+   "motors at lower limit" and "Throttle is at minimum" conditions above. In
+   manual collective modes, the condition is met when collective is at or below
+   :ref:`H_COL_LAND_MIN<H_COL_LAND_MIN>` or the collective stick is low.  In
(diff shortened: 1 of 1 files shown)


bnsgeyer:
@Hwurzburg I changed the wording a bit.  I removed the sentence on mid-collective because I don't think that should be mentioned since it isn't used in the code.

Hwurzburg:
@bnsgeyer still need your approval in order to merge

bnsgeyer (approved):
@Hwurzburg Thanks for teeing this one up! Approved

PR #7907, opened 2026-08-01
build: support Python virtual environments

### LINKED ISSUE #6155: Support virtual environment for python
# Request

Can we have the wiki setup work in a python virtual environment?

https://github.com/ArduPilot/ardupilot_wiki/issues/358#issuecomment-2263025787

I'm having issues with the `flake8` install hosing my ROS 2 install which has flake8 installed as a debian. 

# Expected behavior

```
python3 -m venv .venv
source .venv/bin/activate
./Spinxsetup.sh
python3 update.py --site dev
```

# Actual  behavior

```
python3 -m venv .venv
source .venv/bin/activate
./Spinxsetup.sh
```
```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

Hwurzburg:
resolves #6155
@peterbarker   dont know if this is still needed or not

### DIFF
--- a/Sphinxsetup.sh
+++ b/Sphinxsetup.sh
@@ -7,6 +7,14 @@ if [ "$UID" -eq 0 ]; then
      exit 1
 fi
 
+# pip refuses "--user" when it is run by a virtual environment interpreter, and
+# this script runs with "set -e", so probe the python3 that will actually
+# receive the install rather than trusting $VIRTUAL_ENV, which is only set when
+# an environment has been activated in the shell.
+is_python_venv() {
+    python3 -c 'import sys; raise SystemExit(0 if hasattr(sys, "real_prefix") or sys.prefix != getattr(sys, "base_prefix", sys.prefix) else 1)'
+}
+
 # Check for lsb_release and set distribution variables safely
 if command -v lsb_release &> /dev/null; then
     DISTRIBUTION_ID=$(lsb_release -i -s)
@@ -38,7 +46,11 @@ if [[ "$(uname)" == "Darwin" ]]; then
         python3 -m ensurepip --upgrade || true
     fi
     SCRIPT_DIR=$(dirname $(realpath ${BASH_SOURCE[0]}))
-    python3 -m pip install --user --upgrade -r "$SCRIPT_DIR"/requirements.txt
+    if is_python_venv; then
+        python3 -m pip install --upgrade -r "$SCRIPT_DIR"/requirements.txt
+    else
+        python3 -m pip install --user --upgrade -r "$SCRIPT_DIR"/requirements.txt
+    fi
     echo "Setup completed successfully for macOS!"
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


peterbarker:
> resolves #6155 @peterbarker dont know if this is still needed or not

I don't either.

@Ryanf55 and @khancyr ?

Hwurzburg:
@khancyr should I close this and the associated issue?

Hwurzburg:
@peterbarker @khancyr @Ryanf55 - to answer "is this still needed": yes. Tested just now.

`Sphinxsetup.sh` sets `set -e` on line 2, and `pip install --user` hard-errors inside a venv:

```
$ python3 -m venv testvenv && . testvenv/bin/activate
$ python3 -m pip install --user --upgrade sphinx
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
$ echo $?
1
```

So on any non-noble distro (Ubuntu 22.04, Debian 12, macOS) a user with a venv active cannot run setup at all - the script aborts on that line. The noble branch already installs without `--user`, so this just extends the same behaviour to the case where the user has made their own venv.

This is the minimal fix: it reacts to `$VIRTUAL_ENV` rather than creating or managing a venv, so there is no new state and no change for anyone not using one.

Sphinxsetup.sh
```diff
@@ -38,7 +38,11 @@ if [[ "$(uname)" == "Darwin" ]]; then
         python3 -m ensurepip --upgrade || true
     fi
     SCRIPT_DIR=$(dirname $(realpath ${BASH_SOURCE[0]}))
-    python3 -m pip install --user --upgrade -r "$SCRIPT_DIR"/requirements.txt
+    if [ -n "${VIRTUAL_ENV:-}" ]; then
```
Ryanf55: What sets the VIRTUAL_ENV environment variable?

Hwurzburg: The activation script that `venv`/`virtualenv` generates sets it - sourcing `bin/activate` does `VIRTUAL_ENV=<path>; export VIRTUAL_ENV` and prepends `$VIRTUAL_ENV/bin` to `PATH` (PEP 405, Python >= 3.3). Creating an environment doesn't set it, and pip doesn't either.

Which is a good question to have asked, because it's the wrong thing to test here. `VIRTUAL_ENV` tracks *activation*, but what matters is which interpreter `python3` resolves to. Those come apart in both directions: `uv run`, `tox` and `hatch` run the environment's interpreter without exporting `VIRTUAL_ENV` (so pip would still reject `--user` and the script would still abort), and a stale exported `VIRTUAL_ENV` can name an environment that is no longer on `PATH` (so we'd drop `--user` against the system python for no reason).

I've pushed a version that probes the interpreter instead:

```bash
is_python_venv() {
    python3 -c 'import sys; raise SystemExit(0 if hasattr(sys, "real_prefix") or sys.prefix != getattr(sys, "base_prefix", sys.prefix) else 1)'
}
```

`sys.prefix != sys.base_prefix` is the documented PEP 405 test; the `real_prefix` check covers pre-20 `virtualenv`. Debian/Ubuntu system pythons have equal prefixes, so the non-venv path is unchanged. Tested activated, unactivated-but-on-PATH, stale `VIRTUAL_ENV`, and plain system python.

Not addressed here, and pre-existing: the noble branch drops `--user` unconditionally, which is a separate PEP 668 question.

PR #7920, opened 2026-08-02
4.8:23092/DO_SET_ROI_WPNEXT_OFFSET

Hwurzburg:
Documents MAV_CMD_DO_SET_ROI_WPNEXT_OFFSET (point gimbal at next waypoint with an offset), per ardupilot PR https://github.com/ArduPilot/ardupilot/pull/23092.

Documented as Copter-only since AP_Vehicle::get_wp_location() is currently only overridden by ArduCopter; Plane/Rover/Sub accept the mission item but the mount never receives a target. Also notes the >2MB flash requirement.


rmackay9 (approved):
Looks OK to me.  This is @peterbarker's work I think

PR #7922, opened 2026-08-03
common: document MAV_CMD_DO_SET_REVERSE mission command

### LINKED ISSUE #1640: Rover: rewrite mission command list page
The [Mission Commands](http://ardupilot.org/rover/docs/common-mavlink-mission-command-messages-mav_cmd.html) page is very generic - we should replace it with a page that lists all the mission commands that Rover accepts.

In particular we should list the DO_REVERSE command that is only supported by Rover

Hwurzburg:
Add the Rover-only MAV_CMD_DO_SET_REVERSE mission command to the mission command reference: a new full section describing param1 (0=forward, 1=reverse) plus a list entry in the "Commands supported by Rover" section. Also fixes the source-file link in that section, which pointed at Rover/commands_logic.cpp -- removed/renamed to Rover/mode_auto.cpp.

Fixes #1640

### DIFF
--- a/common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
+++ b/common/source/docs/common-mavlink-mission-command-messages-mav_cmd.rst
@@ -318,7 +318,7 @@ Commands supported by Rover
 ===========================
 
 This list of commands was inferred from the command handler in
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):
Thanks!

PR #7923, opened 2026-08-03
dev: document MAVLink fence/rally upload, log download, calibration

### LINKED ISSUE #1884: Dev: fill out mavlink interface pages
Our developer wiki has a [Mavlink Interface](http://ardupilot.org/dev/docs/mavlink-commands.html) section with some information on what mavlink commands our ArduPilot vehicles support but we should add sections or pages with much more detail of how ArduPilot vehicles can be controlled using MAVlink.  Some things we should add include:

- how to arm/disarm the vehicle
- how to set the vehicle's mode
- how to read and write parameter values
- how to upload and download missions, fences and rallypoints
- how to download logs
- how to make the vehicle take-off and fly/drive to a given location (in Guided mode)

The purpose of these pages is to make it easier for software developers to write ground stations that work with ArduPilot vehicles

Hwurzburg:
Fill the three remaining gaps identified in #1884's "much more detail of how ArduPilot vehicles can be controlled using MAVlink" request:

- mavlink-mission-upload-download.rst: new section explaining that fence and rally points reuse the same mission-item protocol via the mission_type field (MAV_MISSION_TYPE_FENCE / _RALLY, MAVLink 2 only), plus the MAV_CMD_NAV_FENCE_* / MAV_CMD_NAV_RALLY_POINT item types and how ArduPilot selects a rally point on RTL.
- mavlink-log-download.rst (new): LOG_REQUEST_LIST/LOG_ENTRY, LOG_REQUEST_DATA/LOG_DATA, LOG_ERASE, LOG_REQUEST_END, including the disarm-required behaviour, why a trailing zero-count LOG_DATA isn't a reliable end-of-transfer signal, and a pointer to MAVFTP as the faster alternative.
- mavlink-calibration.rst (new): MAV_CMD_PREFLIGHT_CALIBRATION (gyro/baro/trim/simple-accel/force-save, and which fields ArduPilot actually implements vs. the MAVLink spec's superset), the interactive full 6-position accel cal handshake via MAV_CMD_ACCELCAL_VEHICLE_POS, and compass calibration via MAV_CMD_DO_START_MAG_CAL/_ACCEPT_/_CANCEL_ plus MAG_CAL_PROGRESS/_REPORT.

Both new pages added to the MAVLink Interface toctree. Content verified against the ardupilot firmware source (GCS_Common.cpp, AP_Compass_Calibration.cpp, MissionItemProtocol_Fence.cpp, AP_Logger_MAVLinkLogTransfer.cpp, AP_Logger_File.cpp, AP_Rally.cpp), not just the MAVLink message/command names, including a Codex-assisted second-pass review that caught several spec-vs-actual-behaviour mismatches (unimplemented fence inclusion groups, unimplemented PREFLIGHT_CALIBRATION mag-start path, RC-calibrating param4, rally selection depending on RALLY_INCL_HOME/RALLY_LIMIT_KM, LOG_REQUEST_END semantics).

Fixes #1884

### DIFF
--- a/dev/source/docs/mavlink-calibration.rst
+++ b/dev/source/docs/mavlink-calibration.rst
@@ -1,0 +1,162 @@
+.. _mavlink-calibration:
+
+==================
+Sensor Calibration
+==================
+
+This page explains how MAVLink can be used by a ground station or companion computer to trigger sensor calibrations. The user wiki pages for these procedures are here: :ref:`accelerometer <copter:common-accelerometer-calibration>`, :ref:`compass <copter:common-compass-calibration-in-mission-planner>`, :ref:`radio <copter:common-radio-control-calibration>`.
+
+.. note::
+
+   The vehicle must be disarmed before starting a calibration. If
+   armed, ``MAV_CMD_PREFLIGHT_CALIBRATION`` replies with a "Disarm to
+   allow calibration" status text and is rejected, and
+   ``MAV_CMD_DO_START_MAG_CAL`` replies with "Disarm to allow compass
+   calibration". ``MAV_CMD_DO_ACCEPT_MAG_CAL`` and
+   ``MAV_CMD_DO_CANCEL_MAG_CAL`` have no armed-state check.
+
+Simple calibrations
+--------------------
+
+Gyro, barometer, board-level (trim) and simple accelerometer calibrations are each triggered with a single `MAV_CMD_PREFLIGHT_CALIBRATION <https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_CALIBRATION>`__ sent within a `COMMAND_INT <https://mavlink.io/en/messages/common.html#COMMAND_INT>`__ (or ``COMMAND_LONG``). Only one of the fields below should be non-zero in a given message.
+
+.. raw:: html
+
+   <table border="1" class="docutils">
+   <tbody>
+   <tr>
+   <th>Command Field</th>
+   <th>Type</th>
+   <th>Description</th>
+   </tr>
+   <tr>
+   <td><strong>param1</strong></td>
+   <td>float</td>
+   <td>1: gyro calibration</td>
+   </tr>
+   <tr>
+   <td><strong>param2</strong></td>
+   <td>float</td>
+   <td>magnetometer calibration — see below</td>
+   </tr>
+   <tr>
+   <td><strong>param3</strong></td>
+   <td>float</td>
+   <td>1: barometer (and airspeed, if fitted) calibration</td>
+   </tr>
+   <tr>
+   <td>param4</td>
+   <td>float</td>
+   <td>positive value: sets an internal "RC calibrating" flag that blocks arming (arming check fails with "RC calibrating") while true; zero/negative clears it — but only when param1 and param3 are both 0. ArduPilot does not itself run an RC endpoint calibration routine in response to this field. See note below</td>
+   </tr>
+   <tr>
+   <td><strong>x</strong> (param5)</td>
+   <td>int32_t</td>
+   <td>Accelerometer calibration action:
+
+   1: full 6-position calibration (see below)
+
+   2: board-level/trim calibration
+
+   4: simple (single position) accelerometer calibration
+
+   76: force-accept the existing accelerometer calibration as valid, without re-running it
+   </td>
+   </tr>
+   <tr>
+   <td>y (param6)</td>
+   <td>int32_t</td>
+   <td>Vehicle-specific: Copter — 1: CompassMot calibration. Rover — 1: windvane direction calibration, 2: windvane speed calibration. Sub — 1: rejected (MAV_RESULT_UNSUPPORTED), CompassMot is not supported. Plane, AntennaTracker and Blimp: not used.</td>
+   </tr>
+   <tr style="color: #c0c0c0">
+   <td>z (param7)</td>
+   <td>float</td>
+   <td>not used</td>
+   </tr>
+   </tbody>
+   </table>
+
+.. note::
+
+   ``x``/``y`` are ``int32_t`` fields of ``COMMAND_INT``, not ``float``. When
+   this command is sent as ``COMMAND_LONG`` instead, its float param5/param6
+   are cast directly to ``x``/``y`` on receipt.
+
+.. note::
+
+   ArduPilot does not itself run an RC endpoint calibration routine in
+   response to param4 — see the :ref:`radio calibration page
+   <copter:common-radio-control-calibration>` for how RC calibration
+   actually works. The "RC calibrating" flag it does set/clear is
+   only actually evaluated when param1 and param3 are both 0 — a
+   command that also triggers a gyro (param1=1) or barometer
+   (param3=1) calibration returns before reaching the param4 check,
+   so param4 has no effect in that same message. In practice this
+   means: combining a gyro or barometer calibration with param4=0
+   will **not** clear a flag left set by an earlier command — a
+   separate command with param1=0, param3=0 and param4 ≤ 0 (or a
+   reboot) is required to clear it. Conversely, an unrelated
+   accelerometer-calibration command (param5/``x`` set, param1=0,
+   param3=0) with param4 left at its default of 0 **will** clear the
+   flag, even in the middle of an in-progress RC calibration.
+
+ArduPilot only implements one action for the magnetometer's param2
+field: a value of 76 force-accepts the existing compass calibration as
+valid without re-running it (useful after a parameter reload that
+cleared the calibration-valid flags). The MAVLink spec also defines a
+value of 1 to *start* a magnetometer calibration via this same
+command, but ArduPilot does not implement that path — sending param2 =
+1 alone returns ``MAV_RESULT_UNSUPPORTED``. Use the dedicated
+:ref:`compass calibration commands <mavlink-calibration_compass-calibration>` below to actually start a mag calibration.
+
+Full 6-position accelerometer calibration
+------------------------------------------
+
+Sending ``MAV_CMD_PREFLIGHT_CALIBRATION`` with x (param5) = 1 starts
+the interactive full accelerometer calibration, which requires the
+vehicle to be placed in each of 6 orientations in turn. ArduPilot
+drives this using `MAV_CMD_ACCELCAL_VEHICLE_POS <https://mavlink.io/en/messages/ardupilotmega.html#MAV_CMD_ACCELCAL_VEHICLE_POS>`__:
+
+.. note::
+
+   Before starting the position wizard, ArduPilot first runs a gyro
+   calibration internally. If that gyro calibration fails (for
+   example the vehicle is being handled or vibrating), the whole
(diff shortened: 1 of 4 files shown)


dev/source/docs/mavlink-commands.rst
```diff
@@ -22,6 +22,8 @@ ArduPilot supports the `MAVLink protocol <https://mavlink.io/en/>`__ for communi
     Gimbal / Camera Mount <mavlink-gimbal-mount>
     MAVFTP <mavlink-mavftp>
     Mission Upload/Download <mavlink-mission-upload-download>
+    Log Download <mavlink-log-download>
```
rmackay9: nitpick: could we put these in alphabetical order?

rmackay9 (dismissed):
Looks OK to me with the exception of the little nitpick re alphabetical ordering.

In fact, @peterbarker is the one who knows whether the details are correct

Txs!

dev/source/docs/mavlink-commands.rst
Hwurzburg: will do, but I am keeping Mavlink Basics at the top for immediate eye find...and other commands at the bottom

dev/source/docs/mavlink-mission-upload-download.rst
```diff
@@ -110,3 +110,60 @@ The example commands below can be copy-pasted into MAVProxy (aka SITL) to test t
...
+- `MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION <https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION>`__ — one vertex of an inclusion polygon (vehicle must stay inside); at least 3 required, each with param1 set to the total vertex count of that polygon
+- `MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION <https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION>`__ — one vertex of an exclusion polygon (vehicle must stay outside)
+- `MAV_CMD_NAV_FENCE_CIRCLE_INCLUSION <https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_CIRCLE_INCLUSION>`__ — circular area the vehicle must stay inside (param1 = radius in metres)
+- `MAV_CMD_NAV_FENCE_CIRCLE_EXCLUSION <https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_CIRCLE_EXCLUSION>`__ — circular area the vehicle must stay outside (param1 = radius in metres)
+
+There is also a work-in-progress ``MAV_CMD_NAV_FENCE_HOME_CIRCLE_INCLUSION``
+(5005), not yet in the stable MAVLink message set.
```
peterbarker: This work was merged.

rmackay9 (dismissed):

dev/source/docs/mavlink-mission-upload-download.rst
```diff
@@ -110,3 +110,60 @@ The example commands below can be copy-pasted into MAVProxy (aka SITL) to test t
...
+
+The user wiki pages for :ref:`Fence Setup <copter:common-geofencing-landing-page>` and :ref:`Rally Points <copter:common-rally-points>` explain these features from a GCS-user's perspective.
+
+Fence items
+~~~~~~~~~~~
+
+Each fence ``MISSION_ITEM_INT`` commonly uses one of the following
+commands:
+
+- `MAV_CMD_NAV_FENCE_RETURN_POINT <https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_RETURN_POINT>`__ — location the vehicle heads to if it breaches the fence and RTL is the configured fence action (there can be only one)
```
peterbarker: This is not accurate - it's not used generally, only used for two very specific functions.

dev/source/docs/mavlink-calibration.rst
```diff
@@ -0,0 +1,123 @@
...
+
+   2: board-level/trim calibration
+
+   4: simple (single position) accelerometer calibration
+
+   76: force-accept the existing accelerometer calibration as valid, without re-running it
+   </td>
+   </tr>
+   <tr style="color: #c0c0c0">
+   <td>y (param6)</td>
```
peterbarker, line 67: This isn't true, behaviour across vehicles differs.

peterbarker (dismissed):

dev/source/docs/mavlink-mission-upload-download.rst
Hwurzburg: done

dev/source/docs/mavlink-mission-upload-download.rst
Hwurzburg: done

dev/source/docs/mavlink-calibration.rst
Hwurzburg: expanded

peterbarker:
I asked claude+codex to compare this vs the actual code.

```
  Wrong statements about the code

  1. dev/source/docs/mavlink-mission-upload-download.rst:129 — MAVLink 1 guard doesn't cover MISSION_CLEAR_ALL
  The note implies non-mission types are rejected on MAVLink 1. mavlink2_requirement_met() is called from handle_mission_count(), handle_mission_request_list(), handle_mission_request_int(),
  handle_mission_request() and handle_mission_write_partial_list() — but not from MissionItemProtocol::handle_mission_clear_all() (libraries/GCS_MAVLink/MissionItemProtocol.cpp:33-39),
  which goes straight to clear_all_items() and ACKs MAV_MISSION_ACCEPTED. A MAVLink 1 client sending MISSION_CLEAR_ALL with mission_type=MAV_MISSION_TYPE_FENCE silently erases the stored
  geofence.

  2. dev/source/docs/mavlink-log-download.rst:37 — "invalid log id cancels the transfer" is unreachable
  AP_Logger::handle_log_request_data() returns immediately on if (_log_sending_link != nullptr). Since that's non-null for the whole transfer, the packet.id > num_logs || packet.id < 1 →
  end_log_transfer() branch only runs when no transfer is active. A client implementing abort-by-bad-id gets no effect: LOG_DATA keeps streaming, dataflash logging stays disabled, arming
  stays blocked with "Downloading logs". Only LOG_REQUEST_END aborts.

  3. dev/source/docs/mavlink-log-download.rst:27 — zero-count LOG_DATA terminator condition is too narrow
  handle_log_send_data() ends the transfer on nbytes < MAVLINK_MSG_LOG_DATA_FIELD_DATA_LEN || _log_data_remaining == 0, so a short final chunk also terminates with no zero-count follow-up. A
  zero-count LOG_DATA only appears when the requested ofs is already ≥ log size. Implementing end-of-log detection on count==0, as the page's phrasing invites, hangs at 99% on any log whose
  size isn't a multiple of 90.

  4. dev/source/docs/mavlink-calibration.rst:120 — MAG_CAL_REPORT is neither one-shot nor unconditional
  MSG_MAG_CAL_REPORT/MSG_MAG_CAL_PROGRESS are in STREAM_EXTRA3_msgs (GCS_MAVLink_Parameters.cpp), and Compass::send_mag_cal_report() has no already-sent latch — it re-emits for every
  SUCCESS/FAILED compass on each slot. So a client with SRx_EXTRA3=0 sees nothing and appears to hang; a client treating each report as the completion event re-prompts at the EXTRA3 rate.
  (Codex independently confirmed this one.)

  5. dev/source/docs/mavlink-calibration.rst:37 — param2 row marked "not used" contradicts the page's own prose
  The row is styled <tr style="color: #c0c0c0">, the convention other dev MAVLink pages use for genuinely unused fields. But GCS_Common.cpp:4968 acts on it: is_equal(packet.param2, 
  PREFLIGHT_CALIBRATION_MAGNETOMETER_FORCE_SAVE) (76) force-saves the compass calibration — exactly what line 85 of the same page describes.

  Omitted constraints the firmware enforces

  6. dev/source/docs/mavlink-mission-upload-download.rst:143 — polygon vertex ordering invariant missing
  AC_PolyFence_loader.cpp:1044-1065 tracks expected_type_count/expecting_type and rejects the upload with "Received incorrect vertex type" / "Unexpected vertex count" if a differing item
  type or param1 appears mid-run (and "Incorrect item count" at line 1122 if short); circles and return points are rejected mid-run too (lines 1076-1079, 1097-1100). All vertices of one
  polygon must be contiguous and share an identical param1 — stated nowhere on the page. Violating it fails the whole fence upload.

  7. dev/source/docs/mavlink-calibration.rst:48 — param4 is coupled to param1/param3, not independent
  GCS_Common.cpp:4911-4927: param1==1 (gyro) and param3==1 (baro) both return before line 4926's rc().calibrating(is_positive(packet.param4)). Combining a gyro cal with param4=0 leaves
  gcs_is_calibrating set, and AP_Arming.cpp:916-917 then blocks arming with "RC calibrating" until a clean command or reboot. Symmetrically, an accel-cal command with param4=0 silently
  clears the flag mid-RC-calibration.

  8. dev/source/docs/mavlink-calibration.rst:97 — accel cal's gyro precondition missing
  GCS_Common.cpp:4930-4939: for PREFLIGHT_CALIBRATION_ACCELEROMETER_FULL the handler runs AP::ins().calibrate_gyros() first and returns MAV_RESULT_FAILED if it fails, before
  acal_init()/start(). A wizard started on a vehicle being handled gets a bare failure, no MAV_CMD_ACCELCAL_VEHICLE_POS ever arrives, and it waits forever.

  9. dev/source/docs/mavlink-log-download.rst:27 — single-outstanding-request rule missing
  AP_Logger_MAVLinkLogTransfer.cpp:104-116 returns early whenever _log_sending_link != nullptr, staying deliberately silent on the same channel. The page's advice to "repeat LOG_REQUEST_DATA
  with increasing ofs" and re-request missing chunks invites pipelining, which gets neither data nor error. Exactly one LOG_REQUEST_DATA may be outstanding.

  Markup

  10. dev/source/docs/mavlink-calibration.rst:50 (also :55) — :ref: inside a .. raw:: html block
  Raw blocks are emitted verbatim, so :ref:`radio calibration page <copter:common-radio-control-calibration>` is never converted; the browser drops the bogus element and the cell renders as
  "see the :ref:`radio calibration page ` for how RC calibration actually works" with no link. update.py's check_ref_directives() won't catch it, so it ships broken.

  ---
  Two more from Codex that the workflow didn't surface: the same line 120 also claims the report says "whether it was accepted", but the boolean is _cal_saved[compass_id]
  (AP_Compass_Calibration.cpp:348) — MAVLink's autosaved field, distinct from the result status at :347. And mavlink-mission-upload-download.rst:154 omits that FENCE_OPTIONS bit 1
  (AC_Fence.cpp:145) switches inclusion-area behaviour from intersection to union.
```

Please note that we may not be actually compliant with the spec... that's a separate issue I've addressed in various places elsewhere.

Hwurzburg:
@peterbarker all 12 points from your review addressed (the 10 numbered items plus the 2 extra Codex-found ones) — pushed as an amended commit. Re-review requested.

Hwurzburg:
@peterbarker other updates need this as a base....been a few weeks so I am going to merge without your re-review....my robot says its fixed all the points your robot brought up

PR #7924, opened 2026-08-03
rover: document Wheel Rate Control (WRC_ parameters)

### LINKED ISSUE #2109: Rover-4.0 wiki changes
A list of changes we should make for the Rover-4.0 release:

- [ ] complex fences (aka stayout zones, inclusion zones, exclusion zones)
- [ ] consolidate the object avoidance under a landing page
- [ ] mark object avoidance's Dodge method as deprecated for 4.0 and move it to the bottom of the page it's on

Hwurzburg:
Adds the last unaddressed item from #2109's Rover-4.0 checklist. The existing wheel-encoder page only covered using wheel encoders for EKF position/velocity estimation; it never mentioned AP_WheelRateControl, which uses the same encoders to close a PID rate loop around the throttleLeft/throttleRight skid-steering outputs (Rover/Log.cpp, libraries/AP_WheelEncoder/AP_WheelRateControl.cpp).

Also fixes a pre-existing typo on the same page: the WENC2_TYPE bullet linked to the WENC_TYPE anchor instead of WENC2_TYPE.

Content verified against ArduPilot firmware source, including a Codex-assisted second-pass review that caught two real gaps: the encoder "enabled" check has no health/freshness guarantee, and WRC_RATE_MAX=0 does not disable the feature (WRC_ENABLE=0 does).

Fixes #2109

### DIFF
--- a/rover/source/docs/wheel-encoder.rst
+++ b/rover/source/docs/wheel-encoder.rst
@@ -18,7 +18,7 @@ Connection and Setup
 
 - connect motor encoder's A and B outputs to the autopilot (A Pixhawk using AUX OUT 3,4,5 and 6 pins is shown for example above).
 - set those output's ``SERVOx_FUNCTION`` to -1 to allow them to be used as GPIO inputs. See :ref:`common-gpios` for more information. For the example above, set :ref:`SERVO11_FUNCTION<SERVO11_FUNCTION>`, :ref:`SERVO12_FUNCTION<SERVO12_FUNCTION>`, :ref:`SERVO13_FUNCTION<SERVO13_FUNCTION>`, and :ref:`SERVO14_FUNCTION<SERVO14_FUNCTION>` to "-1".
-- set :ref:`WENC_TYPE <WENC_TYPE>` and :ref:`WENC2_TYPE <WENC_TYPE>` to 1 to enable reading from two wheel encoders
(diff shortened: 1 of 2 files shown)


rover/source/docs/wheel-encoder.rst
```diff
@@ -43,6 +43,42 @@ The ``WHEEL_DISTANCE`` MAVLink message shows the total distance travelled by eac
 
 .. image:: ../images/Mavlink-Inspector.jpg
 
+Wheel Rate Control
+==================
+
+For skid-steering vehicles, the wheel encoders can also be used to
```
rmackay9: Txs for this.  I think that for now at least we should only recommend this for balance bots.  I've never found this useful for regular skid-steering rovers.

Sorry to create work but let's move the new contents of this page to the top of the "Tuning the Balance Bot" page https://ardupilot.org/rover/docs/balance_bot-tuning.html and put it in a new section called, "Tuning Wheel Rate Control".  On this page we can leave the Wheel Rate Control section but let's shorten the sentence to just, 

"For Balance bots, the wheel encoders can also be used to close a rate-control loop around each side's throttle output.  See here for more details" and let's make "here for more details" a link to the above "Tuning the Balance Bot" page

Hwurzburg: done

rmackay9 (approved):

PR #7925, opened 2026-08-03
Common: dronebridge button name fixed

rmackay9:
This PR makes these changes to the DroneBridge page:

1. fixes the name of the button to be "RSettings".  I've got the boards and this label is also shown on the top-right of the [official pics for the board](https://drone-bridge.com/products/1_esp32c6-official)
2. advises users to modify the channel and password fields
3. shows how a GCS may be used to modify settings instead of using the device's web page

I updated this PR a few times but I'm all done now


Hwurzburg (approved):

Hwurzburg:
thanks!

PR #7927, opened 2026-08-03
common: add Lanbao PSK-CM8JL65-CC5 lidar setup page

### LINKED ISSUE #2433: Lanbao range finder setup page
It seems like we added a [driver for the Lanbao CM8JL65 lidar](https://github.com/ArduPilot/ardupilot/issues/13249) but didn't create a wiki page (nor parameter descriptions) so it's unlikely users know how to set it up.

Hwurzburg:
Fixes #2433: RNGFND_TYPE=26 was added for this sensor but never got a wiki page, matching the ardupilot#13249 report. The RNGFND_TYPE parameter description has since been fixed on the firmware side (RNGFND_TYPE=26 now shows "LanbaoPSK-CM8JL65-CC5" instead of blank), but the setup page itself was still missing.

Content verified against libraries/AP_RangeFinder/AP_RangeFinder_Lanbao.cpp: serial/UART only, always 115200 baud (set automatically by the driver), and critically the sensor has no way to report "out of range" so the driver hard-caps accepted readings at 6m -- documented as a warning since setting RNGFND1_MAX higher would have no effect.

Range/dimensions/weight and the Seeed Studio product link (now discontinued/out of stock, noted as such) came from PX4's own documentation page for this sensor, cross-checked against the Seeed Studio listing directly.

Added to the Unidirectional Rangefinders list on
common-rangefinder-landingpage.rst.

### DIFF
--- a/common/source/docs/common-lanbao-cm8jl65-lidar.rst
+++ b/common/source/docs/common-lanbao-cm8jl65-lidar.rst
@@ -1,0 +1,32 @@
+.. _common-lanbao-cm8jl65-lidar:
+
+======================
+Lanbao PSK-CM8JL65-CC5
+======================
+
+The Lanbao PSK-CM8JL65-CC5 is a small (38 x 18 x 7mm, <10g) serial (UART) infra-red time-of-flight distance sensor with a stated range of 0.17m to 8m.
+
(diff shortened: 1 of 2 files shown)


rmackay9 (approved):
Txs.  It's come and gone but still good to close the issue and help people who might have purchased it

rmackay9:
I've added a PR to leave it out of the firmware by default for 4.8 https://github.com/ArduPilot/ardupilot/pull/34018

PR #7936, opened 2026-08-04
copter: document how RTL can avoid breaching fences

### LINKED ISSUE #2993: Copter: add instructions to RTL page on how to ensure RTL does not breach fence
ASTM certification requires RTL not breach any geofences (aka stay-in and stay-out zones).  We should add a section describing how to set this up using Dijkstra's avoidance.

Hwurzburg:
Fixes #2993. RTL flies a direct straight-line path back to the return
point by default, which can cross a polygon/circular fence or
stay-out zone since RTL has no path planning enabled by default
(OA_TYPE=0). Points users at the existing Dijkstra's/BendyRuler
avoidance pages (already confirmed applicable to RTL, per their own
"AUTO, GUIDED, and RTL" notes) rather than duplicating their setup
instructions.

### DIFF
--- a/copter/source/docs/rtl-mode.rst
+++ b/copter/source/docs/rtl-mode.rst
@@ -41,6 +41,8 @@ calculated cone height is lower.
 
 If an :ref:`altitude fence <common-geofencing-landing-page>` has been enabled, the RTL climb/return altitude will be limited to be below the fence's maximum altitude.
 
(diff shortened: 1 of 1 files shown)


rmackay9 (commented):
wonderful, thanks!

rmackay9 (approved):

PR #7939, opened 2026-08-04
dev: document pymavlink MAVFTP reference implementation and perf note

### LINKED ISSUE #4688: Dev: document MAVFTP in the mavlink interface section
We are recommending to many users/companies to increase their use of MAVFTP for various bulk uploads/downloads (e.g. parameters, missions, etc) so we should have some documentation in the [MAVLink Interface section](https://ardupilot.org/dev/docs/mavlink-commands.html) to at least help GCS developers get started with implementing this feature.

- document (or link to) how pymavlink can do MAVFTP transfers so that developers can test a known working implementation
- document the messages involved and/or link to the MAVLink specifications but also add warnings that using the MAVLink specification is not sufficient for good performance (apparently pymavlink has many extra optimisations)

Hwurzburg:
Fixes #4688. The page was only 15 lines (link to spec + list of uses)
with neither of the two things the issue asked for:

- a link to a known-working implementation for developers to test
  against (pymavlink's mavftp.py/mavftp_op.py)
- a warning that the bare MAVLink spec sequence isn't sufficient for
  good performance

Checked pymavlink/mavftp.py directly: it pipelines reads via
OP_BurstReadFile instead of one OP_ReadFile per reply, and keeps a
backlog of several in-flight unacknowledged write chunks (write_qsize/
max_backlog settings) instead of a strict request-then-wait loop.
Named both concretely in the warning rather than just saying
"optimisations exist".

### DIFF
--- a/dev/source/docs/mavlink-mavftp.rst
+++ b/dev/source/docs/mavlink-mavftp.rst
@@ -13,3 +13,12 @@ Common uses for using MAVFTP include:
 - :ref:`Uploading firmware to the autopilot <common-install-sdcard>`
 - :ref:`Uploading Lua scripts <copter:common-lua-scripts>`
 - Fast download of parameters, :ref:`onboard logs <copter:common-logs>`, mission command files and :ref:`rally points <copter:common-rally-points>` and :ref:`terrain data <copter:terrain-following>`
+
+Reference Implementation
+========================
+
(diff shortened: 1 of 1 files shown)


rmackay9 (approved):
very nice, thanks!

PR #7953, opened 2026-08-10
Copter: mention AltHold for Save Trim

### LINKED ISSUE #7952: Copter's save-trim also works in AltHold mode
while testing the 4.7 fix for simple mode I noticed that the save trim wiki page (https://ardupilot.org/copter/docs/autotrim.html#save-trim) says the vehicle should be flown in Stabilize (e.g. "fly your copter in Stabilize mode ") mode but it also works in AltHold so we should add this to the sentence

Justine0211:
## Summary

Update the Copter Save Trim instructions to state that the procedure can be performed in either Stabilize or AltHold mode.

## Motivation

The current documentation mentions only Stabilize mode, while Save Trim also works in AltHold mode. This change aligns the user instructions with the supported behavior.

## Testing

Documentation-only change. Reviewed the updated RST sentence for clarity and consistency.

Fixes #7952

### DIFF
--- a/copter/source/docs/autotrim.rst
+++ b/copter/source/docs/autotrim.rst
@@ -58,7 +58,7 @@ Save trim involves essentially transferring your radio transmitter's trims into 
 .. image:: ../images/MP_SaveTrim_Ch7.png
     :target: ../_images/MP_SaveTrim_Ch7.png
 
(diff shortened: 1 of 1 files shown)


rmackay9 (dismissed):
it looks like there's a repeated sentence

Hwurzburg:
@rmackay9 pushed a fix for the repeated sentence — step 3 now reads cleanly with no duplication.

Hwurzburg (approved):

rmackay9:
Thanks!!

PR #7962, opened 2026-08-12
4.8:32389/OptiFlow Filter

Hwurzburg:
Bit 3 (AGL KF for optflow scaling) was added in ArduPilot/ardupilot#32389 and had no wiki coverage beyond the auto-generated parameter description. Bit 2 (terrain alt for optflow above rangefinder range) was also undocumented here, so added both together since they're both relevant to the same section.

Traced against the actual PR source (AP_NavEKF3_OptFlowFusion.cpp): bit 3 requires a working rangefinder, falls back to the normal terrainState-pd estimate if the rangefinder hasn't fused within 5 seconds, and is compiled in wherever optical flow fusion already is.


rmackay9:
sorry and thanks!

Issue #5475, opened 2023-10-10
Inconsistent CAN startup documentation


lida2003:
As I have described in https://discuss.ardupilot.org/t/inconsistent-can-startup-documentation/107538

a) The startup procedure is not what it is in code, 
b) and I didn’t find AP_UAVCAN class also. 

So it’s quite out of date. Hope someone with the knowledge can help to update the wiki.

Thanks.

Hwurzburg:
closed by #7966

PR #7966, opened 2026-08-14
common: correct DroneCAN pool defaults

Tartistbz:
## Summary

Issue #5475 reports that the CAN startup documentation no longer matches the implementation. The documented DroneCAN memory pool sizes were stale, and the parameter group for a second driver was not explicit.

## Changes

- Correct the default DroneCAN pool size from 12 KB to 8 KB.
- Correct the CAN FD default from 24 KB to 16 KB.
- Note that board hardware definitions may override these defaults.
- Clarify that a second driver uses the corresponding `CAN_D2_*` parameters.

## Testing

- `git diff --check`
- Verified the defaults against `libraries/AP_DroneCAN/AP_DroneCAN.cpp` (`DRONECAN_NODE_POOL_SIZE` is 8192 bytes for classic CAN and 16384 bytes for CAN FD).
- Full Sphinx build was not run locally because `sphinx-build` is not available in this environment.

Related to #5475

### DIFF
--- a/common/source/docs/common-canbus-setup-advanced.rst
+++ b/common/source/docs/common-canbus-setup-advanced.rst
@@ -50,7 +50,7 @@ parameter ``CAN_Px_DRIVER``, where x is the number of the CAN port.
 The value of this parameter is the id of driver that will be associated with this
 port (interface).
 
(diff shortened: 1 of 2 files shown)


Hwurzburg (approved):
Verified DroneCAN pool defaults (8 KB / 16 KB CANFD) against libraries/AP_DroneCAN/AP_DroneCAN.cpp on current master, and the CAN_D1_*/CAN_D2_* parameter naming. Numbers and claims are accurate.

PR #7967, opened 2026-08-15
4.8:33947/Simple Mode Fix

Hwurzburg:
Documents ardupilot PR https://github.com/ArduPilot/ardupilot/pull/33947, which fixes a bug where the Simple/Super Simple mode roll/pitch rotation was computed but silently discarded in most flight modes (a regression from the 4.7.0 SI-units migration), and consolidates the transform into the shared pilot-input helpers.

As a result:
- Drift mode now picks up Simple/Super Simple mode support, so it's removed from the "not supported" list on the Simple/Super Simple Modes page.
- Save Trim is confirmed to work while Simple/Super Simple mode is enabled, with a caution about not changing heading between landing and saving.
- Precision Landing's manual reposition input is now rotated too.


rmackay9 (approved):
thanks very much!

PR #7971, opened 2026-08-16
4.8:33792/zero-takeoff-alt-check

Hwurzburg:
Documents the zero-takeoff-altitude mission arming check added in ArduPilot/ardupilot#33792.

`AP_Arming::mission_checks()` now fails arming with "Mission: Zero takeoff altitude" if any takeoff-type mission command has an altitude of exactly 0 (negative altitudes aren't checked). Added to both the Mission: prose subsection and the main "Pre-Arm Failure Messages (All vehicle types)" table, matching how the other Mission: entries are each mirrored in the table.


IamPete1 (dismissed):
Thanks

PR #7972, opened 2026-08-16
common: prearm-safety-checks: fold prose section into the tables, verified against source

Hwurzburg:
Removes the "Other Failure messages" prose section (Barometer, Compass, GPS, INS, Board Voltage, Parameter, Battery/Power Monitor, Airspeed, Logging, Safety Switch, System, Mission, Rangefinder subsections) and verifies every entry against current master (`AP_Arming.cpp` and the vehicle-specific overrides):

- Most entries duplicated a message already correctly documented in one of the three tables above, just under stale or paraphrased wording (e.g. "Ch7&Ch8 Opt cannot be same" was superseded years ago by "Duplicate Aux Switch Options"; "Check Board Voltage" was never the real message, "Board (Xv) out of range 4.3-5.8v" was; "GPS Glitch"/"Alt disparity" were near-misses for the real "GPS glitching"/"Altitude disparity" already in the Copter table).
- Some named messages no longer exist in current source at all: "Need 3D Fix", "Bad Velocity", "Gyro cal failed", "INS not calibrated". "KDECAN Failed" is now dead code — the KDECAN case in AP_Arming's CAN-protocol switch is an empty no-op.
- Two were real, verified, and missing from the tables, so added: "Baro: not healthy" and "Compass not calibrated".
- Fixed "Airspeed 1 not healthy" -> "Airspeed: 1 not healthy" while in there; the actual format string is "Airspeed: %s".


rmackay9 (approved):
LGTM.  I checked various items that are being removed and they seem to be in the table above already. txs!

Issue #2248, opened 2019-11-21
Update Hardware Components Links - Communicating with ODroid via MAVLink


piyushchauhan:
[Communicating with ODroid via MAVLink](http://ardupilot.org/dev/docs/odroid-via-mavlink.html) - 
Update the links of harwares and others parts. There directed to 404 page.

brunoolivieri:
@piyushchauhan thanks for the information. Please, could you propose the updates?

piyushchauhan:
Will do soon.

Hwurzburg:
@piyushchauhan shall I close this? seems like our ODroid pages in the wiki is very outdated and should either be removed to ARCHIVED status or updated.....

Hwurzburg:
Closing this one, as archiving has since answered it.

The page is now **Archived: Communicating with ODroid via MAVLink**, listed under [Archived Topics](https://ardupilot.org/dev/docs/common-archived-topics.html), whose stated purpose is "topics that are no longer considered relevant to most users, but which may have some value to users with old hardware."

The hardware is the reason. The U3 was already marked discontinued when this issue was filed, and the C1+ and XU4 have since left Hardkernel's lineup too. Of the 12 hardkernel.com links on the page, 9 now return 404 and 3 still resolve - and refreshing them is not possible, because the products they pointed at no longer exist to link to.

Rather than keep a parts list pointing at a vendor catalogue that has moved on, the page stays archived as a historical record for anyone still running one of these boards. Dead shop links on an archived page for discontinued hardware are expected rather than a defect.

Feel free to reopen if you feel this has not been addressed.

PR #7973, opened 2026-08-18
docs: archive legacy ODroid companion computer pages

Tartistbz:
## Summary

- Archive the legacy ODroid MAVLink and Samba pages.
- Add warnings about unsupported ODroid hardware and Ubuntu 14.04 / DroneAPI instructions.
- Mark the ODroid entry as archived and point readers to current companion computer documentation.

## Validation

- Ran git diff --check.
- Parsed the changed RST sections with Docutils.
- Verified the Companion Computers cross-reference and navigation label.

Related to #2248

### DIFF
--- a/common/source/docs/common-archived-topics.rst
+++ b/common/source/docs/common-archived-topics.rst
@@ -117,6 +117,7 @@ value to users with old hardware.
     Building for Navio on RPI 2<building-for-navio-on-rpi2>
     Setup the waf Build Environment on Windows10 using WSL <building-setup-windows10>
     Intel Edison <intel-edison>
+    ODroid <odroid-via-mavlink>
     Interfacing with Pixhawk Using the NSH <interfacing-with-pixhawk-using-the-nsh>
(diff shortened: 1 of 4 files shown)


rmackay9:
Thanks for this

I support archiving the Odroid page.  The RPI pages are probably better for helping users of companion computers.

I don't think this is how we do the archiving though.  It should be moved to the Archive section, Archived is added to the title and we should remove it from the companion computer page.  I don't think we normally add an explanation that the page is no longer maintained.

Tartistbz:
Thanks, updated to follow the existing archive structure: the ODroid page is now listed under Archived Topics, removed from Companion Computers, and the extra deprecation notices have been removed.

Hwurzburg:
@Tartistbz sqaush to one commit and I will merge...thanks!

Tartistbz:
Squashed to one commit as requested. Thanks!

Hwurzburg (approved):

rmackay9 (approved):

PR #7981, opened 2026-08-20
common: add firmware update page

peterbarker:
really just here to hold the table at the end showing what versions we support moving to and from

### DIFF
--- a/common/source/docs/common-firmware-update.rst
+++ b/common/source/docs/common-firmware-update.rst
@@ -1,0 +1,41 @@
+.. _common-firmware-update:
+
+================
+Firmware Updates
+================
+
+Flying Stable ArduPilot Point Releases
+--------------------------------------
+
+The ArduPilot project is very careful about what is merged into our "point releases".
+
+Anything that goes into one has either been evaluated as being strictly limited in what impact it can have on existing systems, or as being an important bugfix.
+
+We encourage users to upgrade to any point release which is released as they sometimes contain very important bugfixes.
+
+Parameter Migration
+-------------------
+
+When upgrading you should take a copy of your parameters, and save them with a filename corresponding to the version you are moving from.  If you decide you must return to an earlier firmware these will be invaluable in restoring your vehicle.  However - *do not* ever apply those saved parameter files to newer versions of the firmware!
+
+ArduPilot goes to some effort to make firmware upgrades seamless.  From time to time we need to move parameters around, change their scaling or simply remove them.  We do this automatically on first boot after the firmware is updated.
+
+We can't retain this parameter migration code forever, so there are limits to which versions we guarantee upgrades from and to.  You can still upgrade from older versions to newer versions, but you may find that some functionality changes unexpectedly as you are now using default parameter values rather than your customised values.
+
+If your autopilot falls far behind the modern stable versions you can still get the parameter conversions - you just need to flash intermediate versions - making it a multi-step process.
+
++--------------+--------------------------+
+| New version  | Oldest Migration version |
++==============+==========================+
+| 4.8(latest)  | 4.2                      |
++--------------+--------------------------+
+| 4.7          | 4.2                      |
++--------------+--------------------------+
+| 4.6          | 4.1                      |
++--------------+--------------------------+
+| 4.5          | 4.1                      |
++--------------+--------------------------+
+| 4.4          | 4.1                      |
++--------------+--------------------------+
+
+Note that not every vehicle type has a release for every version number - Sub, for example, has no 4.2, 4.3, 4.4 or 4.6 release.  Use the oldest release your vehicle type has which is no older than the version in the table above; a Sub user on 4.1 would move to 4.5, then to 4.7.
--- a/copter/source/docs/additional-information.rst
+++ b/copter/source/docs/additional-information.rst
@@ -18,4 +18,5 @@ Additional Information and Topics of Interest
     OEM Customization <common-oem-customizations>
     Use Cases and Applications<common-use-cases-and-applications>
     Web Tools <https://firmware.ardupilot.org/Tools/WebTools/>
+    Firmware Update <common-firmware-update>
     Appendix <common-appendix>
--- a/plane/source/docs/additional-information.rst
+++ b/plane/source/docs/additional-information.rst
@@ -15,4 +15,5 @@ Additional Information and Topics of Interest
     OEM Customization <common-oem-customizations>
     Use Cases and Applications<common-use-cases-and-applications>
     Web Tools <https://firmware.ardupilot.org/Tools/WebTools/>
+    Firmware Update <common-firmware-update>
     Appendix <common-appendix>
--- a/rover/source/docs/additional-information.rst
+++ b/rover/source/docs/additional-information.rst
@@ -15,4 +15,5 @@ Additional Information and Topics of Interest
     OEM Customization <common-oem-customizations>
     Use-Cases and Applications <common-use-cases-and-applications>
     Web Tools <https://firmware.ardupilot.org/Tools/WebTools/>
+    Firmware Update <common-firmware-update>
     Appendix <common-appendix>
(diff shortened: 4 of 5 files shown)


peterbarker:
(the numbers were me throwing darts, please suggest alternatives + justifications :-) )

common/source/docs/common-firmware-update.rst
```diff
@@ -0,0 +1,35 @@
...
+
++--------------+--------------------------+
+| New version  | Oldest Migration version |
++==============+==========================+
+| 4.8          | 4.2                      |
++--------------+--------------------------+
+| 4.7          | 4.2                      |
++--------------+--------------------------+
+| 4.6          | 4.1                      |
++--------------+--------------------------+
```
Hwurzburg: 4.8(latest)
have to reformat table, this is what a user sees in firmware server

common/source/docs/common-advanced-configuration.rst
```diff
@@ -19,24 +19,28 @@ tuning options for the vehicle.
     Baro Thrust Compensation <common_baro_thrust_compensation>
     Boat mode <boat-mode>
     Bootloader Update <common-bootloader-update>
+    Firmware Update <common-firmware-update>
```
Hwurzburg: I would not put it on this page..rather additional-information.rst...

Hwurzburg (changes requested):

common/source/docs/common-advanced-configuration.rst
peterbarker: > I would not put it on this page..rather additional-information.rst...

Done

common/source/docs/common-firmware-update.rst
peterbarker: Done

common/source/docs/common-firmware-update.rst
```diff
@@ -0,0 +1,35 @@
...
+If your autopilot falls far behind the modern stable versions you can still get the parameter conversions - you just need to flash intermediate versions - making it a multi-step process.
+
++--------------+--------------------------+
+| New version  | Oldest Migration version |
++==============+==========================+
+| 4.8(latest)  | 4.2                      |
++--------------+--------------------------+
+| 4.7          | 4.2                      |
++--------------+--------------------------+
+| 4.6          | 4.1                      |
```
ES-Alexander, line 34: Sub skipped 4.2, 4.3, 4.4, and 4.6, so should probably have 4.1 -> 4.5 and 4.5 -> 4.7/4.8.
If that's not valid for the other firmware variants then we could differentiate with some `site` filtering.

ES-Alexander (changes requested):

common/source/docs/common-firmware-update.rst
peterbarker: I don't know whether it makes sense for this to be vehicle-specific.

In the linked PR, for example, I'm trying to remove some old conversion code.  The version that was introduced and the version it's leaving aren't changing there.

We should add a note about Sub's missing versions, somehow, I guess?  So a Sub user wanting to go from 4.1 to 4.8 would go via 4.5 not 4.6.  (the table above is specifies the maximum jumps, not the suggested or actual jumps)

peterbarker: Oh, just to emphasise - I opened this PR with "I made these numbers up".  If there's general agreement that we want this then I'll try to come up with some real numbers.

@rmackay9 has previous suggested that a strict 3-version rule is the way to go.  So you could go 4.5->4.8 but not 4.4->4.8 - but you could go 4.5->4.7->4.8

rmackay9:
I'll leave it to @Hwurzburg re whether we need a new page but here are some other pages that users might think are related:

- DOWNLOADS: Firmware https://ardupilot.org/copter/docs/common-downloads_firmware.html
- Loading Firmware https://ardupilot.org/copter/docs/common-loading-firmware-onto-pixhawk.html
- Release procedures https://ardupilot.org/dev/docs/release-procedures.html

rmackay9:
I think we should move the "Parameter Migration" section to the, "Loading Firmware" page

peterbarker:
I asked Claude to flesh out the table for me:
```
● Done. The numbers you threw darts at turned out to be right — here's the
  evidence to back them up in the PR.

  Method: for each release, find the newest parameter-conversion code that has
  been deleted from that branch, then work out which release first shipped that
  conversion. Any firmware older than that release loses those parameters.

  4.8 (master) → oldest 4.2
  - 94ef9886f6e / 58e5e61946d (2026-08-26) "AP_Compass/AP_AHRS: remove from-4.1
    custom-rotation conversion code" — PARAMETER_CONVERSION - Added: Nov-2021,
    first shipped in 4.2.0 (verified absent in Copter-4.1.0, present in
    Copter-4.2.0). Plus everything below.

  4.7 → oldest 4.2
  - 79e97b0864c (2025-07-16) "AP_BattMonitor: remove param conversion to dynamic
    tables required for update from 4.1 to 4.2" —
    convert_dynamic_param_groups(), absent in Copter-4.1.0, present in
    Copter-4.2.0.
  - 175e24d5140 (2025-07-22) "AP_RPM: remove param conversion from 2021" —
    AP_RPM::convert_params(), same story: first shipped 4.2.0.
  - Everything else dropped in 4.7 is pre-4.0: e202a4128cb (Copter, "all in the
    Copter-4.0.0 tag"), fadc2938675 (Plane, "all present in 4.0.0"),
    069939b87c7/7c903adff8a (Rover/Sub ARMING_CHECK, "in 4.0.0").

  4.6 / 4.5 / 4.4 → oldest 4.1
  - e8802d1858f (2023-01-03) "AP_Airspeed: remove old param conversion" — the
    AHRS_WIND_MAX → ARSPD_WIND_MAX conversion (Added: Oct-2020) first shipped in
    4.1.0 (absent in Copter-4.0.7), removed in time for 4.4. That's the binding
    constraint for 4.4–4.6.
  - 4.6 also drops 2fd10d5ff29 (Plane quadplane PIDs, "upgrading from 3.7 to 4.6
    will not work as well") — 3.7-era, doesn't move the floor.
  - Nothing removed between 4.5.7 and 4.6.3 touches anything newer than that,
    and no conversion dated after Oct-2021 is missing from 4.6.3.

  I cross-checked this two ways: ancestry of each removal commit against
  Copter-4.5.7/4.6.3/4.7.0/master, and a per-release census of every
  PARAMETER_CONVERSION - Added: <date> comment still in the tree (the only date
  that disappears 4.6→4.7 besides the 2021 ones is Feb-2022, and that's
  cef48fc432f, which removed comments that were mislabelled — they were default
  values, not conversions).

  Changes to the page: table rows for 4.5 and 4.4 added (both 4.1), and a note
  that the table gives maximum jumps, not vehicle-specific paths — which answers
  ES-Alexander's Sub point without needing [site] filtering: Sub 4.1 → 4.5 →
  4.7 falls straight out of the generic numbers.

  I didn't touch rmackay9's request to move the "Parameter Migration" section
  over to the Loading Firmware page — say the word and I'll do that too.
```

peterbarker:
Henry and I chatted.

We noted that there were several sections missing from one page but were present on another.

@Hwurzburg is going to make a common page to hold some of this stuff (inc. references to the custom build server IIUC)

Hwurzburg:
Per @rmackay9's suggestion above, the "Parameter Migration" section (prose + version table) is now carried in #8002, which collapses the two "Loading Firmware" sidebar entries into a single `common-loading-firmware.rst` page. It lands there as the "Parameter Conversion" subsection under Additional Information, with the wording and table unchanged.

That does leave the section in two places: `common-firmware-update.rst` here still has it. Whichever of these merges second probably wants it dropped from that copy — this PR would then be left with "Flying Stable ArduPilot Point Releases", which #8002 does not cover.

No files overlap between the two PRs, so they won't conflict in git either way.

common/source/docs/common-firmware-update.rst
Hwurzburg: comment added

Hwurzburg:
closing, incorporated in previous wiki prs

PR #7988, opened 2026-08-22
Common: DroneBridge ESP32 gets multivehicle setup instructions

rmackay9:
This adds instructions for using DroneBridge ESP32 for multiple vehicles and/or multiple GCSs.

It took me quite a few attempts to figure out the setup that worked so hopefully it will be useful to others.  I have tested it fairly rigorously so I think it's correct.

I've also built the page locally and it looks OK to me.


Hwurzburg:
thanks!

PR #7997, opened 2026-08-26
Copter: add arc-waypoint to command list

rmackay9:
This adds the new-for-4.7 [ARC_WAYPOINT command](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_ARC_WAYPOINT) to the copter command list.

We should also add it to the more general [mission-command-list page](https://ardupilot.org/copter/docs/common-mavlink-mission-command-messages-mav_cmd.html) but I'll do that as a separate PR.

I have built this locally and it looks OK to me


Hwurzburg:
thanks

PR #8002, opened 2026-08-27
common: consolidate firmware loading onto a single page

Hwurzburg:
The setup sidebar carried two separate "Loading Firmware" entries — one for boards that already have an ArduPilot compatible bootloader, one for boards that do not — which asked the reader to decide which page applied before either page explained how to tell. Both pages also carried their own copies of the firmware download instructions and the beta/developer version descriptions.

### New page

Adds a single **Loading Firmware** page (`common-loading-firmware.rst`) as the sidebar entry, with a local `.. contents::` jump table and four sections:

1. **Download the Firmware** — firmware server navigation, which file type each loading method needs (`.apj` / `_with_bl.hex` / `.abin` / `.bin`), then Stable, Beta, Latest Developer Version and Custom Builds.
2. **Loading the Firmware** — how to tell whether the autopilot has an ArduPilot compatible bootloader, from the board's provenance and from a USB/Device Manager test, then links to the two loading pages for each case.
3. **Testing that it Worked**.
4. **Additional Information** — Parameter Conversion, Loading Firmware via SD Card, and Firmware Feature Limitations with the custom build server.

### Existing pages

Both loading pages are retitled to say which case they cover, and reduced to just the loading procedure — the download, beta/developer, SD card, feature limitation and testing content now lives once on the parent page:

- `common-loading-firmware-onto-pixhawk` → "Loading Firmware to Boards **with** an ArduPilot Compatible Bootloader"
- `common-loading-firmware-onto-chibios-only-boards` → "Loading Firmware to Boards **without** an ArduPilot Compatible Bootloader"

Both keep their existing anchors, so the many board pages that link straight to them are unaffected.

The Zadig / Betaflight Configurator alternate DFU method is dropped, leaving STM32CubeProgrammer as the single documented path, and the two images it used are removed. The external flash / SSBL section stays, since those boards cannot be loaded with STM32CubeProgrammer at all.

The sidebar entries collapse to one in all seven menus (copter, copter tradheli, plane, rover, sub, blimp, planner), and the cross-references that previously offered the reader a choice between the two pages now point at the new page.

### Parameter Conversion

The Parameter Conversion section takes the "Parameter Migration" text and its version migration table from #7981, as @rmackay9 suggested there. The wording and table are @peterbarker's.

This overlaps #7981: `common-firmware-update.rst` there still holds the same section, and would need trimming (or reducing to just its "Flying Stable ArduPilot Point Releases" section) so the text does not live in two places. No files are touched in common with #7981, so the two do not conflict in git.

### Testing

`update.py --fast` builds clean on copter, plane, rover and sub; no new Sphinx warnings.


rmackay9:
It probably makes sense to have a single page but it becomes a big page so I wonder if we should have an index at the top to allow users to more quickly jump to the section they need.

Hwurzburg:
It does....that's the content directive...got to build it to see

ES-Alexander:
Is the intent for this to cover _all_ firmware loading? If so, it seems to be missing any reference to Linux boards, which generally handle things differently (at least as I understand it, though admittedly I only have experience using Linux boards with BlueOS). Not sure if there are other board types that similarly don't follow the described process here 🤷‍♂️

Hwurzburg:
> Is the intent for this to cover _all_ firmware loading? If so, it seems to be missing any reference to Linux boards, which generally handle things differently (at least as I understand it, though admittedly I only have experience using Linux boards with BlueOS). Not sure if there are other board types that similarly don't follow the described process here 🤷‍♂️

no Chibios only...historically Linux boards assume a developer level knowledge and other than listing some Linux boardds, we dont ahve much in the wiki....contributions would be welcome, of course....

common/source/docs/common-loading-firmware.rst
```diff
@@ -0,0 +1,158 @@
...
+ArduPilot goes to some effort to make firmware upgrades seamless.  From time to time we need to move parameters around, change their scaling or simply remove them.  We do this automatically on first boot after the firmware is updated.
+
+We can't retain this parameter migration code forever, so there are limits to which versions we guarantee upgrades from and to.  You can still upgrade from older versions to newer versions, but you may find that some functionality changes unexpectedly as you are now using default parameter values rather than your customised values.
+
+If your autopilot falls far behind the modern stable versions you can still get the parameter conversions - you just need to flash intermediate versions - making it a multi-step process.
+
++--------------+--------------------------+
+| New version  | Oldest Migration version |
++==============+==========================+
+| 4.8(latest)  | 4.2                      |
```
peterbarker:
```suggestion
| 4.8(latest)  | 4.3                      |
```

peterbarker (dismissed):
The other numbers are still made up, but this one is correct

rmackay9:
Thanks for this.  I'm sorry but I'm not a big fan of this change because it moves the most common path that users will be on (e.g. loading ArduPilot onto a board that already has an AP bootloader on it) down at least one level (maybe two levels).

Previously users would most likely be greeted with this view which immediately shows them they first need to plug in a USB cable
image

With this change their instead presented with a complex looking list of links and many lines of text before an image is visible.  If they just scroll down they're most likely misled into thinking they need to manually download a firmware from firmware.ardupilot.org. 
image

Hwurzburg:
@rmackay9 restructured

rmackay9 (approved):
very nice!

PR #8003, opened 2026-08-27
dev: archive the Morse, SCRIMMAGE and AirSim simulator pages

### LINKED ISSUE #8001: Archive Dev Morse Simulator page
Morse isn't actively maintained and I haven't heard of anyone using it for a long time.  In order to reduce complexity let's move the page to the archive
https://ardupilot.org/dev/docs/sitl-with-morse.html

We should probably do the same for the SCRIMMAGE simulator https://ardupilot.org/dev/docs/sitl-with-scrimmage.html

Hwurzburg:
Fixes #8001.

None of these three belongs in the list of simulators a developer would choose from today:

- **Morse** — no longer maintained, and @rmackay9 reports no known users of the ArduPilot backend for it.
- **SCRIMMAGE** — the connection has not been maintained for a long time; its page already warned it may not work with current ArduPilot code.
- **AirSim** — Microsoft has archived the upstream project. Its page already carried a notice saying it was archived and unmaintained, yet the page was still listed among the *active* simulators and the notice claimed AirSim was "still available and supported".

### What changed

Follows the procedure in `common-wiki-editing-archiving` for all three:

- `Archived:` prefix added to each page title.
- The existing unmaintained notices replaced with a warning saying the topic is archived, pointing at maintained alternatives — Webots, Gazebo, and **PteroSim** where an Unreal Engine based simulator is wanted (which is the natural substitute for AirSim readers).
- All three removed from the simulator lists and the menu toctree on the Simulation page, and added to the `dev` section of Archived Topics — so each page stays in exactly one toctree and remains reachable via Appendix → Archived Topics.

Page content is otherwise untouched, so anyone still running one of these simulators keeps the instructions.

Prose mentions of AirSim in `gsoc-ideas-list` and `roadmap` are left alone — neither is a `:ref:` link, and `roadmap` is itself already archived.

### Questions for @rmackay9

The issue named Morse definitely and said *"We should probably do the same for the SCRIMMAGE simulator"* — I have done both, on the strength of the unmaintained warning already on the SCRIMMAGE page. AirSim was added at @Hwurzburg's request. Happy to drop any of the three if you would rather keep it in the active list.

### Testing

Clean `update.py --fast --site dev` build, zero Sphinx warnings. Verified in the rendered HTML that all three titles carry the Archived prefix, all three appear under Archived Topics, the archive warnings resolve their cross-references, and no references to any of the three remain on the Simulation page.

### DIFF
--- a/common/source/docs/common-archived-topics.rst
+++ b/common/source/docs/common-archived-topics.rst
@@ -127,6 +127,9 @@ value to users with old hardware.
     SITL Serial Mapping <sitl-serial-mapping>
     2020/2021 Roadmap <roadmap>
     Trouble Shooting Pixhawk4 Boot <troubleshooting-pixhawkpx4-boot>
+    AirSim Simulator <sitl-with-airsim>
+    Morse Simulator <sitl-with-morse>
+    SCRIMMAGE Simulator <sitl-with-scrimmage>
 [/site]
 [site wiki="copter,rover"]
 .. toctree::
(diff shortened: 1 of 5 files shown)


khancyr (dismissed):
I agree with this, but not to reference Airsim. Airsim is also dead and we should put it in archive too !

rmackay9 (dismissed):

rmackay9:
I don't know much about AirSim so I have no opinion on that but archiving scrimmage and morse sounds good to me, thanks!

Hwurzburg:
adding AirSim to the list

PR #8012, opened 2026-08-29
common: benewake: document I2C connection, fix TF-Luna range

### LINKED ISSUE #7949: Benewake TF02-Pro / TF03 / TFS20-L / TF-Luna / TF-Nova / TF350
[Benewake TF02-Pro / TF03 / TFS20-L / TF-Luna / TF-Nova / TF350](https://ardupilot.org/copter/docs/common-benewake-tf02-lidar.html) - The TF-Luna is capable of IIC communication. Reference https://en.benewake.com/uploadfiles/2024/04/20240426135946148.pdf section 6.3 for instructions. The parameters that I have set are:
RNGFND1_FUNCTION	0
RNGFND1_PIN	-1
RNGFND1_RMETRIC	1
RNGFND1_SCALING	3
RNGFND1_STOP_PIN	-1
RNGFND1_TYPE	25
This has allowed me to get data from them accurately.

Hwurzburg:
Fixes #7949, reported by @Strohtho.

The page described these lidar as serial and CAN only - "proprietary serial and/or CAN interfaces", with every Interface column entry reading UART or CAN. In fact every model on the page except the TF03 and TF350 can be connected over I2C, and ArduPilot has had drivers for that the whole time.

## Two drivers involved

- **TFS20-L** has its own: `RNGFND1_TYPE` = 46 (`BenewakeTFS20L`). This was documented nowhere on the wiki. Worse, the page's serial section told TFS20-L owners to set type 27, which is the TF03 *serial* driver.
- **The rest** share the TFmini Plus I2C command protocol (the `0x5A` command frames in `AP_RangeFinder_Benewake_TFMiniPlus.cpp`) and use `RNGFND1_TYPE` = 25 (`BenewakeTFmini-I2C`). This is what @Strohtho found working on a TF-Luna.

Both are `AP_RangeFinder_Backend_I2C` subclasses defaulting to address 0x10.

## Changes

- New "Connecting via I2C" section between the serial and CAN sections
- Interface column marks I2C on TF02, TF02-Pro, TF02-i, TFS20-L, TF-Luna and TF-Nova; TF03 and TF350 are unchanged as UART, CAN
- Opening sentence mentions I2C

## One thing deliberately not copied from the issue

The report lists `RNGFND1_PIN`, `RNGFND1_RMETRIC`, `RNGFND1_SCALING` and `RNGFND1_STOP_PIN` alongside the settings that worked. Those are analog-only - `ratiometric` and `scaling` are read in `AP_RangeFinder_analog.cpp` and neither I2C backend references them. Rather than documenting them as required, the section carries a note saying they do not apply to an I2C connection.

Parameter names and the type codes 25 and 46 were checked against a generated `apm.pdef.xml`, which is also where the exact GCS string `BenewakeTFmini-I2C` comes from.

## Also: TF-Luna range corrected

The table gave the TF-Luna as `0.80m / 3m`. [Benewake specify](https://en.benewake.com/TFLuna/index.html) "0.2 ~ 8 m" - the old figures look like the accuracy specification ("± 6 cm @ (0.2 ~ 3 m)") mistaken for the detection range. Corrected to `0.20m / 8m`. It is in the same table row this PR already edits, so it is folded in here rather than raised separately.

That same product page lists the TF-Luna's interfaces as "UART, I/O, I²C", which independently confirms the I2C support documented above.

Built with `python3 update.py --fast --site copter`.


rmackay9 (dismissed):
Ideally we would have wiring diagrams for at least one of the lidars capable of I2C but in any case this is an improvement, thanks!

Hwurzburg:
> Ideally we would have wiring diagrams for at least one of the lidars capable of I2C but in any case this is an improvement, thanks!

added

rmackay9 (approved):

rmackay9:
thanks very much!

PR #8033, opened 2026-09-07
common: add MechFlow V1 optical flow and rangefinder sensor page

mechinteldynamics:
Adds a new supported-sensor wiki page for MechFlow V1, a combined PMW3901 optical-flow + VL53L1X ToF rangefinder module that outputs MSP v2 over a single UART. No firmware changes are needed — it uses the existing generic AP_OpticalFlow_MSP / AP_RangeFinder_MSP drivers, the same protocol path already documented on the Mateksys 3901-L0X page, which this page is modeled on.

Disclosure: I'm the manufacturer of this sensor (MechIntel/AMID Systems).

Testing: Bench tested on a custom board (Vajra F4, custom hwdef target only — unmodified mainline ArduPilot Copter).

PR #8036, opened 2026-09-09
common: mechflow-v1: add indoor vs outdoor use warning

mechinteldynamics:
Follow-up to #8033 (merged). Adds a dedicated warning about the ToF rangefinder's reduced range/reliability in direct sunlight — a general laser ToF limitation, not specific to this unit — and clarifies that outdoor altitude hold in direct sun should use a separate sunlight-rated rangefinder as primary, with MechFlow V1 still useful at short range (e.g. landing/take-off) in that setup.


Hwurzburg:
thanks

PR #8018, opened 2026-08-31
Offline wiki

freddygaffey:
adresses #337 

The goal was to make the wiki faster, make it work offline, and make it exportable. This was done by adding modern web techniques to the existing Sphinx build; the pages themselves are unchanged.

You can try it here: https://offline-wiki.pebnum.com/dev/docs/wiki-offline-copies.html

**Everything reader-facing is opt-in.** A service worker is registered only after a reader turns on the *Offline mode* switch on the Offline page, or saves a wiki. A reader who does neither has no worker and browses exactly as today; the site-wide footprint of the PR on an ordinary page is the Offline menu entry, a manifest link and one script tag. A browser test in Chromium, Firefox and WebKit pins that a fresh visit ends with no registration.

What it adds:

- **A service worker** that intercepts requests and answers them from cached files: pages are served from cache and refreshed in the background, images and Sphinx's fingerprinted assets cache-first. Measured on the demo, a page went from about 1,000 ms to 13 ms, roughly 60x.
- **One archive per wiki** (common, copter, plane, ...), built by the build server after Sphinx as a `tar.gz` with a path-to-hash table. Saving a wiki is a single request rather than a crawl of thousands of pages, so readers cannot DoS the site by saving it. Updates fetch only the files whose hash changed.
- **Export to a single HTML file**, for machines where the browser may evict its cache, or when a portable file is simply preferred. It carries the wiki's navigation and full-text search, and is built in the browser from what is already saved.

The branch is ten commits in build order: `build_parameters` (fetch every ref), the archive builder and its tests, two passes over the built output (lossless PNG recompression, lazy YouTube embeds), the template hooks, the service worker, the offline page, the export, and the developer documentation. Commits 4 and 5, and the parameter-page layout rule in commit 6, are the only things that change output for readers who do not opt in.

**Tests:** `npm install && npm test` (panel, worker, exporter, archives, image passes) and `npm run test:browsers` (the same pages read offline in three real browsers).

For the details see `dev/source/docs/wiki-offline-copies.rst`, or https://offline-wiki.pebnum.com/dev/docs/wiki-offline-copies.html

PR #8037, opened 2026-09-09
Declare raw html images to Sphinx: fix the Rpanion logo, add a pre-commit check

freddygaffey:
Fixes the broken Rpanion logo on the commercial support page and adds a check so the same mistake cannot land again.

The page is a raw HTML table of `` tags. Sphinx does not read raw HTML, so every logo also has a hidden `.. image::` directive at the bottom of the file to make Sphinx copy the file into the build. The Rpanion entry was added without one, so every wiki asks for `_images/rpanion.png` and gets a 404. The logo also moves into `images/logos/` with the others.

Also adds a pre-commit hook that requires every `` inside a `.. raw:: html` block to have an image or figure directive of the same file name in the same file. It runs in the existing CI pre-commit job. It fails on the page as it is on master today and passes with the fix.

Requested in https://github.com/ArduPilot/ardupilot_wiki/pull/8018#discussion_r3947735498

Check the discussion link before posting. It should point at khancyr's "this need a check so it never happens again" comment, and I could not confirm the anchor from the API.


khancyr (approved):
LGTM
Working locally

PR #8055, opened 2026-09-12
4.8:33712/TFminiPlus power down

Hwurzburg:
Documents the RNGFNDx_PWRRNG power down support added for the I2C TFmini Plus in ArduPilot/ardupilot#33712 (master / 4.8 only).

- new "Optional Power Saving" section on the Benewake TFmini / TFmini Plus page
- notes the powered down status: distance is not shown on the OSD and consumers fall back to other sources while the lidar is off
- height above terrain source differs per vehicle, so copter/sub (terrain database only) and plane (terrain database, else baro above home or above the landing target while landing) are split with site shortcodes
- "Testing the sensor" underline changed from `=` to `-` so it is no longer nested under the I2C section


IamPete1 (commented):
It does also work for the pre-existing PWM rangefinder, just those are not very common anymore.

Hwurzburg:
there is only one listed in the wiki....

PR #8059, opened 2026-09-12
4.8:33842/QRTL pause time

Hwurzburg:
Documents the new Q_RTL_PAUSE_TIME loiter stage added in ArduPilot/ardupilot#33842 (master / 4.8 only).

- new bullet in the approach phase list: the vehicle holds position above the landing point for Q_RTL_PAUSE_TIME seconds ("Land pause started") before the landing descent, applying to both QRTL and AUTO VTOL landings
- default is 0 (no pause) except on tailsitters, where it defaults to 5 seconds to let the vehicle settle before descending
- QRTL Mode descent description updated to mention the pause
- approach.jpg gains a caption line noting the delayed land descent


IamPete1 (approved):
Thanks!

PR #8065, opened 2026-09-13
Fix Firefox offline downloads and preserve validated PNG caches

tridge:
Fix two problems with offline wiki saves and builds: Firefox can abort an archive download with `Error in input stream`, and clean builds discard the PNG optimisation cache and re-encode unchanged images.

For Firefox, leave same-origin `/offline/` requests to the browser network stack instead of forwarding their response streams through the service worker. This removes the download's dependency on the worker lifetime while preserving the existing routes for saved pages, images and exports. Existing offline archives remain valid.

For PNG optimisation, keep the cache in `<destdir>/offline.cache/images/` so it survives production's checkout cleanup. Reuse results for unchanged source and optimised bytes, and skip the image pass when Pillow is unavailable. Make persistent cache entries safe to reuse:

- Validate cached PNG structure, chunk CRCs and decoded pixels before publication; regenerate damaged or mismatched entries.
- Record “cannot shrink” only after a successful encode. Transient failures preserve the source and are retried later.
- Recheck legacy empty markers once while retaining valid prebuilt PNG cache files.
- Use unique temporary files, sync file data and directory updates on POSIX, and preserve web-readable image permissions.

Document HTTP exclusion and manual cleanup of the build cache. Automatic cache eviction is not added.

Validation:

- Reproduced the original download failure in Firefox 153 and 154. Both completed full Copter, Plane and Shared images saves through a local proxy serving the patched worker; the original worker also failed through that proxy.
- All 111 service-worker checks and 48 PNG checks pass, including corrupt cache recovery, transient failures, legacy markers, concurrent writers and write failures.
- The new `update.py` cache-directory wiring test passes with and without `--destdir` and runs in the normal CI test command.
- All 767 prebuilt compressed PNGs pass the new validation and remain reusable.
- All applicable pre-commit hooks pass.


tridge:
> **Deprecated — see below for the updated review.**

<details><summary>Previous review (2026-09-13, head ee77e0de91)</summary>

**Automated review note — AI-generated (Claude), validated against the live diff.** Please sanity-check before acting.

Reviewed at head `ee77e0de91`. This comment was first written for the on-demand review and was **updated 2026-09-14** after an independent re-review at the same head for the AIReview label. **Verdict: COMMENT — no blockers.** The change since the first version: the fix I suggested for issue 2 was not sufficient (see below), and a new NOTE 3 was added.

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_14_AIReview/devcall_pr_reviews.html#prwiki-8065

Moving the cache under `--destdir` works, and the new tests catch regressions in it. Because the cache now survives every clean build, two failure paths that used to last one night can now last indefinitely.

### ISSUE

**1. A failed encode is recorded as "cannot shrink", permanently.** `scripts/optimise_images.py:48-50` swallows every exception and returns the input, and `:94-103` then writes a `.unchanged` marker.
- Reproduced: with `Image.save` raising `MemoryError` once, a 960-byte image stayed at 960 bytes on every later run, although it shrinks to 422.
- Before this PR, `git clean -x` wiped that entry the next night.
- *Fix:* have `shrink_png()` signal failure (e.g. return `None`), and write a marker only when encoding succeeded.

**2. A cache entry is used without checking it.** `scripts/optimise_images.py:83-85` trusts the cached bytes, and `:107-108` publishes them whenever they are shorter.
- An entry cut to 0 bytes is published as a 0-byte image.
- **Correction to my earlier suggestion:** checking for the PNG signature is not enough. An entry cut to half length (211 of 422 bytes) still starts with the signature, gets published, and fails to decode.
- `_write_atomic()` (`:115-119`) never fsyncs.
- *Fix:* reject an entry that doesn't end with the IEND chunk (`b"IEND\xaeB\x60\x82"`), or decode it. Also fsync before the rename.

### NOTE

**3. (new) `_write_atomic()` uses a fixed `.pngtmp` temp name** (`:117-119`).
- Two concurrent writers can truncate each other's temp file before the rename, leaving a 0-byte entry. Reproduced with two processes.
- Nothing reaches this today:
  - `update.sh:108` holds `build.lck` around the only `update.py` call.
  - The image pass runs once, after all Sphinx builds finish.
  - CI and the Docker script use per-checkout caches.
- `tempfile.mkstemp(dir=path.parent)` makes it moot.

**4. The cache sits inside the web root.**
- `update.sh:222` passes `--destdir /var/sites/wiki/web`.
- The nginx rules documented in `wiki-offline-copies.rst` have no rule for `offline.cache/`.
- The content is public images, so this is clutter rather than a leak. I haven't seen the production config.

**5. The cache has no eviction.** Measured cold: 766 of 1,893 distinct tracked PNGs shrink, for 251.8 MB. Edited or deleted images leave their old copies behind.

**6. The `update.py` wiring is untested.**
- No test imports `update.py`, so reverting `update.py:1369-1370` still passes.
- The "production-style git clean preserves the cache" test passes a cache path outside the repo it cleans, so it cannot fail.
- `test_optimise_images.py` itself does run in CI via `npm test`.

### What was checked
- All 28 checks pass (Python 3.14.4, Pillow 12.1.1), and 6 mutations of the new logic are caught.
- The cache key is a SHA-256 of a version prefix plus the bytes, so edited images miss and duplicates share an entry.
- Clean builds leave `<destdir>/offline.cache` alone, and a missing Pillow returns early.
- Two independent Codex passes: the first raised 1, 3 (web root) and 4 and added 2 and 5; the second raised the temp-name race.
- CI: 4 passing.



</details>

tridge:
**Automated review note — AI-generated (Claude), validated against the live diff (Claude + Codex cross-checked).** Please sanity-check before acting.

Re-reviewed at head `be88e7e233` (you were last told `ee77e0de91`); my earlier comment above is superseded. Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_14_1624/devcall_pr_reviews.html#prwiki-8065

**Previous round: all six points are resolved, and reverting each code fix makes a test fail.** Verdict stays **COMMENT — no blockers**. The new cache validation makes clean builds much slower, and several of its checks have no test.

### Resolved
- **1. Failed encode recorded as "cannot shrink":** `shrink_png()` now returns `None` (`scripts/optimise_images.py:50,55`), and `run()` skips writing a marker (`:124-126`). After a one-off `MemoryError`, the next run shrinks the image.
- **2. Unchecked cache entries:** `_valid_cached_png()` (`:58-76`) checks the IEND chunk, the CRCs and the pixels. Empty, half-length, missing-IEND-CRC, bad-CRC and wrong-pixel entries are all regenerated.
- **3. Fixed temp name:** `mkstemp()` (`:159`) plus fsync. Two processes each wrote the same path 1,500 times: the old writer gave 3,396 incomplete reads, the new one none in 3.15M reads.
- **4. Cache in the web root:** I served the documented nginx block (`wiki-offline-copies.rst:421-422`) on nginx 1.28.3.
  - Every cache path returns 404, including percent-encoded and `..` forms.
  - `/offline/*` still returns 200.
- **5. No eviction:** manual cleanup is now documented (`wiki-offline-copies.rst:360-363`).
- **6. Wiring untested:** `TestImageCacheWiring` fails when `update.py:1369-1370` is reverted, and `npm test` runs it in CI.

### ISSUE
**A. Clean builds now spend minutes validating cache hits.** Each PNG cache hit runs `_valid_cached_png()` (`:112`), about 40-60 ms per hit: `verify()`, two full decodes and two RGBA conversions. The image write at `:146` then does a file fsync and a directory fsync.
- Warm run over the 1,893 unique tracked PNGs (766 hits): **0.6 s at `ee77e0de91`, 37-54 s at this head**. About 32 CPU s of that is validation and the rest fsync; Codex measured 64 s independently.
- CI's full build recompressed 3,211 PNGs. Production runs `git clean -x` every night, so each of those becomes a hit: roughly 2 minutes of validation per build (an estimate, not measured).
- *Fix:* when writing an entry, record its SHA-256; `shrink_png()` has just checked its pixels. On a hit, compare the hash and decode only on a mismatch or for an older entry. Don't fsync the build-tree image at `:146`: it is rebuilt every night.

### NOTE
**B. Several of the new checks are untested.** Each of these mutations leaves all 48 checks passing:
- removing `image.verify()` (`:68`);
- removing the `len(candidate) < len(data)` guard (`:112`);
- removing the RGBA comparison (`:74`);
- removing the directory fsync (`:169-174`);
- ignoring the existing file mode (`:156`).

The "bad CRC" case (`scripts/tests/test_optimise_images.py:285`) corrupts the IHDR CRC, which `Image.open` already rejects. A flipped IDAT CRC loads cleanly and only `verify()` catches it, so that is the case to add.

### Checked, no finding
- **The `frontend/sw.js:763` bypass loses nothing that used to work offline.**
  - Saved-cache keys never start with `/offline/` (`common_offline_unpack.js:204-225`).
  - `NEVER_STORE` (`sw.js:698`) already kept the worker from storing them.
  - The page's fetches of the manifest and file tables all handle a rejected fetch.
  - The cold Codex pass showed a lost `/offline/files/` fallback only with a hand-seeded cache entry that no code writes, so I dropped it.
  - Not tested in Firefox here.
- **Tests:** 48/48 PNG checks and the wiring test pass. `checkDownloadBypass` passes 4/4, and removing the bypass fails all 4. CI: 4 passing.
- **Corpus:** of 1,893 unique PNGs, 766 shrink, 1,125 get a marker and 2 fail to decode.
  - The 2 are AVIF files named `.png` (`images/Teravolt/StrobeIT/StrobeIT_2.png`, `images/Teravolt/AeroNav-1/AeroNav-1_3.png`), retried each build at a cost of milliseconds.
  - No image fails the pixel check, so no image is re-encoded on every build.

PR #8068, opened 2026-09-14
Common: dronebridge update for multi-vehicle setups

rmackay9:
This updates the DroneBridge page's multi-vehicle setup section to use the newly added Air-To-Air feature.


Hwurzburg:
thanks

PR #8070, opened 2026-09-14
OFFLINE-WIKI: bug fixes: Play saved-page videos from the host the live pages use

freddygaffey:
A saved page's video card became a player from youtube-nocookie.com, while the live pages embed from www.youtube.com. Some networks filter the nocookie host and let youtube.com through, and on one of those a saved page kept its "needs a connection" card while online, next to a live page that played. The card now becomes the same player the live page carries, and the reachability probe asks that host. A jsdom suite loads pwa.js with a card and checks the player's host, the Vimeo path, and that an unreachable host leaves the card alone.


tridge:
**Automated review note — AI-generated (Claude), independently cross-checked by a second model and re-verified against the live diff.** Please sanity-check before acting.

Reviewed at head `90f40cd5d0`. Full report: https://uav.tridgell.net/DevCallReviews/2026_09_15_AIReview/devcall_pr_reviews.html#prwiki-8070

**Verdict: APPROVE — no blockers.** What I checked:
- **Live pages:** they really do embed from `www.youtube.com`, so saved pages now match. The pinned sphinxcontrib-youtube fork only uses youtube-nocookie when `privacy_mode` is set, and no page sets it.
- **The new jsdom test:** it fails against the pre-fix `pwa.js`, and against three mutants (host change reverted, probe result ignored, Vimeo branch removed). `test_offline_page.js` still passes.
- **In a browser:** in headless Chromium, a served saved page played the real video. Over `file://` both hosts give Error 153, but `pwa.js` only runs on served pages.

Two optional notes, neither caused by this PR:
- `frontend/js/pwa.js:645`/`653`: the saved player ignores the start time. The card keeps `&t=Ns`, but the embed starts at 0:00, and 16 `.. youtube::` directives use `?start=`. Reading `t=` and adding `?start=N` would fix it.
- `scripts/tests/test_pwa_video.js` isn't in the test-suite list in `dev/source/docs/wiki-offline-copies.rst`.

PR #8074, opened 2026-09-15
OFFLINE-WIKI: bug fixes - Firefox cached storage, replace good page with server error, UI clarity with persistence, refresh during saving

freddygaffey:
Bugfixes
@tridge  Can you AI review this please.

1. A server error replaced a good saved page. With the worker on, a cache-first asset (theme CSS, an image, a fingerprinted script) that came back as a 404 or 503 during a publish window was handed to the page as-is, even when the reader held a working copy in a saved wiki. That is the blank-images symptom you saw on ardupilot.org during a rebuild. Now a non-OK response on that path is answered from the saved copy when one exists, and nothing is stored. With nothing saved, the error passes through as before.
2. "Saving" sat at 100 percent while the panel was still checking. After the last byte of an archive arrives, the panel fetches the file table and checks every entry against it, which on a slow device takes long enough to look stuck. The row now shows "Checking Copter…" at 99 percent for that phase and only reaches 100 percent and "Saved" once the completeness marker is written.
3. The storage line said "temporary" after the browser had granted persistence. The save asks for persistent storage but only waits 1.5 seconds for the answer, because Firefox turns the request into a prompt. If the reader clicked Allow after the save finished, the grant was recorded but the panel kept saying temporary until the next update tick. The late answer now re-renders the storage line at once.
4. A refresh cancelled after it had started writing stayed marked complete. Cancelling a refresh of an already saved wiki part-way through the archive left the old completeness marker in place over a cache that was now a mix of old and new entries. The cancel now unmarks the cache and tells the worker its caches changed, so the wiki shows "Incomplete, save again" rather than being served half-updated.


freddygaffey:
@tridge 
I have been asked to write a blog post about the offline wiki I am writing it as if this has been merged in can you please review it before merging.

AP-Review:
**Automated review note — AI-generated (Claude), cross-checked by a second independent Claude pass against the live diff.** Please sanity-check before acting.

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_22_AIReview/devcall_pr_reviews.html#prwiki-8074

Reviewed at head `c3c25853b0`. **Verdict: COMMENT — nothing here is a correctness defect you introduced.** The three fixes that do ship are real and each is backed by a test that both passes confirmed bites. The substantive problems are in the **PR description**, and the obvious fix for the one real asymmetry was measured to be wrong, so it is raised below as a question rather than a prescription.

### 1. Body item 4's fix is already in master; this PR adds only the test

The body says *"A refresh cancelled after it had started writing stayed marked complete … The cancel now unmarks the cache and tells the worker its caches changed."* That recovery block (`common_offline_page.js:1110-1119` at this head, guard at `:1111`) is present verbatim at the merge-base and landed in `87dafeadb` — `git log -S "err.apVerify || wrote"` finds it and `git merge-base --is-ancestor 87dafeadb 21cd49633` succeeds. Confirmed by revert: with both sources at the merge-base and your tests kept, **both cancel checks pass**, and the only three failures are the other items. The new test does add real incremental cover (the `AbortError`-with-`wrote` path, distinct from the `apVerify` paths already covered), so keep it — your commit message is honest and does not claim item 4; only the body does.

### 2. Body item 1's "an image" is not what the code changes

`frontend/sw.js:589-598` already returns the held copy for any request with **no query string**, *ahead* of the fetch. Probed at both heads: `_images/frame.png` with the server at 503 gives the saved copy at both, with `seen.fetches` **empty** at both — the network is never asked. The `?v=abc` variant gives 503 at the merge-base and the saved copy at your head. So the new block changes behaviour only for **fingerprinted** assets (theme CSS and scripts), which is exactly what the new test covers — and Sphinx `_images/` URLs carry no query. Worth narrowing the claim before it becomes the accepted explanation of the blank-images report, and worth re-checking which asset actually failed (likely a `_static` one). Loose wording rather than wrong.

### 3. A question about the 404/410 asymmetry — and please do **not** take the obvious fix

The new guard at `frontend/sw.js:608` is `!response.ok`, which includes 404 and 410, while elsewhere in the same file a 404/410 is authoritative removal: `maybeRevalidate` deletes the entry at `:568-569` under the comment *"The page is gone upstream; serving it forever would be lying"*, and `staleWhileRevalidate` evicts at `:457-462`. The behaviour change is real, measured through your own `bootWorker` harness against both heads (fingerprinted `_static`, old-build copy saved): 404 → **200 saved copy** at your head against **404** at the merge-base; 410 and 503 likewise; with an incomplete saved wiki, 404 at both.

**But the consequence I was going to raise is refuted.** Against the live server, `theme.css?v=BOGUSFINGERPRINT` returns **200** and `documentation_options.js?v=OLDBUILDHASH` returns **200** — ardupilot.org ignores the fingerprint query and serves old fingerprints rather than 404ing them. A 404 needs the *path* to be gone, and Sphinx static paths are stable; the only ways in are a mid-rsync window or a theme upgrade dropping a file, and in the latter the requester is the reader's own saved old-build page, for which the old bytes are correct.

**And the fix is worse than the problem.** Patching the guard to `response.status >= 500` was tried: 503 still falls back but **404 reverts to passing through** — which deletes half of your own stated fix, since the body says the asset *"came back as a 404 **or** 503 during a publish window"*. It also lands undetected: under that patch the new worker check still passes 3/3 and `test_offline_page.js` still passes 281/281. So: is the asymmetry against `:568`/`:457` deliberate for fingerprinted assets? If so, a one-line comment saying why would stop the next reader filing it.

### 4. Cancel is inert during exactly the phase this PR makes visible

`activeDownload.signal` goes only to `ApUnpack.fetchArchive` (`:1018`). The table fetch at `:1029` has **no `signal`**, and neither the prune nor `storeTable` is abortable. Reproduced by holding `copter-files.json` open: the panel shows `Checking Copter…` with badge `Saving`, bar `checking`, button `Cancel`; pressing Cancel does nothing; and when the table is released the save **completes and the marker is written**. Not a regression — the merge-base has no `signal` there either — but what is new is that this PR *names* the phase ("on a slow device it is long enough to need saying"), turning a hidden window into a visibly non-functional Cancel. Fix: pass `signal: activeDownload && activeDownload.signal` at `:1029` and re-check after `storeTable`, or stop labelling the button Cancel for that phase.

### Notes

- **The "everything else" route has the same shape and is untouched** (`:857-869`) — `searchindex.js` and `objects.inv` return the server's error verbatim on non-OK and consult `heldOffline` only in the `catch`. Probed at 503 with a complete saved copy: both return **503 at both heads**, while the same request with the network *throwing* returns the saved copy at both. **But it is a scope gap rather than something you broke:** that route is deliberately network-first (no cache read on the success path; `searchindex.js` shares one URL across builds), and your body scopes item 1 to a cache-first asset. Mirroring the block there would make a saved build-N index answer a 503 on a live build-N+2 site. For reference the pattern is already established for a different index — `paramIndex` gets it right at `:376` with the `if (fresh && fresh.ok)` at `:379`.
- **Redundant second sweep at `:609`.** For a non-fingerprinted request `heldOffline()` has already run at `:590` and returned falsy; `:609` repeats the whole thing — a fresh `caches.keys()` sweep plus an `isComplete` read and up to three `cache.match` calls per cache. Measured cache-read counts: 3 against 2 with nothing saved, **8 against 5** with one complete saved wiki lacking the path. Gating the block with the `fingerprinted` flag already in scope restores the merge-base counts exactly and still passes 3/3 and 281/281 — tested, not just suggested. Two negligible caveats: a wiki becoming complete between `:590` and `:609`, and a cross-origin `THIRD_PARTY_CACHE` request whose path matches a stored one.
- **`rowProgress(entry.id, 99, 'checking')` puts "Saving" and "checking" side by side** — `:702-708` is the `percent < 100` branch, so the badge reads `Saving` while the bar reads `checking`, and only the status line says "Checking Copter…". Reproduced in the jsdom harness. The lowercase bar text does match house convention (`'done'` beside badge `Saved`) and your test deliberately asserts only that the badge is not `Saved`. **And it cannot get stuck:** the worst case — a failed *refresh* of an already-saved wiki, where `storedIds[id]` is still true so the bar is not cleared — is repainted from truth by `renderWikis()`. If the row is meant to say it is checking, `rowProgress` wants a badge label for that state.
- **Dropped after checking:** the floating `renderStorage()` promise at `:980`. The mechanism is real (`asked` cannot reject, but `renderStorage()` has no internal catch and `caches.keys()`/`estimate()` can) but it is **not introduced here** — the merge-base already has three floating `renderStorage()` calls. The button-state risk was checked and is genuinely absent (`updateSaveState`/`updateExportState` bail early on `activeDownload || activeExport || exportBusy`; `clear.disabled` includes `!!activeDownload`); `updateTotal()` does not bail, but `renderStorage` already runs mid-download at the merge-base via `dropFoldedCaches(entry).then(renderStorage)`, so `:980` adds no new class of mid-download re-render. The Firefox question behind that fix could not be checked empirically here (no browser engine) but is not open: the Storage Standard sets the box mode to persisted *before* `persist()` resolves true.

### Checked and clean

All three shipping fixes have biting cover, confirmed by mutation and revert: deleting the new `sw.js` block makes "a fingerprinted asset the server answers 503 is served from the saved wiki" fail with `status 503`; reverting both sources to the merge-base while keeping your tests gives **278 passed / 3 failed**, and those three are exactly the new behaviours. Full suites at this head: `test_offline_page.js` **281/0**, `test_pwa_video.js` 5 passed, and the python checkers all pass; `test_offline_worker.js`, `test_offline_browsers.js`, `test_offline_export.js` and `test_offline_archives.py` fail or abort **identically at the merge-base** (missing archives / playwright binaries), so environmental. CI green. Zero trailing whitespace on added lines. The new guard correctly excludes `type === 'opaque'` so cross-origin no-cors responses are not mistaken for errors (`opaqueredirect` would slip through, but `cacheFirst` only serves images and static with the default `redirect: 'follow'`); the error response is discarded rather than stored; the held copy is returned through `inflate()`; and `heldOffline` skips caches without a completion marker, so a half-written download cannot answer. Moving `rowProgress(…, 100, 'done')` to after the `COMPLETE_MARKER` put is the right place — the recovery `.catch` sits before it, so a failed verification cannot reach the "done" state.

One last thing: this is a single commit spanning two subsystems (the service worker and the save panel), and since both of the substantive items above sit in the `sw.js` half, a per-subsystem split would make that half reviewable and revertable on its own.

<sub>Reviewed by two independent Claude passes. Codex was unavailable this run (usage limit until 2026-09-27), so the second pass was a fresh-context Claude agent rather than a different vendor.</sub>

PR #8080, opened 2026-09-18
 OFFLINE-WIKI: parameter versions as zstd deltas, single-copy saved wikis, honest video cards, panel status, parallel tests

freddygaffey:
@tridge  Can you AI review please
The TLDR;
- THIS INSTALLS PACKAGES ON THE WIKI BUILD SERVER update.sh now pip-installs requirements.txt we add the zstandard Python package
- We now download all the wiki params: compressed, they come to about 10 MB for all wikis.
- Some bug fixes in the caching and video loading.
- The top bar's Offline item turns green when offline mode is on and the worker is running. The exact colour, and what it should show in the in-between states, still needs to be refined.

Follow-ups to the offline wiki (#8018), tested on the review mirror and on a Windows laptop whose link to ardupilot.org stalls. It includes the four fixes of #8074; close that one or merge it first, either is fine.

## Parameter versions as zstd deltas

Each vehicle builds a full parameter list for every release, 4 to 6 MB apiece and nearly identical from one to the next. Until now the Offline page offered them one by one, each a 300 KB gzipped download.

- Every 4.x stable and the newest beta travel inside each wiki archive: the newest stable as a plain page, the rest as zstd deltas against it, 20 to 40 KB each. On the live site that is 8 to 10 MB for every vehicle against about 60 MB, and it arrives with the archive.
- The build writes each delta with the Python zstandard module (added to requirements.txt). update.sh now installs requirements.txt at the start of every run, before the theme step, never upgrading a satisfied pin and never fatally, so a module a change adds is in place for the first build that needs it. Without the module the archives build with no historical versions and say so.
- Two decoders behind one surface, vendored in frontend/js/zstd-delta.js: zstd compiled to WebAssembly (2 to 4 ms for a 4 MB page) and fzstd, pure JavaScript, taken when WebAssembly is unavailable or the wasm will not load (about 150 ms). Every delta header carries the content hash of the page it rebuilds and both decoders are checked against it. A versioned page is sniffed for the delta magic, so it rebuilds whichever version of the code stored it and however it was marked, and a header from before the hash was added still reads.
- The Offline page's per-version picker is removed; a row says how many versions its archive carries. If even the JavaScript decoder cannot run, the page offers one button that fetches the plain pages instead. The single-file export carries deltas as deltas with the decoder embedded once and rebuilds a version when it is opened, so the file offers every saved version.

## Saved copies

- A saved wiki's pages and files are served in place and never copied into the browsing caches: one copy, kept current by the update check, no promotion or eviction pass, no refresh fetch per view of a saved page. Pages merely browsed keep their stale-while-revalidate refresh and the reload prompt.
- A fingerprinted static file whose saved copy matches its crc32 fingerprint is served at once; before, it waited for the network every time. The match is remembered for the worker's life.
- Every network wait a stored copy can cover is capped at the page route's 5 s; when the browser reports offline and a copy exists the network is not tried first. With nothing stored the network is still waited for.
- A 5xx or 404 during a publish window no longer replaces a good saved copy.
- A check finishes an interrupted refresh (a cache that lost its marker but was complete before, or the shared images any saved wiki needs) instead of reporting up to date beside an incomplete row. A first save the reader cancelled is left for them to decide.

## Panel

- The site's top-menu Offline item is green on every page while the worker is active, with the last check in its tooltip.
- "Offline mode: on" is green only once the worker registration is active, and says "starting" in the gap after opting in.
- Beside the Wiki header: when the saved wikis were last compared with the site and whether that found updates. Stamped only by a check that fetched the manifest and compared every saved wiki's build.
- A "Checking" phase during a save, a late persistence grant recorded, a cancelled refresh unmarked.

## Video cards

The still becomes the player when the video host answers, regardless of navigator.onLine (wrong on Windows behind some adapters); a click while the flag says offline probes the host first; the card reads "Video not loaded", "Connecting to YouTube", or "Video could not be loaded", and never looks like a player.

## Tests

- npm test runs every suite in parallel, one process each, the page suite in four shards (about 30 s on six cores); npm run test:browsers runs Chromium, Firefox and WebKit one per process.
- New coverage: build-side unit tests; the archive check that every delta names its base, carries the page hash and rebuilds; decoder tests for both paths and the fallback; an upgrade-compatibility matrix in the page, worker and export suites (marked or unmarked entry, hashed or hashless header, either decoder, a real page at a versioned path, an update replacing an old-format entry, an export over a mix); stalled-network, fingerprint and single-copy checks in the worker; fallback, panel status and video card tests.
- Each new check was proven with a mutation run. Full build, all suites and all three browser engines pass, and the differential update was validated end to end in Chromium against a changed build: two changed files fetched loose, no archive, the new page and the rebuilt version served offline afterwards.


AP-Review:
> **Deprecated — see below for the updated review.**

<details><summary>Previous review (2026-09-18)</summary>

**Automated review note — AI-generated (Claude), validated against the live diff.** Please sanity-check before acting.

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_19_AIReview/devcall_pr_reviews.html#prwiki-8080

Reviewed at head `88c750e256`.

### Verdict: REQUEST CHANGES

New to the report this run; ~6000 changed lines across 36 files, mostly JavaScript and Python tooling rather than ReST. The vendoring is unusually clean and is worth saying so: `zstd.wasm` is byte-identical to `@bokuweb/zstd-wasm@0.0.27`, the glue and the fzstd copy each differ from upstream in one and two trivial hunks, and both added licence files name the exact versions vendored. What blocks it is a swallowed-error path in the exporter that desynchronises the page index from the HTML block ids — after one failed rebuild, every later page in an exported single-file wiki renders the *previous* page's body — plus a trivial ruff failure. Underneath that sits a coherent theme worth fixing together: the JavaScript decoder verifies neither the zstd frame checksum nor a content hash, and there are two paths where it is the only decoder in play.

CI at this head: ✗ 1 failing, 4 passing.

### Findings

- **BUG** — `common/source/_static/common_offline_export.js:270` — **A swallowed delta failure shifts every later page's block id, so the exported file shows the wrong page.** The pages loop uses the **pages-array** index `i` for block ids (`DOC.deltaBlock(i,...)` at `:269`, `DOC.pageBlock(i,...)` at `:296`, both emitting `id="p"`), but the viewer resolves a path through `byPath`, built in `common_offline_document_builder.js:80` as `D.pages.forEach(function(p,i){byPath[p.p]=i;})` — the **payload** index — and then does `document.getElementById("p"+i)` (`:242`, `:256`). Those agree only if every entry of `pages` pushes exactly one `index` entry. On master the loop had no `catch`, so a read failure rejected the whole export; this PR adds the first swallowing handler, which does `done++` with no `index.push` while `i` keeps advancing. After one swallowed failure at `pages[k]`, payload position `k` maps to `pages[k+1]`, which wrote `id="p(k+1)"`. **Reproduced by exporting**: the first page after the omitted delta resolves to a **missing block**, and the next displays the preceding page's body under the correct title and breadcrumb; later delta and base lookups break too. Reachable whenever `ApUnpack.readFrom` rejects for a versioned parameter page: base page missing from the saved cache, malformed delta, or hash mismatch (`common_offline_unpack.js:335`). The `rawBlock(j,...)` id at `:311` is a pages index too, read back as `"r"+byPath[pg.b]`, so it rides on the same broken invariant. Fix: key the blocks on the payload position. In both branches capture `var slot = index.length;` immediately before the `index.push`, pass `slot` to `DOC.deltaBlock`/`DOC.pageBlock`, and for the base compute its slot by scanning `index` for `pg.p === basePath.replace(/\.html?$/, '')` instead of scanning `pages` (`:307`).

- **ISSUE** — `frontend/js/zstd-delta.js:1186` — **The JavaScript decoder never checks the zstd frame checksum the build writes, and returns a silently wrong page about 38 % of the time under corruption.** `scripts/build_offline_artifacts.py`'s `delta_against` sets `write_checksum=1`, and the fixture frame carries it (frame header descriptor `0xa4`: single_segment, **checksum**, no dictID). `patchJs` loops `while (!st.l)` and returns; it never reads `st.c` and never verifies the trailing XXH64. The only guard is the length check at `:1200`. Measured by running both decoders out of this branch against `scripts/tests/fixtures/delta-page.zst` + `delta-base.html`: corrupting **only the 4 trailing checksum bytes** gives wasm `zstd delta did not apply (code -22)` while JS returns 232285 bytes byte-identical to `delta-page.html` with no error; across 600 random single-bit flips in the compressed payload (skipping the 9-byte header and 4-byte trailer) wasm threw 593-594 and returned a wrong page **0** times, while JS threw 366-368 and returned a **silently wrong page 226-227 times** (both passes ran this independently with different seeds and agree to within a few counts). `sw.js:392` and `common_offline_unpack.js:335` do catch this with the `APDELTA1` content hash, so the two main read paths are safe — but see the next two findings for where that backstop is absent. `test_zstd_delta.js:73-74` already documents the gap in a comment.

- **ISSUE** — `common/source/_static/common_offline_document_builder.js:246` — **The exported single-file document rebuilds deltas with the JS decoder and checks nothing.** To scope it precisely, per the second pass: the exporter *does* verify the original entry on the way out, through `ApUnpack.readFrom` — the gap is the **subsequent reconstruction inside the exported file**, where running the real `showDelta()` against a corrupted fixture painted incorrect content with no error. `ApZstd.init(null)` forces the JavaScript decoder — reasonable, since a single HTML file cannot load a `.wasm` — which per the finding above has no checksum. And no content hash is checked either: `common_offline_export.js:265-269` has `delta.hash` in hand from `ApUnpack.deltaHeader` and discards it, carrying only `delta.frame`. `dev/source/docs/wiki-offline-copies.rst`'s new paragraph states "Either way the rebuilt page is checked against the hash in the header before it is served" — that sentence is not true of the export it describes two sentences later. Fix: put `delta.hash` into the index entry alongside `d`/`b` and have `showDelta` compare the first 8 bytes of `crypto.subtle.digest('SHA-256', out)` before painting, the same derivation as `contentHash` in `common_offline_unpack.js:191-198`, which matches `hashlib.sha256(data).hexdigest()[:16]` at `scripts/build_offline_artifacts.py:350`.

- **ISSUE** — `frontend/sw.js:318` — The header hash is optional — `if (hash !== undefined && !/^[0-9a-f]{16}$/.test(hash))` here and `if (head.hash && ...)` at `common_offline_unpack.js:188,335` — with a comment saying "Written before the hash was added: rebuilt, but not checked against it." The hashless form is deliberately accepted for compatibility, and the second pass narrows the consequence correctly: **wasm still verifies a checksum-bearing frame**, so the zero-verification case is the **JS fallback** rather than every reader. Confirmed by running the real unpacker with JS selected: a hashless entry against a wrong base returned HTTP 200 carrying incorrect content. If an unhashed `APDELTA1` ever shipped, bump the magic to `APDELTA2` and require the hash there; if it never shipped, just require it.

- **ISSUE** — `frontend/js/zstd-delta.js:1217` — The output allocation is driven unchecked by the delta's own frame header. `size i32`, so those come back as −1/−2), but there is no upper bound: a frame declaring up to `0x7FFFFFFF` drives a 2 GB `malloc` whose return is not checked before `_ZSTD_decompress_usingDict(dctx, out, size, ...)` writes to it. `src`/`dict` at `:1211`/`:1213` also leak if the `HEAPU8.set` between them throws, since the `try/finally` only opens at `:1220`. Contained inside the wasm sandbox, but an archive is something a reader can be handed by a third party. A bound from the code's own constants would be `sw.js:838`'s `STORE_LIMIT_BYTES = 12 * 1024 * 1024`; the new `wiki-offline-copies.rst` text puts real pages at 4-6 MB, so `if (size  STORE_LIMIT_BYTES) throw` plus `if (!out) throw` covers it. The JS side has the same unbounded allocation at `:1191` but raises `RangeError` rather than corrupting anything.

- **ISSUE** — `scripts/tests/test_update_build.py:58` — CI is red on `pre-commit` for a trivial lint: ruff `E741 Ambiguous variable name: 'l'`, three occurrences, at lines 58, 73 and 89. **"Everything else in pre-commit passed" is not true** — flake8 fails on the same three lines, and the executable-shebang hook fails for `scripts/tests/run_all.js` and `scripts/tests/test_pwa_nav.js`. The other PR checks do pass, and `build-linux (3.10)` (27m36s), `build-docker`, `shellcheck` and `Trigger` are all green — note `build-linux` does run `npm test`, which is now `run_all.js`, so the new suites really are exercised in CI.

- **NOTE** — `frontend/sw.js:685` — The fingerprint CRC does not match Sphinx's, latently. Sphinx computes `?v=` as `f'{zlib.crc32(content):08x}'` where `content = file.read_bytes().translate(None, b'\r')` — **carriage returns stripped first** (sphinx 9.1.0, `sphinx/builders/html/_assets.py:130,135`) — while `crc32()` here runs over the stored bytes verbatim. Demonstrated: for `b"body {\r\n  color: red;\r\n}\r\n"` Sphinx emits `670a9b73` and this code computes `7f6dff25`. Scanned and **none** contains a CR byte — 16 repository `_static`/frontend JS and CSS files, 146 installed Sphinx files and 11 theme-checkout files — so it is correct today. One gap: no installed `sphinx_rtd_theme` package was present on the second pass's machine, so the theme half rests on the checkout rather than the installed package. If one ever acquires CRLF, `heldMatchingFingerprint` returns undefined and `cacheFirst` falls through to the network — a 5 s stall (`NETWORK_TIMEOUT_MS`, `sw.js:54`) per such asset, with the saved copy never used while online. One-line fix: strip `0x0d` before the CRC.

- **ISSUE** — `frontend/js/ZSTD-LICENSE:3` — **The zstd licence material is incomplete.** The file supplies attribution and links but omits the BSD redistribution conditions and disclaimer that the upstream zstd licence requires distributors to reproduce, and the bokuweb `master/LICENSE` URL it references returns 404. Include the applicable licence texts alongside the vendored assets. (`FZSTD-LICENSE` is fine — it matches the published package exactly.)

- **NOTE** — `frontend/js/zstd-delta.js:378` — The vendored fzstd no longer runs in strict mode. A `"use strict"` directive only takes effect as the first statement of the function body, and the inserted `var exports = {};` demotes it to a no-op expression statement. Upstream fzstd runs strict. Swap the two lines.

- **NOTE** — `scripts/tests/run_all.js:10` — Three CLI flags silently do the wrong thing. The docstring documents `--all`, but `:66` tests `only === 'all'`, which comes from `--only all` (what `package.json` uses), so `--all` silently runs just the unit suites. `--page-shards` with no value gives `Number(undefined)` = `NaN`, so `Array.from({length: NaN})` is `[]`, the largest suite vanishes and the run still reports "all suites passed". `--jobs` with no value gives `jobsMax = NaN`, so `running.size < NaN` is false and the pump never starts — and here the second pass corrected the first: it does **not** hang, because an unresolved Promise does not keep Node alive. It prints `NaN at a time`, runs no tests, and **exits 0** — a silent pass, which is arguably worse than a hang. A `Number.isFinite` guard falling back to the default in each `opt()` covers all three.

- **NOTE** — `update.sh:203` — Two stacked soft failures can publish archives with no historical parameter versions and only a log line. `python3 -m pip install --user -q -r wiki-requirements.txt || progress "requirements install failed; ..."` is deliberately non-fatal, and `add_param_versions` (`scripts/build_offline_artifacts.py:504-507`) catches `ImportError` on `zstandard` and returns `[]`. On a host where `pip install --user` hits PEP 668 both layers absorb it and the build succeeds having quietly dropped the feature this PR exists for. Suggest failing the build when `zstandard` is missing **and** `param_versions_for()` found versions to carry. (The production build host could not be checked from here, so the outcome is unconfirmed; the code path is verified.) Relatedly, `:509-510` drops every version with no `log()` at all when a wiki has betas but no stable — unlike the `ImportError` path, which does log.

- **NOTE** — `frontend/js/pwa.js:827` — `window.open` after an async probe is likely popup-blocked. The new capture-phase click handler calls `event.preventDefault()` and then opens inside `hostReachable(href).then(...)`, i.e. after a network round-trip and outside the user-gesture task. Chrome's 5 s transient activation window may let it through; Firefox and Safari generally will not, in which case an external link clicked while `navigator.onLine === false` but the host actually reachable does nothing at all — strictly worse than the error page it replaces. Not tested in a real browser. The usual workaround is to open the window synchronously and navigate or close it on the probe's result.

- **NOTE** — `frontend/sw.js:14` — `importScripts` injects roughly 70 unscoped globals into the worker (`Module`, `err`, `key`, `buffer`, `init`, `run`, `abort`, `quit_`, `HEAPU8`, ...). All 59 top-level names in `sw.js` were diffed against them and there is **no collision today** — but `err` and `buffer` are names a service worker plausibly reaches for, and because `importScripts` runs before `sw.js`'s body, a future top-level `const err` would be a redeclaration `SyntaxError` that kills the entire worker, not just the decoder. Worth a comment at the import site, or wrapping the glue.

<sub>Both a primary read and an independent second pass were run; where the two disagreed the disagreement is stated in the report rather than averaged away. Anything marked *unconfirmed* was not reproduced — please push back on it.</sub>


</details>

AP-Review:
> **Deprecated — see below for the updated review.**

<details><summary>Previous review (2026-09-19)</summary>

**Automated review note — AI-generated (Claude), cross-checked by an independent Codex pass against the live diff.** Please sanity-check before acting.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_19_1329/devcall_pr_reviews.html#prwiki-8080

### Re-reviewed at head `9009370894`: **REQUEST CHANGES → COMMENT.** Thirteen of the fourteen findings are closed.

My earlier comment above is superseded. Taking the previous round's 14 findings in order, because that is what you are waiting to read.

A note on sizing first, in case it comes up elsewhere: the branch was rebased, so the three-dot compare reports the whole PR. Against the two merge bases (`8f87afecb2` → `fb4c010d98`) your real delta is **1000 diff lines across 18 files**. And CI went from **1 failing / 4 passing** to **5/5 green** — `Trigger`, `build-docker`, `build-linux (3.10)`, `pre-commit`, `shellcheck`.

#### Resolved

1. **BUG — the export block-id shift.** Fixed, and fixed exactly the way it needed to be. `common_offline_export.js:263-264` and `:286-287` now capture `var slot = index.length;` immediately before each `index.push` and pass `slot` to `DOC.deltaBlock`/`DOC.pageBlock`; `:308-316` finds the base page's slot by scanning **`index`** for `e.p === bare` rather than scanning `pages`; and the viewer still builds `byPath` from the payload index (`common_offline_document_builder.js:80`), so the two now agree by construction and a swallowed delta can no longer shift every later block. The `leftOut` counter (`:235`, `:272`, `:355`) makes the omission visible instead of silent, and the comment at `:260-262` states the invariant where the next person will look for it. I checked the one thing that could still be wrong: `:311` keeps the *last* `e.p === bare` match, which would matter if two index entries shared `p` — they cannot, because a delta entry's `p` is the versioned path and the base's is the bare one.

2. **ISSUE — the exported single-file document checked nothing.** Fixed in both halves. `delta.hash` is now carried into the index entry (`h: delta.hash`), and the exported viewer checks it:
   ```js
   if(!pg.h||!(window.crypto&&crypto.subtle))throw new Error("no hash to check the rebuilt page against");
   return crypto.subtle.digest("SHA-256",out).then(function(d){
   if(hex16(d)!==pg.h)throw new Error("the rebuilt page does not match its hash");
   ```
   `hex16` emits 8 bytes as 16 hex characters, which matches `contentHash` and the Python `hexdigest()[:16]`. It **fails closed** when there is no hash or no SubtleCrypto, which is the stronger of the two available choices and more than I asked for.

3. **ISSUE — the header hash was optional.** Now required and refused outright: `frontend/sw.js:388` and `common_offline_unpack.js:327` both `throw` on a hashless entry. You took the "just require it" option rather than bumping the magic, which is the simpler correct answer.

4. **ISSUE — the wasm output allocation was unbounded and leaked on throw.** Both halves fixed at `zstd-delta.js:1221-1248`: one `alloc()` helper that throws when `malloc` returns 0, `size <= 0` refused, `size > MAX_PAGE_BYTES` refused with `:1148` naming `sw.js`'s `STORE_LIMIT_BYTES` as where the number comes from, and a single `try/finally` freeing `out`, `src` and `dict` — which closes the leak on a throwing `HEAPU8.set` as well.

5. **ISSUE — the ruff/flake8/shebang pre-commit failure.** Green in CI, and `pre-commit` was the one red check last round.

6. **ISSUE — the incomplete zstd licence.** `frontend/js/ZSTD-LICENSE` now reproduces the three BSD redistribution conditions and the "AS IS"/DISCLAIMED warranty text, and the 404 URL is gone.

7. **NOTE — the fingerprint CRC did not strip carriage returns as Sphinx does.** Fixed at `sw.js:687-692`, `if (bytes[i] === 0x0d) { continue; }`. This was the latent one; it is now correct rather than accidentally correct.

8. **NOTE — the vendored fzstd had lost strict mode.** The two lines are swapped, and I confirmed the semantics rather than assuming them: a node probe of both shapes gives *directive first → strict (ReferenceError on an undeclared assignment)*, *directive second → not strict*. So it is a real fix, not a cosmetic one.

9. **NOTE — three `run_all.js` flags silently doing the wrong thing.** All three fixed: `:29-30` is now `Number.isFinite(n) && n >= 1 ? Math.floor(n) : dflt` applied to both `--jobs` and `--page-shards`, and `:32` is `args.includes('--all') ? 'all' : opt('only','unit')` so `--all` works as documented.

10. **NOTE — `window.open` after an async probe.** Fixed with the standard pattern: the tab is opened synchronously inside the click and then navigated or closed on the probe's answer, with `tab.opener = null` and a fallback if the synchronous open is itself blocked. Neither pass browser-tested it, so that rests on the pattern being known-good rather than on an observation.

11. **NOTE — the `importScripts` globals.** A comment at the import site plus a `try/catch`, which is one of the two options I offered. The collision hazard itself is unchanged, as that finding said it would be.

#### Still open, and one place I think you are right

12. **ISSUE — the JS decoder still verifies no zstd frame checksum.** `patchJs` still loops `while (!st.l)` and never reads `st.c`. Re-raising it as a known property rather than as a defect, because the two paths I named last round as *lacking a backstop* are both now closed (items 2 and 3), so there is no longer a consumer that can serve a JS-decoded page without a content-hash check. And you documented it instead of leaving it implicit — `wiki-offline-copies.rst:388-391` now says "a delta without that hash is refused rather than served; the JavaScript decoder verifies no checksum of its own, so the hash is what stands between a corrupt delta and a wrong page." That is the honest version of the sentence I called untrue last round, and it closes that half too.

13. **NOTE — the stacked soft failures in the archive build. I think your answer is the right one and I am dropping the ask.** The substance of my complaint was that the feature could vanish silently; that is fixed. `build_offline_artifacts.py` gains an `error()` helper writing to stderr with an ERROR prefix, the zstandard path now names the count dropped ("none of the N parameter versions found are carried; the archive is published without them"), and the previously *completely* silent no-stable-release path now logs too. You chose loud-but-non-fatal — `error()`'s own docstring says "something the build carries on past" — where I had asked for a hard failure. Not failing an entire wiki build over one optional feature is a defensible call, and it is now visible in CI, so I would leave it as you have it. `update.sh:203` is untouched, which follows from the same decision.

#### What is left — two things, neither blocking

**1. The JS decoder still allocates from the untrusted frame header *before* the new bound.**
`frontend/js/zstd-delta.js:1194-1201`. The order is the problem:
```js
function patchJs(delta, base) {
  var head = ApFzstd.rzfh(delta, 1);            // this call already allocates
  if (typeof head !== 'object' || !(head.u > 0)) { throw ... }
  if (head.u > MAX_PAGE_BYTES) { throw ... }    // the bound is checked after
  var out = new Uint8Array(base.byteLength + head.u);
```
Inside `rzfh`, with `w == 1`, `:487` is `var buf = new u8((w == 1 ? (fss || ws) : w ? 0 : ws) + 12);` — that is `fss + 12`, where `fss` is the frame's own declared content size. The only guard there is `if (ws > 2145386496) err(1)`, which bounds the *window*, not `fss`.

Worth being precise, because my two passes reached opposite conclusions on this and I settled it by measurement. Feeding a 12-byte frame declaring `0x7f000000` and intercepting the `Uint8Array` constructor:
```
declared content size in the frame: 2130706432
allocation sizes requested, in order: [ 2130706444 ]
=> the oversize allocation happens BEFORE the MAX_PAGE_BYTES check
```
So the check at `:1199` does fire, and it does prevent the *second*, larger allocation at `:1202` — which is why one pass read this as fully resolved. But the first allocation has already been requested, and the new test named *"a frame claiming 2 GB is refused before the page is allocated"* does not assert what its name says: it checks only the eventual error.

**Not a blocker.** The failure is a `RangeError` or an allocation failure — the page does not rebuild, rather than rebuilding wrongly — which is what I said of the JS side last round too. The wasm path *is* properly bounded, and I re-confirmed `MAX_PAGE_BYTES` still equals `sw.js`'s `STORE_LIMIT_BYTES`. The fix is to read the declared size from the header bytes and check it before the call that allocates the window, rather than after.

**2. The base page's `rawBlock` id is the one part of your fix that no test covers.**
`common_offline_export.js:311`. Reverting just that lookup to the pre-fix `pages.forEach(...)` leaves the export suite **entirely green** while producing a genuinely broken file: the emitted document carried `id="r7"` where the shell computes `"r"+byPath[pg.b]` = `r6`, and a real reader gets *"This parameter list could not be rebuilt in this file."* The cause is `test_offline_export.js:345-351` — `bootShell()` feeds the shell *synthetic* raw bodies instead of reading the exported file's `r` blocks, so it cannot see a raw-base mis-numbering. Both of my passes found this independently.

The shipped code is correct; the coverage is missing, and it is missing on half of the fix you just made — so a future refactor of the `needRaw` loop can silently break every delta-held version in the export with both suites green. The block-id check at `:671-676` already reads the file with `fs.readFileSync`; extending it to the `r` ids closes it.

#### Also worth a look, none of it blocking

- **From an insecure origin the export now refuses every delta-held version, where before it painted one.** `common_offline_document_builder.js:251` throws when `!(window.crypto&&crypto.subtle)`. Exercised in Chromium from three origins: `file://` and `http://localhost` are secure contexts and rebuild fine; `http://<LAN-IP>/` is not, `crypto.subtle` is `undefined`, and the page reads *"could not be rebuilt: no hash to check the rebuilt page against"* — which blames a missing hash when the hash is present and it is SubtleCrypto that is absent. Failing closed is right; the message is not. The documented path (double-click from a USB stick) is `file://` and works, so this only bites someone putting the export on a plain-HTTP share. Separating the two causes would make the message actionable.

- **`leftOut` is counted and then thrown away** (`:355`), so a reader is never told a version was dropped from their file — the only trace is a `console.warn`, while the on-screen message reports the reduced page count as if nothing were missing. Appending to that message when `r.leftOut` is non-zero closes it.

- **The `</script` block escaping is not reversible.** `pageBlock`/`rawBlock` escape with `replace(/<\/(script)/gi, '<\\/$1')` and `unblock` reverses it, but an *already*-escaped `<\/script` is left alone on the way in and then has its backslash eaten on the way out. Latent — a repository grep finds the literal only in the builder and its own test, neither of which becomes a page block. For a base page in a `rawBlock` it fails safe: the bytes then miss their SHA-256 and the version shows an error rather than wrong content.

- Smaller: `parameterVersions` still returns a `drop` key that is never assigned into and no longer read (dead at the previous head too); in the no-toctree fallback `buildNav` is still passed the cache-derived `pages` rather than `index`, so a left-out version can appear as a dead sidebar link (latent — production pages all carry a toctree); `test_build_offline_artifacts.py:178-183` only passes when `zstandard` is installed and should `skipTest` instead, as the youtube suite already does for sphinx; `frontend/js/zstd.wasm` is committed mode 100755 (also true before, and `check-executables-have-shebangs` only inspects text files); and `--only bogus` still silently runs the unit set.

#### How this was checked

Every resolution above was verified against the code at `9009370894` rather than taken from a commit message, and the two that turn on semantics rather than on text were probed rather than argued: the strict-mode swap with a node check of both statement orders, and the `hex16`/`contentHash`/`hexdigest()[:16]` agreement traced across the Python and both JavaScript sides. The rebase-only test was run with the empty-patch guard in place (old patch 7868 lines, new 8227, neither empty) so a failed fetch could not masquerade as "nothing changed".

<sub>A primary read plus an independent second pass; where they disagreed the disagreement is stated above rather than averaged away. Anything marked unconfirmed was not reproduced — please push back on it.</sub>


</details>

AP-Review:
**Automated review note — AI-generated (Claude), cross-checked by an independent Codex pass against the live diff.** Please sanity-check before acting.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_19_2337/devcall_pr_reviews.html#prwiki-8080

**Re-reviewed at head `5d1ed58c29`** (previously `9009370894`) — my earlier comment above is superseded.

### Verdict: COMMENT

**Every finding from the previous round is closed, including both that were left open** — and you proved one of them the right way, by mutation-testing the new test rather than just adding it. The delta is nine files and every one targets something from that list.

The verdict isn't APPROVE only because a **cold** review of this head — run deliberately with no sight of my findings, because a clean APPROVE is the least-validated verdict this process produces — reproduced three defects in a real browser. None of them corrupts data: the integrity design held under everything thrown at it.

CI at this head: 5/5 green.

### Previous round — every finding accounted for

| Previous finding | Status | Evidence at `5d1ed58c29` |
|---|---|---|
| **The JS decoder allocates from the untrusted frame header *before* the new bound** | **RESOLVED** | `zstd-delta.js:1195-1216` adds `declaredSize()`, checked before `ApFzstd.rzfh()`. I verified the parse field by field against fzstd's own `rzfh` — `at = 6 - single + dictbytes` matches `bt = 6 - ss; bt += db`, `width` matches `fsb`, the `+256` for `sizeFlag === 1` matches `+ ((fcf == 1) && 256)` — so the declared size can't diverge upward. And the test isn't vacuous: the same Proxy instrumentation on a *valid* fixture records 3 allocations, largest 465,360 bytes. An independent pass ran 11,264 synthetic header probes; largest trapped first allocation was 12,582,924 bytes, the page limit plus 12 bookkeeping bytes. |
| **The base page's `rawBlock` id is the one part of the fix no test covers** | **RESOLVED, proven by mutation** | `test_offline_export.js:679-687` parses the `r(\d+)` ids out of the written file and compares them with the base page's index positions. Reverting just that lookup to the pre-fix `pages.forEach(...)` form gives *FAIL — blocks 42 for index 41*, exactly the regression demonstrated by hand last round. |
| **From an insecure origin the export blames a missing hash** | **RESOLVED** | `common_offline_document_builder.js:252-253` splits the two causes; covered by a new check at `test_offline_export.js:325-333`. |
| **`leftOut` is counted and then thrown away** | **RESOLVED** | `common_offline_page.js:1348-1353` appends *", N parameter version(s) left out"*, with new tests for both the N>0 and N=0 wordings. |
| **The `</script` block escaping is not reversible** | **RESOLVED, proven by mutation** | The new round-trip check at `test_offline_export.js:245-259` is real: reverting all three regexes to the old form gives *FAIL — a page block closes nothing early and reads back byte for byte*. |
| **Five smaller items** — dead `drop` key, `buildNav` passed cache-derived pages, `--only bogus` silently running the unit set, `zstd.wasm` at mode 100755, the Python test not skipping without zstandard | **all RESOLVED** | On `buildNav` I checked the field use rather than assuming: `listNav`/`listOrder` read only `p.path` and strip a `.html?` suffix bare index paths don't carry, so nothing is lost. The mode change is invisible in a patch since the blob is unchanged — the git tree gives 100755 at the old head and 100644 here. |
| **The JS decoder verifies no zstd frame checksum** | **STILL OPEN**, as a known property | Unchanged and, as last round, acceptable: every consumer checks the `APDELTA1` content hash and `wiki-offline-copies.rst:388-391` states it. No action asked. |

### New findings

- **ISSUE** — [`frontend/sw.js:12-18`](https://github.com/ArduPilot/ardupilot_wiki/blob/5d1ed58c29/frontend/sw.js#L12-L18) — **a transient failure to load the decoder disables delta reconstruction permanently for that worker.** `importScripts('/js/zstd-delta.js')` is in a `try/catch`, so a one-off 503 or a blip at install time is swallowed with a `console.warn` and installation succeeds without `ApZstd`. Reproduced in Chromium by returning 503 for the first decoder request: a worker restart *and* an explicit `registration.update()` both left the decoder undefined, with no further request for it.
  **Your comment at `:9-11` anticipates the case** — "a worker may not import a new script once it is installed… those versions are simply not held; everything else is unaffected" — and that's right for a decoder that is genuinely absent. What it doesn't cover is that the only cause here is *transience* (it's a first-party asset that always ships), and that nothing ever retries: `sw.js` is byte-identical, so the browser's update check finds no difference and keeps the crippled worker until `CACHE_VERSION` or the file changes. A lazy retry isn't available to you — `importScripts` is only legal during initial evaluation — so the honest options are to let the install fail so the browser retries, or to have the page detect a decoder-less worker and re-register.

- **ISSUE** — [`common_offline_document_builder.js:258`](https://github.com/ArduPilot/ardupilot_wiki/blob/5d1ed58c29/common/source/_static/common_offline_document_builder.js#L258) — **the stale-paint guard compares a raw route against a normalised path, so a *successful* rebuild can leave "Rebuilding this parameter list…" on screen.** `current()` (`:89`) returns the location hash as-is, while `path` comes from `lookup(raw)` at `:261` and carries no `.html` suffix — `:379` strips `\.html?$` when resolving, which is what stored paths look like. Opening `#/rover/docs/parameters-Rover-stable-V4.6.0.html` therefore fails the equality test and the paint is skipped; reopening works only because the result is cached. Normalise both sides through the same function before comparing.

- **ISSUE** — [`common_offline_document_builder.js:259`](https://github.com/ArduPilot/ardupilot_wiki/blob/5d1ed58c29/common/source/_static/common_offline_document_builder.js#L259) — **the failure path has no stale-paint guard at all**, so a late error overwrites whatever the reader has navigated to. The success path on the line above guards with `if(current()===path)`; the `.catch` paints unconditionally. Found independently by both passes, and reproduced: delaying the digest, navigating from a parameter version to the homepage, then resolving with a mismatching hash replaced the homepage with the old version's error. Fix is the guard the line above already uses.

- **NOTE** — `common_offline_export.js:309-315` — latent under-reporting in the new `leftOut` accounting. When a carried delta's base page isn't in the index or among the saved pages, `if (slot === -1 || !entry) { return undefined; }` skips its `rawBlock` silently: the delta entry stays, the reader gets "could not be rebuilt in this file", `leftOut` isn't incremented and nothing is warned — so the Saved message reports a whole copy. Not reachable through the UI today. A `console.warn` plus `leftOut++` would match what the failed-rebuild branch at `:270-273` already gets.

- **NOTE** — `build_offline_artifacts.py:510` — a missing `zstandard` logs an error, publishes reduced artifacts and exits zero. A fixture with an index and two stable parameter versions produced an archive containing only the index, while its manifest advertised three pages. Deliberate and tested, so a note — but a build that ships a manifest describing pages it didn't write is a shape CI can't detect. A non-zero exit, or a manifest-vs-content assertion at the end of the build, would close it.

- **NOTE** — `scripts/tests/test_offline_page.js` — seven bare `{ … }` blocks in `main()` (`:575, 932, 1017, 1044, 2715, 2730, 2746`) sit outside `section()`, and `section()` is what `--shard i/n` filters on, so every shard re-runs them. Measured: unsharded 323 checks, the four shards 86 + 92 + 74 + 98 = 350, with exactly 9 checks running 4 times each. Nothing is lost — the union covers all 322 distinct names — but the split isn't a partition and the per-shard counts mislead.

- **NOTE** — `dev/source/docs/wiki-offline-copies.rst:388-395` describes the hash check but not the constraint it implies: SubtleCrypto is required, so a delta-held version opens from `file://`, `https://` or localhost but refuses on a plain-`http://` LAN share. The in-file message now says so; one sentence here would stop it surprising someone. Also trivial: `test_zstd_delta.js:119` hardcodes `12 * 1024 * 1024` rather than deriving it from `MAX_PAGE_BYTES`.

### What was checked and held

**The integrity design is the claim that most needed testing, and it held.** The cold pass drove wrong bases, wrong hashes, missing bases, truncated frame payloads and unavailable SubtleCrypto through both decoders and found **no path to a silently wrong parameter table** — every one was refused visibly. There's also no delta *chain* to worry about, which was the first thing I looked for: `build_offline_artifacts.py:502-540` writes every carried version as a delta against the single newest-stable base named in the `APDELTA1` header, so one corrupt delta costs exactly one version and fails closed. Reconstruction is bounded at both ends, the wasm path frees in one `finally`, and `ApZstd.init` memoises via `ready` so repeated version switches don't re-instantiate the module. One honest limit: the 12 MB cap bounds the *first* allocation, not total memory — the combined base/output buffer and other copies are additional.

Ran: a full rover wiki build from this head in a fresh venv, then `test_zstd_delta.js` (17 checks), `test_offline_export.js rover` (118), the page suite unsharded (323) and as 4 shards (350), and the whole unit set — all green except `worker` and `archives`, which fail loudly for want of a full `update.py` build and are environmental, not yours. `run_all.js` does exit 1 when a suite fails. Two mutation tests and one instrumentation probe, to show the new coverage isn't decorative. **Not done:** the Playwright browser suites weren't run here.

<sub>Both a primary read and an independent cold second pass were run; the cold pass produced the first and third findings above and argued for a harsher verdict than I adopted. Anything marked *unconfirmed* was not reproduced — please push back on it.</sub>

